PR #7, opened 2015-10-24
[RFC] resize-rootfs: recipe to resize the rootfs on first boot.

koenkooi:
I haven't actually run tested this yet, only inspected buildhistory.

Signed-off-by: Koen Kooi <koen@dominion.thruhere.net>

### DIFF
--- a/recipes-bsp/firstboot/resize-rootfs.bb
+++ b/recipes-bsp/firstboot/resize-rootfs.bb
@@ -1,0 +1,18 @@
+SUMMARY = "Resize rootfs to fill the complete partition"
+
+LICENSE = "MIT"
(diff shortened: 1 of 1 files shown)


ndechesne:
hi,
i don't think we should assume in this recipe that rootfs is  mount as /dev/disk/by-partlabel/rootfs. This is true for db410c currently, but the recipe isn't meant to be specific for this board. 

we should be able to find where / is mounted dynamically, instead no?

koenkooi:
I think so, /dev/rootfs should work for that. I'll take a look next week

fboudra:
on Debian, I'm going toward a resize-helper script and a systemd service to run it on first boot. It will be generic for any board and won't assume 410c specifically.

I think we should adopt a common approach.

ndechesne:
I am closing it, since we now have a different (better) solution coming from meta-96boards.

PR #9, opened 2015-12-03
skales-native: depend on -native recipes for python & dtc

avoltz:
This fixes a build error I ran into testing jethro on the dragonboard.

dtbTool (provided by skales) uses dtc. The skales recipe depends on dtc, but not the native version. If dtc-native is not built/installed to the sysroot before virtual/kernel deploy is run, the build fails. Since skales is a native recipe, it seems correct to use -native deps.

I reproduced the error by simply running the deploy recipe from a clean state. 'bitbake -c deploy virtual/kernel'

### DIFF
--- a/recipes-devtools/skales/skales_git.bb
+++ b/recipes-devtools/skales/skales_git.bb
@@ -8,8 +8,6 @@ LIC_FILES_CHKSUM = "file://mkbootimg;beginline=3;endline=29;md5=114b84083e657f38
 
 DEPENDS = "python dtc"
 
-inherit native
-
 SRCREV = "15ece94f09f534105f28c944d0030f8953e8267f"
 PV = "0.1+git${SRCPV}"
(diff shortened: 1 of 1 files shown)


avoltz:
Renaming the recipe seems fine too, if skales is meant to be used on the target. I wasn't sure if it ever had a purpose on target. Though I guess at the very least it would be useful for building the same qcom image on the board--a strange but valid use case.

I think if you drop -native from the recipe name, it would need RDEPENDS for python and dtc for the target case.

Do you want me to try these changes and rework the fix?

ndechesne:
if you can try that, this would be nice.. thanks!

avoltz:
Made your suggested changes and retested. It fixes the bug I was running into.

Let me know what you think

ndechesne:
thanks, i merged it!

PR #12, opened 2016-03-31
Fixes for booting core-image-sato built with musl/clang on dragon board

kraj:


recipes-graphics/xorg-xserver/xserver-xf86-config/10-preload-modules.conf
```diff
@@ -0,0 +1,10 @@
+Section "Module"
+    Load "fb"
+    Load "shadow"
+    Load "int10"
+    Load "vbe"
+    Load "vgahw"
+    Load "exa"
+    Load "freedreno"
```
ndechesne: hmm. the whole idea of recent xserver is to autodetect and automatically load the right hw module at run time. I am not sure I understand why we need that.. 

also in the other commit you add -modesetting , so why would we have both freedreno and modesetting

kraj: yes good catch. I have updated the pu branch which removes them. Please take a look again

recipes-graphics/xorg-xserver/xserver-xf86-config_0.1.bbappend
```diff
@@ -0,0 +1 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
```
ndechesne: should we do that only for a selected set of machines? with this change we are impacting all BSP that might have meta-qcom in bblayers..

ndechesne: and do we really have to override that file? can't we provide an additional conf file with the missing modules that we need for our platforms instead?

kraj: OK fair enough. Update the pu branch

kraj: override is suited here, since the files from OE-core are for qemu targets. and they have modules which needs to be removed when loading on dragoboard.

kraj:
@ndechesne anything more needed from my side ?

recipes-graphics/xorg-xserver/xserver-xf86-config_0.1.bbappend
ndechesne: right. i missed the fact that it was a machine override.. do we need similar override for ifc6410?

kraj: I did not test it on that board. but yes it will be needed if its using same GPU, however that could be done separately, there should be no regression seen for that board. Only musl based x images wont start but that was the case anyway.

conf/machine/dragonboard-410c.conf
```diff
@@ -12,6 +12,7 @@ XSERVER_OPENGL ?= "xf86-video-freedreno \
                   "
 
 XSERVER ?= "xserver-xorg \
+            xserver-xorg-module-libint10 \
```
ndechesne: sorry for being so slow to reply these days.. is there a specific module that uses this one? in other words should we do a RDEPENDS at the right place instead?

recipes-graphics/xorg-driver/xf86-video-freedreno_1.4.0.bb
```diff
@@ -4,6 +4,10 @@ SUMMARY = "X.Org driver for Adreno mobile GPUs"
 
 #FIXME:
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+SRC_URI += "file://0001-fbmode_display-Add-a-weak-definition-of-fd_bo_from_f.patch \
```
ndechesne: this doesn't seem to be the right fix. that shows a problem in libdrm in fact.  in fact, when libdrm is built without kgsl backend enabled (which is our case by default) this function is not defined in libdrm, hence your problem.

can you try this libdrm patch instead:
http://hastebin.com/oqofamazix

if you confirm it works (it should) i will send it out upstream, and we will need it in oe-core until we upgrade libdrm.

conf/machine/dragonboard-410c.conf
kraj: If you remove it.then graphics doesnt come up. This is minimum set that is required. I dont remember which other module has dep on it. But it sure is needed

recipes-graphics/xorg-driver/xf86-video-freedreno_1.4.0.bb
kraj: I am away from home until friday so cant test it however, I think this definition would work. It should probably be weak. I can test it upon return.

ndechesne: I don't think we need a weak synbol. when HAVE_KGSL is defined the function is implemented. the problem was that if HAVE_KGSL is not defined then it's missing. so the issue you found with xf86-video-freedreno is a 'real' bug since the function is undefined (though we don't use it in our use case anyways). so the fix is to make sure that this function is defined when HAVE_KGSL is not defined.

kraj: alright, so I dropped the old patch and tried this patch, it worked fine with musl image core-image-sato booted ok. So I updated the pull request with patch applied to libdrm via a bbappend. When this gets into libdrm we can drop it but until then its applied via this layer.

ndechesne:
@kraj : https://cgit.freedesktop.org/mesa/drm/commit/?id=739b8886025b31f2f9d09a535c8ff3670bb1d036

kraj:
@ndechesne bravo

PR #13, opened 2016-05-06
alsa-lib: fix patch

twoerner:
Upstream has added support for another hardware platform so this patch needs a
small adjustment to continue to apply.

Signed-off-by: Trevor Woerner <twoerner@gmail.com>


ndechesne:
thanks, btw, I also merged the same patch in the krogoth branch which is now available.

PR #15, opened 2016-07-11
libdrm bbappend: patch fails to apply to 2.4.68, lock bbappend to OE-core version 2.4.67

koenkooi:
Signed-off-by: Koen Kooi <koen@dominion.thruhere.net>


ndechesne:
i had a similar patch , which I apparently forgot to push.. i just did it..

PR #18, opened 2016-08-28
Lk bootloader builds

madisongh:
This patch series is for building the lk bootloader for the dragonboard-410c, along with the toolchain pieces required to get that build to work.

If accepted, I'll submit a further pull request to bring in the rest of the bits needed to build a complete fastboot-able set of binaries for complete initialization of the Dragonboard's eMMC.

### DIFF
--- a/recipes-bsp/lk/lk-linaro-qcomlt_git.bb
+++ b/recipes-bsp/lk/lk-linaro-qcomlt_git.bb
@@ -1,0 +1,47 @@
+DESCRIPTION = "Little Kernel bootloader for MSM platforms"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=5a1abdab641eec675725c843f43f03af"
+
+SRC_URI = "git://${LK_GIT_REPO};branch=${SRCBRANCH}"
+LK_GIT_REPO ??= "git.linaro.org/landing-teams/working/qualcomm/lk.git;protocol=http"
+SRCBRANCH = "release/LA.BR.1.2.4-00310-8x16.0"
+SRCREV = "2740fc8aeb78bb2e012f63f6d500f3133139c504"
+PV = "1.2.4-00310+${SRCPV}"
+
+COMPATIBLE_MACHINE = "(dragonboard-410c)"
+MAKE_TARGET = ""
+MAKE_ARGS = ""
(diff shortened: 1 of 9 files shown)


ndechesne:
hi, thanks for the patches! I have been willing to intregrate LK for some time.. however i am wondering if we really need to go that far and include the arm-non-eabi cross compiler. can't we get LK to build with the regular toolchain instead ? (OE can build uboot or grub for example..)

i am a bit worried about the maintenance of adding the extra cross compiler in a BSP layer.. 

what do you think?

madisongh:
I'd like to avoid the extra toolchain recipes, too, but I wasn't sure we could coax the aarch64 toolchain into producing 32-bit baremetal code, which is what we need for lk.   I'll look into and see.

koenkooi:
According to the TCWG in linaro, you can't generate 32bit code with an aarch64 toolchain :/

What could be used is something like https://github.com/96boards/oe-rpb-manifest/pull/45 , but that would be distro specific. The 96boards hikey machine has a similar problem and does fetch an external toolchain: https://github.com/96boards/meta-96boards/blob/master/recipes-bsp/l-loader/l-loader_git.bb

madisongh:
Yes, I found the same - two different back-ends, so at least for now the aarch64 toolchain can't generate aarch32 code.  I found the meta-ti layer also pulls in a prebuilt 32-bit toolchain for its recipes that need 32-bit baremetal code generation, although it has a separate recipe for it.

Another approach would be to use multilib and have the recipe depend on the 32-bit flavor of the toolchain, but that carries with it a bunch of other undesirable baggage, and I'm not sure that the 32-bit Linux ABI-targeted toolchain would generate the correct baremetal code for LK's purposes anyway.

ndechesne:
well, if we have to get into multilib, then i would probably reconsider my initial statement ;-)

my main fear/worry is the maintenance cost for keeping the 32 bit non linux cross compiler in meta-qcom.. i still feel that it's wrong to have it there... @koenkooi is there a better place for that, or do you think it's _ok_ to keep it in meta-qcom?

koenkooi:
I'd go for a 32bit helper machine, since toolchain maintenance is not something you want to do. But from a 'one command to rule them all', but extra toolchain can be considered more user friendly.

rsalveti:
Pre-built toolchain would be easier to maintain at least, and could already unblock this merge (depends on Nico).

PR #19, opened 2016-08-28
Replace overrides in kernel recipes with machine/SoC settings

madisongh:
I put these changes together to make it easier to add a new target and to add bbappends to customize these recipes where needed, and to do some cleanup to eliminate redundant code and make version changes detectable at build time when SRCREV changes (by using SRCPV in PV).

I've been able to test these changes with my db410c, but couldn't test with others as I don't have the hardware.

### DIFF
--- a/conf/machine/include/qcom-apq8016.inc
+++ b/conf/machine/include/qcom-apq8016.inc
@@ -25,6 +25,14 @@ PREFERRED_PROVIDER_virtual/libgles2 ?= "mesa"
 PREFERRED_PROVIDER_virtual/xserver ?= "xserver-xorg"
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-linaro-qcomlt"
 
+KERNEL_DEFCONFIG ?= "${S}/arch/arm64/configs/defconfig"
+KERNEL_IMAGETYPE = "Image"
+
+BOOTIMG_PAGE_SIZE ?= "2048"
+BOOTIMG_BASE_ADDR ?= "0x80000000"
+DT_IMAGE_BASE_NAME ?= "dt-${KERNEL_IMAGE_BASE_NAME}"
+DT_IMAGE_SYMLINK_NAME ?= "dt-${KERNEL_IMAGE_SYMLINK_NAME}"
+
 # Fastboot expects an ext4 image, which needs to be 4096 aligned
 IMAGE_FSTYPES_append = " ext4.gz"
 IMAGE_ROOTFS_ALIGNMENT = "4096"
--- a/conf/machine/include/qcom-apq8064.inc
+++ b/conf/machine/include/qcom-apq8064.inc
@@ -23,6 +23,14 @@ PREFERRED_PROVIDER_virtual/libgles2 ?= "mesa"
 PREFERRED_PROVIDER_virtual/xserver = "xserver-xorg"
 PREFERRED_PROVIDER_virtual/kernel = "linux-linaro-qcomlt"
 
+KERNEL_DEFCONFIG ?= "${S}/arch/arm/configs/qcom_defconfig"
+KERNEL_IMAGETYPE = "zImage"
+
+BOOTIMG_PAGE_SIZE ?= "2048"
+BOOTIMG_BASE_ADDR ?= "0x80200000"
+DT_IMAGE_BASE_NAME ?= ""
+DT_IMAGE_SYMLINK_NAME ?= ""
+
 # Fastboot expects an ext4 image, which needs to be 4096 aligned
 IMAGE_FSTYPES_append = " ext4.gz"
 IMAGE_ROOTFS_ALIGNMENT = "4096"
--- a/recipes-kernel/linux/linux-linaro-qcomlt-tracking_4.6.bb
+++ b/recipes-kernel/linux/linux-linaro-qcomlt-tracking_4.6.bb
@@ -11,11 +11,10 @@ LOCALVERSION ?= "-linaro-lt-qcom"
 SRCBRANCH ?= "integration-linux-qcomlt"
 SRCREV ?= "5574089362de2fb0ebbc0c4d9bd48203b59b7b1b"
 SRC_URI = "${LINUX_LINARO_QCOM_GIT};nobranch=1"
+PV .= "+git${SRCPV}"
 
 COMPATIBLE_MACHINE = "(apq8064|apq8016|apq8096)"
 
-KERNEL_DEFCONFIG_aarch64 ?= "${S}/arch/arm64/configs/defconfig"
-KERNEL_DEFCONFIG_apq8064 ?= "${S}/arch/arm/configs/qcom_defconfig"
 KERNEL_CONFIG_FRAGMENTS += "${S}/kernel/configs/distro.config"
 
 # append DTB, since bootloader doesn't support DTB
@@ -30,8 +29,3 @@ do_compile_append_apq8064() {
 
 # Wifi firmware has a recognizable arch :(
 ERROR_QA_remove = "arch"
-
-QCOM_BOOTIMG_ROOTFS_dragonboard-410c = "mmcblk0p10"
-QCOM_BOOTIMG_ROOTFS_dragonboard-820c = "sde18"
-QCOM_BOOTIMG_ROOTFS_ifc6410 = "mmcblk0p12"
-QCOM_BOOTIMG_ROOTFS_sd-600eval = "mmcblk0p12"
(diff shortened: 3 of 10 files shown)


conf/machine/include/qcom-apq8016.inc
```diff
@@ -25,6 +25,14 @@ PREFERRED_PROVIDER_virtual/libgles2 ?= "mesa"
 PREFERRED_PROVIDER_virtual/xserver ?= "xserver-xorg"
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-linaro-qcomlt"
 
+KERNEL_DEFCONFIG ?= "${S}/arch/arm64/configs/defconfig"
```
ndechesne: I really think that the defconfig is a kernel recipe variable, and does not belong to the machine. e.g. it's an implementation detail for the kernel recipe.

conf/machine/include/qcom-apq8016.inc
```diff
@@ -25,6 +25,14 @@ PREFERRED_PROVIDER_virtual/libgles2 ?= "mesa"
...
+KERNEL_IMAGETYPE = "Image"
+
+BOOTIMG_PAGE_SIZE ?= "2048"
+BOOTIMG_BASE_ADDR ?= "0x80000000"
+DT_IMAGE_BASE_NAME ?= "dt-${KERNEL_IMAGE_BASE_NAME}"
```
ndechesne: this variable was not duplicated for each new machines. why do you think it's better to move it in the soc conf file?

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -1,64 +1,44 @@
...
-    depends = d.getVar("DEPENDS", True)
-    depends = "%s skales-native" % depends
-    d.setVar("DEPENDS", depends)
+DEPENDS += "skales-native"
+
+BOOT_IMAGE_BASE_NAME ?= "boot-${KERNEL_IMAGE_BASE_NAME}"
+BOOT_IMAGE_SYMLINK_NAME ?= "boot-${MACHINE}"
+
+python __anonymous() {
+    if d.getVar('SERIAL_CONSOLES', True):
```
ndechesne: i am not strictly opposed to this piece of change, but I don't think it's needed in this patch, and this can be done separately later.

I think there are good things in this patch, to avoid duplication across machines. and it would be simpler to review/merge if the patch was just replacing variables/removing duplication only. for SERIAL_CONSOLES, i think we can keep the current implementation, and improve it later.

conf/machine/include/qcom-apq8064.inc
```diff
@@ -23,6 +23,14 @@ PREFERRED_PROVIDER_virtual/libgles2 ?= "mesa"
 PREFERRED_PROVIDER_virtual/xserver = "xserver-xorg"
 PREFERRED_PROVIDER_virtual/kernel = "linux-linaro-qcomlt"
 
+KERNEL_DEFCONFIG ?= "${S}/arch/arm/configs/qcom_defconfig"
+KERNEL_IMAGETYPE = "zImage"
+
+BOOTIMG_PAGE_SIZE ?= "2048"
+BOOTIMG_BASE_ADDR ?= "0x80200000"
+DT_IMAGE_BASE_NAME ?= ""
```
ndechesne: is that really needed? since we don't need these vars, maybe it's even better/simpler to not define them at all.

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -1,64 +1,44 @@
...
+    else:
+        d.setVar('BOOTIMG_DT_ARG', '')
 }
 
-QCOM_BOOTIMG_ROOTFS ?= "undefined"
+BOOTIMG_CMDLINE ?= "${BOOTIMG_CMDLINE_ROOT} ${BOOTIMG_CMDLINE_CONSOLE}"
 
-# set output file names
-DT_IMAGE_BASE_NAME = "dt-${PKGE}-${PKGV}-${PKGR}-${MACHINE}-${DATETIME}"
-DT_IMAGE_BASE_NAME[vardepsexclude] = "DATETIME"
```
ndechesne: I think this change is lost in the patch, can you check b6d82b37dd4330d79450f0ae2936a0eb1dac34c4 , if i read that properly your patch would reintroduced what I had fixed, no?

recipes-kernel/linux/linux-linaro-qcomlt-tracking_4.6.bb
```diff
@@ -30,8 +29,3 @@ do_compile_append_apq8064() {
 
 # Wifi firmware has a recognizable arch :(
 ERROR_QA_remove = "arch"
-
-QCOM_BOOTIMG_ROOTFS_dragonboard-410c = "mmcblk0p10"
```
ndechesne: i really like the removal of these vars here, it was on my todo list..

ndechesne:
thanks for the patch.. I really like where it's going. i tried to make a few comments. let me know what you think.

madisongh:
Thanks for the review.  I'll take a whack at making the changes more incremental and see how to address the specific issues you pointed out.

conf/machine/include/qcom-apq8016.inc
madisongh: But it varies based on MACHINE (or SoC in this case), and I was looking to consolidate all the required settings for adding a new machine to the machine config files.  Then again, it is rather specific to the linux-linaro-qcom set of recipes, so maybe it does make sense to keep it back there and use the overrides.

conf/machine/include/qcom-apq8016.inc
madisongh: An oversight on my part.  Thanks.

conf/machine/include/qcom-apq8064.inc
madisongh: Yep, could do it that way.

recipes-kernel/linux/linux-qcom-bootimg.inc
madisongh: I changed the definition to use ${KERNEL_IMAGE_BASE_NAME}, which already has that flag set.  That was more in keeping with what I'd seen in other BSP layers (and would have the device tree filename follow any custom setting of KERNEL_IMAGE_BASE_NAME that a user might set), but perhaps was a bit gratuitous.  I could put back the original definition, in which case we would need to set the vardepsexclude flag again.

ndechesne: i was wondering if that would be enough to rely on vardepsexclude from the other variable. but I wasn't sure if that would work. My comment was just to make sure that you tried this out. I think it is good to use KERNEL_IMAGE_BASE_NAME, though this can be done in a separate patch.

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -1,64 +1,44 @@
...
-    echo "This is not an initrd" > ${B}/initrd.img
-
-    mkbootimg --kernel ${B}/arch/${ARCH}/boot/Image \
-              --ramdisk ${B}/initrd.img \
-              --output ${DEPLOYDIR}/${BOOT_IMAGE_BASE_NAME}.img \
-              --dt ${DEPLOYDIR}/${DT_IMAGE_BASE_NAME}.img \
-              --pagesize 2048 \
-              --base 0x80000000 \
-              --cmdline \
-              "root=/dev/${QCOM_BOOTIMG_ROOTFS} rw rootwait console=${ttydev},${baudrate}n8"
```
ndechesne: just found out that 'rw' got lost in your patch.

madisongh:
Closing this, as I've started on a new, more incremental, series.

recipes-kernel/linux/linux-qcom-bootimg.inc
madisongh: OK, noted, but isn't rw the default anyway?

ndechesne: I am seeing this, when booting a fresh rootfs without 'rw':

[    5.636019] EXT4-fs (mmcblk0p10): mounted filesystem with ordered data mode. Opts: (null)
[    5.639481] VFS: Mounted root (ext4 filesystem) readonly on device 179:10.

madisongh: Huh, OK, I've usually seen it the other way round - having to use ro to get a read-only rootfs, with read/write the default.  I'll make sure that doesn't get lost in the updated series.

PR #20, opened 2016-09-02
Kernel recipe patches v2, part 1 - reduce duplication

madisongh:
Moves settings and a function _append that were duplicated across the kernel recipes into common .inc files.


ndechesne:
thanks. much easier to review, so far ! note this is likely breaking the recipe linux-linaro-lt-freedreno, (since it includes the same .inc) which we used long time ago. but don't worry about that, i have been willing to nuke that recipe for some time, so I will go ahead and remove it. if we need to add it back later (or add any QCOM Android based kernel) we will implement differently. This recipe is not used anymore, afaik.

i will look at the series again tonight, and will likely merge it as-is.

PR #21, opened 2016-09-04
Kernel recipe patches v2,  part 2 - bootimg-related cleanup

madisongh:


recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -6,38 +6,24 @@ QCOM_BOOTIMG_ROOTFS_dragonboard-820c = "sde18"
 QCOM_BOOTIMG_ROOTFS_ifc6410 = "mmcblk0p12"
 QCOM_BOOTIMG_ROOTFS_sd-600eval = "mmcblk0p12"
 
+QCOM_BOOTIMG_BUNDLE_DT = "1"
+QCOM_BOOTIMG_BUNDLE_DT_apq8064 = "0"
```
ndechesne: if these variables are machine specific, i would prefer to set them in the machine conf file. unlike the defconfig there are not 'kernel' variable, and are really specific to the machine (and the bootloaders). So we could set QCOM_BOOTIMG_KERNEL_BASE in the machine (.conf or .inc). While at it, i would put the page size there too. Even thought it's 2048 in both cases, let's assume it can change at some point. 

QCOM_BOOTIMG_ROOTFS can also be put in the machine conf file, for the same reason.

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -17,6 +17,18 @@ DT_IMAGE_SYMLINK_NAME = "dt-${KERNEL_IMAGE_SYMLINK_NAME}"
 BOOT_IMAGE_BASE_NAME = "boot-${KERNEL_IMAGE_BASE_NAME}"
 BOOT_IMAGE_SYMLINK_NAME = "boot-${KERNEL_IMAGE_SYMLINK_NAME}"
 
+# append DTB if bootloader doesn't support DTB
```
ndechesne: i don't think this patch is needed:

1. we append DTB to zImage because it's a bootloader limitation on all apq8064 devices (at least for now). unless you want to reuse this trick on a non-8064 platform, i think it's simpler / more obvious to keep an _apq8064 override. If we ever have a 8064 platform with an improved bootloader (which can use the DT image), we can revisit

2. I don't think it is correct to assume that QCOM_BOOTIMG_BUNDLE_DT == 0 means we need to append the DTB. For example, we can also append the DTB even with a DT image bundled. So if we needed to make a if statement, we would need another parameter.

madisongh: OK, thanks for clarifying.

ndechesne:
merged a more recent version, so closing

PR #22, opened 2016-09-05
Kernel recipe patches, part2 - bootimg-related patches, v2

madisongh:
This replaces #21, modified based on the comments there:
- dropped the do_compile_append change
- moved the variable settings to the machine/SoC conf files
- added a variable for the page size

### DIFF
--- a/conf/machine/include/qcom-apq8064.inc
+++ b/conf/machine/include/qcom-apq8064.inc
@@ -26,3 +26,7 @@ PREFERRED_PROVIDER_virtual/kernel = "linux-linaro-qcomlt"
 # Fastboot expects an ext4 image, which needs to be 4096 aligned
 IMAGE_FSTYPES_append = " ext4.gz"
 IMAGE_ROOTFS_ALIGNMENT = "4096"
+
+QCOM_BOOTIMG_KERNEL_BASE ?= "0x80200000"
+QCOM_BOOTIMG_PAGE_SIZE ?= "2048"
+QCOM_BOOTIMG_BUNDLE_DT ?= "0"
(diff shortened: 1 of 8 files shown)


conf/machine/include/qcom-apq8064.inc
```diff
@@ -26,3 +26,7 @@ PREFERRED_PROVIDER_virtual/kernel = "linux-linaro-qcomlt"
 # Fastboot expects an ext4 image, which needs to be 4096 aligned
 IMAGE_FSTYPES_append = " ext4.gz"
 IMAGE_ROOTFS_ALIGNMENT = "4096"
+
+QCOM_BOOTIMG_KERNEL_BASE ?= "0x80200000"
+QCOM_BOOTIMG_PAGE_SIZE ?= "2048"
+QCOM_BOOTIMG_BUNDLE_DT ?= "0"
```
ndechesne: nitpick.. do we really need to set BUNDLE_DT to 0, since we later check if it's set to 1, maybe we can leave it undefined?

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -1,43 +1,18 @@
 DEPENDS += "skales-native"
 
-QCOM_BOOTIMG_ROOTFS ?= "undefined"
```
ndechesne: maybe we can leave the ?= to undefined, with a comment that this is expected to be set in the MACHINE conf file. It will be more obvious for anyone making a new machine.

ndechesne:
merged a more recent version, so closing

PR #2, opened 2015-10-17
dragonboard-410c: recommend wifi modules

koenkooi:
This makes wifi work out of the box:

root@dragonboard-410c:~# dmesg | grep wcn | grep -v Cap
[    0.224776] irq: no irq domain found for /soc/wcnss-smp2p/slave-kernel !
[    2.074374]  remoteproc0: Direct firmware load for wcnss.mdt failed with error -2
[    4.953539] wcnss: IRIS Reg: 0x51120004
[    5.165072]  remoteproc0: Booting fw image wcnss.mdt, size 7260
[    8.576187] wcnss: received WCNSS_CBC_COMPLETE_IND from FW
[    8.577049] wcn36xx-msm a000000.qcom,wcn36xx: wcn36xx_msm_probe initialized
[    8.596453] wcn36xx wcn36xx: Direct firmware load for wlan/macaddr0 failed with error -2
[    8.596545] wcn36xx wcn36xx: Failed (-2) to read macaddressfile wlan/macaddr0, using a random address instead
[    8.606677] wcn36xx: mac address: 00:0a:f5:3e:dd:52
[    9.019923] wcn36xx: firmware WLAN version 'WCN v2.0 RadioPhy vRhea_GF_1.12 with 19.2MHz XO' and CRM version 'CNSS-PR-2-0-1-2-c1-00010'
[    9.020020] wcn36xx: firmware API 1.5.1.2, 41 stations, 2 bssids
[   48.928342] wcn36xx: ERROR SMD_EVENT (259) not supported
[  112.929315] wcn36xx: ERROR SMD_EVENT (259) not supported

After configuring connman:

root@dragonboard-410c:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host
       valid_lft forever preferred_lft forever
2: sit0@NONE: <NOARP> mtu 1480 qdisc noop
    link/sit 0.0.0.0 brd 0.0.0.0
3: wlan0: <BROADCAST,MULTICAST,UP,LOWER_UP8000> mtu 1500 qdisc mq qlen 1000
    link/ether 00:0a:f5:3e:dd:52 brd ff:ff:ff:ff:ff:ff
    inet 172.20.0.139/24 brd 172.20.0.255 scope global wlan0
[...]

PR #23, opened 2016-09-05
Kernel recipe patches, part2 bootimg-related cleanup, v3

madisongh:
#22 with comments incorporated

### DIFF
--- a/conf/machine/dragonboard-410c.conf
+++ b/conf/machine/dragonboard-410c.conf
@@ -19,4 +19,6 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     ${@'firmware-qcom-dragonboard410c' if d.getVar('ACCEPT_EULA_dragonboard-410c', True) == '1' else ''} \
 "
 
+QCOM_BOOTIMG_ROOTFS ?= "mmcblk0p10"
+
 CMDLINE = "console=ttyMSM0,115200n8 root=/dev/mmcblk0p10 rootwait"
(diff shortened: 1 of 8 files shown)


ndechesne:
merged a slightly modified version (i have only move the symlink creation for the DT image in the last commit).

ndechesne:
btw, thanks a lot for your patches!

ndechesne:
and btw (again), I have boot tested images on DB820c and IFC6410.

ndechesne:
hi, i just realized that there is a small issue with 589bbe8, before that patch, the symlink for the boot image was boot-${MACHINE}.img, and now that it has become boot-${KERNEL_IMAGE_SYMLINK_NAME}.img which corresponds to boot-${KERNEL_IMAGETYPE}-${MACHINE}.img.. Was it really intentional? I can fix it up in 2 ways:
1. change it back to what it was
2. make an additional symlink, in case you really wanted to have the one you did..

madisongh:
Sorry, that difference slipped my mind when I put together the patches.  Including the image type in the symlink is more aligned with how naming works for other platforms, and allows for the possibility that at some point there will be multiple kernel image types.

I think the second approach would be better, to have the backward compatibility in naming but making it easier to implement later-stage image creation/packaging that could work across multiple platforms without having to special-case the naming difference.

ndechesne:
i prefer #2 as well, i will do that quickly.

ndechesne:
now that I look even closer at it.. KERNEL_IMAGE_SYMLINK_NAME definition has changed in master (after krogoth), in 849b67b2e4820564b5e5c9bd4bb293c44351c5f3. So we only need the fixup in krogoth branch in fact..

PR #30, opened 2016-10-29
gstreamer1.0-plugins-v4l2: Remove introspection patch forced by OE-Core

kraj:
OE-Core tries to patch every plugin thats including the .inc file
from OE-core, this however assumes that common repo is at same srcrev
for external plugins. Unfortunately this is not the case with this
v4l2 plugin. Therefore we remove the OE-Core patch and instead apply
similar patch but stored locally in qcom layer which applies cleanly
on the srcrev used by v4l2 plugin for common repo

Signed-off-by: Khem Raj raj.khem@gmail.com


ndechesne:
sorry about the issue.. as discussed earlier , this plugin belongs to the past ;-) the future is the upstream v4l2 decoder driver from -good. In fact this plugin probably doesn't even work with the current version of gst in oe/master.. so I will just remove it from the tree (in the master branch). 

in the next few weeks, i will update the BSP (kernel) so that the v4l2 kernel driver will work with upstream gst.

so , i am closing this PR at the same time..

PR #32, opened 2016-11-24
qrtr: fix installation in /lib64

koenkooi:
Signed-off-by: Koen Kooi <koen.kooi@linaro.org>


ndechesne:
thanks for the fix, I prefer to fix upstream directly, just sent this out

https://github.com/andersson/qrtr/pull/2

will upgrade the recipe once it is merged.

ndechesne:
I pushed 990dbe3f5e200de1cd7c3c6c68d4a2cb6b9795eb, it should fix this issue (in a better manner).

PR #33, opened 2017-03-02
410c: disable schedutils

plasticchris:
Without it `bitbake core-image-minimal` errors:

```error: ‘__NR_sched_setattr’ undeclared (first use in this function)```

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -19,3 +19,5 @@ BBFILES += "${@' '.join('${LAYERDIR}/%s/recipes*/*/*.bbappend' % layer \
 # Add layer-specific bb files too
 BBFILES += "${@' '.join('${LAYERDIR}/%s/recipes*/*/*.bb' % layer \
               for layer in BBFILE_COLLECTIONS.split())}"
+
+EXTRA_OECONF_append_util-linux = " --disable-schedutils" 

### CODE REVIEW

**Comment by ndechesne** [comment]
hmm. at first sight, it doesn't seem to be the right place for such a change.. is that a recent build failure? is that only when building with a board from the meta-qcom BSP?

ideally i would like to avoid such generic change. we are not supposed to make a change that impacts every machine in a BSP layer.

**Comment by plasticchris** [comment]
I'm building core-image-minimal for the dragonboard 410c qcom-oe-manifest when I get this error. I'm guessing the scheduler syscall is not supported by this chip? I thought this might be a common thing across QC chips so this is why I put it here. If there's a better place, no worries. But as is it doesn't build from a fresh checkout so I put up a patch to make people aware.

**Comment by plasticchris** [comment]
```
repo manifest -r
<?xml version="1.0" encoding="UTF-8"?>
<manifest>
  <remote fetch="https://github.com" name="github"/>
  
  <default remote="github" revision="master" sync-j="4"/>
  
  <project name="openembedded/bitbake" path="bitbake" revision="3ce6169afa646ef2b847e5fbabfe0191c93928b7" upstream="master"/>
  <project name="openembedded/meta-openembedded" path="layers/meta-openembedded" revision="131255dd561225d7e4ebbf5155420a685dfebf6e" upstream="master"/>
  <project name="openembedded/openembedded-core" path="layers/openembedded-core" revision="0116e4814aa55feb0f8944deb98e39b4534bab0c" upstream="master"/>
  <project name="plasticchris/meta-qcom" path="layers/meta-qcom" revision="f7c41cd90c852a5192c6e377309d8eb46a938c20" upstream="master">
    <linkfile dest="setup-environment" src="../../.repo/manifests/setup-environment"/>
  </project>
</manifest>
```

plasticchris:
I redirected the meta-qcom layer to add this change, but otherwise it should be the same as your repo.

PR #34, opened 2017-04-06
Fixes for master

kraj:


ndechesne:
thanks Khem. I will push the qmic patch only, manually. I am hoping it will be merged quickly upstream as well, so that we can simply upgrade the recipe.

for the alsa-lib patch, i would like to keep it, we are still using master branch for morty.. i would like to avoid diverging the branch as much as possible..

kraj:
No you are not using master. you have branched meta-qcom for morty already see

https://github.com/ndechesne/meta-qcom/tree/morty

ndechesne:
well, i know what i do ;-) the 2 branches exist to create , but they are the same. and as i said I will try to branch off as late as possible..

kraj:
hmmm, this is causing dangling bbappend warning in my world builds, so that time has come where you treat them differently.

PR #37, opened 2017-04-09
Play well with other BSP layers

kraj:


kraj:
I run world builds and this module starts building for non-fredreno machines too resulting in build failures.

ndechesne:
ok.. I merged it.. i am going to start maintaining morty on its own now ;-)

PR #38, opened 2017-04-14
dragonboard-820c: create 32-bit machine configuration

pramodgurav:
This 32-bit machine is not intended to build 32-bit linux kernel,
so it uses "linux-dummy" for virtual/kernel provider, and doesn't
include the kernel into the rootfs image.

Signed-off-by: Pramod Gurav <pramod.gurav@linaro.org>

### DIFF
--- a/conf/machine/dragonboard-820c-32.conf
+++ b/conf/machine/dragonboard-820c-32.conf
@@ -1,0 +1,21 @@
+#@TYPE: Machine
+#@NAME: dragonboard-820c-32
+#@DESCRIPTION: 32-bit Machine configuration for the DragonBoard 820c (96boards), with Qualcomm Snapdragon 820 APQ8096.
+
+require conf/machine/include/qcom-apq8096.inc
+
+MACHINE_FEATURES = "usbhost usbgadget alsa screen wifi bluetooth ext2"
+
+# Building 32-bit kernel is not supported.
+PREFERRED_PROVIDER_virtual/kernel = "linux-dummy"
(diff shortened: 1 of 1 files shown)


ndechesne:
I don't see how you are building a 32 bit user space with this machine... you only removed the build of the kernel, but you are still building 64 bit user space, as far as I can see.

pramodgurav:
Hi Nico,
Honestly I took reference from db410c-32 bit machine file.
Could you please tell me what changes are required to build a 32bit machine for db820c.

ndechesne:
have a look at "diff -u dragonboard-410c.conf dragonboard-410c-32.conf". the most important change is this one:

-require conf/machine/include/arm/arch-armv8.inc
+require conf/machine/include/tune-cortexa8.inc

pramodgurav:
Hi Nico,
So I should include conf/machine/include/arm/arch-armv8.inc for 32 bit 820c conf?
For that I will have to change conf/machine/include/qcom-apq8096.inc so that inclusion of conf/machine/include/arm/arch-armv8.inc moves to 64 bit db820 conf.

Is this right approach Nico?

ndechesne:
please check 9aff746f12528b2497f10ea898549a3959734807, and do something similar.
you need to include arch-armv8.inc in db820c conf and tune-cortexa8.inc for db820c-32

pramodgurav:
Thanks Nico,
I am making the necessary changes and creating a pull request after testing.

PR #39, opened 2017-04-17
dragonboard-820c: create 32-bit machine configuration

pramodgurav:
This 32-bit machine is not intended to build 32-bit linux kernel,
so it uses "linux-dummy" for virtual/kernel provider, and doesn't
include the kernel into the rootfs image.

Signed-off-by: Pramod Gurav <pramod.gurav@linaro.org>


ndechesne:
hi, it looks better now. I would like to give it a try before merging it, i am in vacation next 3 days. i will try to look at it soon'ish.

pramodgurav:
Thanks Nico!

lumag:
@ndechesne do we still want to support this? We still have 32-bit db410c machine config.

PR #40, opened 2017-04-17
linux-linaro-qcom: Fix build with gcc7

kraj:
Backport a patch from upstream


ndechesne:
@kraj thanks. i merged it. however I noticed it was merged in 4.9 stable. I am looking at merging stable-4.9 into our release branch soonish. so i will revert this change once this is done.

kraj:
yes its back ported into 4.9.20, you can unbolt it when you upgrade past that version.

PR #47, opened 2017-10-17
linux-linaro-qcomlt: stop using linux-dtb.inc

rsalveti:
Device tree support is now handled by the kernel class.


ndechesne:
thanks. so this is the first change that requires to fork master/rocko vs pyro branch, sadly. but I agree it's needed...

PR #49, opened 2017-11-08
Bootable SD card image - master

madrypl:
Tested with yesterday openembedded-core's master-next. Built from scratch and everything was OK.


conf/machine/dragonboard-410c.conf
```diff
@@ -21,3 +21,11 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
...
 
 QCOM_BOOTIMG_ROOTFS ?= "mmcblk0p10"
+
+# Define rootfs partiton (kernel argument)
+SD_QCOM_BOOTIMG_ROOTFS ?= "mmcblk1p7"
+
+# Assemble SD card
+IMAGE_FSTYPES_append = " wic"
+WKS_FILE = "dragonboard410c-sd.wks"
+WKS_FILE_DEPENDS = "firmware-qcom-dragonboard410c-bootloader-sdcard linux-linaro-qcomlt"
```
ndechesne: i missed that earlier. why do we have linux-linaro-qcomlt here? a $MACHINE conf file should not depends on an hard coded kernel recipe. Users of the BSP layer should be able to change the kernel.

madrypl:
This is because linux-linaro-qcom target builds boot.img which is needed to build SD card image.
I guess boot.img is not default kernel recipe's artifact and moreover is created in on_deploy task. Probably WIC is trying to build image earlier after kernel and rootfs are built, but before deploy. So, on this stage zImage(or whatever format) and rootfs are ready, but still in their work dirs.

Will look at this closer.


Dnia 8 listopada 2017 22:01:17 CET, Nicolas Dechesne <notifications@github.com> napisał(a):
>ndechesne commented on this pull request.
>
>
>
>> @@ -21,3 +21,11 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
> "
> 
> QCOM_BOOTIMG_ROOTFS ?= "mmcblk0p10"
>+
>+# Define rootfs partiton (kernel argument)
>+SD_QCOM_BOOTIMG_ROOTFS ?= "mmcblk1p7"
>+
>+# Assemble SD card
>+IMAGE_FSTYPES_append = " wic"
>+WKS_FILE = "dragonboard410c-sd.wks"
>+WKS_FILE_DEPENDS = "firmware-qcom-dragonboard410c-bootloader-sdcard
>linux-linaro-qcomlt"
>
>i missed that earlier. why do we have linux-linaro-qcomlt here? a
>$MACHINE conf file should not depends on an hard coded kernel recipe.
>Users of the BSP layer should be able to change the kernel.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/ndechesne/meta-qcom/pull/49#pullrequestreview-75246444

-- Wysłane za pomocą K-9 Mail.

madrypl:
Hi,
I'm sorry it took so long, but didn't have spare time to manage it. 
Finally, checked it once more and you're right. This dependency is not required at all. 
PR updated.

ndechesne:
Many thanks for your patience and good work. I am merging this now!
We need to monitor when the WIC patches get merged into the stable branches, so that we can apply these patches in meta-qcom release branches!

madrypl:
Thank you for support. I'll do my best to monitor stable branches and if I see all is merged then will port meta-qcom as well.

PR #50, opened 2017-11-24
dragonboard-410c.conf: use wic.gz by default

rsalveti:
Prefer the compressed version of the sdcard image by default as it is
substantially smaller.


ndechesne:
thanks.. it should have been done like that initially ;-)

PR #53, opened 2017-12-06
dragonboard-410c.conf: add wic.bmap image type

zumbi:
It would allow us to avoid uncompressing wic image every time
we update SD card with new image using bmap-tools.

Signed-off-by: Héctor Orón Martínez <hector.oron@collabora.co.uk>


ndechesne:
thanks for the PR. do you think keep wic.gz and wic.bmap makes sense or should we just keep wic.bmap? i am not using wic much, so i don't know what most users 'really care about'.

zumbi:
Hello Nicolas,
In short, keep both *wic.gz and *wic.bmap, the latter is complementary to the former.
bmap file is XML file which contain mapped areas and some extra information, it is very light weight respect size. If you want to know more I would recommend you to read https://source.tizen.org/documentation/reference/bmaptool/introduction
Thanks for considering

PR #56, opened 2018-01-15
recipes-kernel/linux: Add linux-linaro-qcomlt-integration recipe

alimon:
Add the recipe in order to be able build integration kernel for testing,
new variables was add to set the git branch and revision of the build.


recipes-kernel/linux/linux-linaro-qcomlt-integration_git.bb
```diff
@@ -0,0 +1,21 @@
...
+DESCRIPTION = "Linaro Qualcomm Landing team Integration Kernel"
+
+inherit pythonnative
+
+require recipes-kernel/linux/linux-linaro-qcom.inc
+require recipes-kernel/linux/linux-qcom-bootimg.inc
+
+LOCALVERSION ?= "-linaro-lt-qcom-integration"
+
+LINUX_LINARO_QCOM_SRCBRANCH ?= "integration/auto/v4.15-2018-01-15-20-34-03"
```
ndechesne: this is going to change every day/week. i don't think we should set that here. we should either use AUTOREV but in that case we should make sure the recipe is disabled by default , so that we don't do network git/fetch by default

alimon: Yes, my idea is to have a job to set conf/auto.conf with:

PREFERRED_PROVIDER_virtual/kernel="linux-linaro-qcomlt-integration"

We can use ${AUTOREV} for LINUX_LINARO_QCOM_SRCREV but we need to specify certain branch, i suggest creating/setting SRCBRANCH "integration/master" and use ${AUTOREV} into SRCREV.

PR #59, opened 2018-02-07
firmware-qcom: use nonarch_base_libdir instead of /lib

rsalveti:
Hardcoded /lib is not compatible with usrmerge.

### DIFF
--- a/recipes-bsp/firmware/firmware-qcom-dragonboard410c_1032.1.bb
+++ b/recipes-bsp/firmware/firmware-qcom-dragonboard410c_1032.1.bb
@@ -19,23 +19,23 @@ do_compile() {
 }
 
 do_install() {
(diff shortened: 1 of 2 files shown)


ndechesne:
thanks. i also cherry-picked on rocko branch, btw. 

i noticed that you kept the sd600eval firmware as they are. i think it's fine. i have more changes planned for 600 coming soon in this area.. so they can stay as they are for now.

rsalveti:
> i noticed that you kept the sd600eval firmware as they are. i think it's fine. i have more changes planned for 600 coming soon in this area.. so they can stay as they are for now.

Cool, indeed, had the impression it wasn't required but I guess I was looking for dragonboard targets. Will propose a similar change there so we can them all in sync.

PR #68, opened 2018-04-16
layer.conf: add LAYERSERIES_COMPAT markup

rsalveti:
Allows the user to easily identify if the layer is compatible with
oe-core.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -8,6 +8,7 @@ BBFILES += "${LAYERDIR}/recipes-*/*/*.bb \
 BBFILE_COLLECTIONS += "qcom"
 BBFILE_PATTERN_qcom := "^${LAYERDIR}/"
 BBFILE_PRIORITY_qcom = "5"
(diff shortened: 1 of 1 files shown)


ndechesne:
i haven't followed up closely on this thing upstream... why do we mark 'sumo' in the master branch, is that how it should be done?

rsalveti:
> i haven't followed up closely on this thing upstream... why do we mark 'sumo' in the master branch, is that how it should be done?

Yeah, all layers tracking master (development) are now compatible with sumo (and a similar change would probably be required/good to have for the rocko branch).

The upstream thread around the subject: http://lists.openembedded.org/pipermail/openembedded-core/2018-April/149705.html

rsalveti:
The warning you get when using meta-qcom with latest oe-core master:

> WARNING: Layer qcom should set LAYERSERIES_COMPAT_qcom in its conf/layer.conf file to list the core layer names it is compatible with.

PR #69, opened 2018-05-14
Enable lab-bjorn to run straight off qcom/rocko

andersson:
This bumps the bootrr recipe to use the latest commit from the bootrr project, fixes kernel module auto loading and makes it possible to automatically run the bootrr board scripts in a LAVAless environment.

### DIFF
--- a/recipes-test/bootrr/bootrr.bb
+++ b/recipes-test/bootrr/bootrr.bb
@@ -4,7 +4,7 @@ SECTION = "test"
 LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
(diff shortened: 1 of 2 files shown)


recipes-test/bootrr/bootrr.bb
```diff
@@ -4,7 +4,7 @@ SECTION = "test"
 LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
-SRCREV = "365a5ee5fec5f6c28bf05764d0b9e16fe2b8e560"
+SRCREV = "${AUTOREV}"
```
ndechesne: this change is problematic. it is very much not recommended to use AUTOREV in 'real' layers... while convenient, it breaks a few things that any layer are expected not to break:
* reproducible builds
* build without network

AUTOREV forces bitbake to access network during the recipes parsing stage...

how often do you expect bootrr to be updated?

PR #76, opened 2018-08-31
Add diag tool into testing ramdisk image

alimon:


recipes-test/diag/diag_git.bb
```diff
@@ -0,0 +1,23 @@
...
+
+do_compile () {
+	oe_runmake
+}
+
+do_install () {
+	oe_runmake install 'DESTDIR=${D}'
+}
+
+FILES_${PN} = "/usr/bin/diag-router /usr/bin/send_data"
```
ndechesne: this is not needed.

alimon: I removed the FILES_PN set.

PR #77, opened 2018-09-17
recipes-test: Remove bootrr{,-init} from initramfs-bootrr-image

alimon:
The bootrr and init script is now handled inside the CI
integration/mainline [1] jobs so don't need to have the into meta-qcom
layer.

The CI integration job now gets the last bootrr version and set the
initscript needed for boot the board [2][3].

[1] https://ci.linaro.org/job/lt-qcom-linux-integration/
[2] https://git.linaro.org/ci/job/configs.git/commit/?id=c7627a865
[3] https://git.linaro.org/ci/job/configs.git/commit/?id=132f0e8d0


alimon:
I plan to create a non machine specific test ramdisk and rename the recipe to don't suggest directly  bootrr because will be used for other purposes like functional testing.

PR #83, opened 2018-11-08
Update skales SRC_URI to fetch via HTTP

evadeflow:
As written, the SRC_URI for the skales recipe causes the fetch step to fail for developers operating behind corporate HTTP proxies.  This commit should make the recipe Just Work t.m. for a larger number of users.

### DIFF
--- a/recipes-devtools/skales/skales_git.bb
+++ b/recipes-devtools/skales/skales_git.bb
@@ -11,7 +11,7 @@ DEPENDS = "python dtc"
 SRCREV = "6eac9e943de53c4aaaede3697e9226a47686fe25"
 PV = "1.5.0+git${SRCPV}"
 
-SRC_URI = "git://codeaurora.org/quic/kernel/skales"
+SRC_URI = "git://source.codeaurora.org/quic/cc-qrdk/oss/tools/skales;protocol=http;branch=skales/master"
 
 S = "${WORKDIR}/git"
 


evadeflow:
Sorry, I should explain the `branch=skales/master` bit. Without it, I get an error saying:

```
WARNING: skales-1.5.0+gitAUTOINC+6eac9e943d-r0 do_fetch: Failed to fetch URL git://source.codeaurora.org/quic/cc-qrdk/oss/tools/skales;protocol=http, attempting MIRRORS if available
ERROR: skales-1.5.0+gitAUTOINC+6eac9e943d-r0 do_fetch: Fetcher failure: Unable to find revision 6eac9e943de53c4aaaede3697e9226a47686fe25 in branch master even from upstream
```

I assume this is because—when using HTTP protocol—revisions are resolved on a per-branch basis, but I can't be certain. (I only know it doesn't work for me, with HTTP, unless I specify the branch...)

recipes-devtools/skales/skales_git.bb
```diff
@@ -11,7 +11,7 @@ DEPENDS = "python dtc"
 SRCREV = "6eac9e943de53c4aaaede3697e9226a47686fe25"
 PV = "1.5.0+git${SRCPV}"
 
-SRC_URI = "git://codeaurora.org/quic/kernel/skales"
+SRC_URI = "git://source.codeaurora.org/quic/cc-qrdk/oss/tools/skales;protocol=http;branch=skales/master"
```
ndechesne: you are also changing the source tree here, why? quick/kernel/skales was the real 'upstream' even though it unlikely to be maintained anymore.. but quic/cc-qrdk/tools/skales is something I wasn't aware of...

ndechesne:
fixed in 623ca9feb60fc1ec04a64ce802f1486fb64b0eea

PR #85, opened 2018-12-20
layer.conf: Don't take distro policies in hand

kraj:
Distro's should decide via bblayers.conf on adding the
layers, BSPs layers should provide for the machine policies


ndechesne:
I don't understand here. I am not adding new layers on behalf of distro, I am only tweaking BBFILES to selectively include some bbappends *based* on the list of layers that the user decided to pull into the build. 

and i am doing that exactly to provide machine policies, e.g. for gpsd bbappend file. if meta-oe is not included in the build, I cannot provide gpsb bbappend. but if meta-oe is used, then I need gpsd bbappend so that gpsd works on machines defined in meta-qcom.

I would have agreed if you complained that gpsd bbappend modifies SRC_URI regardless of machine, looking at that, i think this is a bug here.. but not what you are reporting.

kraj:
well I was looking at what it was doing to my BBFILES, anyway using dynamic layers is better option. I have fixed the pull for that. 

generic modification to gpsd did not catch my eye before but you are right. this is not a good thing to do. So may be you should either propose it to meta-oe or make this apply using SOC_FAMILY overrides for machines where it matters

ndechesne:
d'oh... BBFILES_DYNAMIC is really nice! I didn't know about it ;) 

can you please squash both commits? Also why did you create dymamic-layers folder? it's not really needed, or is it a convention that others use generally?

ndechesne:
as for the GPSd patch, we are planning to get it merged upstream eventually, it's not the right thing to merge it in meta-oe for now. But I will fix it to use SOC_FAMILY.

ndechesne:
@kraj , looking at how I manage SOC_FAMILY.. do you think it's a good idea to create a "qcom:" SOC_FAMILY? or would that be too generic?

kraj:
> d'oh... BBFILES_DYNAMIC is really nice! I didn't know about it ;)
> 
> can you please squash both commits? Also why did you create dymamic-layers folder? it's not really needed, or is it a convention that others use generally?

squashed and pushed. Creating dynamic-layers folder is a convention not a need. It does keep all
external tweaks in one place so kind of nice.

kraj:
> @kraj , looking at how I manage SOC_FAMILY.. do you think it's a good idea to create a "qcom:" SOC_FAMILY? or would that be too generic?

we should not create it if this is then used to customize recipes for meta-qcom, but it can be created
if all qcom SOCs share something e.g. same graphics driver or things like that. gpsd example does fit into that category. but it can be abused so my suggestion would be to use other knobs first if possible

ndechesne:
thanks! sorry about the delay.

PR #90, opened 2019-02-02
wcnss-config: Do not use tag in SRCREV

kraj:
Using tags in SRCREV would force bitbake to poke at remote repo all the
time since tags are floating pointers, therefore encode the
corresponding SHA to avoid build failures when NO_NETWORK is set


ndechesne:
thanks! i pushed to thud as well, just in case.

PR #91, opened 2019-02-08
Update default tune configuration based on latest OE-Core

ricardosalveti:
Latest OE-Core removed the arch-armv8 tune file and added a new set of tune files cortex A cores.

@kraj another heads up.


kraj:
This is good change.

ndechesne:
@ricardosalveti what testing have you done? on both 410c and 820c?

ricardosalveti:
@ndechesne yes, both db410c and db820c.

PR #92, opened 2019-02-12
Improvements initramfs-bootrr-image, systemd, autologin and rename

alimon:


alimon:
@ndechesne I added only systemd because adding packagegroup-core-boot it will end adding kernel and modules to the image and increase the size from 14MB -> 42MB.

ndechesne:
adding systemd in an image is still wrong.. since it's really the distro that gets to decide the init system. 

i think something like this should be better:

--- a/recipes-test/images/initramfs-bootrr-image.bb
+++ b/recipes-test/images/initramfs-bootrr-image.bb
@@ -2,7 +2,7 @@ DESCRIPTION = "Small ramdisk image for running bootrr"

 PACKAGE_INSTALL = " \
     ${ROOTFS_BOOTSTRAP_INSTALL} \
-    busybox \
+    packagegroup-core-boot \
     base-passwd \
     devmem2 \
     diag \
@@ -15,10 +15,11 @@ PACKAGE_INSTALL = " \
     rmtfs \
     tcpdump \
     udev \
 "

 # Do not pollute the initrd image with rootfs features
-IMAGE_FEATURES = ""
+IMAGE_FEATURES = "debug-tweaks"

 export IMAGE_BASENAME = "initramfs-bootrr-image"
 IMAGE_LINGUAS = ""
@@ -31,4 +32,4 @@ inherit core-image
 IMAGE_ROOTFS_SIZE = "8192"
 IMAGE_ROOTFS_EXTRA_SPACE = "0"

-BAD_RECOMMENDATIONS += "busybox-syslog"
+NO_RECOMMENDATIONS = "1"

The modules come as RRECOMMENDS, so by disable RECOMMENDATIONS, we can build the image with systemd (or sysvinit based on DISTRO) without getting a large image with module.

recipes-test/images/initramfs-bootrr-image.bb
```diff
@@ -33,3 +33,14 @@ IMAGE_ROOTFS_SIZE = "8192"
...
+# Enable local auto-login of the root user (local = serial port and
+# virtual console by default, can be configured).
+LOCAL_GETTY ?= " \
+    ${IMAGE_ROOTFS}${systemd_system_unitdir}/serial-getty@.service \
+    ${IMAGE_ROOTFS}${systemd_system_unitdir}/getty@.service \
+"
+local_autologin () {
+    sed -i -e 's/^\(ExecStart *=.*getty \)/\1--autologin root /' ${LOCAL_GETTY}
+}
+ROOTFS_POSTPROCESS_COMMAND += "${@bb.utils.contains('IMAGE_FEATURES', 'debug-tweaks', 'local_autologin;', '', d)}"
```
ndechesne: since we are adding debug-tweaks in IMAGE_FEATURES we should simply add local_autologin to ROOTFS_POSTPROCESS_COMMAND

recipes-test/images/initramfs-bootrr-image.bb
```diff
@@ -33,3 +33,14 @@ IMAGE_ROOTFS_SIZE = "8192"
 IMAGE_ROOTFS_EXTRA_SPACE = "0"
 
 BAD_RECOMMENDATIONS += "busybox-syslog"
+
+# Enable local auto-login of the root user (local = serial port and
+# virtual console by default, can be configured).
+LOCAL_GETTY ?= " \
```
ndechesne: nice, so that works fine and it doesn't depend on SERIAL_CONSOLE at all?

recipes-test/images/initramfs-bootrr-image.bb
ndechesne: well, actually, we should to local_autologin only if we use systemd, that would be a better check. i still think that we should be able to build with oenodistro (instead of rpb) and get a functional image with sysvinit for example.

alimon: Ok, I will test for systemd init manager.

recipes-test/images/initramfs-bootrr-image.bb
alimon: No, it  enables autologin in all consoles.

recipes-test/images/initramfs-bootrr-image.bb
alimon: I made your suggested changes and the size remains the same after set NO_RECOMMENDATIONS.

PR #93, opened 2019-02-19
recipes-kernel: Add linux-firmware_git.bbappend for dragonboard820c

alimon:
Downgrade to stable version of ath10k firmware for dragonboard820c,
RM.4.4.1.c2-00057-QCARMSWP-1 version is causing a HW reboot [1]
use RM.4.4.1-00079-QCARMSWPZ-1 instead.

A new version that fixed the problem is available [2] but not landed
linux-firmware yet.

[1] https://validation.linaro.org/scheduler/job/1906358#L2444
[2] https://github.com/kvalo/ath10k-firmware/blob/master/QCA6174/hw3.0/4.4.1/firmware-6.bin_WLAN.RM.4.4.1-00132-QCARMSWP-1


alimon:
@ndechesne I know is hacky but any other option?, may be add a full recipe.

PR #97, opened 2019-03-10
[RFC] conf: Bundle DTB with kernel (QCOM_BOOTIMG_BUNDLE_DT)

mrchapp:
By setting QCOM_BOOTIMG_BUNDLE_DT to 1, the boot image will
append the device tree file to the kernel. Otherwise, booting
fails in LAVA with the following:
  Parsed command exited 1.
  action: boot-fastboot
  command: ['nice', 'lxc-attach', '-n', 'lxc-target-623344', '--', 'fastboot', '-s', '26e32e60', 'boot', '/lava-lxc/boot--4.20+git0+0f7c162c1d-r0-dragonboard-410c-20190227221652-91-91.img']
  message: Command '['nice', 'lxc-attach', '-n', 'lxc-target-623344', '--', 'fastboot', '-s', '26e32e60', 'boot', '/lava-lxc/boot--4.20+git0+0f7c162c1d-r0-dragonboard-410c-20190227221652-91-91.img']' returned non-zero exit status 1
  output: < waiting for 26e32e60 >
  downloading 'boot.img'...
  OKAY [  0.593s]
  booting...
  FAILED (remote: dtb not found)
  finished. total time: 1.185s
  return code: 1
[...]
  [870] [870] fastboot: processing commands
  [1040] [1040] fastboot: getvar:slot-count
  [1040] [1040] fastboot: getvar:slot-suffixes
  [1050] [1050] fastboot: download:00a07000
  [1650] [1650] fastboot: boot
  [2230] [2230] DTB offset is incorrect, kernel image does not have appended DTB

The Rocko branch does set it to 1, but the definition is gone
from Sumo, Thud and master.


mrchapp:
Probably related to how we're building in LKFT (Skales?).

I'm requesting comments for this as I'd like to know if this gets in the way of how things how are done for QCOM, and thus if we should add QCOM_BOOTIMG_BUNDLE_DT somewhere in our kernels or distro configurations.

If approved, this should be propagated to Thud and Sumo.

ndechesne:
this is more or less a revert of b97de48b1f58e339b2d40098d3369acf9b94b30f. we do not support skales and/or DT image any longer. Setting BUNDLE_DT does not append the DTB to the kernel image. it creates a special "DT" image which includes DTBs and puts that DT image inside the boot.img,

when BUNDLE_DT is not set, we simply append the DTB to the kernel image, and the bootloader will find it and pass it to the kernel. 

if it doesn't work for you, it probably means you are not using the latest bootloader , we updated the bootloader in Septembre 2017 to support this use case.

Can you check the bootloader version? it should be printed in the first ~200 lines on the UART

mrchapp:
Thanks for the prompt feedback.

Is this it?:
```
S - QC_IMAGE_VERSION_STRING=BOOT.BF.3.0-00286
S - IMAGE_VARIANT_STRING=HAAAANAZA
S - OEM_IMAGE_VERSION_STRING=CRM
S - Boot Config, 0x000002e1
S - Core 0 Frequency, 0 MHz
```
Here's a LAVA job that failed: https://lkft.validation.linaro.org/scheduler/job/623344

Not sure where (or what recipe) to look for -- I diffed on meta-{rpb,96boards,linaro} between Rocko and Sumo and could not find anything that looked relevant.

Two things worth nothing, though, is we include "recipes-kernel/linux/linux-qcom-bootimg.inc" in the LKFT kernels [1], and we do have a consideration for Skales [2], so it is possible that we are missing a change of a different nature elsewhere.

[1] https://github.com/Linaro/meta-lkft/blob/rocko/recipes-kernel/linux/machine-specific-hooks.inc
[2] https://github.com/Linaro/meta-lkft/blob/rocko/recipes-kernel/linux/linux-generic-stable-rc_5.0.bb#L111

ndechesne:
ah.. so you use MACHINE from meta-qcom, but with a different kernel, i see. 

i think you are missing  
do_compile_append() from linux-linaro-qcom.inc

this is where we append the DTB to the kernel image.

i checked your bootloader and it's good version

mrchapp:
That worked. We're now setting BUNDLE_DT to 0 and appending the DTB, even in Rocko: https://github.com/Linaro/meta-lkft/pull/25

PR #101, opened 2019-05-15
mesa: Add backport of freedreno dmabuf modifiers patch

petegriffin:
This is required to get weston-simple-dmabuf-drm to work
and also now required by Chromium v73 ozone-wayland backend.

It is present in mesa 19.0, backporting to Thud which is
on 18.1.9. This is intended as a tempoary fix as this patch
has also been sent to oe-core ml
http://lists.openembedded.org/pipermail/openembedded-core/2019-May/282314.html

Signed-off-by: Peter Griffin <peter.griffin@linaro.org>

### DIFF
--- a/recipes-graphics/mesa/mesa_%.bbappend
+++ b/recipes-graphics/mesa/mesa_%.bbappend
@@ -3,6 +3,12 @@ GALLIUMDRIVERS_append_apq8064 = ",freedreno"
 GALLIUMDRIVERS_append_apq8016 = ",freedreno"
 GALLIUMDRIVERS_append_apq8096 = ",freedreno"
 
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI_append_apq8064 = " file://0001-freedreno-add-query-for-dmabuf-modifiers.patch"
+SRC_URI_append_apq8016 = " file://0001-freedreno-add-query-for-dmabuf-modifiers.patch"
+SRC_URI_append_apq8096 = " file://0001-freedreno-add-query-for-dmabuf-modifiers.patch"
+
 PACKAGECONFIG_append_apq8064 = " \
     gallium \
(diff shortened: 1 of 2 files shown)


ndechesne:
the patch file is missing!

petegriffin:
@ndechesne oh dear..not my day. I just updated and pushed again

recipes-graphics/mesa/mesa_%.bbappend
```diff
@@ -3,6 +3,12 @@ GALLIUMDRIVERS_append_apq8064 = ",freedreno"
 GALLIUMDRIVERS_append_apq8016 = ",freedreno"
 GALLIUMDRIVERS_append_apq8096 = ",freedreno"
 
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI += " \
```
ndechesne: sorry to notice that just now.. but that would break the yocto compatible 'rules'. e.g. adding meta-qcom in bblayers.conf *without* selecting any of its machine shouldn't impact the build. 

here adding meta-qcom and building for x86 would impact mesa, since we unconditionally modify SRC_URI and patch it. so we would need this instead:

SRC_URI_append_apq8064 = "  file://0001-freedreno-add-query-for-dmabuf-modifiers.patch"
SRC_URI_append_apq8016 = "  file://0001-freedreno-add-query-for-dmabuf-modifiers.patch"
SRC_URI_append_apq8096 = "  file://0001-freedreno-add-query-for-dmabuf-modifiers.patch"

It is a but ugly, i know.. and i want to introduce a more generic 'qcom' or 'snapdragon' flag eventually.. but that's the only solution for now.

petegriffin: @ndechesne no problem, is good to know what the OE best practice is.  I just updated to what you suggested.

ndechesne: thanks!

PR #108, opened 2019-07-12
Linux kernel upgrade and mesa/xorg drivers for DB845c

alimon:


recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -10,6 +10,8 @@ BOOT_IMAGE_BASE_NAME = "boot-${KERNEL_IMAGE_NAME}"
 BOOT_IMAGE_SYMLINK_NAME = "boot-${KERNEL_IMAGE_LINK_NAME}"
 SD_BOOT_IMAGE_BASE_NAME = "boot-sd${KERNEL_IMAGE_NAME}" 
 SD_BOOT_IMAGE_SYMLINK_NAME = "boot-sd-${KERNEL_IMAGE_LINK_NAME}"
+KERNEL_CMDLINE_dragonboard-845c = "root=/dev/${1} rw rootwait console=${ttydev},${baudrate}n8 clk_ignore_unused pd_ignore_unused"
```
ndechesne: KERNEL_CMDLINE_append_dragonboard-845 instead?

recipes-graphics/mesa/mesa_%.bbappend
```diff
@@ -17,3 +18,8 @@ PACKAGECONFIG_append_apq8096 = " \
     gallium \
     ${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'xa', '', d)} \
 "
+
```
ndechesne: we need to update xserver-xorg_%.bbappend as well.

recipes-kernel/linux/linux-qcom-bootimg.inc
alimon: Right, better to use append, changing...

recipes-graphics/mesa/mesa_%.bbappend
alimon: Ok

ndechesne:
pushed to master as well.

PR #109, opened 2019-07-20
db845c update kernel and qrtr/rmtfs upgrades

alimon:


recipes-support/rmtfs/rmtfs_git.bb
```diff
@@ -18,6 +18,11 @@ S = "${WORKDIR}/git"
 
 EXTRA_OEMAKE = "'LDFLAGS=${TARGET_LDFLAGS} -L${STAGING_LIBDIR} -lqrtr -ludev'"
 
+do_install_prepend_dragonboard-845c () {
```
ndechesne: why is that needed? what happens if we don't have this patch? 
we added that in b03212707cec5c49bde4e1d6afb58513e1731318, and I would like to understand how to deal with 8016 and 845. If we need different files for each platform, we probably should use apq8016/rmtfs.service instead of patching them in the recipe.

alimon: The kernel config that I compiled the module as part of the kernel, I can change it to compile as a module and yes will be different _pas intead of _pil.

ndechesne: everything that can be a module should be a module in general. so if you can make it a module then it's better. 
looks like there is a newer version of rmtfs that we should use that would automatically start the modem. however i don't think we've started the modem yet, only adsp for audio.

alimon: @ndechesne Ok changed to use modules and don't enable rmtfs yet.

recipes-multimedia/alsa/alsa-lib_1.1.%.bbappend
```diff
@@ -0,0 +1,7 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${BPN}:"
+
+SRC_URI += " \
```
ndechesne: we should not modify SRC_URI without an override. and that change makes alsa-lib package MACHINE_ARCH.

alimon: @ndechesne Just update the patch to only apply when is sdm845 or apq8096.

ndechesne:
pushed to master as well

PR #113, opened 2019-08-29
qdl: needs pkgconfig and udev

kraj:
Fixes build issues e.g.
| qdl.c:43:10: fatal error: 'libudev.h' file not found
| #include <libudev.h>
|          ^~~~~~~~~~~

/bin/sh: pkg-config: command not found
patch.c:33:10: fatal error: libxml/parser.h: No such file or directory
   33 | #include <libxml/parser.h>
      |          ^~~~~~~~~~~~~~~~~

### DIFF
--- a/recipes-devtools/qdl/qdl_git.bb
+++ b/recipes-devtools/qdl/qdl_git.bb
@@ -5,7 +5,9 @@ SECTION = "devel"
 LICENSE = "BSD-3-Clause"
 LIC_FILES_CHKSUM = "file://qdl.c;beginline=1;endline=31;md5=1c7d712d897368d3d3c161e5493efc6a"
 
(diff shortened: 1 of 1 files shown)


ndechesne:
@kraj , it broke qdl-native which now depends on udev-native. i will add a class-target to fix it. Do you know why we don't provide udev-native? is that a specific limitation?

kraj:
> @kraj , it broke qdl-native which now depends on udev-native. i will add a class-target to fix it. Do you know why we don't provide udev-native? is that a specific limitation?

hmm, yeah udev is specific to target interpretation so wont make much sense for native but I guess for now we can limit the include to `class-target` specific and you might also want to check the nativesdk version if you support that

PR #114, opened 2019-08-30
qdl: Fix QA Issue: No GNU_HASH in the ELF binary

kraj:
This is because of LDFLAGS and CFLAGS being overwritten in Makefile as a
result we missing hash-style option

see
https://errors.yoctoproject.org/Errors/Details/267081/

### DIFF
--- a/recipes-devtools/qdl/files/0001-Makefile-Use-pkg-config-for-libxml2-detection.patch
+++ b/recipes-devtools/qdl/files/0001-Makefile-Use-pkg-config-for-libxml2-detection.patch
@@ -12,20 +12,24 @@ Signed-off-by: Nicolas Dechesne <nicolas.dechesne@linaro.org>
  Makefile | 4 ++--
  1 file changed, 2 insertions(+), 2 deletions(-)
 
-diff --git a/Makefile b/Makefile
-index 0a934ff..f18393d 100644
 --- a/Makefile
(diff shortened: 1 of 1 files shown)


recipes-devtools/qdl/files/0001-Makefile-Use-pkg-config-for-libxml2-detection.patch
```diff
@@ -12,20 +12,24 @@ Signed-off-by: Nicolas Dechesne <nicolas.dechesne@linaro.org>
...
  
  SRCS := firehose.c qdl.c sahara.c util.c patch.c program.c ufs.c
--- 
-2.23.0
-
+ OBJS := $(SRCS:.c=.o)
+ 
+ $(OUT): $(OBJS)
+-	$(CC) -o $@ $^ $(LDFLAGS)
++	$(CC) $(CFLAGS) -o $@ $^ $(LDFLAGS)
```
ndechesne: why do you need this one?

kraj: strictly speaking, this part is not essential

ndechesne: ok, i can merge as it is. thanks!

PR #120, opened 2019-10-28
DB845c bluetooth

alimon:


alimon:
Cherrypick'ed  in warrior.

PR #123, opened 2019-11-19
DB845c warrior WiFi and Audio

alimon:


conf/machine/dragonboard-845c.conf
```diff
@@ -16,6 +16,8 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     kernel-modules \
     ${@bb.utils.contains('DISTRO_FEATURES', 'bluetooth', 'linux-firmware-qca', '', d)} \
     ${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'mesa-driver-msm', '', d)} \
+    pd-mapper \
```
ndechesne: why not in MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS in sdm845.inc like we do for qrtr and rmtfs. any reason why it should be handled differently?

recipes-bsp/firmware/firmware-qcom-dragonboard845c_20190529180356-v4.bb
```diff
@@ -3,9 +3,13 @@ DESCRIPTION = "QCOM Firmware for DragonBoard 845c"
 LICENSE = "Proprietary"
 LIC_FILES_CHKSUM = "file://LICENSE.qcom.txt;md5=cbbe399f2c983ad51768f4561587f000"
 
-SRC_URI = "https://releases.linaro.org/96boards/dragonboard845c/qualcomm/firmware/RB3_firmware_${PV}.zip"
+SRC_URI = " \
+	https://releases.linaro.org/96boards/dragonboard845c/qualcomm/firmware/RB3_firmware_${PV}.zip \
+	git://github.com/andersson/linux-firmware.git;branch=db845c-wifi;name=andersson;destsuffix=git \
```
ndechesne: i believe the files in Bjorn's PR are coming from the RB3_firmware package. they might have been post processed, but they are from the same package. we should use only the RB3_firmware package.

ndechesne:
I suppose that it supersedes #122 , if so, please close it.

conf/machine/dragonboard-845c.conf
alimon: Ok, I will update it.

recipes-bsp/firmware/firmware-qcom-dragonboard845c_20190529180356-v4.bb
alimon: Most of the files comes from RB3_firmware package, only boad-2.bin was built by Bjorn based on RB3_firmware package.

alimon: @andersson Any other details that I miss?

PR #127, opened 2019-11-30
tqftpserv: Fix build on musl

kraj:
Proposed upstream see

https://github.com/andersson/tqftpserv/pull/4

but I wonder if such recipes should be in meta-oe and not stashed into  BSP layers

### DIFF
--- a/recipes-support/tqftpserv/tqftpserv/0001-include-limits.h-for-PATH_MAX.patch
+++ b/recipes-support/tqftpserv/tqftpserv/0001-include-limits.h-for-PATH_MAX.patch
@@ -1,0 +1,25 @@
+From c21250e08cc5e6815340244693437e6372b5eca4 Mon Sep 17 00:00:00 2001
+From: Khem Raj <raj.khem@gmail.com>
+Date: Fri, 29 Nov 2019 17:52:33 -0800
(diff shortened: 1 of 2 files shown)


ndechesne:
@kraj, these recipes only make sense on qcom chipset (not even all of them), that's why there are here. they are brought with MACHINE_ESSENTIAL_EXTRA when needed. isn't that the right approach?

kraj:
> @kraj, these recipes only make sense on qcom chipset (not even all of them), that's why there are here. they are brought with MACHINE_ESSENTIAL_EXTRA when needed. isn't that the right approach?

the fact that it was building for `MACHINE = "qemuarm"` seems it is really not machine-specific
so maybe you should limit it via COMPATIBLE_HOST

ndechesne:
it was building for a world build? that's fine to me.. i was saying that it's added in the image by default only if building for a qcom platform which requires it.

kraj:
> it was building for a world build? that's fine to me.. i was saying that it's added in the image by default only if building for a qcom platform which requires it.

adding to an image is optional, agreed, however the recipe is not marked as such to be QCA specific
so if they are using something thats SOC specific and are not functional on non QCA SOCs
then its best in the interest of users to mark them SOC specific.

PR #128, opened 2019-12-02
conf/machine: Update rootfs partition in dragonboard-410c

alimon:
New bootloader comes with new partitions to store rmtfs firmware [1].

[1] https://git.linaro.org/landing-teams/working/qualcomm/db-boot-tools.git/commit/?id=04008fd6570b566041e6ded5df738068416607e3


alimon:
Support for use PARTLABEL in root cmdline is pending, we will need to use a ramdisk in OE images.

Targets thud, warrior and zeus.

PR #135, opened 2020-01-29
Qcom master python3 fixes

petegriffin:
Couple of fixes for the recent python2 removal as it is EOL


ndechesne:
thanks for the PR.. now that I am forced to look into that, I think i am going to do things slightly differently.. I will reuse your last patch, but I will remove dependency on pythonnative. i think it's not needed, and for what we are doing, we can use the host python instead.. i will test that and merge if that works.

petegriffin:
btw there is also another issue with meta-qcom

ERROR: No recipes available for:
  /srv/builds/peter.griffin/rpb-master/build-rpb-wayland/conf/../../layers/meta-qcom/recipes-multimedia/alsa/alsa-lib_1.1.%.bbappend

I updated the bbappend to match latest alsa-lib package name, but the patches don't apply any longer.

ndechesne:
hi Pete, I am closing this PR, since I just pushed an update that includes one of your patch, and a few others. I also fixed the recipe for alsa UCM files, though it's not perfect yet.. at least it will build for now.

PR #137, opened 2020-02-05
Removed deprecated configs

alimon:


ndechesne:
thanks, pushed to zeus and warrior, as well.

PR #141, opened 2020-02-19
FastRPC recipe

alimon:


conf/machine/include/qcom-apq8016.inc
```diff
@@ -23,6 +23,7 @@ PREFERRED_PROVIDER_virtual/xserver ?= "xserver-xorg"
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-linaro-qcomlt"
 
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
+    fastrpc \
```
ndechesne: fastrpc is not an 'essential' package, in fast adsprpc wasn't either.
qrtr, rmtfs, .. are essential since they are needed to boot all cores on the chipset, but fastrpc should be installed only in images when it's needed.

PR #142, opened 2020-03-11
Add fastrpc

alimon:


ndechesne:
I am not sure I want to merge the last 2 patches. I don't want to make assumptions on meta-rpb here.. it's just a BSP layer. can you remove them for now, so that we can at least merge the first 2?

alimon:
@ndechesne Just removed the dynamic-layers related patches.

PR #147, opened 2020-03-26
Support qcom-armv8a MACHINE multiple Android boot images with consolidated userspace

alimon:
The new MACHINE holds configuration to install QCOM userspace tools and firmware for Dragonboards.

The different settings need by board (DTB) like roofs, base address and page size are handled on linaro-linux-qcomlt include files and the changes are did in a manner that are backwards compatible.

There is a lack of support of uboot (db{410,820}c and WIC images (db410c) because the current code base only handles one build configuration, further investigation is needed.


lumag:
First patches look good,
> Acked-by: Dmitry-Baryshkov <dmitry.baryshkov@linaro.org>

I've concerns about the last patch, it looks a bit messy, having all those conditions. Let's check, if we can improve it.
- I'll take a look during the onto making multi-dtb code simpler (I started looking onto that for RB5, but abandoned because it was decided to be unnecessary).
- Then we can move all per-dtb settings to qcom-armv8a.conf by turning them into flags. This way the iteration code will become free from switch case.

PR #150, opened 2020-05-13
recipes-connectivity: Add networkmanager-qcom to set DB845c eth mac

alimon:
The dragonboard 845c dosen't come with unique MAC address on ethernet so
use Network Manager feature to set it preserving across connections.


alimon:
Targets zeus and dunfell too.

ndechesne:
old and abandoned.

PR #153, opened 2020-05-25
gpsd: Rework PDS service patch changes in gpsd_log

alimon:
Due to changes on argument order of gpsd_log and use GPSD_LOG
too see rev 81257aeeaf.


alimon:
@ndechesne This patch is only for master in dunfell gpsd is 3.19 and is not affected.

ndechesne:
thanks, nice debug!

PR #154, opened 2020-05-28
pd-mapper: Fix build with musl

kraj:


kraj:
@ndechesne it will be good to do a musl build for meta-qcom for accepting patches so we can catch trivial issues like this

ndechesne:
thanks @kraj for the advice.. I have now integrated a CI job for meta-qcom on the Linaro CI.. It was long overdue, that was the bug that pushed me to do it.. 
https://ci.linaro.org/job/lt-qcom-openembedded-meta-qcom-master-premerge/
I am building for a couple of machines from meta-qcom , with DISTRO=poky, and building with glibc and musl.. hopefully we should catch problems before you do now.

PR #162, opened 2020-06-17
Sm8250 support

lumag:


conf/machine/sm8250-mtp.conf
```diff
@@ -0,0 +1,26 @@
...
+
+MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
+    kernel-modules \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'wifi', 'linux-firmware-ath10k', '', d)} \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'bluetooth', 'linux-firmware-qca', '', d)} \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'mesa-driver-msm', '', d)} \
+"
+# linux-firmware-qcom-adreno-a650 
+
+# /dev/sda1 is 'rootfs' partition after installing the latest bootloader package from linaro
```
ndechesne, line 22: I doubt this is correct.

conf/machine/sm8250-mtp.conf
```diff
@@ -0,0 +1,26 @@
...
+MACHINE_FEATURES = "usbhost usbgadget alsa screen wifi bluetooth ext2"
+
+KERNEL_IMAGETYPE ?= "Image.gz"
+KERNEL_DEVICETREE ?= "qcom/sm8250-mtp.dtb"
+
+SERIAL_CONSOLE ?= "115200 ttyMSM0"
+
+MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
+    kernel-modules \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'wifi', 'linux-firmware-ath10k', '', d)} \
```
ndechesne, line 16: is the wifi firmware included in this package?

recipes-kernel/linux/linux-linaro-qcomlt_5.7.bb
```diff
@@ -13,7 +13,7 @@ LOCALVERSION ?= "-linaro-lt-qcom"
 SRCBRANCH ?= "release/qcomlt-5.7"
 SRCREV ?= "ba75ed77ea9ee88706bd1b78512f4f86040feb6c"
 
-COMPATIBLE_MACHINE = "(apq8016|apq8096|sdm845)"
+COMPATIBLE_MACHINE = "(apq8016|apq8096|sdm845|sm8250)"
```
ndechesne, line 16: does 5.7 boot fine on the MTP?

lumag: Yes, it does.
```
uname -a
Linux sm8250-mtp 5.7.2+linaro #1 SMP PREEMPT Tue Jun 16 13:00:35 UTC 2020 aarch64 GNU/Linux
```

conf/machine/sm8250-mtp.conf
lumag: How can I verify, which partitions is it?

conf/machine/sm8250-mtp.conf
lumag: This was a c&p from `dragonboard-845c`, which included `l-f-ath10k and l-f-sm845-modem`. Dropping this for now till we get firmware packages sorted out.

conf/machine/sm8250-mtp.conf
ndechesne: by default, we don't generate an initramfs, and we boot directly from kernel into user space. So we use QCOM_BOOTIMG_ROOTFS to set root= in the boot arg. So we need to pick one of the partition on the device that we want to use to the rootfs. 

For Dragonboard, we usually release our own GPT and we create a partition called 'rootfs'. But for the MTP, we don't do that, so we need to pick one of the existing partition. We usually pick userdata partition (because it's typically large, and it's safe to erase if needed). So you need to find out which block device corresponds to userdata by poking at the GPT on the device.

lumag: Updated to point to userdata partition

PR #165, opened 2020-06-20
Remove venus fw install from firmware-qcom-dragonboard845c recipe

olivierschonken:
The linux-firmware package complains that venus fw is supplied by both
linux-firmware and firmware-qcom-dragonboard845c.

The linux-firmware recipe is actually ambiguous in the sense that it doesn't
list the venus-5.2 package in the recipe, but does install it alongside all
the other firmware in the do_install part of the recipe.

Collected errors:
 * check_data_file_clashes: Package linux-firmware wants to install file /workdir/build/tmp-glibc/work/dragonboard_845c-linux/rb3-image-dev/9999.1-r0/rootfs/lib/firmware/qcom/venus-5.2/venus.b00
	But that file is already provided by package  * firmware-qcom-dragonboard845c
 * check_data_file_clashes: Package linux-firmware wants to install file /workdir/build/tmp-glibc/work/dragonboard_845c-linux/rb3-image-dev/9999.1-r0/rootfs/lib/firmware/qcom/venus-5.2/venus.b01
	But that file is already provided by package  * firmware-qcom-dragonboard845c
 * check_data_file_clashes: Package linux-firmware wants to install file /workdir/build/tmp-glibc/work/dragonboard_845c-linux/rb3-image-dev/9999.1-r0/rootfs/lib/firmware/qcom/venus-5.2/venus.b02
	But that file is already provided by package  * firmware-qcom-dragonboard845c
 * check_data_file_clashes: Package linux-firmware wants to install file /workdir/build/tmp-glibc/work/dragonboard_845c-linux/rb3-image-dev/9999.1-r0/rootfs/lib/firmware/qcom/venus-5.2/venus.b03
	But that file is already provided by package  * firmware-qcom-dragonboard845c
 * check_data_file_clashes: Package linux-firmware wants to install file /workdir/build/tmp-glibc/work/dragonboard_845c-linux/rb3-image-dev/9999.1-r0/rootfs/lib/firmware/qcom/venus-5.2/venus.b04
	But that file is already provided by package  * firmware-qcom-dragonboard845c
 * check_data_file_clashes: Package linux-firmware wants to install file /workdir/build/tmp-glibc/work/dragonboard_845cp-linux/rb3-image-dev/9999.1-r0/rootfs/lib/firmware/qcom/venus-5.2/venus.mdt
	But that file is already provided by package  * firmware-qcom-dragonboard845c

Signed-off-by: Olivier Schonken <olivier.schonken@gmail.com>

### DIFF
--- a/recipes-bsp/firmware/firmware-qcom-dragonboard845c_20190529180356-v4.bb
+++ b/recipes-bsp/firmware/firmware-qcom-dragonboard845c_20190529180356-v4.bb
@@ -27,14 +27,12 @@ do_compile() {
 
 do_install() {
     install -d ${D}${nonarch_base_libdir}/firmware/
-    install -d ${D}${nonarch_base_libdir}/firmware/qcom/venus-5.2/
     install -d ${D}${nonarch_base_libdir}/firmware/qcom/sdm845
 
     install -m 0444 ./17-USB3-201-202-FW/K2026090.mem ${D}${nonarch_base_libdir}/firmware/
     install -m 0444 ./18-adreno-fw/a630_zap*.* ${D}${nonarch_base_libdir}/firmware/qcom/
     install -m 0444 ./20-adsp_split/firmware/adsp*.* ${D}${nonarch_base_libdir}/firmware/qcom/sdm845
     install -m 0444 ./21-cdsp_split/firmware/cdsp*.* ${D}${nonarch_base_libdir}/firmware/qcom/sdm845
-    install -m 0444 ./33-venus_split/venus.* ${D}${nonarch_base_libdir}/firmware/qcom/venus-5.2/
 
     install -d ${D}${nonarch_base_libdir}/firmware/ath10k/WCN3990/hw1.0/
     install -m 0444 ./board-2.bin ${D}${nonarch_base_libdir}/firmware/ath10k/WCN3990/hw1.0/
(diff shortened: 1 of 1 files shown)


ndechesne:
thanks for the detailed report. We missed that in the recent firmware changes. But it's not that simple.. we've noticed that the venus-5.2 firmware shipped in linux-firmware (upstream) is no longer working (in fact we even suspect it might never have worked..). It was pushed by QCOM , but they might have pushed a wrong version. it explains why in oe-core recipe we didn't create the venus-5.2 package in the linux-firmware recipe (unlike the other versions). 

So for dragonboard845c machine, we need venus-5.2 but we need the version which is provided by meta-qcom. 

And now of course, i can see why there is the conflict that you found.. Let's check with @alimon how he tested that in the first place. Then we need to figure out how to fix the issue. Right now i think the only real fix would be to remove the venus-5.2 from linux-firmware recipe (in OE core).. but I would like to hear from @alimon first.

Note that we are trying to understand why it fails with the firmware from linux-firmware.. but we haven't found the reason yet. and we might just send a PR to update the firmware..

olivierschonken:
No problem, this was the quick fix to get it to complete the build.  My previous fix was to checkout meta-qcom from before the linux-firmware dependency merge.  Will revert back to that for the interim until linux-firmware is fixed.

ndechesne:
if you just need to fix the build issue (and don't care about working video codecs on 845c board) then you can use this patch for now. If you care about video codecs on 845c, you should instead remove the firmware from the OE-core recipe. 

But I can't merge this change since video codecs wouldn't work on 845c..

ndechesne:
hmm. i think I could merge a patch with a bbappend for linux-firmware (for 845c) that removes the firmware. If you get a chance to try that, please submit it, otherwise I can have a look later.

alimon:
@olivierschonken In our RPB builds it select by default firmware-qcom-dragonboard845c recipe so I didn't see that issue before, We are aware of linux-firmware venus packages are not working. A partial fix should be remove them as you send the PR.

https://github.com/ndechesne/meta-qcom/pull/166

olivierschonken:
With https://github.com/ndechesne/meta-qcom/pull/165 merged as a partial fix, this PR is no longer needed. Closing

PR #171, opened 2020-07-06
RB5 and integration backports

alimon:


ndechesne:
I am going to merge anyways... but please note that the kernel recipes should be *before* the new machines, since after #1 or after #3 it won't build..

PR #172, opened 2020-07-06
recipes-kernel: Linux 5.7 bump to 5.7.7 (21bb88052948)

alimon:
Changes,

21bb88052948 venus: firmware: Set virtual address ranges
2d8435ad822f firmware: qcom_scm: Add memory protect virtual address ranges
4130227f0f7e Merge tag 'v5.7.7' into release/qcomlt-5.7
fb6c79052380 Linux 5.7.7


ndechesne:
so, with this change we can now properly package the venus firmware in linux-firmware recipe, and then get rid of d1352e81c7973930876c4363824eafd24d87e5ab, is that right?

alimon:
> so, with this change we can now properly package the venus firmware in linux-firmware recipe, and then get rid of [d1352e8](https://github.com/ndechesne/meta-qcom/commit/d1352e81c7973930876c4363824eafd24d87e5ab), is that right?

@ndechesne For venus-5.2 is right, I didn't test venus-5.4 but I guess is the same case.

PR #178, opened 2020-08-26
firmware-qcom-rb5: package QRB5165 RB5 firmware files

lumag:
Add recipe extracting firmware files from Qualcomm's binary releases and
packaging them into proper binary package.

### DIFF
--- a/recipes-bsp/firmware/firmware-qcom-rb5_1.0.bb
+++ b/recipes-bsp/firmware/firmware-qcom-rb5_1.0.bb
@@ -1,0 +1,69 @@
+# Provide base URI of NHLOS_Binaries.zip and adreno_1.0_qrb5165_rb5.tar.gz
+# files.  Use "file://" if those files are copied into
+# recipes-bsp/firmware/files/ directory.
+# NHLOS_URI ?= "file://"
+# ADRENO_URI ?= "file://"
+
+DESCRIPTION = "QCOM Firmware for Qualcomm Robotics RB5 platform"
+
+LICENSE = "Proprietary"
+
+# There is no license file in the archive
+#LIC_FILES_CHKSUM = "file://license.txt;md5="
+ERROR_QA_remove = "license-checksum"
+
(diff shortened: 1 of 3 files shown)


recipes-bsp/firmware/firmware-qcom-rb5_1.0.bb
```diff
@@ -0,0 +1,61 @@
+# Copy/link NHLOS_Binaries.zip and adreno_1.0_qrb5165_rb5.tar.gz into
+# recipes-bsp/firmware/files/ directory or provide full URI below.
+# Then uncomment the RB5_HAVE_FIRMWARE line
+NHLOS_URI ?= "file://"
```
ndechesne, line 4: i think we could have something simpler here.

first we could create a single firmware tarball that includes everything (like RB3_firmware.zip file for db845c), instead of NHLOS_URI and ADRENO_URI, e.g. we just combine them into one file. When the vendor makes a release of the firmware, it will most likely be a single file (like RB3).

So we could set here
RB5_FIRMWARE_URI ?= ""
and remove RB5_HAVE_FIRMWARE. 

RB5_FIRMWARE_URI can be set in local.conf as file://xxxx.zip or even http://www.xxx.com/foobar.zip in case we want to archive it somewhere.

lumag: @ndechesne my goal was to use original release (from chipcode) as much as possible. When the vendor releases the firmware, this recipe will have to be adapted anyway, to use correct paths inside the tarball.

ndechesne: ok, fair enough.. can you then do

NHLOS_URI ?= ""
ADRENO_URI ?= ""

instead, and remove RB5_HAVE_FIRMWARE?

lumag: @ndechesne done

ndechesne:
thanks!

PR #180, opened 2020-09-02
Use venus 5.2 from linux-firmware for DB845c and replace dhcp-client with dhcpcd

alimon:
A fix was add on mainline to support linux-firmware binary,

https://git.linaro.org/landing-teams/working/qualcomm/kernel.git/commit/?h=release/qcomlt-5.7&id=21bb88052948b35bdce926f301f2ba7970040812


alimon:
@ndechesne Failed because removal of dhcp-client, I changed to use dhcpcd.

PR #182, opened 2020-09-03
Improve testimages not install QCOM tools by default

alimon:


recipes-graphics/mesa/mesa_%.bbappend
```diff
@@ -11,3 +11,6 @@ PACKAGECONFIG_append_apq8064 = "${PACKAGECONFIG_FREEDRENO}"
 PACKAGECONFIG_append_apq8096 = "${PACKAGECONFIG_FREEDRENO}"
 PACKAGECONFIG_append_sdm845 = "${PACKAGECONFIG_FREEDRENO}"
 PACKAGECONFIG_append_sm8250 = "${PACKAGECONFIG_FREEDRENO}"
+
```
ndechesne, line 14: this does not look right to me. And it impacts the YP compatible status (e.g. yocto-check-layer script will fail with this change). In a BSP layer you can change only settings for the machines you own.

again, if you need that, you can do these changes in your 'layer' or in your own local.conf or auto.conf.

recipes-bsp/firmware/firmware-qcom-dragonboard410c_1034.2.1.bb
```diff
@@ -31,15 +31,11 @@ do_install() {
     ::image/modem.* ::image/mba.mbn ::image/wcnss.* ${D}${nonarch_base_libdir}/firmware/qcom/msm8916
 
     install -d ${D}${sysconfdir}/
-    install -m 0644 LICENSE ${D}${sysconfdir}/QCOM-LINUX-BOARD-SUPPORT-LICENSE
+    install -m 0644 LICENSE ${D}${sysconfdir}/QCOM-LINUX-BOARD-SUPPORT-LICENSE-${PN}_${PV}
```
ndechesne, line 34: perhaps we need only $PN, not $PV?

recipes-bsp/firmware/firmware-qcom-dragonboard410c-bootloader-sdcard_17.09.bb
```diff
@@ -9,7 +9,7 @@ SRC_URI = "https://releases.linaro.org/96boards/dragonboard410c/linaro/rescue/17
 SRC_URI[md5sum] = "e15da2a623442d66587aea506599fb69"
 SRC_URI[sha256sum] = "9885f915ebd4986432340cf1d03b8fd2bfdd97ad6a4a7466200fddbe41cdcf5c"
 
-COMPATIBLE_MACHINE = "(dragonboard-410c)"
+COMPATIBLE_MACHINE = "(dragonboard-410c|qemuarm|qemuarm64)"
```
ndechesne, line 12: a better fix might be to remove the COMPATIBLE_MACHINE, no? Maybe it was misused in the first place. the recipe can 'build' for any machine.

alimon: Ack, I limited to qemu only but that's better since are binares can be installed in any machine.

recipes-bsp/firmware/firmware-qcom-dragonboard410c_1034.2.1.bb
alimon: Is fine to me to not have the PV.

recipes-graphics/mesa/mesa_%.bbappend
alimon: Right, this will affect any build that includes meta-qcom. So will remove it and enable in my config.

recipes-test/images/initramfs-test-image.bb
```diff
@@ -24,6 +24,7 @@ PACKAGE_INSTALL = " \
     udev \
     usbutils \
     wpa-supplicant \
+    ${INITRAMFS_IMAGE_EXTRA_INSTALL} \
```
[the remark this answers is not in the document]
alimon: Ok, Will use PACKAGE_INSTALL.

PR #184, opened 2020-09-07
recipes-bsp/firmware: Compatible machine and License install

alimon:
- Remove COMPATIBLE_MACHINE to allow be installed in any machine.

- Remove conflicts with linux-firmware-qcom-license:
  When try to install together this tree packages fails due to each of
  them provide license so use linux-firmware one since we are using
  linux-firmware when available.

- Install each license with ${PN} suffix to avoid install
  conflicts.

### DIFF
--- a/recipes-bsp/firmware/firmware-qcom-dragonboard410c-bootloader-sdcard_17.09.bb
+++ b/recipes-bsp/firmware/firmware-qcom-dragonboard410c-bootloader-sdcard_17.09.bb
@@ -9,7 +9,6 @@ SRC_URI = "https://releases.linaro.org/96boards/dragonboard410c/linaro/rescue/17
 SRC_URI[md5sum] = "e15da2a623442d66587aea506599fb69"
 SRC_URI[sha256sum] = "9885f915ebd4986432340cf1d03b8fd2bfdd97ad6a4a7466200fddbe41cdcf5c"
 
-COMPATIBLE_MACHINE = "(dragonboard-410c)"
 PACKAGE_ARCH = "${MACHINE_ARCH}"
 
 S = "${WORKDIR}"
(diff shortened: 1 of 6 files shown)


kraj:
@ndechesne @alimon this pull removed the `COMPATIBLE_MACHINE` which opened up these recipes to more than arm machines, e.g. qemumips is failing e.g.

https://errors.yoctoproject.org/Errors/Details/498777/

ndechesne:
@kraj , I continue to be impressed with your CI system ;) 

I think the right fix is what you did in https://github.com/ndechesne/meta-qcom/pull/186, right?
While the firmware are not needed on other machines, they are not "incompatible", e.g. they won't break other machines should they install them. 

we are building 'single' image for all the qcom machines, so we needed to be able to install all our firmware recipes at once.

kraj:
right, this fixes one issue to and reveals another. Which I posted another pull, but perhaps you should just build all meta-qcom recipes for a non-arm device like qemux86 or qemumips to ensure that you are not adding recipes into common namespace which will step on other arches/machines.

ndechesne:
i am currently building
            name: IMAGES
            default: 'core-image-base core-image-weston core-image-x11 initramfs-test-image'

for each of these:
            name: MACHINE
            values:
                - 'dragonboard-410c'
                - 'dragonboard-845c'
                - 'qrb5165-rb5'

are you suggesting i just add qemux86 here? If you caught this issue with the firmware, it probably means you are doing a world build, right? I am not sure i can afford a 'world' build with the CI capacity i have (now).

kraj:
Yes, I am doing world builds as part of Yoe Distro ( which you guys should use :) ), but you can just do world build for your layer along. Look into `meta-openembedded/meta-oe/recipes-core/packagegroups/packagegroup-meta-oe.bb` and `meta-openembedded/meta-oe/recipes-core/images/meta-oe-image.bb`, you can also define such an image for meta-qcom and test every recipe from your layer for machines you use plus qemumips or some such. You really don't have to build whole world, just meta-qcom world.

PR #193, opened 2020-10-08
Support multiple device tree files per machine

lumag:


lumag:
@ndechesne @alimon I'd really ask to get this into master (and then dunfell). Otherwise testing analog audio with Linaro-built images would be impossible for me.

alimon:
@lumag LGTM, @ndechesne can you merge?

recipes-kernel/linux/linux-linaro-qcom.inc
```diff
@@ -81,12 +81,29 @@ do_configure_prepend() {
         oe_runmake -C ${S} O=${B} savedefconfig && cp ${B}/defconfig ${WORKDIR}/defconfig.saved
 }
 
+# Kernel image might be restored from sstate and thus be appended with the DTB.
+# In order not to end up with several DTB images appended to the kernel image,
+# remove and rebuild it.
+do_compile_prepend() {
```
ndechesne, line 87: are you sure about that? 
I think it was a potential bug *before* this change, but not with the current patch, since we no longer override this file

ndechesne: hmm. i don't even understand how it could happen. do_compile is not a sstate enabled task. only do_deploy. so we couldn't have restored the file with sstate. what am i missing?

lumag: I have seen this several times when switching back and forth between multi-dtb enabled and not-enabled trees, but it might be related with me sporadically doing strange things. I can try reproducing the issue. If that fails, I'll just drop this chunk.

ndechesne: it might happen if you ran tasks individually, like

bitbake -fc compile

back and forth between the 2 branches, i can see that. since do_compile_append, before your patch would override this file. 
so in the worst case it is a bug that is no longer there with your new implementation, since you don't override that file anymore, since you use it multiple times for each dtb..

i would say, please remove this chunk, and I will merge the PR!

lumag: @ndec done

PR #198, opened 2020-11-03
recipes-kernel: Add Linux 5.9 support for apq8016|apq8096|sdm845

alimon:

### DIFF
--- a/recipes-kernel/linux/linux-linaro-qcomlt_5.9.bb
+++ b/recipes-kernel/linux/linux-linaro-qcomlt_5.9.bb
@@ -8,7 +8,11 @@ require recipes-kernel/linux/linux-linaro-qcom.inc
 require recipes-kernel/linux/linux-qcom-bootimg.inc
 
 LOCALVERSION ?= "-linaro-lt-qcom"
-SRCBRANCH ?= "release/rb5/qcomlt-5.9"
-SRCREV ?= "6d5a9a5da79684f69e4c66a7cf9108ab4e77025f"
 
-COMPATIBLE_MACHINE = "(sm8250)"
+SRCBRANCH ?= "release/qcomlt-5.9"
+SRCREV ?= "fca727995a6800a2c0f07313e3e0917cffdf0e36"
+
+SRCBRANCH_sm8250 = "release/rb5/qcomlt-5.9"
+SRCREV_sm8250 = "6d5a9a5da79684f69e4c66a7cf9108ab4e77025f"
+
+COMPATIBLE_MACHINE = "(apq8016|apq8096|sdm845|sm8250)"


recipes-kernel/linux/linux-linaro-qcomlt_5.9.bb
```diff
@@ -8,7 +8,11 @@ require recipes-kernel/linux/linux-linaro-qcom.inc
...
-SRCREV ?= "6d5a9a5da79684f69e4c66a7cf9108ab4e77025f"
 
-COMPATIBLE_MACHINE = "(sm8250)"
+SRCBRANCH ?= "release/qcomlt-5.9"
+SRCREV ?= "fca727995a6800a2c0f07313e3e0917cffdf0e36"
+
+SRCBRANCH_sm8250 = "release/rb5/qcomlt-5.9"
+SRCREV_sm8250 = "6d5a9a5da79684f69e4c66a7cf9108ab4e77025f"
+
+COMPATIBLE_MACHINE = "(apq8016|apq8096|sdm845|sm8250)"
```
ndechesne, line 18: do you really want to support sm8250/rb5 here? have you discussed with @lumag  about that?

alimon: @ndechesne I did in this way because Linux 5.9 recipe exist before with release/rb5/qcomlt-5.9, I set default to release/qcomlt-5.9 and override the same values for sm8250.

alimon: @lumag Is this recipe currently in use for RB5?, I see latest hash for release/rb5/qcomlt-5.9 is 0e85e8ccbd1c0d8e6a995a6ceb1a63b808b76adc.

lumag: This recipe for rb5's 5.9 was used for the release. I'm definitely fine with this PR.

ndechesne: my recommendation here is the following:
let's rename the existing linux-linaro-qcomlt_5.9.bb to linux-linaro-qcomlt-rb5_5.9.bb and use that recipe for RB5 machine.
then we can introduce this new flavor of 5.9 only for the other targets.

then for 5.10, we will see how to handle that.

ndechesne: @lumag anibal will backport this recipe in dunfell branch for the RB3 release. so we need to make sure that we don't break RB5 builds with both master and dunfell branches.

lumag: The RB5-specific 5.9 recipe is a part of dunfell branch also. As I said above, I'm fine with this patch. Renaming old recipe and adding the new one would also be fine with me. I'd prefer the current solution though: all qcomlt 5.9 releases are in a single recipe, no need to play with `PREFERRED_PROVIDER_virtual/kernel`.

ndechesne: argh. i am sorry.. i think i read the patch too quickly.. the recipe takes care of the 2 different commits for RB5 vs the others! so the patch is rather good in fact!

will check once more and merge it!

PR #208, opened 2020-11-26
RB5 updates for Dunfell

lumag:


recipes-multimedia/alsa/alsa-ucm-conf_1.2.%.bbappend
```diff
@@ -1,6 +1,6 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${BPN}:"
 
-SRC_URI_append_sdm845 = " \
+SRC_URI_append = " \
```
ndechesne, line 3: we should not do that. If someone adds meta-qcom in BBLAYERS, but don't build for a QCOM machine, they should not be impacted by this change.

lumag: In the past I also preferred to have as fine-grained control as possible. However now I prefer to have all patches enabled for as broad audience as possible, to ensure that they do not break a thing. If you absolutely insist, I can make them enabled only for sdm845 and sm8250. However such fine-grained enablements make e.g. multi-target support harder (consider #147)

ndechesne: I know.. i also know that some users complain when they build with meta-qcom for non qcom hardware since our patches are getting into their build.. 

for the multi target support, we will most likely need a broader 'soc family'. e.g. _qcom instead of _<soc>. 

It's not that I want per soc config, it's just that I think we should not include our patches for non QCOM platform. if you want to create a new _qcom override, I am fine with that too! then we can include all *our* patches with _qcom.

lumag: Sounds like a good idea. I'll work on it separately.

PR #209, opened 2020-11-26
Qualcomm family override

lumag:


conf/machine/include/qcom-apq8064.inc
```diff
@@ -1,4 +1,4 @@
-SOC_FAMILY = "apq8064"
+SOC_FAMILY = "apq8064:qcom"
```
ndechesne, line 1: hmm. shouldn't it be qcom:apq8064, e.g. from more generic to less generic?

conf/machine/include/qcom-apq8016.inc
```diff
@@ -1,4 +1,4 @@
-SOC_FAMILY = "apq8016"
+SOC_FAMILY = "apq8016:qcom"
```
ndechesne, line 1: perhaps we can create include/qcom.inc with:

SOC_FAMILY = "qcom"

ndechesne:
it looks much better with a single override!

conf/machine/include/qcom-apq8016.inc
lumag: I'd probably pick up a patch from #147

ndechesne: makes sense. yes.

conf/machine/include/qcom-apq8064.inc
lumag: done, in `qcom-common.inc`

conf/machine/include/qcom-apq8016.inc
lumag: done

PR #211, opened 2020-11-27
recipes-graphics: mesa_git bump to rev aed8d30b50

alimon:


lumag:
Reviewed-by: Dmitry Baryshkov

ndechesne:
does the build fail if we pass the PACKAGECONFIG?

lumag:
> does the build fail if we pass the PACKAGECONFIG?

This might be a cleaner solution:
```
PLATFORMS_remove = "drm surfaceless"
```

ndechesne:
> > does the build fail if we pass the PACKAGECONFIG?
> 
> This might be a cleaner solution:
> 
> ```
> PLATFORMS_remove = "drm surfaceless"
> ```

probably. That said I was interesting to understand why we needed this change. I have now looked at the upstream commit, and I understand. But perhaps a better commit message would be helpful.

alimon:
@ndechesne  @lumag I updated the commit to remove platforms instead.

PR #222, opened 2020-12-10
Enable WiFi on RB5

lumag:
Two patches to fix WiFi on RB5 board. Note, firmware files are not provided by this PR, they will come separately.


recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -10,6 +10,7 @@ SD_BOOT_IMAGE_BASE_NAME = "boot-sd${KERNEL_IMAGE_NAME}"
 SD_BOOT_IMAGE_SYMLINK_NAME = "boot-sd-${KERNEL_IMAGE_LINK_NAME}"
 KERNEL_CMDLINE = "root=${1} rw rootwait console=${ttydev},${baudrate}n8"
 KERNEL_CMDLINE_append_dragonboard-845c = " clk_ignore_unused pd_ignore_unused"
+KERNEL_CMDLINE_append_qrb5165-rb5 = " pcie_pme=nomsi"
```
ndechesne, line 13: is that a temporary solution? how can we get rid of it eventually?

lumag: Not until upstream driver supports working with less than 32 MSI vectors (Mani told that there are plans to support this, but no particular schedule for that). So I'd assume this is semi-permanent for now.

PR #223, opened 2020-12-10
RFC: Use pil-squashed firmware for RB5

lumag:


lumag:
@ndechesne I've rebased this to remove conflict and dropped for now mounting removal patch. However pil-squashing of 'mounted' firmware does not happen for now.

PR #224, opened 2020-12-11
Mesa git update to rev 533f6debb17b02ee9f92f51c19a2da9c0700bca1 and fix OSMESA set

alimon:


recipes-graphics/mesa/mesa_git.bb
```diff
@@ -10,10 +10,11 @@ SRC_URI = "git://gitlab.freedesktop.org/mesa/mesa.git;protocol=https \
            "
 LIC_FILES_CHKSUM = "file://docs/license.rst;md5=9aa1bc48c9826ad9fdb16661f6930496"
 
-SRCREV = "aed8d30b507568b7fc0f32afca012f8def5aca16"
+SRCREV = "533f6debb17b02ee9f92f51c19a2da9c0700bca1"
 #SRCREV_sm8250 = "${AUTOREV}"
 
 PLATFORMS_remove = "drm surfaceless"
+PACKAGECONFIG[osmesa] = "-Dosmesa=true,-Dosmesa=false"
```
ndechesne, line 17: it does not look like the right thing to do. we should add -Dosmesa=false in the build options instead, no? Why are we creating a new PACKAGECONFIG in a bbappend?

alimon: @ndechesne This PACKAGECONFIG comes from the original mesa.inc in OE-Core, mesa_git inherits it enabling build master.  

https://git.openembedded.org/openembedded-core/tree/meta/recipes-graphics/mesa/mesa.inc#n176

ndechesne: if it's already defined in mesa.inc, why do we need it here then?

alimon: Because the -Dosmesa changed in recent mesa master rev 533f... so classic isn't a supported value now only true and false.

alimon: https://gitlab.freedesktop.org/mesa/mesa/-/commit/ee802372180

ndechesne: ok, got it! thanks!

PR #229, opened 2020-12-16
Add rdeps on perl and python3

kraj:
Use python3 explicitly.

I have not runtime tested it, but it should work out. Please review


lumag:
RDEPENDS patch is good, another one is not.
Thank you for noticing this, I'm working on a correct patch now.

kraj:
ok please cherry pick the first one and we can then close this PR, provided you have fix for not needing `/usr/bin/python`

lumag:
@kraj see #230

PR #233, opened 2020-12-21
recipes-bsp: firmware-qcom-dragonboard845c fix installation of renesas

alimon:
When renesas firmware loader landed upstream [1] there is a common name
for the firmware so install the K2026090.mem as renesas_usb_fw.mem.

[1] https://github.com/torvalds/linux/commit/a66d21d7dba84deeaf3b296c43eafc11094b6f09


alimon:
@ndechesne Still not getting ROM flashed with this changes, I will sync with Vinod about it.

PR #236, opened 2020-12-29
Backport two patches from master

lumag:


ndechesne:
I applied locally, rebased and pushed. thanks!

PR #239, opened 2021-01-08
RB5: use Thundercomm firmware release for the firmware-qcom-rb5 recipe

lumag:
This is an initial preview to let us test the firmware packages.


recipes-bsp/firmware/firmware-qcom-rb5_20201229182615-v1.bb
```diff
@@ -0,0 +1,71 @@
...
+SRC_URI[md5sum] = "11af73d4e90ed82e6d40b90f1a99ab2f"
+SRC_URI[sha256sum] = "5a349eed4569a67e5c41d528e312039c005dc7e9dc43e014c6dbdfa85169c90e"
+
+S = "${WORKDIR}/RB5_firmware_${PV}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+VENUS_FW = "vpu-1.0"
+
+# Firmware archive contains some directories marked as read-only. Unzip will
```
ndechesne, line 28: you can assume that we will create the ZIP file ourselves, so we should fix that problem in the ZIP before we publish to linaro.org. 
It is unlikely that we will get a 'working' ZIP file from TCOM, they might release us patched files on top of their initial release. in fact we don't know. 

Check the README in the (private) firmware tree, I have an instruction to "Create Linaro firmware archive"

lumag: @ndechesne I'll drop this from the final revision then. For now it's just easier to use the original file.

recipes-bsp/firmware/firmware-qcom-rb5/generate_board-2_json.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
ndechesne, line 1: do we really need a duplicate file, we have similar script for 845c, right? Should that belong to qca-swiss-army recipe instead?

lumag: Ideally they can probably go into the `qca-swiss-army-knife` repo itself, but the upstream doesn't seem too responsive

ndechesne: right. we can at least put it in the recipe, so that we don't duplicate for each firmware recipe?

lumag:
This PR is now updated to use the firmware package available once it is handled by systems team.

lumag:
@ndechesne this can be merged now

PR #259, opened 2021-02-02
Initial machine support for SDX55

Mani-Sadhasivam:
Add machine support for SDX55 MTP and Telit FN980

Signed-off-by: Manivannan Sadhasivam <manivannan.sadhasivam@linaro.org>


lumag:
@alimon @fboudra this build failure looks like another git fetcher failure happening after stretch -> buster switch. Would it be possible and proper to drop all git2 cached downloads from our builders so that we won't have such issues in future?

lumag:
run tests

PR #261, opened 2021-02-03
qrtr and rmtfs updates

alimon:


ndechesne:
NAK. this is not the right way. SRCPV will construct a string that includes the commit already. so we should not update PV directly. and if the PR server is started, it will even automatically increment the version. 

e.g. I have this:
qrtr_0.0+0+983b223423-r0_cortexa53.ipk

the first '0.0' comes from PV, then "+0+983b223423-r0" comes from SRCPV (which comes from bitbake getrev()). If you were to rebuild the same recipe over with PR server the next version would be +1+983b223423-r0, .. .

I would be ok to change 0.0 to 0.2 or 0.3, that is fine.

recipes-support/qrtr/qrtr_git.bb
```diff
@@ -7,10 +7,10 @@ LIC_FILES_CHKSUM = "file://LICENSE;md5=15329706fbfcb5fc5edcc1bc7c139da5"
 
 inherit systemd
 
-SRCREV = "983b223423f3ec36f231455b6a93bc2a10921794"
+SRCREV = "9dc7a88548c27983e06465d3fbba2ba27d4bc050"
 SRC_URI = "git://github.com/andersson/${BPN}.git;branch=master;protocol=https"
 
-PV = "0.0+${SRCPV}"
+PV = "0.3+47g9dc7a88"
```
lumag, line 13: Ideally one should also set PE = "1" here and in the other recipe.

lumag, line 13: Also fixing `SRCREV` in the `PV` introduces a point of failure if one updates `SRCREV` but not `PV`.
I'd set `PV` to `"0.0+47+${SRCPV}"`.

As a second thought, why do you need to add this at all? We do not have package feeds, so we do not care about package versioning. And if one cares, he should use PRSERV instead, which would generate unique PR values.

ndechesne: why?

ndechesne: as i said in the comment, we don't even want/need the +47 that is a "point of failure" too, since it means 47 ahead of tag 0.3.. so it needs to be updated each time.

lumag: @ndec the proposal changes versioning scheme, so it'd be better to increase epoch.

alimon: I would like to maintain the same versions across Debian/OE but I agree that's not the best way to do it.

@lumag Indeed we don't have package feeds but once in a while when updating qrtr/rmtfs we have version backwards error that's we don't care because no package feeds but still is an error.

@lumag @ndechesne What about at least 0.3+${SRCPV}?, that will lead to errors of version going backwards too but we will ignore them.

lumag:
Acked-by: Dmitry Baryshkov

PR #266, opened 2021-02-09
Automatically update LT9611UXC firmware

lumag:


ndechesne:
isn't that too much buggy? I have several concerns:
* it will impact every boot, forever, even though it's needed at max once.
* if the firmware is wrong, and display doesn't work, do you expect the user needs to reboot to get a working display after firmware upgrade?
* I don't think graphical.target is right. you need display even for a non graphical console, no?

Should we simply provide the script without the systemd service and document that? could the kernel driver take care of that and check if the version is 'right' and load the firmware ?

lumag:
- There is a firmware version check, so firmware update will happen only once
- It depends. Next modeset (e.g. running the gdm) will get the user the working display
- True. What about basic.target?  Or getty.target?
- I thought about that. However if the user w/o serial cable flashes our board, he will have 'it does not work experience'. With this service the user will get 'reboot once, works' experience.

ndechesne:
how about we try to use ConditionFirstBoot then, so that the service is only ever used once? 
for the target, multi-user.target or getty.target seems reasonable choices

lumag:
ok

recipes-bsp/lt9611uxc-upgrade/lt9611uxc-upgrade/lt9611uxc.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Update LT9611UXC Firmware
+ConditionOnFirstBoot=true
```
ndechesne, line 3: so does that actually work? it will do it just once?

PR #268, opened 2021-02-10
Sumo db845c

roxell:


ndechesne:
my only concern here is that if anyone tries to build meta-qcom/sumo on db845c it will not work since the kernel recipe is too old.. I am going to merge that anyways.. we will deal with the problem if it happens (which is very unlikely).

PR #273, opened 2021-02-16
recipes-kernel/linux/linux-linaro-qcomlt-dev: use AUTOREV

ndechesne:
This kernel recipe builds the QCLT integration branch which is not a
*relase* branch, but is regularly rebased. With a fixed SRCREV the
build will often break. Instead let's use AUTOREV. While it's not
recommended, DEFAULT_PREFERENCE will ensure that this recipe is not
used unless it's explicitely required.

### DIFF
--- a/recipes-kernel/linux/linux-linaro-qcomlt-dev.bb
+++ b/recipes-kernel/linux/linux-linaro-qcomlt-dev.bb
@@ -12,14 +12,15 @@ require recipes-kernel/linux/linux-linaro-qcom.inc
 require recipes-kernel/linux/linux-qcom-bootimg.inc
 
 SRCBRANCH ?= "integration-linux-qcomlt"
-#SRCREV ?= "${AUTOREV}"
-SRCREV = "7b0c8f86e1ef16d58c582de5a0f0331f82640096"
 
-LINUX_VERSION = "5.10-rc+"
+# Set default SRCREV. it is statically set to the korg v3.7 tag, and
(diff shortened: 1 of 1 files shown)


lumag:
This would still cause parsing errors if the network is unavailable even if -dev is not selected.

ndechesne:
right. Looking at oe-core, they do this trick:

# Set default SRCREVs. Both the machine and meta SRCREVs are statically set                                                                                                                                                                                                                                                  
# to the korg v3.7 tag, and hence prevent network access during parsing. If                                                                                                                                                                                                                                                  
# linux-yocto-dev is the preferred provider, they will be overridden to                                                                                                                                                                                                                                                      
# AUTOREV in following anonymous python routine and resolved when the                                                                                                                                                                                                                                                        
# variables are finalized.                                                                                                                                                                                                                                                                                                   
SRCREV_machine ?= '${@oe.utils.conditional("PREFERRED_PROVIDER_virtual/kernel", "linux-yocto-dev", "${AUTOREV}", "29594404d7fe73cd80eaa4ee8c43dcc53970c60e", d)}'
SRCREV_meta ?= '${@oe.utils.conditional("PREFERRED_PROVIDER_virtual/kernel", "linux-yocto-dev", "${AUTOREV}", "29594404d7fe73cd80eaa4ee8c43dcc53970c60e", d)}'

that would work for us, right?

lumag:
@ndechesne yes, this should work.

PR #277, opened 2021-02-23
Initramfs  to load modules/firmware and mount rootfs

alimon:


alimon:
@ndechesne This is based upon @lumag PR, https://github.com/ndechesne/meta-qcom/pull/232 .

recipes-kernel/images/initramfs-rootfs-image.bb
```diff
@@ -0,0 +1,10 @@
+require initramfs-base-image.bb
+
+DESCRIPTION = "Ramdisk image for releases"
+
+# Add systemd etc/initrd-release to instruct systemd mount rootfs and do switch_root
+initrd_release () {
+    mkdir -p ${IMAGE_ROOTFS}/etc
+    touch ${IMAGE_ROOTFS}/etc/initrd-release
+}
+ROOTFS_POSTPROCESS_COMMAND += "${@oe.utils.conditional('VIRTUAL-RUNTIME_init_manager', 'systemd', 'initrd_release;', '', d)}"
```
lumag, line 10: Just wandered if the initrams image using `initramfs-framework` would be significantly smaller.

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -38,7 +38,7 @@ python do_qcom_img_deploy() {
     kernel_output_dir = d.getVar("KERNEL_OUTPUT_DIR")
     kernel_imagedest = d.getVar("KERNEL_IMAGEDEST")
     kernel = os.path.join(B, "kernel-dtb")
-    definitrd = os.path.join(B, "initrd.img")
+    definitrd = os.path.join(d.getVar("DEPLOY_DIR_IMAGE"), "initramfs-rootfs-image-" + d.getVar("MACHINE") + ".cpio.gz")
```
lumag, line 41: This way we'd loose an ability to build non-initramfs-enabled image. I think it should be fine for now.

lumag (approved):
Looks good to me!
Acked-by: Dmitry Baryshkov <dmitry.baryshkov@linaro.org>

recipes-kernel/linux/linux-qcom-bootimg.inc
lumag: A second thought: we can drop `QCOM_BOOTIMG_BUNDLE_INITRAMFS` completely, require valid `INITRAMFS_IMAGE` and use it through the recipe. For the cases when initramfs is not required we can add `initramfs-dummy-image` providing `This is not an initramfs` file.

recipes-kernel/images/initramfs-base-image.bb
```diff
@@ -0,0 +1,26 @@
+DESCRIPTION = "Base ramdisk image for release and running tests"
```
ndechesne, line 1: it's close to core-image-minimal-initramfs.bb from oe-core. but not quite the same. do we need to diverge from the oe-core recipe?

recipes-kernel/linux/linux-qcom-bootimg.inc
ndechesne: indeed we need to support users who will build without an initramfs.

ndechesne (commented):
I don't like a 60MB initramfs.. that's too big..

recipes-kernel/linux/linux-qcom-bootimg.inc
alimon: > A second thought: we can drop `QCOM_BOOTIMG_BUNDLE_INITRAMFS` completely, require valid `INITRAMFS_IMAGE` and use it through the recipe. For the cases when initramfs is not required we can add `initramfs-dummy-image` providing `This is not an initramfs` file.

@lumag That sound a good idea, The initramfs by default is because the intended is to have minimal builtin kernel drivers and everything else should be modules so initramfs is needed to load that modules before the whole system.

recipes-kernel/images/initramfs-base-image.bb
alimon: @ndechesne Yes, is close, I will try to use core-image-minimal-initramfs and see how it goes.

alimon: @ndechesne It fails to build because depends on grub, after remove grub from RDEPENDS it works but dosen't has any modules/firmware inside.

 https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-core/initrdscripts/initramfs-live-install_1.0.bb#L10

alimon:
> I don't like a 60MB initramfs.. that's too big..

@ndechesne Ack, that's because all firmware/modules built are installed inside the ramdisk.

Modules  52M     ./lib/modules
Firmware 48M     ./lib/firmware

We will need a manner to set which ones are needed to both for example UFS in DB845c.

alimon:
> > I don't like a 60MB initramfs.. that's too big..
> 
> @ndechesne Ack, that's because all firmware/modules built are installed inside the ramdisk.
> 
> The uncompressed ramdisk is 169M where,
> 
> Modules 52M ./lib/modules
> Firmware 48M ./lib/firmware
> 
> We will need a manner to set which ones are needed to both for example UFS in DB845c.

@ndechesne Removing the firmware and modules the ramdisk size is ~9MB compressed.

ndechesne:
> > > I don't like a 60MB initramfs.. that's too big..
> > 
> > 
> > @ndechesne Ack, that's because all firmware/modules built are installed inside the ramdisk.
> > The uncompressed ramdisk is 169M where,
> > Modules 52M ./lib/modules
> > Firmware 48M ./lib/firmware
> > We will need a manner to set which ones are needed to both for example UFS in DB845c.
> 
> @ndechesne Removing the firmware and modules the ramdisk size is ~9MB compressed. We can start from there and then figure out which/how-to add needed firmware/modules in maintainable way.

debian maintains a list of modules to include in their initramfs.. 
https://salsa.debian.org/kernel-team/initramfs-tools/-/blob/master/hook-functions#L600
not sure how that can be applied to OE.. but perhaps we are doing something generic and usptreamable here..

recipes-kernel/linux/linux-qcom-bootimg.inc
lumag: Yes. Just consider supporting custom kernel configs. However with `initramfs-dummy-image` we can drop duplicated code.

alimon:
> > > > I don't like a 60MB initramfs.. that's too big..
> > > 
> > > 
> > > @ndechesne Ack, that's because all firmware/modules built are installed inside the ramdisk.
> > > The uncompressed ramdisk is 169M where,
> > > Modules 52M ./lib/modules
> > > Firmware 48M ./lib/firmware
> > > We will need a manner to set which ones are needed to both for example UFS in DB845c.
> > 
> > 
> > @ndechesne Removing the firmware and modules the ramdisk size is ~9MB compressed. We can start from there and then figure out which/how-to add needed firmware/modules in maintainable way.
> 
> debian maintains a list of modules to include in their initramfs..
> https://salsa.debian.org/kernel-team/initramfs-tools/-/blob/master/hook-functions#L600
> not sure how that can be applied to OE.. but perhaps we are doing something generic and usptreamable here..

I added NO_RECOMMENDATIONS to initramfs-rootfs-image to avoid install all firmware/modules.

lumag:
This is now implemented in an alternative way.

PR #282, opened 2021-03-06
Improve fastrpc-OE integration

lumag:
- mount DSP partition to /usr/lib/rfsa
- provide systemd units for *dsprpcd
- fix libraries packaging

### DIFF
--- a/recipes-support/fastrpc/fastrpc/mount-dsp.sh
+++ b/recipes-support/fastrpc/fastrpc/mount-dsp.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+
+set -e
+
+if [ -r /sys/devices/soc0/machine ] ; then
+	MACHINE=`cat /sys/devices/soc0/machine`
+	case $MACHINE in
+		SM8250|QRB5160)
+			WHAT=/lib/firmware/qcom/sm8250/dspso.bin
+			;;
+	esac
+fi
+
+if [ -z "$WHAT" -o ! -r "$WHAT"] ; then
+	if [ -h /dev/disk/by-partlabel/dsp_a ] ; then
+		WHAT=/dev/disk/by-partlabel/dsp_a
+	else
+		WHAT=/dev/disk/by-partlabel/dsp
+	fi
+fi
+
+if [ -e "$WHAT" ] ; then
+	mount $WHAT /usr/lib/rfsa -o ro
+else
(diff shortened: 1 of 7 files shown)


recipes-support/fastrpc/fastrpc_git.bb
```diff
@@ -6,15 +6,32 @@ LICENSE = "BSD-3-Clause"
...
+SYSTEMD_SERVICE_${PN} = "usr-lib-rfsa.service"
+
+do_install_append() {
+    install -d ${D}${libdir}/rfsa
+
+    install -d ${D}${systemd_unitdir}/system
+    install -m 0644 ${WORKDIR}/usr-lib-rfsa.service ${D}${systemd_unitdir}/system
+
+    install -d ${D}${sbindir}
+    install -m 0755 ${WORKDIR}/mount-dsp.sh ${D}${sbindir}
```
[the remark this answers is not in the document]
lumag: Because this way they can be shared between RB5, RB3 and other boards.

recipes-support/fastrpc/fastrpc/mount-dsp.sh
```diff
@@ -0,0 +1,22 @@
...
+			;;
+	esac
+fi
+
+if [ -z "$WHAT" -o ! -r "$WHAT"] ; then
+	if [ -h /dev/disk/by-partlabel/dsp_a ] ; then
+		WHAT=/dev/disk/by-partlabel/dsp_a
+	else
+		WHAT=/dev/disk/by-partlabel/dsp
+	fi
```
ndechesne, line 19: there is no fallback here. e.g. what if we are not on 8250, and there is no dsp_a and no dsp partition? e.g. like on db410c, this service will fail on boot?

lumag: True, because there is nothing to be mounted, isn't it?
I can change the mount script to return silently if `$WHAT` does not exist.

lumag:
I'll check fastrpc on db410c and report back

recipes-support/fastrpc/fastrpc/mount-dsp.sh
ndechesne: yes. a failed systemd service isn't a good thing.

lumag: Fixed the service not to fail on db410c. Proper 410c support is still TBD

PR #289, opened 2021-04-01
qrb5165-rb5: use new rootfs partition

lumag:
Latest RB5 bootloader package provides updated partition scheme. It
removes Android partitions and adds single rootfs partition at
/dev/sda1. Switch to new rootfs partition.

### DIFF
--- a/conf/machine/qrb5165-rb5.conf
+++ b/conf/machine/qrb5165-rb5.conf
@@ -22,9 +22,8 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
 "
 # linux-firmware-qcom-adreno-a650 
 
-# 'userdata' is sda8 with older firmware/GPT and sda6 with newer firmware.
-# Allow kernel to resolve it on it's own.  Wipe it and use for our build.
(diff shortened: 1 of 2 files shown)


conf/machine/qrb5165-rb5.conf
```diff
@@ -22,9 +22,8 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
 "
 # linux-firmware-qcom-adreno-a650 
 
-# 'userdata' is sda8 with older firmware/GPT and sda6 with newer firmware.
-# Allow kernel to resolve it on it's own.  Wipe it and use for our build.
-QCOM_BOOTIMG_ROOTFS ?= "PARTLABEL=userdata"
+# /dev/sda1 is 'rootfs' partition after installing the latest bootloader package from linaro
+QCOM_BOOTIMG_ROOTFS ?= "/dev/sda1"
```
ndechesne, line 27: why not PARTLABEL=rootfs?

lumag: Because if we are to use the `rootfs` partition, we can be pretty sure that it is on `/dev/sda1`.

PR #302, opened 2021-05-21
recipes-test: Add new utilities to 'initramfs-test-image.bb'

bhupesh-sharma:
Add following utilities to 'initramfs-test-image.bb':

-> For kexec/kdump analysis:
   - crash
   - makedumpfile
   - kexec
-> iozone3: for filesystem performance analysis.
-> cryptsetup: for crypto benchmarks.
-> util-linux: for 'gdisk', 'taskset', etc.
-> coreutils: mainly for 'dd' (as busybox 'dd' does not show performance
              figures).

Addition of the above tools results in an approximate increase of
6M in the resulting rootfs size (cpio.gz archive)

Before:
-------
$ ls -lah initramfs-test-image-qcom-armv8a-20210521045425.rootfs.cpio.gz
19M

After:
------
$ ls -lah initramfs-test-image-qcom-armv8a-20210521070020.rootfs.cpio.gz
25M

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-full-image.bbappend
+++ b/dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-full-image.bbappend
@@ -1,0 +1,8 @@
+PACKAGE_INSTALL += " \
+    crash \
+    devmem2 \
+    iozone3 \
+    libgpiod \
+    libgpiod-tools \
+    makedumpfile \
+"
--- a/recipes-test/images/initramfs-test-full-image.bb
+++ b/recipes-test/images/initramfs-test-full-image.bb
@@ -1,0 +1,37 @@
+require recipes-test/images/initramfs-tiny-image.bb
+
+DESCRIPTION = "Relatively larger ramdisk image for running tests (bootrr, etc)"
+export IMAGE_BASENAME = "initramfs-test-full-image"
+
+PACKAGE_INSTALL += " \
+    bluez5 \
+    coreutils \
+    dhcpcd \
+    diag \
+    e2fsprogs \
+    e2fsprogs-e2fsck \
+    e2fsprogs-mke2fs \
+    e2fsprogs-resize2fs \
+    e2fsprogs-tune2fs \
+    ethtool \
+    gptfdisk \
+    iw \
+    kexec-tools \
+    lava-test-shell \
+    libdrm-tests \
+    lsof \
+    ncurses \
+    ncurses-terminfo \
+    ncurses-terminfo-base \
+    pciutils \
+    pd-mapper \
+    qrtr \
+    rmtfs \
+    rt-tests \
+    stress-ng \
+    tqftpserv \
+    usbutils \
+    util-linux \
+    util-linux-chrt \
+    wpa-supplicant \
+"


recipes-test/images/initramfs-test-image.bb
```diff
@@ -5,6 +5,9 @@ export IMAGE_BASENAME = "initramfs-test-image"
 
 PACKAGE_INSTALL += " \
     bluez5 \
+    coreutils \
+    crash \
+    cryptsetup \
```
ndechesne, line 10: cryptsetup is already in the image, see 78e2dacd0ebcf6719ab122ed65975351e3bd87db.

in the main bb file you should only add packages which come from openembedded-core, if you include packages from other layers, we need to use the ./dynamic-layers/ subfolder instead. meta-qcom *must* build with just oe-core enabled.

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-image.bbappend
```diff
@@ -1,4 +1,9 @@
 PACKAGE_INSTALL += " \
+    coreutils \
```
ndechesne, line 2: the bbappend in dynamic-layers is used when meta-oe is added to the user's build. This is to ensure that if someone does not build with meta-oe, then the build won't fail. 

so you need to add packages from meta-oe in this file, but packages from oe-core can be in the main bb recipe, they don't need to be here.

bhupesh-sharma: Ok, new version v2 has been posted.

recipes-test/images/initramfs-test-image.bb
```diff
@@ -14,14 +17,18 @@ PACKAGE_INSTALL += " \
     e2fsprogs-tune2fs \
     ethtool \
     gptfdisk \
+    iozone3 \
     iw \
+    kexec \
```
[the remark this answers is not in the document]
bhupesh-sharma: Ok, new version v2 has been posted.

recipes-test/images/initramfs-test-image.bb
bhupesh-sharma: Ok, new version v2 has been posted.

recipes-test/images/initramfs-test-image.bb
```diff
@@ -17,10 +17,16 @@ PACKAGE_INSTALL += " \
     iw \
     lava-test-shell \
     libdrm-tests \
+    lsof \
+    ncurses \
+    ncurses-terminfo \
```
lumag, line 22: do you need full terminfo?

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-image.bbappend
```diff
@@ -1,4 +1,14 @@
...
+    crash \
     cryptsetup \
     devmem2 \
+    iozone3 \
+    kexec \
+    libgpiod \
```
lumag, line 8: I think libgpiod would be picked up by libgpiod-tests, wouldn't it?

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-image.bbappend
```diff
@@ -1,4 +1,14 @@
...
+    libgpiod-tools \
+    makedumpfile \
+    mbw \
+    util-linux \
```
lumag, line 12: I thought you said you need only `chrt` util.

recipes-test/images/initramfs-test-image.bb
bhupesh-sharma: yes, cgps needs the same.

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-image.bbappend
bhupesh-sharma: qcom uses some user-sapce apps based on libgpiod for gps enablement. Idea is to let them run the same, if they cross compile and add it to the initramfs_xxx.cpio.gz

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-image.bbappend
bhupesh-sharma: Strange that the pull request doesn't show the updated commit log :(
I added the details there:
-> From 'meta-oe' layer:
    - lsof: list open files.
    - ncurses: for enabling cgps like tools on target.
    - rt-test: for 'lmbench', etc.
    - stress-ng: to load and stress a system.

-> From 'openembedded-core' layer:
    - crash: for analyzing a vmcore obtained from crash'ed kernel.
    - libgpiod-tools / libgpiod: for gpio set/get, etc.
    - mbw: for benchmarking.
    - util-linux-chrt: for 'chrt'

bhupesh-sharma: util-linux: for 'gdisk', 'taskset', etc.

bhupesh-sharma: May be I need to update the commit log again :(

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-image.bbappend
ndechesne: unless github is playing with me, I still see the same issue in v2.

recipes-test/images/initramfs-test-image.bb
```diff
@@ -17,10 +17,16 @@ PACKAGE_INSTALL += " \
...
+    ncurses-terminfo-base \
     pciutils \
     pd-mapper \
     qrtr \
     rmtfs \
+    rt-tests \
```
ndechesne, line 28: do you really need rt-tests, and not just lmbench? I don't even see how lmbench is pulled into the image

recipes-test/images/initramfs-test-image.bb
lumag: Interesting. I'm using cgps with just ncurses-terminfo-base, no need for ncurses of full ncurses-terminfo.

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-image.bbappend
lumag: I'd prefer to see individual packages instead of full `util-linux` suite here. Not to mention that `gdisk` comes from `gptfdisk` package, rather than `util-linux`.

recipes-test/images/initramfs-test-image.bb
bhupesh-sharma: Yes, they are needed for cyclictest as well, which are used for running low-latency / rt checks.

recipes-test/images/initramfs-test-image.bb
bhupesh-sharma: Strange. May be you are using some other packages added to your receipe as well.

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-image.bbappend
bhupesh-sharma: There are several of those. It would be easier to add the full package rather than individual ones and then realize that some default busybox utility doesn't work well whereas the 'util-linux' one does (I have already landed into such situations before). Anyways the overall size increment is not too much for the whole 'util-linux' package.

lumag: My opinion would be that we should still keep it small. Getting test utils is fine. Getting individual software packages too. But getting a bunch of software 'just in case' generally is not from my point of view.

ndechesne:
we've discussed with @bhupesh-sharma , and agreed to introduce initramfs-test-full-image with 'more' utilities. So that we have:
1. initramfs-tiny
2. initramfs-test
3. initramfs-test-full 
Where we incrementally add more packages, so that users can pick the appropriate image they need.

recipes-test/images/initramfs-test-image.bb
bhupesh-sharma: As @ndechesne mentioned, I have pushed 'initramfs-test-full' related changes in the latest PR version.

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-full-image.bbappend
```diff
@@ -0,0 +1,14 @@
+PACKAGE_INSTALL += " \
+    coreutils \
```
ndechesne, line 2: coreutils is in OE-core, it should not be in the bbapend file. only packages which are built from recipes from meta-oe should be here.

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-full-image.bbappend
```diff
@@ -0,0 +1,14 @@
...
+    crash \
+    cryptsetup \
+    devmem2 \
+    iozone3 \
+    kexec \
```
ndechesne, line 7: kexec is OE-core, no?

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-full-image.bbappend
```diff
@@ -0,0 +1,14 @@
...
+    libgpiod \
+    libgpiod-tools \
+    makedumpfile \
+    mbw \
```
ndechesne, line 11: mbw is in master only, so it will fail when backporting in dunfell. we need the recipe in meta-backports (https://git.linaro.org/openembedded/meta-backports.git/). so the backport won't be trivial.

dynamic-layers/openembedded-layer/recipes-test/images/initramfs-test-full-image.bbappend
```diff
@@ -0,0 +1,14 @@
...
+    util-linux \
```
ndechesne, line 12: oe-core, not meta-oe

recipes-test/images/initramfs-test-image.bb
lumag: Nope. Please recheck w/o ncurses-terminfo. TERM=linux is included in the ncurses-terminfo-base.

bhupesh-sharma: I tried it earlier, but it did not work for me. Hmm.. Let me try again But anyways we have these in the 'initramfs-test-full-image.bb ' instead of 'initramfs-test-image.bb', so increase size shouldn't be an issue.

ndechesne:
I am going to merge as is, since we really need it merged now. we will do further improvements later (ncurses and some cleanup such as building on top of initramfs-test instead of tiny).

PR #308, opened 2021-06-02
lt9611uxc-upgrade script: Only cat firmware version when it exists

olivierschonken:
Update script to only execute cat in directories where lt9611uxc_firmware
exists.

On my system both module and 5-002b are symbolic links, thus the previous
script would cause the lt9611uxc firmware update systemd unit to fail because
the script would try to access lt9611uxc_firmware in the module directory and
return a failure which would cause the systemd service to be listed under
'systemctl list-units --failed'

Signed-off-by: Olivier Schonken <olivier.schonken@gmail.com>


lumag:
Reviewed-by: Dmitry Baryshkov <dmitry.baryshkov@linaro.org>

PR #309, opened 2021-06-02
Fixes for dsp partition mount

lumag:


recipes-support/fastrpc/fastrpc/mount-dsp.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+modprobe socinfo || true
+
```
bryanodonoghue, line 6: Tested-by: Bryan O'Donoghue <bryan.odonoghue@linaro.org>

recipes-support/fastrpc/fastrpc/mount-dsp.sh
```diff
@@ -13,7 +13,14 @@ if [ -r /sys/devices/soc0/machine ] ; then
 	esac
 fi
 
-if [ -z "$WHAT" -o ! -r "$WHAT"] ; then
+if [ -z "$WHAT" -o ! -r "$WHAT" ] ; then
```
ndechesne, line 16: do we really need to keep this 'corner' case? when is that really useful?

lumag: Which corner case? No dspso image? Back then we added it for all the boards which do not have dspso.bin file for whatever reason. So I think it is still useful.

lumag:
> Shoudln't this be SM8250|QRB5160|QRB5165 ?

There is no QRB5160 as far as I remember. it was purely my mistake.

recipes-support/fastrpc/fastrpc/mount-dsp.sh
ndechesne: but now dspso.bin is part of the firmware release. so when are we supposed to fallback?

lumag: On rb3/db410c/etc

ndechesne: lol.. sorry.. i forgot about these other boards :)

PR #314, opened 2021-07-09
qrb5165-rb5: drop lt9611uxc-upgrade

lumag:
Stop installing lt9611uxc-upgrade package, it was found to be confusing
to users. It will be replaced by the notice in the release notes.


ndechesne:
will you remove the recipe as well?

lumag:
> will you remove the recipe as well?

done in both pull requests

PR #321, opened 2021-07-22
conf/machine: Set default SERIAL_CONSOLES in qcom-common.inc

alimon:


ndechesne:
it would be more interesting to explain *why* you are making this change in the commit, rather than telling what you are doing, since we can see that in the patch. so why do we need this change?

alimon:
> it would be more interesting to explain _why_ you are making this change in the commit, rather than telling what you are doing, since we can see that in the patch. so why do we need this change?

@ndechesne I updated the commit msg.

PR #324, opened 2021-07-27
linux-linaro-qcom: update LOCALVERSION

ndechesne:
Fixes: 7829b0f0f691 (linux-linaro-qcomlt: move common settings to the
include file)

In this commit, some common config were moved into the shared include
file, however LOCALVERSION was not updated based on its value in the
kernel recipe. Restore LOCALVERSION to what it was before this patch.


ndechesne:
run tests

ndechesne:
run tests

PR #327, opened 2021-07-28
Kernel recipe structure changes and Linux 5.13 for dunfell

alimon:


ndechesne:
hmm. are you sure this is correct? we don't have 7829b0f0f691 in dunfell branch. so your patch 2 here does not look right. or we need to backport 7829b0f0f691 too. can you please check?

alimon:
@ndechesne right we don't have 7829b0f in the branch, i can cherry-pick too.

ndechesne:
well, with 675a8e5181b9d89f551220115cb011c96d44e8d9, i don't think it can work, so yes we need it

alimon:
@ndechesne I picked all related changes from @lumag to have the same structure changes that are in master.

ndechesne:
thanks! I've pushed ede2390278bf (linux-linaro-qcom: update LOCALVERSION) as well, since it was missing from this PR. Looks like we are aligned now.

PR #232, opened 2020-12-21
Support bundling initramfs into boot image

lumag:
This builds on top of #193, adding support for creation of kernel+initramfs bundles.

PR #330, opened 2021-08-02
Boot image with the bundled initramfs support

lumag:
A version of #232 with `QCOM_BOOTIMG_BUNDLE_INITRAMFS` removed.

### DIFF
--- a/recipes-kernel/linux/linux-qcom-bootimg.inc
+++ b/recipes-kernel/linux/linux-qcom-bootimg.inc
@@ -1,5 +1,17 @@
 QIMG_DEPLOYDIR = "${WORKDIR}/qcom_deploy-${PN}"
 
+# Define INITRAMFS_IMAGE to create kernel+initramfs Android boot images in
+# addition to default boot images. For example add the following line to your
+# conf/local.conf:
+#
+# INITRAMFS_IMAGE = "initramfs-kerneltest-image"
+#
+
+python __anonymous () {
+    if d.getVar('INITRAMFS_IMAGE') != '':
+        d.appendVarFlag('do_qcom_img_deploy', 'depends', ' ${INITRAMFS_IMAGE}:do_image_complete')
+}
+
 python do_qcom_img_deploy() {
     import shutil
     import subprocess
@@ -11,6 +23,17 @@ python do_qcom_img_deploy() {
     else:
         qcom_deploy_dir = d.getVar("QIMG_DEPLOYDIR")
         image_dir = d.getVar("DEPLOY_DIR_IMAGE")
+
+    initrd = None
+    if d.getVar('INITRAMFS_IMAGE') != '':
+        initrd_image_name = d.getVar("INITRAMFS_IMAGE_NAME")
+        baseinitrd = os.path.join(d.getVar("DEPLOY_DIR_IMAGE"), initrd_image_name)
+        for img in (".cpio.gz", ".cpio.lz4", ".cpio.lzo", ".cpio.lzma", ".cpio.xz", ".cpio"):
+            if os.path.exists(baseinitrd + img):
+                initrd = baseinitrd + img
+                break
+        if not initrd:
+            bb.fatal("Could not find initramfs image %s for bundling" % d.getVar("INITRAMFS_IMAGE"))
 
     B = d.getVar("B")
     D = d.getVar("D")
@@ -47,16 +70,29 @@ python do_qcom_img_deploy() {
         def getVarDTB(name):
             return d.getVarFlag(name, dtb_name) or d.getVar(name)
 
-        def make_image(template, rootfs):
-            output = os.path.join(qcom_deploy_dir, template % (dtb_name, kernel_image_name))
-            output_link =  os.path.join(qcom_deploy_dir, template % (dtb_name, kernel_link_name))
+        def make_image_internal(output, output_link, rootfs, initrd = definitrd):
             subprocess.check_call([mkbootimg,
                 "--kernel", kernel,
-                "--ramdisk", definitrd,
+                "--ramdisk", initrd,
                 "--output", output,
                 "--pagesize", getVarDTB("QCOM_BOOTIMG_PAGE_SIZE"),
                 "--base", getVarDTB("QCOM_BOOTIMG_KERNEL_BASE"),
                 "--cmdline", "root=%s rw rootwait %s %s" % (rootfs, consoles, getVarDTB("KERNEL_CMDLINE_EXTRA") or "")])
+            if os.path.exists(output_link):
+                os.unlink(output_link)
+            os.symlink(os.path.basename(output), output_link)
+
+        def make_image(template, rootfs):
+            output = os.path.join(qcom_deploy_dir, template % (dtb_name, kernel_image_name))
+            output_link =  os.path.join(qcom_deploy_dir, template % (dtb_name, kernel_link_name))
+            make_image_internal(output, output_link, rootfs)
+            return output
+
+        def make_initramfs_image(template, rootfs, initrd, initrd_image_name):
+            output = os.path.join(qcom_deploy_dir, template % (initrd_image_name, dtb_name, kernel_image_name))
+            output_link =  os.path.join(qcom_deploy_dir, template % (initrd_image_name, dtb_name, kernel_link_name))
+            make_image_internal(output, output_link, rootfs, initrd)
+            output_link =  os.path.join(qcom_deploy_dir, template % ("initramfs", dtb_name, kernel_link_name))
             if os.path.exists(output_link):
                 os.unlink(output_link)
             os.symlink(os.path.basename(output), output_link)
@@ -79,11 +115,17 @@ python do_qcom_img_deploy() {
         if not os.path.exists(output_img):
             os.symlink(os.path.basename(output), output_img)
 
+        if initrd:
+            make_initramfs_image("boot-%s-%s-%s.img", rootfs, initrd, d.getVar("INITRAMFS_IMAGE"))
+
         sd_rootfs = getVarDTB("SD_QCOM_BOOTIMG_ROOTFS")
         if sd_rootfs:
             output = make_image("boot-sd-%s-%s.img", sd_rootfs)
             if not os.path.exists(output_sd_img):
                 os.symlink(os.path.basename(output), output_sd_img)
+
+            if initrd:
+                make_initramfs_image("boot-sd-%s-%s-%s.img", rootfs, initrd, d.getVar("INITRAMFS_IMAGE"))
 }
 
 do_qcom_img_deploy[depends] += "skales-native:do_populate_sysroot"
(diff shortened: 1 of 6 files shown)


bhupesh-sharma:
would it possible to add some comments / documentation on how to use this option in your PR?

bhupesh-sharma (commented):
We will need to add initramfs-test-full-image.bb to these as well.

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -79,11 +101,17 @@ python do_qcom_img_deploy() {
         if not os.path.exists(output_img):
             os.symlink(os.path.basename(output), output_img)
 
+        if initrd:
+            make_image("boot-initramfs-%s-%s.img", rootfs, initrd)
```
ndechesne, line 105: we probably want to have the INITRAMFS_IMAGE in the filename that we build, to avoid overriding the boot image when building multiple of them.

lumag: You can have only a single image in the `INITRAMFS_IMAGE`. Thus you can not build 'multiple of them'.

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -1,5 +1,17 @@
 QIMG_DEPLOYDIR = "${WORKDIR}/qcom_deploy-${PN}"
 
+# Define INITRAMFS_IMAGE to create kernel+initramfs Android boot images in
+# addition to default boot images. For example add the following line to your
+# conf/local.conf:
```
lumag, line 5: @bhupesh-sharma here is the documentation (in the source code). Just define the `INITRAMFS_IMAGE`.

recipes-kernel/linux/linux-qcom-bootimg.inc
ndechesne: you can build several of them calling bitbake multiple times, e.g. changing the value INITRAMFS_IMAGE in between. it's a valid scenario.  for example we will want to build initram-kerneltest-image and initramfs-kerneltest-full-image very soon.

lumag:
> We will need to add initramfs-test-full-image.bb to these as well.

I've changed `initramfs-test-image` in a way to disable usage of bbappends.

recipes-kernel/linux/linux-qcom-bootimg.inc
lumag: @ndechesne the the images from the previous release would be removed from the deploy dir during preparation stage (as stale objects). Or are you thinking about boot images stashed somewhere?

ndechesne: we will indeed publish the images on linaro.org. 
and i think it makes sense to identify what we built by filename, using one or another initramfs will really make a very different boot-initramfs-xxx.img file.

ndechesne: well actually.. it is going to make things a bit more complicated in our release build scripts.. since we will need to run the build in 2 steps now, with and without INITRAMFS_IMAGE..

lumag: I see... I'll extend the code to include initramfs image name.
Actually no. The image w/o initramfs is still generated in the same pass. See the build script (or the output dir).

lumag: Done

recipes-test/images/initramfs-test-image.bb
```diff
@@ -24,3 +24,11 @@ PACKAGE_INSTALL += " \
...
     wpa-supplicant \
 "
+
+# We'd like to include extra packages provided by layers which we do not depend
+# on. This can be handled by .bbappends, but then image recipes including this
+# one would not get all these tools. So simulate dynamic bbappend here.
+#
+# Unfortunately BBFILES_DYNAMIC isn't honored by include directive, so include
+# files have to be put into the main layer's dir.
+include ${@ " ".join(map(lambda layer: "recipes-test/images/initramfs-test-image-" + layer + ".inc", d.getVar('BBFILE_COLLECTIONS').split()))}
```
ndechesne, line 34: this is a complicated hack... how about we do something simpler:

PACKAGE_INSTALL += " \
    ${@bb.utils.contains("BBFILE_COLLECTIONS", "meta-oe", "XXX YYY ZZZ", "", d)} \
    ${@bb.utils.contains("BBFILE_COLLECTIONS", "meta-networking", "XXX YYY ZZZ", "", d)} \
"

it is also a 're-implementation' of BBFILES_DYNAMIC, but isn't that simpler?

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -32,9 +75,9 @@ python do_qcom_img_deploy() {
     else:
         bb.fatal("Unuspported ARCH %s" % arch)
 
-    if os.path.exists(output_img):
+    if os.path.lexists(output_img):
```
ndechesne, line 78: why this change? it might be a good change, but it's not related to $subject?

recipes-kernel/linux/linux-qcom-bootimg.inc
bhupesh-sharma: As @ndechesne mentioned on irc, just setting it on the command-line is not working. Have you tested that use-case?

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -24,6 +46,27 @@ python do_qcom_img_deploy() {
     output_img =  os.path.join(qcom_deploy_dir, "boot-%s.img" % (kernel_link_name))
     output_sd_img =  os.path.join(qcom_deploy_dir, "boot-sd-%s.img" % (kernel_link_name))
 
+    def make_image(template, dtb_name, rootfs, initrd=definitrd):
+        baseinitrd = os.path.basename(initrd)
+        output = os.path.join(qcom_deploy_dir, template % {'dtb': dtb_name, 'kernel': kernel_image_name, 'initrd': baseinitrd})
+        output_link = os.path.join(qcom_deploy_dir, template % {'dtb': dtb_name, 'kernel': kernel_link_name, 'initrd': baseinitrd})
+        output_initramfs_link = os.path.join(qcom_deploy_dir, template % {'dtb': dtb_name, 'kernel': kernel_link_name, 'initrd': 'initramfs'})
```
ndechesne, line 53: why that?

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -24,6 +46,27 @@ python do_qcom_img_deploy() {
     output_img =  os.path.join(qcom_deploy_dir, "boot-%s.img" % (kernel_link_name))
     output_sd_img =  os.path.join(qcom_deploy_dir, "boot-sd-%s.img" % (kernel_link_name))
 
+    def make_image(template, dtb_name, rootfs, initrd=definitrd):
+        baseinitrd = os.path.basename(initrd)
+        output = os.path.join(qcom_deploy_dir, template % {'dtb': dtb_name, 'kernel': kernel_image_name, 'initrd': baseinitrd})
```
ndechesne, line 51: it is a bit confusing since initrd is not 'used' in the template for the regular image.. it wasn't obvious at first sight. i am finding this code less and less obvious to read (and to maintain).. not sure what to do.

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -47,21 +90,6 @@ python do_qcom_img_deploy() {
         def getVarDTB(name):
             return d.getVarFlag(name, dtb_name) or d.getVar(name)
 
-        def make_image(template, rootfs):
```
ndechesne, line 50: why did you move the function?

recipes-kernel/linux/linux-qcom-bootimg.inc
ndechesne: setting in .conf file works. it's setting on the command line which does not work.

bhupesh-sharma: Yep, I tried it with .conf as well, but the query is whether that is the expected way. If yes, probably we can better document it.

lumag: Of course it is. Typically you never ever define variables in the shell environment. In fact the environment is sanitized before being fed into called programs. You have to whiltelist the variable before bitbake would accept it.

recipes-test/images/initramfs-test-image.bb
lumag: Yep. I was just thinking about files. Will do it this way.

recipes-kernel/linux/linux-qcom-bootimg.inc
lumag: Ah. I was fighting with broken symlinks not being updated at some point. I'll drop this (or move to a separate patch).

recipes-kernel/linux/linux-qcom-bootimg.inc
lumag: To get `boot-initramfs-apq8016-sbc-dragonboard-410c.img` links in addition to `boot-initramfs-kerneltest-image-apq8016-sbc-dragonboard-410c.img`. Having explicit initramfs name is good, but I want also to have a simple/readable symlink.

recipes-kernel/linux/linux-qcom-bootimg.inc
lumag: Yes. I'm also a bit unhappy about it. Let me see if I can improve it.

recipes-kernel/linux/linux-qcom-bootimg.inc
lumag: a part of my symlinks fight. Will move it back.

recipes-kernel/linux/linux-qcom-bootimg.inc
lumag: @ndechesne cleaned things up a bit.

recipes-test/images/initramfs-test-image.bb
lumag: Done

ndechesne:
thanks for the clean up! It looks better now. I am going to merge this PR, since we need it right away.

However, I think we should discuss about what we have done here.. the more i look into it, the more I think something is wrong with the whole idea of building the boot image in the kernel recipe. It was convenient initially since the boot image only contains 'artifact' built by the kernel recipe, but as it gets more and more complicated, we should change that. 

we could move the boot image creation to a custom 'image' class , or probably even better investigate how we could use WIC since its purpose is exactly to build images from all the build artifacts, once all recipes are built.

We are building lots of different artifact:
* kernel and DTBs, modules
* images (ext4, ubifs, android sparse images which is not supported in OE now, cpio)
* boot images with different purpose: 1) to load kernel and jump to rootfs, 2) boot image with initramfs for testing (and no rootfs), 3) hopefully soon boot images with initramfs which pivot into an actual rootfs...

lumag:
@ndechesne point noted. I will think whether we can improve this by shifting into initramfs recipes.

The major problem is that if one would like to build a kernel image w/o the initramfs, he'd still call bitbake virtual/kernel rather than some bitbake initramfs-dummy. Thus we'd probably end up with some code duplication.

PR #344, opened 2021-08-24
gpsd-machine-conf: provide gpsd config for qcom machines

lumag:
Provide gpsd config adding pds:// devices.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-navigation/gpsd/gpsd-machine-conf/gpsd.pds
+++ b/dynamic-layers/openembedded-layer/recipes-navigation/gpsd/gpsd-machine-conf/gpsd.pds
@@ -1,0 +1,5 @@
+START_DAEMON="true"
+GPSD_OPTIONS=""
+DEVICES="pds://0"
+USBAUTO="true"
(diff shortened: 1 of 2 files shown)


ndechesne:
I think you are overriding the default gpsd config file here, and you are doing it for *all* machines, not just 410c, not just meta-qcom machines, right? that does not seem right to me.

lumag:
No, it is overridden for all qcom machines (see `do_install:append:qcom`).

lumag:
@ndechesne @andersson  related question: should there be a PDS service on db845c? I don't see one on my board.

ndechesne:
ok, not *all* machines.. but at least all 'qcom' machines.. not sure it's great. For example, on the ADP SA8155P we know that we want to support an external GPS.. can we not support multiple config? or perhaps just provide the config file, and not override the default to make it simpler for users who want to use PDS://?

lumag:
@ndechesne the default is an empty config. So we can probably drop this. In fact we should probably check why 'gpsdctl add pds://0' might result in an error rather than in addition of the device.

alimon:
@lumag Also we can use, 

$ gpsdctl add pds://any

Should work by default getting the frist device available.

PR #349, opened 2021-09-07
Sa8155p dunfell

ndechesne:


ndechesne:
@bhupesh-sharma @lumag , so this is what I ended with for the dunfell backports.. i think it has everything we need for SA8155p-adp and for the initramfs bundling.. there are some quirks in some backports.. please review.

ndechesne:
I file https://lists.openembedded.org/g/openembedded-core/message/155767 in the process.

PR #354, opened 2021-09-08
layer.conf: set INITRAMFS_IMAGE for sa8155p machine

bhupesh-sharma:
Set INITRAMFS_IMAGE as "initramfs-kerneltest-full-image"
for sa8155p machine .

### DIFF
--- a/conf/machine/sa8155p-adp.conf
+++ b/conf/machine/sa8155p-adp.conf
@@ -4,6 +4,9 @@
 
 require conf/machine/include/qcom-sa8155p.inc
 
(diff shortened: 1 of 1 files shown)


alimon:
@bhupesh-sharma LGTM.

ndechesne:
AS per @lumag , this should be in sa8155p-adp.conf instead. In fact , in its current form, it doesn't even set it correctly (since we have changed the override syntax in master, e.g. we should use : instead of _ now. But regardless, it should be in the machine conf.

PR #357, opened 2021-09-11
Add bootrr to meta qcom

bhupesh-sharma:
This is a RFC patch which allows adding bootrr to the meta-qcom layer as a receipes-test component.
To do:
- I am using my github tree as the SRC_UI, instead of Bjorn's official tree as PR raised against the bootrr tree is still under review <https://github.com/andersson/bootrr/pull/38>
- SRCBRANCH, etc should also be changed in the final version of this patch to point to the 'bootrr/master' branch.

I tested the resulting bootrr board helper files on sm8150-mtp and sa8155-adp boards with this change by running them as-

# qcom,sm8150-mtp

### DIFF
--- a/recipes-test/bootrr/bootrr_git.bb
+++ b/recipes-test/bootrr/bootrr_git.bb
@@ -1,0 +1,16 @@
+SUMMARY = "simple low-level testing tool for qcom boards"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=987293312a134ab40eec5f3d446cfaff"
+
+SRCREV = "d2329902b701e0efa345628471d0d275d5a5835a"
+SRC_URI = "\
+	git://github.com/andersson/bootrr.git \
+	file://bootrr-auto-switch-to-using-sh.patch \
+"
+
+S = "${WORKDIR}/git"
+
+do_install() {
+	oe_runmake install 'DESTDIR=${D}'
+}
--- a/recipes-test/bootrr/bootrr/bootrr-auto-switch-to-using-sh.patch
+++ b/recipes-test/bootrr/bootrr/bootrr-auto-switch-to-using-sh.patch
@@ -1,0 +1,22 @@
+From ef8073c29988c1cf46221e3062811f19e051b62d Mon Sep 17 00:00:00 2001
+From: Dmitry Baryshkov <dmitry.baryshkov@linaro.org>
+Date: Sat, 2 Oct 2021 16:36:47 +0300
+Subject: [PATCH] bootrr-auto: switch to using sh
+
+There is nothing bash-specific in bootrr-auto, we can safely use /bin/sh
+instead. Verified with dash and with busybox sh.
+
+Signed-off-by: Dmitry Baryshkov <dmitry.baryshkov@linaro.org>
+---
+ helpers/bootrr-auto | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/helpers/bootrr-auto b/helpers/bootrr-auto
+index 672c590..607be0a 100755
+--- a/helpers/bootrr-auto
++++ b/helpers/bootrr-auto
+@@ -1,4 +1,4 @@
+-#!/bin/bash
++#!/bin/sh
+ 
+ $(pwd)/helpers/bootrr-generic-tests
--- a/recipes-test/images/initramfs-test-full-image.bb
+++ b/recipes-test/images/initramfs-test-full-image.bb
@@ -4,6 +4,7 @@ DESCRIPTION = "Relatively larger ramdisk image for running tests (bootrr, etc)"
 
 PACKAGE_INSTALL += " \
     bluez5 \
+    bootrr \
     coreutils \
     dhcpcd \
     diag \


recipes-test/bootrr/bootrr_git.bb
```diff
@@ -0,0 +1,26 @@
+SUMMARY = "simple low-level testing tool for qcom boards"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=987293312a134ab40eec5f3d446cfaff"
+
+SRCREV = "999c82b3b006004ce0f05821358132cdab2ebee6"
+SRC_URI = "git://github.com/bhupesh-sharma/bootrr.git;protocol=https"
```
ndechesne, line 7: if we want to add a recipe for bootrr, we need to use upstream. however we decided in the past not to include bootrr into meta-qcom since it was typically used when doing kernel tests, outside of the OE build process. we can change that.. but that means keeping bootrr recipe up-to-date..

recipes-test/bootrr/bootrr_git.bb
```diff
@@ -0,0 +1,26 @@
+SUMMARY = "simple low-level testing tool for qcom boards"
+
+LICENSE = "BSD-3-Clause"
```
ndechesne, line 3: upstream does not have a LICENSE file, needs to be fixed first.

recipes-test/bootrr/bootrr_git.bb
```diff
@@ -0,0 +1,26 @@
...
+LIC_FILES_CHKSUM = "file://LICENSE;md5=987293312a134ab40eec5f3d446cfaff"
+
+SRCREV = "999c82b3b006004ce0f05821358132cdab2ebee6"
+SRC_URI = "git://github.com/bhupesh-sharma/bootrr.git;protocol=https"
+SRCBRANCH = "license-bootrr"
+SRC_URI[sha256sum] = "5035b9984326325ebb9c58e91bf474a6b4f0a334c5720c97263b01d9588cd9fd"
```
ndechesne, line 9: not needed when using git:// protocol

recipes-test/bootrr/bootrr_git.bb
```diff
@@ -0,0 +1,26 @@
...
+
+S = "${WORKDIR}/git"
+
+do_install() {
```
ndechesne, line 13: why not using the install target rule from the upstream Makefile?

recipes-test/bootrr/bootrr_git.bb
```diff
@@ -0,0 +1,26 @@
...
+		install -m 0755 $file ${D}${bindir}/$(basename $file)
+	done
+	
+	for file in $(ls helpers/*)
+	do
+		install -m 0755 $file ${D}${bindir}/$(basename $file)
+	done
+}
+
+RDEPENDS_${PN} = "bash"
```
ndechesne, line 26: is it true that we need bash?

recipes-test/bootrr/bootrr_git.bb
lumag: We can include default recipe + add support for the `devupstream` class, which would allow developers to use `AUTOREV` if devupstream is selected (like we did for the `linux-linaro-qcomlt-dev`).

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: Yes @ndechesne , similar to pm-qa (http://cgit.openembedded.org/meta-openembedded/tree/meta-oe/recipes-test/pm-qa/pm-qa_git.bb?h=master#n66), bootrr also needs 'bash'

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: Ok, will fix this in v2.

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: Ok, will try with the upstream Makefile rules by default and update the PR accordingly.

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: Right, I have opened a PR: https://github.com/andersson/bootrr/pull/38, for bootrr upstream (for adding the LICENSE file).

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: @ndechesne - ask per our last discussion, I think adding the 'bootrr' receipe to meta-qcom, makes for good test coverage on several qcom boards.

@lumag - Ok will check the 'linux-linaro-qcomlt-dev' implementation.

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: Since @andersson accepted the bootrr related PR, updated v2 sent out.

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: The github page is 'https://github.com/andersson/bootrr.git'. I tried without the above helpers, but the .bb bitbake failed.

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: I tried the upstream Makefile, but installation requires installing the boards/* and helpers/* file, into the /usr/bin directory, so we need a customized installation rule here.

ndechesne: I don't understand. The upstream Makefile can be configured to install anywhere. Am i missing somethign?

lumag: As we are packaging this, may I suggest pushing all data files to /usr/share and all helpers to /usr/libexec?

recipes-test/bootrr/bootrr_git.bb
```diff
@@ -0,0 +1,26 @@
...
+SRCREV = "d2329902b701e0efa345628471d0d275d5a5835a"
+SRC_URI = "git://github.com/andersson/bootrr.git;protocol=https"
+SRCBRANCH = "master"
+SRC_URI[sha256sum] = "d372b56c5709ad3c76a8225e6bfd17fb283a34c4823891d8f16f88da3a34d985"
+
+S = "${WORKDIR}/git"
+
+do_install() {
+        mkdir -p ${D}${bindir}
+	for file in $(ls boards/*)
```
lumag, line 15:
```sh
for file in boards/*
```
no need to complicate this with the subshell and ls.

recipes-test/bootrr/bootrr_git.bb
```diff
@@ -0,0 +1,26 @@
...
+	do
+		install -m 0755 $file ${D}${bindir}/$(basename $file)
+	done
```
lumag, line 18: Or even simpler:
```sh
install -m 0755 boards/* ${D}${bindir}/
```

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: > As we are packaging this, may I suggest pushing all data files to /usr/share and all helpers to /usr/libexec?

No, that would be different from how bootrr Makefile packages stuff. Let's use the default Makefile configuration rules as @ndechesne suggested.

recipes-test/bootrr/bootrr_git.bb
```diff
@@ -0,0 +1,17 @@
+SUMMARY = "simple low-level testing tool for qcom boards"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=987293312a134ab40eec5f3d446cfaff"
+
+SRCREV = "d2329902b701e0efa345628471d0d275d5a5835a"
+SRC_URI = "git://github.com/andersson/bootrr.git;protocol=https"
+SRCBRANCH = "master"
+SRC_URI[sha256sum] = "d372b56c5709ad3c76a8225e6bfd17fb283a34c4823891d8f16f88da3a34d985"
```
lumag, line 9: SRCURI should not be necessary. Also SRCBRANCH can become a `;branch=master` part of SRC_URI

lumag: Moreover IIRC master is a default branch so it should not be necessary at all. 
Also I checked the oe-core layer. Out of 70 recipes just 17 have 'protocol=https', so we should be able to drop that too.

recipes-test/bootrr/bootrr_git.bb
lumag: @andersson @ndechesne I meant something like https://github.com/andersson/bootrr/pull/39

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: Fixed in new version. We can use the default bootrr Makefile rules directly.

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: > As we are packaging this, may I suggest pushing all data files to /usr/share and all helpers to /usr/libexec?

No, that would be different from how bootrr Makefile packages stuff. Let's use the default Makefile configuration rules as @ndechesne suggested.

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: Fixed in new version. We can use the default bootrr Makefile rules directly.

recipes-test/bootrr/bootrr_git.bb
bhupesh-sharma: Right, but I see a error if SRC_URI[sha256sum] is not defined.

bhupesh-sharma: Posted a new version which fixes these cosmetic changes as well.

recipes-test/bootrr/bootrr_git.bb
```diff
@@ -0,0 +1,15 @@
...
+SRCREV = "d2329902b701e0efa345628471d0d275d5a5835a"
+SRC_URI = "git://github.com/andersson/bootrr.git"
+
+S = "${WORKDIR}/git"
+
+do_install() {
+	oe_runmake install 'DESTDIR=${D}'
+}
+
+RDEPENDS:${PN} += "bash"
```
ndechesne, line 15: are you sure about this one? we've been bootrr in our CI without bash, i am pretty sure.

bhupesh-sharma: Strange as I get :
ERROR: bootrr-git-r0 do_package_qa: QA Issue: /usr/bin/bootrr-auto contained in package bootrr requires /bin/bash, but no providers found in RDEPENDS:bootrr? [file-rdeps]

without this RDEPENDS on 'bash'.

bhupesh-sharma: Hi @lumag , can you please post a formal PR for bootrr with andersson/bootrr@ef8073c. I can ack the same. Thanks

bhupesh-sharma: Done. Have sent a revised version, with  andersson/bootrr@ef8073c, for the rest of the changes (altering the layouts of the final executables), I think lets wait for @andersson 's comemnts.

PR #364, opened 2021-09-25
Few fixes for db820c support

lumag:


recipes-kernel/linux-firmware/linux-firmware_%.bbappend
```diff
@@ -2,3 +2,6 @@ inherit update-alternatives
 
 ALTERNATIVE:${PN}-ath11k = "qca6390-board2"
 ALTERNATIVE_LINK_NAME[qca6390-board2] = "/lib/firmware/ath11k/QCA6390/hw2.0/board-2.bin"
+
+ALTERNATIVE:${PN}-ath10k = "qca6174-board2"
```
ndechesne, line 6: do we really need this ALTERNATIVE variables here (for both 820 and 845)? this recipe would break yocto-check-layer. you added it here, just in case linux-firmware (upstream) starts packaging this file, right? should we not skip that and let it fail if/when usptream is updated?

lumag: Upstream provides `board-2.bin`s which do not contain necessary board data (for db820c and for rb5). Thus we are providing overrides. When I was first implementing this I was trying to choose between `ALTERNATIVE` and `RREPLACES`. I choose former so that users of the firmware package can switch to the base version.

This differs from the db845c support, where board data did not exist at all, so we are providing the file w/o any overrides.

I will try asking @kvalo again if we can get board data for our boards included into linux-firmware

ndechesne: ok, i will merge anyways, since it's not a big issue. at some point I would like to remain compliant with yocto-check-layer.. hopefully sometimes..

PR #365, opened 2021-09-29
initramfs-test-full: Add 'fastrpc' and 'remoteproc' capabilities

bhupesh-sharma:
Add 'fastrpc' and 'remoteproc' support in
'initramfs-test-full-image.bb', which provides
support for aDSP, cDSP and sDSP FW loading
and systemd service spawing for the same.

### DIFF
--- a/recipes-test/images/initramfs-test-full-image.bb
+++ b/recipes-test/images/initramfs-test-full-image.bb
@@ -13,6 +13,8 @@ PACKAGE_INSTALL += " \
     e2fsprogs-resize2fs \
     e2fsprogs-tune2fs \
     ethtool \
+    fastrpc \
+    firmware-qcom-${MACHINE} \
     gptfdisk \
     iw \
     hdparm \


recipes-test/images/initramfs-test-full-image.bb
```diff
@@ -13,6 +13,8 @@ PACKAGE_INSTALL += " \
     e2fsprogs-resize2fs \
     e2fsprogs-tune2fs \
     ethtool \
+    fastrpc \
+    firmware-qcom-${MACHINE} \
```
lumag, line 17: NAK, this can break too easily.
Also not that for some machines *DSP firmware is provided by linux-firmware rather than by the board firmware package.

Also note that there is a separate `fastrpc-systemd` containing unit files to start *dsprpc daemons.

bhupesh-sharma: Sure, may be I will customize this only for MACHINE=sa8155p and resend, as there are no plans currently to add these FW files to linux-firmware.

ndechesne: don't we usually include the firmware packages in the $MACHINE .conf file instead? why should it be different here? check MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS in other $MACHINE.

bhupesh-sharma: @ndechesne , yes that's I meant by keeping it only machine specific for sa8155p (in reply to @lumag 's comments). I will send the changed version soon as v2.

lumag:
With `fastrpc` being part of `initramfs-test-image` I think we can close this pull-request. @bhupesh-sharma ?

lumag:
Closing. fastrpc support has been merged in and the firmware doesn't belong to this image.

PR #367, opened 2021-09-30
Drop support for db600c/sd-600eval

lumag:
Support for these boards was dropped from the mainline kernel at 5.0 (at 2018) with the following message:

> This patch removes support for the APQ8064 based Arrow SD600 eval
    board.  This board was never sold publicly and had very limited
    distribution.  As such, we are removing this board and no longer
    going to support it.
    
  Drop support for the boards from meta-qcom too.


lumag:
Note: support for IFC6410 is still in place. The board does not use the firmware-qcom-sd-600eval recipe, it mounts firmware partition instead.

PR #371, opened 2021-10-17
Split huge firmware packages

lumag:
With the inclusion of the dspso.bin images, board firmware packages are becoming quite huge. Split them into a pile of smaller packages, allowing one to control precisely which files go into the rootfs (esp. initramfs).


recipes-bsp/firmware/firmware-qcom.inc
```diff
@@ -4,6 +4,45 @@ FILES:${PN} += "${nonarch_base_libdir}/firmware/"
...
+    for pkg in d.getVar("SPLIT_FIRMWARE_PACKAGES").split():
+        # Depend on the main package to get the license file
+        d.appendVar("RDEPENDS:" + pkg, " " + pn)
+        # and append the INSANE_SKIP of the main package to pass QA
+        d.appendVar("INSANE_SKIP:" + pkg, " " + insanes)
+        # If it's a legacy package, depend on non-legacy files
+        if pkg.endswith("-legacy"):
+            d.appendVar("RDEPENDS:" + pkg, " " + pkg[:-7])
+    if d.getVar("FW_QCOM_SOC") == "unset" and d.getVar("SPLIT_FIRMWARE_PACKAGES") != "":
+        bb.warn("%s: split firmware-qcom packages engaged, but FW_QCOM_SOC is not defined" % pn)
```
ndechesne, line 43: it's an error if FW_QCOM_SOC is not set, not just a warning?

recipes-bsp/firmware/firmware-qcom.inc
```diff
@@ -4,6 +4,45 @@ FILES:${PN} += "${nonarch_base_libdir}/firmware/"
...
+SPLIT_FIRMWARE_PACKAGES ?= ""
+
+PACKAGE_BEFORE_PN += "${SPLIT_FIRMWARE_PACKAGES}"
+# RRECOMMEND all non-legacy packages split from this recipe
+RRECOMMENDS:${PN} += "${@ ' '.join(filter(lambda p: not p.endswith('-legacy'), d.getVar('SPLIT_FIRMWARE_PACKAGES').split())) }"
+
+# Default settings for several split packages
+FILES:${PN}-dspso += "${FW_QCOM_PATH}/*/*dspso.bin"
+FILES:linux-firmware-qcom-${FW_QCOM_SOC}-audio = "${FW_QCOM_SOC_PATH}/adsp.mbn ${FW_QCOM_SOC_PATH}/adsp*.jsn"
+FILES:linux-firmware-qcom-${FW_QCOM_SOC}-audio-legacy = "${FW_QCOM_SOC_PATH}/adsp.mdt ${FW_QCOM_SOC_PATH}/adsp.b*"
```
ndechesne, line 21: not sure if the suffix 'legacy' is appropriate. maybe 'split' (as opposed to squashed)?

recipes-bsp/firmware/firmware-qcom-dragonboard410c.inc
```diff
@@ -32,3 +34,10 @@ do_install() {
 }
 
 FILES:${PN} += "/boot/modem_fsg"
+
+SPLIT_FIRMWARE_PACKAGES = " \
+    linux-firmware-qcom-${FW_QCOM_SOC}-modem \
```
ndechesne, line 39: the main recipe is 'board' dependent e.g. firmware-qcom-dragonboard410c, but you create sub packages which are 'SOC' specific. How would you support 2 boards with the same SoC which have different firmware? e.g. RB3 and the Lenovo C630?

ndechesne: It's much better without 'SOC' in the variable name. You will have a problem with the '/' though, since FW_QCOM_SOC will need to convert / into - maybe.. but that's a problem for later i suppose.. 

I will run some more tests, but it looks like something which I should merge now.

PR #387, opened 2021-11-12
initramfs-rootfs-image: add initramfs that pivots into rootfs

lumag:
Add small initramfs image that will detect, mount and switch into
specified rootfs. At this moment this recipe does not pull any kernel
modules on its own, users of the initramfs have to do that on their own.

### DIFF
--- a/recipes-kernel/images/initramfs-qcom-image.bb
+++ b/recipes-kernel/images/initramfs-qcom-image.bb
@@ -1,0 +1,5 @@
+require initramfs-rootfs-image.bb
+
+DESCRIPTION = "Ramdisk image for pivoting into rootfs extended to boot Qualcomm boards"
+
+PACKAGE_INSTALL += "packagegroup-qcom-boot"
--- a/recipes-kernel/linux/linux-qcom-bootimg.inc
+++ b/recipes-kernel/linux/linux-qcom-bootimg.inc
@@ -144,3 +144,6 @@ addtask do_qcom_img_deploy_setscene
 do_qcom_img_deploy[dirs] = "${QIMG_DEPLOYDIR} ${B}"
 do_qcom_img_deploy[cleandirs] = "${QIMG_DEPLOYDIR}"
 do_qcom_img_deploy[stamp-extra-info] = "${MACHINE_ARCH}"
+
+# We do not need kernel image in /boot, these images are flashed into separate partition.
+RDEPENDS:${KERNEL_PACKAGE_NAME}-base = ""
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -17,4 +17,6 @@ BBFILES_DYNAMIC += " \
     openembedded-layer:${LAYERDIR}/dynamic-layers/openembedded-layer/*/*/*.bbappend \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bb \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bbappend \
+    meta-linux-mainline:${LAYERDIR}/dynamic-layers/meta-linux-mainline/*/*/*.bb \
+    meta-linux-mainline:${LAYERDIR}/dynamic-layers/meta-linux-mainline/*/*/*.bbappend \
 "
--- a/conf/machine/qcom-armv8a.conf
+++ b/conf/machine/qcom-armv8a.conf
@@ -9,7 +9,7 @@ EXTRA_IMAGECMD:ext4 += " -b 4096 "
 # Support for dragonboard{410, 820, 845}c, rb5
 KERNEL_IMAGETYPE ?= "Image.gz"
 SERIAL_CONSOLE ?= "115200 ttyMSM0"
-KERNEL_DEVICETREE ?= "qcom/apq8016-sbc.dtb qcom/apq8096-db820c.dtb qcom/sdm845-db845c.dtb qcom/qrb5165-rb5.dtb qcom/sm8250-rb5-dvt.dtb"
+KERNEL_DEVICETREE ?= "qcom/apq8016-sbc.dtb qcom/apq8096-db820c.dtb qcom/sdm845-db845c.dtb qcom/qrb5165-rb5.dtb"
 
 QCOM_BOOTIMG_PAGE_SIZE[apq8016-sbc] = "2048"
 QCOM_BOOTIMG_ROOTFS = "/dev/sda1"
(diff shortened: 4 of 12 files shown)


lumag:
@alimon @andersson wanted to share this, it is a small (~3.2MiB) initramfs that just probes drivers and then pivots into rootfs. I think we should keep it and add another `initramfs-qcom-rootfs-image` that will depend on platform-specific modules and firmware.

ndechesne:
I like that a lot. I would like to have an initramfs by default since we are building our boot images anyways.. But that means we need to figure out how to bring enough kernel modules into the initramfs, like Debian does for example.. is there really no good example of how to do that with OE?

lumag:
@ndechesne as I wrote, the simplest way would be to RRECOMMEND on the list of modules and firmware packages required to boot the platform. Ideally we should define a list of modules on a per-machine basis (like we do for the firmware).

lumag:
@alimon @ndechesne @andersson 
I've updated this PR with the setup that allowed me to pivot into rootfs using mainline (stable) kernel.
To repeat my tests, add following lines to `conf/local.conf`:

```python
PREFERRED_PROVIDER_virtual/kernel = "linux-stable"
INITRAMFS_IMAGE = "initramfs-qcom-image"
```

Then `bitbake virtual/kernel` and boot using `boot-initramfs-FOO-qcom-armv8a.img`, replacing FOO with machine-specific DTB name. Machine-specific builds should also work, replace `qcom-armv8a` with the OE machine name.

Notes:
- I've tested this only on RB5, eMMC-based boards (db410c, db820c) would require changes.
- initramfs is kept small. I did not include DRM modules (or firmware), just bare minimum. This can be changed later.

lumag:
As a bonus point, this initramfs image only takes 3.3M when compressed:
```
$ du -Dh tmp-rpb-glibc/deploy/images/qcom-armv8a/initramfs-qcom-image-qcom-armv8a.cpio.gz 
3.3M	tmp-rpb-glibc/deploy/images/qcom-armv8a/initramfs-qcom-image-qcom-armv8a.cpio.gz
```

lumag:
run tests

lumag:
Confirmed this to work on db410c and db820c

recipes-kernel/images/initramfs-qcom-image.bb
```diff
@@ -0,0 +1,10 @@
+require initramfs-rootfs-image.bb
+
+DESCRIPTION = "Ramdisk image for pivoting into rootfs extended to boot Qualcomm boards"
+
+PACKAGE_INSTALL_ATTEMPTONLY = " \
+    kernel-module-phy-qcom-qmp \
+    kernel-module-qcom-pon \
+    kernel-module-qnoc-sm8250 \
+    kernel-module-ufs-qcom \
+"
```
lumag, line 10: This doesn't seem to work as expected, so I'd switch to separate packagegroup later.

lumag: fixed the issue by creating separate packagegroup

lumag:
@ndechesne @alimon ok, this looks ready from my point of view. I'd like to hear your comments now :-)

recipes-kernel/linux/linux-stable.inc
```diff
@@ -12,4 +12,3 @@ SCMVERSION ?= "y"
 SRC_URI = "git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git;branch=linux-${PV}.y"
 
 S = "${WORKDIR}/git"
-
```
ndechesne: oops.

recipes-kernel/linux/linux-qcom-bootimg.inc
```diff
@@ -144,3 +144,6 @@ addtask do_qcom_img_deploy_setscene
 do_qcom_img_deploy[dirs] = "${QIMG_DEPLOYDIR} ${B}"
 do_qcom_img_deploy[cleandirs] = "${QIMG_DEPLOYDIR}"
 do_qcom_img_deploy[stamp-extra-info] = "${MACHINE_ARCH}"
+
+# We do not need kernel image in /boot, these images are flashed into separate partition.
```
ndechesne, line 148: the problem with this change is that we are making this change for all users of meta-qcom. 
Why do we have the base package installed in the first place, is that standard thing?

lumag: Yes, it is a standard thing, as typically users would like to have a kernel image in /boot, so that grub/U-Boot/whatever can find it. When using qcom-bootimg we do not need it though. I could not decide whether to put it here or into qcom-common.inc, so after flipping a coin it ended up here.

recipes-kernel/linux/linux-stable.inc
lumag: fixed

recipes-kernel/linux/linux-stable.inc
```diff
@@ -0,0 +1,14 @@
+# Copyright (C) 2014-2021 Linaro
```
ndechesne, line 1: I don't think we should have linux-stable in meta-qcom. there are better places for such recipe elsewhere. meta-linux-mainline..

lumag: I didn't know about the `meta-linux-mainline` :-D
I needed it as a way to test the initramfs. I can drop the respective commit (or just keep the .bbappend adding qcom-bootimg to linux-stable/linux-mainline, if the layer is enabled).

lumag: @ndechesne updates this PR to support meta-linux-mainline

ndechesne:
run tests

ndechesne:
run tests

ndechesne:
run tests , yes again!

ndechesne:
Sorry about the delay.. I reviewed that again, and I think it's good. however the PR has grown into several things now.. Can you please have several PRs, so that we can start merging them:

1. patches 1-4 which implement the core of what this PR is about
2. 5-6 are 2 independent fixes
3. 7-8 is a different PR to bring meta-mainline support facilities
4. 9-10 is a special feature (on top of 1-4)

If i understand correctly if a user simply sets 
INITRAMFS_IMAGE = "initramfs-qcom-image"
in its config, then our initramfs will be built automatically (as per linux-qcom-bootimg.inc) and it will be used for the boot image.. that seems to do what we are looking for.. 

so we will need to update out build scripts for our release to set INITRAMFS_IMAGE to support this boot method in our releases. Correct?

lumag:
@ndechesne 
I'm going to split the PR into several smaller PRs.

Regarding the `INITRAMFS_IMAGE`. Yes, you are correct. However it's a separate change. Let's perform it later, after the rest of PRs are merged.

lumag:
I've split into #428, #429, #430. The last two patches depend on `initramfs-rootfs-image`, so they will have to wait.

lumag:
Replaced with #431

PR #388, opened 2021-11-14
firmware-wcn6855: provide firmware for wcn6855 wifi chip

lumag:
Firmware for the WCN6855 is not yet provided in the linux-firmware.
Until it is provided by the linux-firmware, package it separately.

### DIFF
--- a/recipes-bsp/firmware/firmware-wcn6855_git.bb
+++ b/recipes-bsp/firmware/firmware-wcn6855_git.bb
@@ -1,0 +1,41 @@
+SUMMARY = "Firmware files for Qualcomm/Atheros WCN6855 SoC"
+
+LICENSE = "Firmware-qualcommAthos_ath10k"
+LIC_FILES_CHKSUM = "file://LICENSE.QualcommAtheros_ath10k;md5=cb42b686ee5f5cb890275e4321db60a8"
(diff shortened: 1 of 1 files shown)


recipes-bsp/firmware/firmware-wcn6855_git.bb
```diff
@@ -0,0 +1,41 @@
+SUMMARY = "Firmware files for Qualcomm/Atheros WCN6855 SoC"
+
+LICENSE = "Firmware-qualcommAthos_ath10k"
+LIC_FILES_CHKSUM = "file://LICENSE.QualcommAtheros_ath10k;md5=cb42b686ee5f5cb890275e4321db60a8"
+NO_GENERIC_LICENSE[Firmware-qualcommAthos_ath10k] = "LICENSE.QualcommAtheros_ath10k"
+
+SRC_URI = "git://chromium.googlesource.com/chromiumos/third_party/linux-firmware;protocol=https;branch=master"
```
ndechesne, line 7: in the past we've pulled 'pre release' of firmware from https://github.com/kvalo/ath11k-firmware, is the firmware not available there instead?

lumag: The firmware itself is there, but not the board files. So I deiced to pull up working combo from the ChromeOS repo.

PR #391, opened 2021-11-18
Rework initramfs-test-image family of images

lumag:
- remove duplication between `test` and `test-full` images
- handle `PACKAGE_INSTAL_foo-layer` automatically rather than listing them manually.


ndechesne:
really nice clean up ;)

PR #400, opened 2021-12-08
initramfs-test-image: Add a couple of networking related utils - conn…

bhupesh-sharma:
…man & phytool

Add a couple of networking related utils - connman & phytool
to 'initramfs-test-image' to allow easier network configuration
(auto eth address assignment) and quering / setting phy registers
(via phytool).

### DIFF
--- a/recipes-test/images/initramfs-test-image.bb
+++ b/recipes-test/images/initramfs-test-image.bb
@@ -34,7 +34,9 @@ PACKAGE_INSTALL_openembedded-layer += " \
 "
 
 PACKAGE_INSTALL_networking-layer += " \
+    connman \
     iperf2 \
     iperf3 \
+    phytool \
     tcpdump \
 "


recipes-test/images/initramfs-test-image.bb
```diff
@@ -34,7 +34,9 @@ PACKAGE_INSTALL_openembedded-layer += " \
 "
 
 PACKAGE_INSTALL_networking-layer += " \
+    connman \
```
lumag, line 37: why do you need connman in the test image? systemd-network is already taking caring of network interfaces IIRC

recipes-test/images/initramfs-test-image.bb
```diff
@@ -34,7 +34,9 @@ PACKAGE_INSTALL_openembedded-layer += " \
...
     iperf2 \
     iperf3 \
+    phytool \
```
lumag, line 40: This chunk looks good to me

recipes-test/images/initramfs-test-image.bb
ndechesne: agreed. I don't think we want nor need conman in the 'simple' test images. and we use nm in our OE RPB images.

bhupesh-sharma: Sure, I was having some issues with auto IP addr assignment for eth interfaces and read in the yocto documentation that connman can assign IP addr to both wired and wireless interfaces at boot automatically. However, later on during my testing I found it was counter-productive for testing with static IP addr.

So, I will remove connman from the list of packages suggested to be added to initramfs test image.

PR #402, opened 2021-12-09
Add openssh master

bhupesh-sharma:


ndechesne:
this PR includes patches from another PR. Please check your PR before submitting them.

recipes-test/images/initramfs-test-image.bb
```diff
@@ -37,6 +37,7 @@ PACKAGE_INSTALL_networking-layer += " \
     connman \
     iperf2 \
     iperf3 \
+    openssh \
```
ndechesne, line 40: what is the impact on image size? 
Do you need both client and server? 
Can you compare with dropbear, wouldn't that be enough?

bhupesh-sharma: Dropbear is a nice suggestion,  that results in a initramfs-test-image.cpio.gz which is 1M smaller than the one with openssh included. I will update the PR accordingly. Thanks.

PR #404, opened 2021-12-14
initramfs-test-image: Add a few utils

bhupesh-sharma:
Add the following utilities to 'initramfs-test-image'
to allow easier board development:

- dropbear (for ssh and scp services),
- strace (qcom requested this to be added to allow tracing kernel func
  calls),
- phytool (allows ethernet phy registers to be read / write).

This causes the initramfs-test-image.rootfs.cpio.gz size to be increased
by 0.5M.

(cherry picked from commit 9724a4e9acad343781b1e1997d43b37d1c06bc19)

### DIFF
--- a/recipes-test/images/initramfs-test-image.bb
+++ b/recipes-test/images/initramfs-test-image.bb
@@ -6,6 +6,7 @@ PACKAGE_INSTALL += " \
     bluez5 \
     dhcp-client \
     diag \
+    dropbear \
     e2fsprogs \
     e2fsprogs-e2fsck \
     e2fsprogs-mke2fs \
@@ -20,6 +21,7 @@ PACKAGE_INSTALL += " \
     pd-mapper \
     qrtr \
     rmtfs \
+    strace \
     tqftpserv \
     usbutils \
     wpa-supplicant \
@@ -36,6 +38,7 @@ PACKAGE_INSTALL_openembedded_layer += " \
 PACKAGE_INSTALL_networking_layer += " \
     iperf2 \
     iperf3 \
+    phytool \
     tcpdump \
 "
 


recipes-test/images/initramfs-test-image.bb
```diff
@@ -6,6 +6,7 @@ PACKAGE_INSTALL += " \
     bluez5 \
     dhcp-client \
     diag \
+    dropbear \
```
lumag, line 9: It would be good to define some boundaries between `initramfs-test` and `initramfs-test-full`, otherwise it's hard to decide whether an utility should enter the former or the later one. @ndechesne any ideas?

ndechesne: It is indeed loose.. I never really liked the initramfs-test-full that much.. we just had to do it for the SA8155p 1.0 release when flashing a rootfs was not working.. but i think the idea was:

initramfs-test-image includes utilities that a developer require to do kernel dev in a 'good' environment, and do basic tests.
initramfs-test-full-image includes additional tools to run a full regression test suite

in that regards i think this PR should be ok. 

However I am interested in the actual image size increase (space used in RAM), not the file size of the compressed image. @bhupesh-sharma can you please provide that value?

bhupesh-sharma: Hello @ndechesne , sorry for the late reply (I was caught up in other things). I had a look at the actual image size increase (space used in RAM) and it is also less than 1MB with all 3 utils, namely dropbear, strace and phytool added to the initramfs-test-image image.

So, I think, we should be ok with this change, especially since we have customer requests to get these enabled in the smaller initramfs images (so that they can easily ssh into the board, exercise ethernet phy regs and also trace the kernel syscalls when they execute different user-space tools).

PR #408, opened 2021-12-21
sdx55: switch to main linux-linaro-qcoml kernel recipe

ndechesne:
There is enough upstream support for sdx55, we no longer need to use a
custom kernel recipe.


conf/machine/sdx55-telit-fn980.conf
```diff
@@ -19,4 +19,3 @@ QCOM_BOOTIMG_ROOTFS ?= "ubi0:system"
 
 SERIAL_CONSOLES = "921600;ttyMSM0"
```
lumag, line 21: extra empty line

ndechesne: fixed.

PR #409, opened 2021-12-22
Move sa8155p adp to 5.16 dunfell

bhupesh-sharma:


recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
```diff
@@ -6,5 +6,6 @@ require recipes-kernel/linux/linux-linaro-qcom.inc
 # SRCBRANCH set to "release/qcomlt-5.15" in linux-linaro-qcom.inc
 SRCREV = "43aa1d1f5fcf5d68f07821b0b3a9314f7c7af649"
 
+SRC_URI:append:sa8155p += "file://0001-Revert-kbuild-Enable-DT-schema-checks-for-.dtb-targe.patch"
```
lumag, line 9: Nak, you can pull this patch into your branch directly (Or merge newer 5.16-rc, which has already got the respective revert).

recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
```diff
@@ -0,0 +1,10 @@
+# Copyright (C) 2021 Linaro
+# Released under the MIT license (see COPYING.MIT for the terms)
+
+require recipes-kernel/linux/linux-linaro-qcom.inc
+
+# SRCBRANCH set to "release/qcomlt-5.15" in linux-linaro-qcom.inc
+SRCREV = "43aa1d1f5fcf5d68f07821b0b3a9314f7c7af649"
+
+SRCBRANCH:sa8155p = "release/sa8155p-adp/qcomlt-5.16"
+SRCREV:sa8155p = "78b7b519e9ae3afeb32b134956b397b2f246e0b0"
```
lumag, line 11: This needs to be updated to follow commit destined the master branch

recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
ndechesne: Well, I prefer we take this patch here actually. It's an OE issue if the kernel does not build with this patch, so it's better to have the patch in the recipe. 
I would have preferred to merge a more recent -rc too, but it's too late since we want to make a release (yesterday!)

lumag: @ndechesne ok, I do not want to delay the yesterday's release :-D
But judging from Rob's patch (and it getting merged into mainline) it's not an OE issue (or not only an OE issue), but a generic trouble.

bhupesh-sharma: Heh, the PR was due yesterday, but somehow bitbake on the my newly updated Fedora 35 host is slower than what I was seeing with Fedora 34 (it is still under debug), so the PR was posted today morning only.

Also just wanted to note that we decided to use the 5.16-rc3 version as the base for the adp release one month or so ago. So its too late to update to a new rc :P. I will probably make this similar to the PR, I posted for the master branch as we need to revert Rob's patch for the adp board at both the places for now.

recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
bhupesh-sharma: Sure, I am posting an update shortly.

lumag:
@bhupesh-sharma just update the cherry-picked commit instead of adding new one on top of the previous one.

bhupesh-sharma:
> @bhupesh-sharma just update the cherry-picked commit instead of adding new one on top of the previous one.

@lumag , sure updated PR on the way.

lumag:
It seems 5.16 is too new for the cryprodev from dunfell.

ndechesne:
ok.. i am merging it. I think the PR is safe, and the failure is not related to this patch.

PR #410, opened 2021-12-22
linux: Add new .bb files for release/sa8155p-adp/qcomlt-5.16

bhupesh-sharma:

### DIFF
--- a/recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
+++ b/recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
@@ -4,5 +4,6 @@
 require recipes-kernel/linux/linux-linaro-qcom.inc
 
 COMPATIBLE_MACHINE = "(sa8155p)"
+SRC_URI:append = " file://0001-Revert-kbuild-Enable-DT-schema-checks-for-.dtb-targe.patch"
 SRCBRANCH = "release/sa8155p-adp/qcomlt-5.16"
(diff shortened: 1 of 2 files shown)


recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
```diff
@@ -0,0 +1,10 @@
+# Copyright (C) 2021 Linaro
+# Released under the MIT license (see COPYING.MIT for the terms)
+
+require recipes-kernel/linux/linux-linaro-qcom.inc
+
+# SRCBRANCH set to "release/qcomlt-5.15" in linux-linaro-qcom.inc
```
ndechesne, line 6: this is wrong. 5.16 is only intended for sa8155p machine, so it should have a COMPATIBLE_MACHINE variable, so that it's not used for other machines.

recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
```diff
@@ -0,0 +1,10 @@
...
+SRCREV = "43aa1d1f5fcf5d68f07821b0b3a9314f7c7af649"
+
+SRCBRANCH:sa8155p = "release/sa8155p-adp/qcomlt-5.16"
```
ndechesne, line 9: we should clean up that, since we only use this kernel recipe for sa8155p.

lumag:
Looks good to me and can be merged once the build succeeds

PR #422, opened 2022-02-17
linux: Add new .bb file for release/sa8155p-adp/qcomlt-5.16

bhupesh-sharma:
Create a new .bb file for honister buils for sa8155p adp board
which already supports kernel version 5.16.

### DIFF
--- a/recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
+++ b/recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
@@ -1,0 +1,9 @@
+# Copyright (C) 2021 Linaro
+# Released under the MIT license (see COPYING.MIT for the terms)
+
+require recipes-kernel/linux/linux-linaro-qcom.inc
+
+COMPATIBLE_MACHINE = "(sa8155p)"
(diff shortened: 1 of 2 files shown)


recipes-kernel/linux/linux-linaro-qcomlt_5.16.bb
```diff
@@ -0,0 +1,7 @@
+# Copyright (C) 2021 Linaro
+# Released under the MIT license (see COPYING.MIT for the terms)
+
+COMPATIBLE_MACHINE = "(sa8155p)"
+SRC_URI:append = " file://0001-Revert-kbuild-Enable-DT-schema-checks-for-.dtb-targe.patch"
```
lumag, line 5: Can you merge this commit into the kernel tree instead?
In the end if was reverted upstream too.

bhupesh-sharma: We can do it for the next version (its already on my queue). So far we are looking at providing a honister build to customer which has very few changes as compared to the available dunfell support. Once they shift to honister, I plan to make the tree more cleaner and resembling latest upstream kernels.

lumag:
LGTM

ndechesne:
run tests

PR #430, opened 2022-03-23
Linux mainline support

lumag:
Add support for kernels from `meta-linux-mainline`. This allows using `linux-mainline` and `linux-stable` as providers for `virtual/kernel`.


lumag:
This depends on #429 , so it will have to wait a bit.

lumag:
run tests

PR #438, opened 2022-04-11
Package firmware for Nexus and Pixel phones and tablets

lumag:


ndechesne:
I merged too quickly.. you are adding a dependency to android-tools-native from meta-oe in the recipe.. yocto-check-layer fails.

lumag:
Let me take a quick glance and fix it.

lumag:
@ndechesne should be fixed by https://github.com/ndechesne/meta-qcom/pull/448

PR #439, opened 2022-04-11
Additional ifc6410/apq8064 support

lumag:


ndechesne:
Don't you think it's better to make a generic 32-bit platform instead?

lumag:
> Don't you think it's better to make a generic 32-bit platform instead?

I thought about this. However there are too many differences. For example sdx55 platforms use ubifs with params which differ even between boards supported by us (mtp and fn980). `MACHINE_FEATURES` differ too much, etc.

It might be sensible to create a generic phone/desktop 32-bit platform (8x60, 8960, 8064, 8x74, 8084, odd 8916 stuck with 32-bit kernel). At this moment meta-qcom supports only two such devices (ifc6410, Nexus 7). It would be straightforward to support Nexus 4 (apq8064, not yet supported upstream) and Nexus 5 (msm8974, supported upstream). If might be easy and interesting to suppot Nexus 6 (apq8084, not supported upstream), ifc6540 (apq8084, only basic support) or QS600 (apq8064,  supported upstream).

But all this ends up with the question if working on such support is worth efforts spent during the work. Frankly speaking I don't see a direct benefit. Most of desktop users will probably stick to PmOS. apq8064 platforms were interesting to me as a good example of mdp4 platform. The rest isn't that interesting. Maybe I'll work on Nexus 5 or ifc6540 at some point (as an example of early mdp5), but that's definitely is not going to happen in the next few months. However if somebody needs this, I can spend some spare time and assist/fix issues.

lumag:
@ndechesne What do you think about `qcom-krait` ?

lumag:
@ndechesne updated the PR

recipes-test/images/initramfs-firmware-image.bb
```diff
@@ -1,13 +1,21 @@
 DESCRIPTION = "Tiny ramdisk image with firmware files"
 
-PACKAGE_INSTALL = " \
+# Do not install anything by default
+PACKAGE_INSTALL = ""
+
+PACKAGE_INSTALL:arm64 = " \
```
ndechesne, line 6: should we create qcom-armv8a and qcom-armv7a OVERRIDE instead of using :arm64 and :qcom-krait? krait and kryo really refer to the Arm CPU in the SoC, so it doesn't make sense to me to use krait/kryo to manage the soc firmware. 

I agree that we need a mechanism to group all the 32 bit or 64 bit Snapdragon platforms.. and differentiate them from modem platforms (SDX), there isn't a 64 bit SDX yet.. but most likely it will exist sooner or later..  since we decided to use qcom-armv8a and qcom-armv7a MACHINE to identify a 64 bit or 32 bit snapdragon, using that for OVERRIDE makes sense.

lumag: done

PR #446, opened 2022-04-26
initramfs-test-image: enable gpsd support

lumag:
Bundle gpsd and gps-utils into initramfs-test-image. This cause
compressed image growth of 1.6 MiB.


ndechesne:
well.. where do we set the limit? the description for this image is 'a small ramdisk'. Are we not adding too much here? I understand it's convenient, but how often do we really need gpsd in there? perhaps test-full instead?

lumag:
Ack, test-full seems fine to me.
Maybe we'd need some light utility for working/validating GPS on db410c (and hopefully db820c)

lumag:
Switched to initramfs-test-full-image

PR #449, opened 2022-05-02
initramfs-test-image: Add a few utils

bhupesh-sharma:
Add the following utilities to 'initramfs-test-image'
to allow easier board development:

- cmake
- make
- git


ndechesne:
NAK. It's a test initramfs, we can't add things like that which has an impact (size) on everyone.
We also agreed to include an indication about the image size increase in such PR.

lumag:
I tend to agree here. if you need anything, you can compile it on the host and then sideload it by concatenating several cpio.gz images for the initramfs. Closing now.

PR #451, opened 2022-05-10
gpsd: add versioned patch support

lumag:
PDS patches for GPSd 3.23 and (pending) 3.24 are not compatible. Add two
versions of the patch so that it is possible to test both GPSd versions.


kraj:
please merge this change as corresponding meta-oe [change](https://git.openembedded.org/meta-openembedded/commit/?id=94f4f39e74b9ab289dec0d75c95c281a1e2fe195) is in master now.

PR #453, opened 2022-05-11
initramfs-test-image: enable read-edid package

lumag:
To assist in checking EDID reads and modesetting enable the read-edid
package. The image growth is 9 KiB.

### DIFF
--- a/recipes-test/images/initramfs-test-image.bb
+++ b/recipes-test/images/initramfs-test-image.bb
@@ -46,6 +46,7 @@ PACKAGE_INSTALL_openembedded-layer += " \
     devmem2 \
     lmsensors-config-libsensors \
     lmsensors-sensors \
+    read-edid \
 "
 
 PACKAGE_INSTALL_networking-layer += " \


bhupesh-sharma:
> To assist in checking EDID reads and modesetting enable the read-edid package. The image growth is 9 KiB.
> 
> Signed-off-by: Dmitry Baryshkov [dmitry.baryshkov@linaro.org](mailto:dmitry.baryshkov@linaro.org)

I think we discussed its useful to add such utils to rpb-test images instead of the initramfs images in the past (and NAK'ed a few PRs already along the same lines). So may be we can move this to rpb-test images instead as EDID utils are required only for specific testing and perhaps not best suited for the generic initramfs-test-image [may be we need a initramfs-graphics-test-image instead :) ].

ndechesne:
Well, we nak'd adding cmake, make and git into the initramfs which totally makes sense, not only it would have increased the size a lot, but it really does not make sense to expect compiling from an intiramfs. 
in this case it's a useful debug tool used for display testing, and the size increase is negligible.

bhupesh-sharma:
> t make sense to expect compiling from an intiramfs.
> in this case it's a useful debug tool used for display testing, and the size increase is negligible.

make and cmake doesn't increase the overall size more than some Kbs :)

ndechesne:
> > t make sense to expect compiling from an intiramfs.
> > in this case it's a useful debug tool used for display testing, and the size increase is negligible.
> 
> make and cmake doesn't increase the overall size more than some Kbs :)

cmake seems to be ~30MB.. so perhaps not. That said the information about the size increase was not in the commit log, so we wouldn't know.

PR #456, opened 2022-05-20
Fix alternative wcn6855 firmware packaging

lumag:
Firmware from linux-firmware has issues on sm8450/wcn6856 platforms
(inability to load board file). Allow installing alternative firmware
files.

### DIFF
--- a/recipes-bsp/firmware/firmware-wcn6855_git.bb
+++ b/recipes-bsp/firmware/firmware-wcn6855_git.bb
@@ -19,14 +19,25 @@ do_compile() {
 	:
 }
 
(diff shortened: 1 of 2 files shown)


ndechesne (commented):
$subject is not great.. it's not a workaround to pass yocto-check-layer, it's a bug fix to ensure meta-qcom follows YP compatible guidelines !

lumag:
> $subject is not great.. it's not a workaround to pass yocto-check-layer, it's a bug fix to ensure meta-qcom follows YP compatible guidelines !

@ndechesne, reworded the commit subj & message.

lumag:
gracious ping

PR #457, opened 2022-05-20
Add bootrr to initramfs-test-image

lumag:


recipes-test/bootrr/bootrr_git.bb
```diff
@@ -3,10 +3,9 @@ SUMMARY = "simple low-level testing tool for qcom boards"
 LICENSE = "BSD-3-Clause"
 LIC_FILES_CHKSUM = "file://LICENSE;md5=987293312a134ab40eec5f3d446cfaff"
 
-SRCREV = "d2329902b701e0efa345628471d0d275d5a5835a"
+SRCREV = "bd84c8a164e9c603db85781cfa019b516a6d2ded"
 SRC_URI = "\
 	git://github.com/andersson/bootrr.git;branch=master;protocol=https \
-	file://bootrr-auto-switch-to-using-sh.patch \
```
ndechesne, line 9: please remove the patch too (and explain why the patch was removed, was it merged upstream)?

recipes-test/bootrr/bootrr_git.bb
```diff
@@ -3,7 +3,8 @@ SUMMARY = "simple low-level testing tool for qcom boards"
 LICENSE = "BSD-3-Clause"
 LIC_FILES_CHKSUM = "file://LICENSE;md5=987293312a134ab40eec5f3d446cfaff"
 
-SRCREV = "bd84c8a164e9c603db85781cfa019b516a6d2ded"
+REV = "bd84c8a164e9c603db85781cfa019b516a6d2ded"
```
ndechesne, line 6: what's the reason why you added REV here?

lumag: @ndechesne it is used in `SRCREV:class-devupstream` override

recipes-test/bootrr/bootrr_git.bb
lumag: yes, it was merged upstream. Will do.

recipes-test/bootrr/bootrr_git.bb
```diff
@@ -12,6 +13,12 @@ S = "${WORKDIR}/git"
 
 PV = "0.0+git${SRCPV}"
 
+# Set PREFERRED_VERSION_bootrr = "upstream%" to select git version
+BBCLASSEXTEND = "devupstream:target"
+PV:class-devupstream = "upstream+git${SRCPV}"
+# Use AUTOREV only if the devupstream is selected as preferred
+SRCREV:class-devupstream = '${@oe.utils.conditional("PREFERRED_VERSION_bootrr", "upstream%", "${AUTOREV}", "${REV}", d)}'
```
ndechesne, line 22: oh. i see. are you implying that AUTOREV will be 'parsed/run' even if we don't use the 'upstream' flavor?

ndechesne: are we not supposed to set PREFERRED_PROVIDER to bootrr-upstream to select it? Why are you using PREFERRED_VERSION?

lumag: Unfortunately yes. The `devupstream` adds another variant/kind of the recipe (like `-native`), which is parsed, but not selected by default. So the `${AUTOREV}` will be expanded unless it is protected by the 'check selected' hack.

lumag: Because there is no separate provider, etc. It's just a virtual  recipe with another version.

ndechesne: ok.. i've never used devupstream class.. i would like to try it and understand how it works. if you want to merge the other patches, please split the PR into 2, so that we can merge the first part. or if you are ok, please give me a day or 2 to test devupstream before merging..

lumag: @ndechesne done

PR #458, opened 2022-05-23
bootrr: use devupstream class to package tip revision

lumag:
Add the following line to the local.conf to switch to the tip revision
of the bootrr:

PREFERRED_VERSION_bootrr = "upstream%"


lumag:
Converted to the draft to prevent accidentially merging this PR.

lumag:
OK, as long as we do not have any release for bootrr, it might be easier to just use `SRCREV:pn-bootrr = "${AUTOREV}"` in local.conf

PR #470, opened 2022-07-05
kirkstone: linux: Create 5.19 kernel based recipe for sa8155p adp

bhupesh-sharma:
Create 5.19 kernel based recipe for sa8155p adp
board. The rational behind the same is that
kirkstone support for sa8155p adp board would always
try to follow the latest upstream kernel version.

### DIFF
--- a/recipes-kernel/linux/linux-linaro-qcomlt_5.19.bb
+++ b/recipes-kernel/linux/linux-linaro-qcomlt_5.19.bb
@@ -1,0 +1,5 @@
+require recipes-kernel/linux/linux-linaro-qcom.inc
+
+COMPATIBLE_MACHINE = "(sa8155p)"
+SRCBRANCH = "sa8155p-adp/qcomlt-track-upstream"
+SRCREV = "ecd82903edb65c18195c83e697cfcfd0815cf02a"


recipes-kernel/linux/linux-linaro-qcomlt_5.19.bb
```diff
@@ -0,0 +1,5 @@
+require recipes-kernel/linux/linux-linaro-qcom.inc
+
+COMPATIBLE_MACHINE = "(sa8155p)"
+SRCBRANCH = "sa8155p-adp/qcomlt-track-upstream"
```
lumag, line 4: I'd suggest to use the standard branch name here: `release/sa8155p-adp/qcomlt-5.19`.
For the 'latest upstream' we have the `linux-linaro-qcomlt-dev` recipe.

ndechesne: not sure why we even need/want this recipe in the first place.

ndechesne: but why kirkstone, not master then?

bhupesh-sharma: @ndechesne : Ok, let me use master in v2

lumag: @bhupesh-sharma so,  what for do you need this branch/recipe? For the latest integration we have the separate recipe. Once selected for `virual/kernel` it will default to the latest integration branch. For the latest upstream it's better to use linux-stable kernels.

bhupesh-sharma: Superseeded by PR: https://github.com/ndechesne/meta-qcom/pull/472

PR #472, opened 2022-07-12
master: linux: Create 5.19 kernel based recipe for sa8155p adp

bhupesh-sharma:
Create 5.19 kernel based recipe for sa8155p adp
board. The rational behind the same is that
master support for sa8155p adp board would always
try to support the latest upstream kernel version
tested on the board.

So, while by default the releases for adp will still
be based on '5.15 + OE dunfell', the newer snapshots will
also be available based on '5.19 + OE master'.


bhupesh-sharma:
Self NACK'ing this one. This will be used to point to qcom as a reference for writing sample recipes though.

PR #479, opened 2022-08-26
pulseaudio: fix use of overrides

ndechesne:
Fixes e5a2eb2f4409 (pulseaudio: guard the qcom-specific workaround
with 'qcom' override).

The order of the overrides was incorrect resulting in build failures.

Reported-by: Leonardo Sandoval <leonardo.sandoval@linaro.org>


lumag:
LGTM.

PR #481, opened 2022-09-05
CI: tuxsuite: Remove wrong entry in artifacts field

vishalbhoj:
Signed-off-by: Vishal Bhoj <vishal.bhoj@linaro.org>

### DIFF
--- a/.github/workflows/build-template.yml
+++ b/.github/workflows/build-template.yml
@@ -63,8 +63,7 @@ jobs:
           target: ${{inputs.images}}
           bblayers_conf:
             - BBLAYERS += '../$(echo ${{github.repository}} | cut -d'/' -f2)/'
-          artifacts:
-            - \$DEPLOY_DIR
+          artifacts: []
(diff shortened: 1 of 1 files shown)


vishalbhoj:
artifacts list now expects a list of right directory path relative to deploy_dir path. By default it publishes complete deploy directory

lumag:
@vishalbhoj The commit doesn't tell why the entry is wrong. Also the original intent was obviously to keep deployed files in artifacts. If it is a default, it should be mentioned in the commit message.

ndechesne:
Why are you limiting the artifacts to be in DEPLOY_DIR? there could be other useful artifacts, such as log, stamps, .. or binary files when debugging specific issues.

vishalbhoj:
@lumag I have improved the commit message to describe the right behavior.

@ndechesne We will support other artifacts outside of deploy_dir in future as planned.

lumag (approved):

PR #486, opened 2022-10-21
recipes-test: Add mybw recipe

superna9999:
mybw is a small utility to check memory performance.

Signed-off-by: Neil Armstrong <neil.armstrong@linaro.org>


ndechesne:
thanks. please add it to https://github.com/ndechesne/meta-qcom/blob/master/recipes-test/images/initramfs-test-image.bb

PR #489, opened 2022-11-08
8450 hdk

ndechesne:
This


lumag:
LGTM.
The only issue I observe is that this will break building older kernels since they lack the `sm8450-hdk.dts`.

ndechesne:
Right. But there isn't much we can do about it. New kernel will bring new hardware, or do you have any specific idea in mind?

lumag:
Extending kernel-dtb.bbclass to support optional dtbs? But this sounds like
a big horrible hack.
So let's merge this and forget about old kernels.

PR #491, opened 2022-11-09
Rework Nexus & Pixel firmware

lumag:
- Use rust-android-sparse to move pixel firmware out of dynamic layer
- Change firmware paths to follow upstream


recipes-bsp/firmware-nexus/firmware-qcom-pixel.inc
```diff
@@ -62,6 +62,7 @@ do_install() {
     ls ${B}/firmware/*.jsn && install -m 0644 ${B}/firmware/*.jsn ${D}${FW_QCOM_PATH}
 
     install -m 0644 ${B}/firmware/a[0-9]*.fw* ${D}${FW_QCOM_PATH}
+    ls ${B}/firmware/a[0-9]*_gmu.bin && install -m 0644 ${B}/firmware/a[0-9]*_gmu.bin ${D}${FW_QCOM_PATH}
```
ndechesne, line 65: can we install the ones we really want, instead of installing everything and removing the duplicates? that seems more logical to me, so that we control what gets packaged better.

lumag: done

PR #496, opened 2022-11-24
qcom-sdxmdm: add the machine conf governing SDX and MDM platforms

lumag:
Follow the qcom-armv8a and qcom-armv7a lead and provide a unified machine configuration for Qualcomm modem platforms (including, but not limited to sdx55, sdx65 and mdm9615).


lumag:
@Mani-Sadhasivam , @superna9999 , could you please also give it a glance?

ndechesne:
if we do such a generic machine, I think we at least need to include the underlying arch, eg. armv7a in the name. We also need to pick sdx or mdm, i believe they are just 2 marketing names for the same thing, no?

lumag:
These names were used for different generations of modems. I'm a little bit hesistant to include `armv7a` into the machine name. But to be fully future proof, maybe it should be done. What about `qcom-armv7a-modem`?

Mani-Sadhasivam:
LGTM!

ndechesne:
Where are planning to set PREFERRED_PROVIDER_virtual/kernel ?

lumag:
> Where are planning to set PREFERRED_PROVIDER_virtual/kernel ?

The preference for the `virtual/kernel` is set in the qcom-common.inc

ndechesne:
> > Where are planning to set PREFERRED_PROVIDER_virtual/kernel ?
> 
> The preference for the `virtual/kernel` is set in the qcom-common.inc

ouch.. sorry, i missed that you included this file!

PR #497, opened 2022-12-07
Add support for packaging firmware from proprietary archives 

lumag:
Sometimes it's necessary to build the image using firmware files from vendor-provided `NON-HLOS.bin`, `proprietary.tar.gz` or `adreno.tar.gz` (e.g. see #366 ). Add include files and example recipes to build firmware for ifc6410 and ifc6560 boards.


lumag:
@ndechesne your opinion? This makes it easier to build images for images requiring proprietary firmware (see #366). I have included several recipes which I have locally to demonstrate possible usage.

PR #500, opened 2023-01-04
Add SM8550 debugcc support

superna9999:
Back/forwardport SM8550 support pushed at https://github.com/andersson/debugcc/pull/12

Signed-off-by: Neil Armstrong <neil.armstrong@linaro.org>


lumag:
debugcc was updated to latest revision, which includes sm8550 support. Closing now.

PR #503, opened 2023-02-24
mesa: enable tools package

lumag:
Mesa can provide several tools which simplify debugging mesa and the kernel driver. Enable them to be built into the mesa-tools package.


ndechesne:
Why in meta-qcom, and not in the main oe-core recipe?

lumag:
> Why in meta-qcom, and not in the main oe-core recipe?

What exactly? The build fix or the tools enablement?

ndechesne:
I thought both. But I looked into it a bit more.. the patch probably should go to oe-core, the PACKAGECONFIG maybe not..

lumag:
Done, the musl fixup was accepted into oe-core.

PR #520, opened 2023-06-19
Add qbootctl package

superna9999:
Add qbootclt package to control the boot A/B slot flags.

### DIFF
--- a/recipes-support/qbootctl/files/0001-Fix-to-uint32_t-has-not-been-declared.patch
+++ b/recipes-support/qbootctl/files/0001-Fix-to-uint32_t-has-not-been-declared.patch
@@ -1,0 +1,49 @@
+From 26d488b38a0357468b8657f283ad24a433e30beb Mon Sep 17 00:00:00 2001
+From: alefnode <adrian.campos@teachelp.com>
+Date: Thu, 4 May 2023 09:15:29 +0200
+Subject: [PATCH] Fix to << uint32_t has not been declared >>
+
+Link: https://gitlab.com/sdm845-mainline/qbootctl/-/commit/df63d7c21c19a3e3afb41d029b97b9d068880484
+Signed-off-by: Neil Armstrong <neil.armstrong@linaro.org>
+---
+ bootctrl_impl.cpp | 1 +
+ gpt-utils.h       | 1 +
+ qbootctl.cpp      | 1 +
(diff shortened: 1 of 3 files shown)


superna9999:
If we flash the bootimage on recent platform with A/B enabled, if the boot isn't marked successful for N times, the boot will fail until `fastboot set_active a` is called, which can be unconvenient,

lumag:
Could you please add a `.service`, calling `qbootctl` to mark it as successful?

superna9999:
Yep it was my next plan, but I'm unsure what would be the condition, the grub one doesn't see very good since it requires a login (https://github.com/rhboot/grub2/blob/7de33f4072abf5f14c4fdad6e566c58bdbf3b26e/docs/grub-boot-success.timer)

superna9999:
OK i"ll use the same as the systemd one https://github.com/endlessm/systemd/blob/master/units/systemd-bless-boot.service.in

lumag:
`boot-complete.target` looks good.

superna9999:
Done

superna9999:
Updated recipe to 0.1.2 release since master has some bugs

lumag:
@superna9999 I tried enabling qbootctl, and it fails:
```
| ../git/qbootctl.cpp: In function 'int get_slot_info(slot_info*)':
| ../git/qbootctl.cpp:105:9: error: 'uint32_t' was not declared in this scope
|   105 |         uint32_t active_slot = impl->getActiveBootSlot();
|       |         ^~~~~~~~
| ../git/qbootctl.cpp:30:1: note: 'uint32_t' is defined in header '<cstdint>'; did you forget to '#include <cstdint>'?
|    29 | #include "bootctrl.h"
|   +++ |+#include <cstdint>
|    30 |
| ../git/qbootctl.cpp:107:15: error: 'active_slot' was not declared in this scope
|   107 |         slots[active_slot].active = true;
|       |               ^~~~~~~~~~~

superna9999:
Indeed, it requires https://gitlab.com/sdm845-mainline/qbootctl/-/commit/df63d7c21c19a3e3afb41d029b97b9d068880484 for GCC13, adding it to the recipe.

superna9999:
Done, added the right fix

lumag:
Validated and merged

PR #541, opened 2023-11-14
linux-yocto: add support for QCM6490 RB3 board

adhudase:
Add necessary configuration and kernel changes to the linux-yocto recipe to enable support for Qualcomm QCM6490 RB3 board.

This enables boot to shell on QCM6490 RB3 board with initramfs.


recipes-kernel/linux/linux-yocto/bsp/qcom-armv8a/qcom-qcm6490.cfg
```diff
@@ -0,0 +1,19 @@
+# SPDX-License-Identifier: MIT
+
+CONFIG_PINCTRL_SC7280=y
+CONFIG_PINCTRL_SC7280_LPASS_LPI=m
+CONFIG_SND_SOC_SC7280=m
+CONFIG_INTERCONNECT_QCOM_SC7280=y
+
+CONFIG_SC_CAMCC_7280=y
```
lumag, line 8: =m

recipes-kernel/linux/linux-yocto/bsp/qcom-armv8a/qcom-qcm6490.cfg
```diff
@@ -0,0 +1,19 @@
...
+CONFIG_SC_DISPCC_7280=y
+CONFIG_SC_GPUCC_7280=y
+CONFIG_SC_LPASS_CORECC_7280=y
```
lumag, line 11: =m

recipes-kernel/linux/linux-yocto/bsp/qcom-armv8a/qcom-qcm6490.cfg
```diff
@@ -0,0 +1,19 @@
...
+CONFIG_SC_VIDEOCC_7280=y
```
lumag, line 12: =m

conf/machine/qcom-armv8a.conf
```diff
@@ -29,6 +30,7 @@ QCOM_BOOTIMG_ROOTFS[qrb2210-rb1] ?= "PARTLABEL=userdata"
 QCOM_BOOTIMG_ROOTFS[qrb4210-rb2] ?= "PARTLABEL=userdata"
 SD_QCOM_BOOTIMG_ROOTFS[apq8016-sbc] ?= "/dev/mmcblk1p7"
 KERNEL_CMDLINE_EXTRA[sdm845-db845c] ?= "clk_ignore_unused pd_ignore_unused"
+KERNEL_CMDLINE_EXTRA[qcm6490-rb3] ?= "pcie_pme=nomsi earlycon"
```
lumag, line 33: - earlycon should not be necessary for the board to boot, so it should be dropped here (it can go into your local.conf)
- what is the reason for pcie_pme=nomsi? I think there was an issue with the ath11k driver, but it was solved ages ago.

conf/machine/qcom-armv8a.conf
```diff
@@ -18,6 +18,7 @@ KERNEL_DEVICETREE ?= " \
     qcom/qrb4210-rb2.dtb \
     qcom/qrb5165-rb5.dtb \
     qcom/sm8450-hdk.dtb \
+    qcom/qcm6490-rb3.dtb \
```
lumag, line 21: This should be conditional on linux-yocto. This DT is not supported in `linux-linaro-qcomlt`.

recipes-kernel/linux/linux-yocto/0007-FROMGIT-phy-qcom-qmp-ufs-Add-Phy-Configuration-suppo.patch
```diff
@@ -0,0 +1,193 @@
...
+Add SC7280 specific register layout and table configs.
+
+Co-developed-by: Manish Pandey <quic_mapa@quicinc.com>
+Signed-off-by: Manish Pandey <quic_mapa@quicinc.com>
+Signed-off-by: Nitin Rawat <quic_nitirawa@quicinc.com>
+Git-commit: 8abe9792d1ff7e60f911b56e8a2537be7e903576
+Git-repo: https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git .
+
+Change-Id: I72ba081502ec5e8e47d16cc45e70d265cc31c9a4
+Signed-off-by: Manish Pandey <quic_mapa@quicinc.com>
```
lumag, line 16: - Drop the `Git-commit` and `Git-repo` tags, you have `Upstream-Status` instead
- Drop the empty line betwen the tags, they should come as a single chunk with no interruptions
- Drop the Change-Id tag, there is no Gerrit to track them

recipes-kernel/linux/linux-yocto/0008-PENDING-arm64-dts-qcom-Remove-voltage-vote-support-f.patch
```diff
@@ -0,0 +1,47 @@
...
+ 
+ 	status = "okay";
+ };
++
++&vreg_l7b_2p9 {
++	/delete-property/regulator-min-microvolt;
++	/delete-property/regulator-max-microvolt;
++	regulator-allow-set-load;
++	regulator-allowed-modes = <RPMH_REGULATOR_MODE_LPM RPMH_REGULATOR_MODE_HPM>;
++};
```
lumag, line 37: I don't think we can take these two patches w/o discussion at the mailing list. Min and max voltages are required anyway, even if you don't touch the regulators. They provide the sane values for the rest of the regulator subsystem.

Moreover, the regulators were defined in this DT file, so you can change them directly instead of patching them through the `&vreg_l7b_2p9 { };` statement.

recipes-kernel/linux/linux-yocto/0016-PENDING-arm64-dts-qcom-Add-qcm6490-rb3-support.patch
```diff
@@ -0,0 +1,354 @@
...
++#include <dt-bindings/pinctrl/qcom,pmic-gpio.h>
++#include <dt-bindings/regulator/qcom,rpmh-regulator.h>
++#include "qcm6490.dtsi"
++#include "pm7250b.dtsi"
++#include "pm7325.dtsi"
++#include "pm8350c.dtsi"
++#include "pmk8350.dtsi"
++
++/ {
++	model = "Qualcomm Technologies, Inc. QCM6490 RB3";
```
lumag, line 56: This is not the board name, isn't it?

recipes-kernel/linux/linux-yocto/0020-PENDING-arm64-dts-qcm6490-rb3-Remove-voltage-voting-.patch
```diff
@@ -0,0 +1,58 @@
...
+rails. Due to which the initial voltage set by bootloader
+is overridden by regulator framework with min voltage specified
+on regulator registration.
+
+Fix this temporarily by removing voltage voting support, which
+will prevent regulator framework overriding the voltage set by
+bootloader for QC6490 RB3 Platform.
+
+This commit will be reverted once voltage voting support is added
+in USB driver.
```
lumag, line 17: Ah. I think I now understand your issue. Please change the min and max values to the values that are able to sustain the USB.

recipes-kernel/linux/linux-yocto/0021-PENDING-clk-qcom-gcc-Enable-the-force-mem-core-for-U.patch
```diff
@@ -0,0 +1,211 @@
+From b9b99e5d0a54edc73620a1522ad3d1a4a6d115fb Mon Sep 17 00:00:00 2001
+From: Taniya Das <quic_tdas@quicinc.com>
+Date: Mon, 30 Oct 2023 23:24:19 +0530
+Subject: [PATCH 21/27] PENDING: clk: qcom: gcc: Enable the force mem core for
+ UFS ICE clock
+
+Enable the force mem core for UFS ICE clock. Update the gdsc
+transition delays to the recommended values for functional correctness.
```
lumag, line 8: The commit message should tell why, not what. In other words it should describe why the change was done, not the contents of the patch. And it anyway should be split into two patches.

recipes-kernel/linux/linux-yocto/0022-PENDING-arm64-dts-qcm6490-Update-the-protected-clock.patch
```diff
@@ -0,0 +1,80 @@
...
++			<GCC_USB30_SEC_MOCK_UTMI_CLK_SRC>,
++			<GCC_USB30_SEC_MOCK_UTMI_POSTDIV_CLK_SRC>, <GCC_USB30_SEC_SLEEP_CLK>,
++			<GCC_USB3_SEC_PHY_AUX_CLK>, <GCC_USB3_SEC_PHY_AUX_CLK_SRC>,
++			<GCC_USB3_SEC_PHY_COM_AUX_CLK>, <GCC_USB3_SEC_PHY_PIPE_CLK>,
++			<GCC_USB3_SEC_PHY_PIPE_CLK_SRC>, <GCC_CFG_NOC_LPASS_CLK>,
++			<GCC_MSS_GPLL0_MAIN_DIV_CLK_SRC>, <GCC_MSS_CFG_AHB_CLK>,
++			<GCC_MSS_OFFLINE_AXI_CLK>, <GCC_MSS_SNOC_AXI_CLK>,
++			<GCC_MSS_Q6_MEMNOC_AXI_CLK>, <GCC_MSS_Q6SS_BOOT_CLK_SRC>,
++			<GCC_SEC_CTRL_CLK_SRC>, <GCC_WPSS_AHB_CLK>,
++			<GCC_WPSS_AHB_BDG_MST_CLK>, <GCC_WPSS_RSCP_CLK>;
```
lumag, line 45: Ugh. The `qcm6490-fairphone-fp5` lists less clocks here. Is there any reason for that?

recipes-kernel/linux/linux-yocto/0004-FROMLIST-arm64-dts-qcom-Add-qcm6490-dts-file.patch
```diff
@@ -0,0 +1,483 @@
...
++
++		trusted_apps_mem: trusted_apps@c1800000 {
++			reg = <0x0 0xc1800000 0x0 0x3900000>;
++			no-map;
++		};
++	};
++};
++
++&video_mem {
++	reg = <0x0 0x8a700000 0x0 0x500000>;
```
lumag, line 471: This is not correct. The node name is `memory@8b200000`. So you have to delete the old node and add a new one instead (this applies to other reserved memory nodes).

recipes-kernel/linux/linux-yocto/0012-FROMLIST-dma-heap-Add-proper-kref-handling-on-dma-bu.patch
```diff
@@ -0,0 +1,148 @@
...
+Also moves minor tracking into the heap structure so
+we can properly free things.
+
+Change-Id: Ibb88d8af4fe67133873f8e88b85f5a12d4a112ef
+Signed-off-by: John Stultz <jstultz@google.com>
+Signed-off-by: T.J. Mercier <tjmercier@google.com>
+Signed-off-by: Yong Wu <yong.wu@mediatek.com>
+[Yong: Just add comment for "minor" and "refcount"]
+Link: https://lore.kernel.org/lkml/20230911023038.30649-3-yong.wu@mediatek.com/
+Patch-mainline: linux-media @ 09/11/23, 02:30
```
lumag, line 21: -ENOSUCHTAG

lumag (changes requested):
I took both -yocto and patch review.
The largest issue is the split. You have submitted a lot of changes in a single chunk. Instead split them into single atomic commits doing just a single item: patch linux-yocto, change machine config, etc.

Second. Please split your 26-patch series into smaller chunks too. You have a mixture of boards support, generic changes, temporary changes, debugging changes, etc.
For example, if you take away the patch 10 (definitely a temporary BSP item), it should not cause any changes to e.g. dma-heap patches.

lumag:
For the patch-related comments. You don't have to fix issues that I pointed out for the patch contents. Please pass them to your kernel team. However please fix the patch formatting issues (tags, etc).

conf/machine/include/qcom-common.inc
```diff
@@ -18,7 +18,7 @@ PREFERRED_PROVIDER_virtual/libgl ?= "mesa"
 PREFERRED_PROVIDER_virtual/libgles1 ?= "mesa"
 PREFERRED_PROVIDER_virtual/libgles2 ?= "mesa"
 PREFERRED_PROVIDER_virtual/xserver ?= "xserver-xorg"
-PREFERRED_PROVIDER_virtual/kernel ?= "linux-linaro-qcomlt"
+PREFERRED_PROVIDER_virtual/kernel ?= "linux-yocto"
```
lumag, line 21: This chunk is a definiteve no-go. linux-yocto doesn't support any of machines except qcom-armv8a.

adhudase: May I know how are you building linux-yocto for RB5? Thanks

lumag: By setting the `PREFERRED_PROVIDER_virtual/kernel` in the `local.conf`

conf/machine/qcom-armv8a.conf
sbanerjee-quic: Hi @lumag 
How do we share this machine conf for all the existing RBX and new RB3-Gen2

conf/machine/include/qcom-common.inc
sbanerjee-quic: By when we can set  PREFERRED_PROVIDER_virtual/kernel as linux-yocto? We understood that all boards can be supported on linux-yocto.

lumag: You can set it for the `qcom-armv8a` machine.

conf/machine/qcom-armv8a.conf
lumag: > This should be conditional on linux-yocto

So, I suppose, something like `KERNEL_DEVICETREE:append:pn-linux-yocto` += "qcom/qcom6490-rb3.dtb"`

recipes-kernel/linux/linux-yocto/0008-PENDING-arm64-dts-qcom-Remove-voltage-vote-support-f.patch
adhudase: >>I don't think we can take these two patches w/o discussion at the mailing list. Min and max voltages are required >>anyway, even if you don't touch the regulators. They provide the sane values for the rest of the regulator subsystem.

UFS 2.x / 3.x have different voltage requirements -- Bootloader dynamically sets the proper voltage based on UFS2.x/3.x
There is a limitation in regulator framework -- where during regulator registration it will set the min voltage specified in DT -- which might over-ride the bootloader set voltage causing UFS failure.
Also, for both of these regulators UFS is the only client on those loads and it does not vote for voltage from it's driver..
in DT bindings as well, we see regulator-min/max DT properties are optional properties.
hence we have removed this properties to fix the above issue.

>>Moreover, the regulators were defined in this DT file, so you can change them directly instead of patching them >>through the &vreg_l7b_2p9 { }; statement.

We''ll update accordingly.

recipes-kernel/linux/linux-yocto/0020-PENDING-arm64-dts-qcm6490-rb3-Remove-voltage-voting-.patch
adhudase: yes, we'll update min and max values.

recipes-kernel/linux/linux-yocto/0008-PENDING-arm64-dts-qcom-Remove-voltage-vote-support-f.patch
lumag: > > > I don't think we can take these two patches w/o discussion at the mailing list. Min and max voltages are required >>anyway, even if you don't touch the regulators. They provide the sane values for the rest of the regulator subsystem.
> 
> UFS 2.x / 3.x have different voltage requirements -- Bootloader dynamically sets the proper voltage based on UFS2.x/3.x There is a limitation in regulator framework -- where during regulator registration it will set the min voltage specified in DT -- which might over-ride the bootloader set voltage causing UFS failure. Also, for both of these regulators UFS is the only client on those loads and it does not vote for voltage from it's driver.. in DT bindings as well, we see regulator-min/max DT properties are optional properties. hence we have removed this properties to fix the above issue.

Then the UFS driver should set correct values during the probe.

conf/machine/include/qcom-common.inc
```diff
@@ -18,7 +18,7 @@ PREFERRED_PROVIDER_virtual/libgl ?= "mesa"
 PREFERRED_PROVIDER_virtual/libgles1 ?= "mesa"
 PREFERRED_PROVIDER_virtual/libgles2 ?= "mesa"
 PREFERRED_PROVIDER_virtual/xserver ?= "xserver-xorg"
-PREFERRED_PROVIDER_virtual/kernel ?= "linux-yocto"
+PREFERRED_PROVIDER_virtual/kernel ?= "linux-linaro-qcomlt"
```
lumag: Clean branch, please, without merges and reverts.

recipes-kernel/linux/linux-yocto/0006-PENDING-dt-bindings-arm-qcom-Add-QCM6490-RB3-board.patch
```diff
@@ -0,0 +1,29 @@
+From 6b76519b3c3a7d825e5e5770ba7e38ff59a65aa4 Mon Sep 17 00:00:00 2001
+From: Naina Mehta <quic_nainmeht@quicinc.com>
+Date: Tue, 17 Oct 2023 13:29:51 +0530
+Subject: [PATCH 06/26] PENDING: dt-bindings: arm: qcom: Add QCM6490 RB3 board
```
lumag, line 4: This doesn't look as a separate series. It starts with the "PATCH 06/26". Separate series start with "PATCH 01"

lumag (changes requested):

lumag:
- Drop the merge and revert, please. We do not have a requirement that you start at the top of the tree.
- Split the patchset into separate series. I literally mean that. Separate subdirs, that will probably make life easier (and will also simplify handling of the patches). The patches 01-26 do not look like a separate series.

adhudase:
Sory, this seems getting dirty. Can we close this request and create new one? Thanks

conf/machine/qcom-armv8a.conf
```diff
@@ -6,6 +6,8 @@ MACHINE_FEATURES = "usbhost usbgadget alsa screen wifi bluetooth ext2"
 # UFS partitions in 820/845/RB5 setup with 4096 logical sector size
 EXTRA_IMAGECMD:ext4 += " -b 4096 "
 
+PREFERRED_PROVIDER_virtual/kernel = "linux-yocto"
+
 # Support for dragonboard{410, 820, 845}c, rb5
```
lumag, line 11: ?=, please. Allow users to override this value.

recipes-kernel/linux/linux-yocto/bsp/qcom-armv8a/qcom-qcm6490.cfg
```diff
@@ -0,0 +1,19 @@
+# SPDX-License-Identifier: MIT
+
+CONFIG_PINCTRL_SC7280=y
+CONFIG_PINCTRL_SC7280_LPASS_LPI=m
+CONFIG_SND_SOC_SC7280=m
+CONFIG_INTERCONNECT_QCOM_SC7280=y
+
+CONFIG_SC_CAMCC_7280=m
+CONFIG_SC_DISPCC_7280=y
+CONFIG_SC_GPUCC_7280=y
```
lumag, line 10: GCC?

recipes-kernel/linux/linux-yocto/0001-FROMLIST-arm64-dts-qcom-Use-QCOM_SCM_VMID-defines-fo.patch
```diff
@@ -0,0 +1,384 @@
+From f401b67ba892963d08ae0a51a46e47df55d5c011 Mon Sep 17 00:00:00 2001
+From: Luca Weiss <luca.weiss@fairphone.com>
+Date: Thu, 5 Oct 2023 16:47:31 +0530
+Subject: [PATCH 01/26] FROMLIST: arm64: dts: qcom: Use QCOM_SCM_VMID defines
+ for qcom,vmid
```
lumag, line 5: This is still /26. So the series looks incomplete now.

lumag (changes requested):
Please move series to subdirs to ease maintenance.

Other than that, few minor comments left.

conf/machine/qcom-armv8a.conf
adhudase: sure

recipes-kernel/linux/linux-yocto/bsp/qcom-armv8a/qcom-qcm6490.cfg
adhudase: All of the above selects GCC. 
e.g.
config SC_DISPCC_7280
        ...
        select SC_GCC_7280

recipes-kernel/linux/linux-yocto/0001-FROMLIST-arm64-dts-qcom-Use-QCOM_SCM_VMID-defines-fo.patch
adhudase: ah! sorry, will correct and re-post.

adhudase:
By subdirs, do you mean something like devicetree, bindings, drivers etc.? or something like - boards support, generic changes, temporary changes, debugging changes, etc. you mentioned earlier?

recipes-kernel/linux/linux-yocto/bsp/qcom-armv8a/qcom-qcm6490.cfg
lumag: I'd prefer to be explicit here. In-ketrnel dependencies can change.

lumag:
> By subdirs, do you mean something like devicetree, bindings, drivers etc.? or something like - boards support, generic changes, temporary changes, debugging changes, etc. you mentioned earlier?

Yes. So that each group of patches is visually distinct.

conf/machine/qcom-armv8a.conf
adhudase: just observed that, with ?=, its going back to linaro-qcomlt

adhudase:
I've created these four subdirs - 
common-platform, 
idp, rb3gen2,
generic-kernel-6.5

if this looks okay, I can push the patches.

lumag:
> generic-kernel-6.5



> common-platform,
> idp, rb3gen2,
> generic-kernel-6.5

Well, `generic-kernel-6.5` defintely doesn't look ok. Upgrading to 6.6 would cause unnecessary renames.
I'd propose something simpler:

- generic-drivers
- qcm6490-drivers
- qcm6490-dtsi
- qcm6490-board-dts
- hacks (all temporary and workaround patches go to this subdir).

conf/machine/qcom-armv8a.conf
lumag: So, you need to shuffle things around.

conf/machine/qcom-armv8a.conf
lumag: As a second thought, please keep this conditional for now, until we fully phase out linux-linaro-qcomlt kernel recipe.

recipes-kernel/linux/linux-yocto/0001-FROMLIST-dma-heap-Add-proper-kref-handling-on-dma-bu.patch
```diff
@@ -0,0 +1,146 @@
...
+able to properly handle the references to the heaps
+
+Also moves minor tracking into the heap structure so
+we can properly free things.
+
+Change-Id: Ibb88d8af4fe67133873f8e88b85f5a12d4a112ef
+Signed-off-by: John Stultz <jstultz@google.com>
+Signed-off-by: T.J. Mercier <tjmercier@google.com>
+Signed-off-by: Yong Wu <yong.wu@mediatek.com>
+[Yong: Just add comment for "minor" and "refcount"]
```
lumag, line 19: JFYI: it should be `added`. Also such comments usually go before S-o-b.

recipes-kernel/linux/linux-yocto_6.5.bbappend
```diff
@@ -1,4 +1,36 @@
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}/qcm6490-dtsi:"
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}/qcm6490-board-dts:"
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}/hacks:"
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}/qcm6490-drivers:"
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}/generic-drivers:"
```
lumag, line 5: Please keep subdirs as a part of the patch name.

lumag (changes requested):

lumag (changes requested):
Now that the series are sorted out, coud you pelase reorder into the sensible order:

- generic-drivers
- qcm6490-drivers
- qcm6490-dtsi
- qcm6490-board-dts
- hacks

hashim-shiraz:
> Now that the series are sorted out, coud you pelase reorder into the sensible order:
> 
> * generic-drivers
> * qcm6490-drivers
> * qcm6490-dtsi
> * qcm6490-board-dts
> * hacks

Can we have an alternate name for hacks which I guess would raise questions. How about workarounds (or any other name) ?

lumag:
> Can we have an alternate name for hacks which I guess would raise questions. How about workarounds (or any other name) ?

Sure, why not.

lumag (approved):

PR #548, opened 2023-11-18
Efi boot support

quic-vkraleti:
To support systemd-boot as a boot manager, need to create an esp.bin with uki, uefi stub and other essential packaged into it. 
This is achieved in below patches.
1. Extend systemd-boot recipe through a bbappend to provide Ukify tool.
2. Add linux-qcom-efiimg bbclass to generate uki and in turn pack this uki into esp.bin
3. Inherit linux-qcom-efiimg bbclass in linux-yocto & linux-linaro recipes to generate esp.bin

### DIFF
--- a/classes/linux-qcom-bootimg.bbclass
+++ b/classes/linux-qcom-bootimg.bbclass
@@ -39,7 +39,7 @@ python do_qcom_img_deploy() {
     B = d.getVar("B")
     D = d.getVar("D")
     kernel_output_dir = d.getVar("KERNEL_OUTPUT_DIR")
-    kernel_imagedest = d.getVar("KERNEL_IMAGEDEST")
+    kernel_dtbdest = d.getVar("KERNEL_DTBDEST")
     kernel = os.path.join(B, "kernel-dtb")
     definitrd = os.path.join(B, "initrd.img")
     mkbootimg = os.path.join(d.getVar("STAGING_BINDIR_NATIVE"), "skales", "mkbootimg")
@@ -105,7 +105,7 @@ python do_qcom_img_deploy() {
         with open(kernel, 'wb') as wfd:
             with open(os.path.join(kernel_output_dir, kernel_name), 'rb') as rfd:
                 shutil.copyfileobj(rfd, wfd)
-            with open(os.path.join(D, kernel_imagedest, dtb), 'rb') as rfd:
+            with open(os.path.join(D, kernel_dtbdest, dtb), 'rb') as rfd:
                 shutil.copyfileobj(rfd, wfd)
 
         rootfs = getVarDTB("QCOM_BOOTIMG_ROOTFS")
--- a/classes/linux-qcom-bootimg.bbclass
+++ b/classes/linux-qcom-bootimg.bbclass
@@ -132,7 +132,7 @@ python do_qcom_img_deploy() {
 do_qcom_img_deploy[depends] += "skales-native:do_populate_sysroot"
 do_qcom_img_deploy[vardeps] = "QCOM_BOOTIMG_PAGE_SIZE QCOM_BOOTIMG_KERNEL_BASE KERNEL_CMDLINE_EXTRA QCOM_BOOTIMG_ROOTFS"
 
-addtask qcom_img_deploy after do_populate_sysroot do_packagedata bundle_initramfs before do_deploy
+addtask qcom_img_deploy after do_populate_sysroot do_packagedata do_bundle_initramfs before do_deploy
 
 # Setup sstate, see deploy.bbclass
 SSTATETASKS += "do_qcom_img_deploy"
--- a/classes/uki.bbclass
+++ b/classes/uki.bbclass
@@ -1,0 +1,90 @@
+# This bbclass repacks kernel image as a UKI (Unified Kernel Image)
+# Composed by
+#   - an UEFI stub, from systemd-boot
+#   - the kernel Image
+#   - an initramfs
+#   - other metadata like kernel cmdline
+
+DEPENDS:append = " \
+            systemd-boot-native \
+            python3-native \
+            python3-pefile-native \
+            "
+
+require conf/image-uefi.conf
+
+inherit python3native
+
+do_install[depends] += " \
+    systemd-boot:do_deploy \
+    "
+do_install[depends] += "${@ '${INITRAMFS_IMAGE}:do_image_complete' if d.getVar('INITRAMFS_IMAGE') else ''}"
+do_uki[dirs] = "${B}"
+do_install[postfuncs] += "do_uki"
+
+python do_uki() {
+    import glob
+    import subprocess
+
+    # Construct the ukify command
+    ukify_cmd = ("ukify build")
+
+    deploy_dir_image = d.getVar('DEPLOY_DIR_IMAGE')
+
+    # Ramdisk
+    if d.getVar('INITRAMFS_IMAGE'):
+        initrd_image = d.getVar('INITRAMFS_IMAGE') + "-" + d.getVar('MACHINE')
+        baseinitrd = os.path.join(d.getVar('INITRAMFS_DEPLOY_DIR_IMAGE'), initrd_image)
+        for img in (".cpio.gz", ".cpio.lz4", ".cpio.lzo", ".cpio.lzma", ".cpio.xz", ".cpio"):
+            if os.path.exists(baseinitrd + img):
+                initrd = baseinitrd + img
+                break
+        if not initrd:
+            bb.fatal("initrd= must be specified to create unified kernel image.")
+        ukify_cmd += " --initrd=%s" % initrd
+
+    # Kernel Image.
+    # Note: systemd-boot can't handle compressed kernel image.
+    kernel_image = d.getVar('B') + "/" + d.getVar('KERNEL_OUTPUT_DIR') + "/Image"
+    kernel_version = d.getVar('KERNEL_VERSION')
+    if not os.path.exists(kernel_image):
+        bb.fatal("linux= must be specified to create unified kernel image.")
+    ukify_cmd += " --linux=%s --uname %s" % (kernel_image, kernel_version)
+
+    # Kernel cmdline
+    cmdline = "%s" %(d.getVar('KERNEL_CMDLINE_EXTRA'))
+    ukify_cmd += " --cmdline='%s'" % (cmdline)
+
+    # Architecture
+    target_arch = d.getVar('EFI_ARCH')
+    ukify_cmd += " --efi-arch %s" % target_arch
+
+    # Stub
+    stub = "%s/linux%s.efi.stub" % (deploy_dir_image, target_arch)
+    if not os.path.exists(stub):
+        bb.fatal("stub must be specified to create unified kernel image %s" %stub)
+    ukify_cmd += " --stub %s" % stub
+
+    # Custom UKI name
+    output_dir = d.getVar('D') + "/" +  d.getVar('EFILINUXDIR')
+    os.makedirs(output_dir, exist_ok=True)
+    output = output_dir + "/" + d.getVar('EFILINUXIMG')
+    ukify_cmd += " --output=%s" % output
+
+    # Set env to determine where bitbake should look for dynamic libraries
+    env = os.environ.copy() # get the env variables
+    env['LD_LIBRARY_PATH'] = d.expand("${RECIPE_SYSROOT_NATIVE}/usr/lib/systemd:${LD_LIBRARY_PATH}")
+
+    # Run the ukify command
+    subprocess.check_call(ukify_cmd, env=env, shell=True)
+}
+
+# Support uki as a package
+python () {
+    if not bb.data.inherits_class('nopackages', d):
+        d.appendVar("PACKAGES", " ${KERNEL_PACKAGE_NAME}-uki")
+}
+
+FILES:${KERNEL_PACKAGE_NAME}-uki = " \
+    /${EFILINUXDIR}/${EFILINUXIMG} \
+"
--- a/recipes-core/systemd/systemd-boot-native_254.4.bb
+++ b/recipes-core/systemd/systemd-boot-native_254.4.bb
@@ -1,0 +1,15 @@
+require recipes-core/systemd/systemd.inc
+
+inherit native
+
+RRECOMMENDS:${PN} += "python3-pefile-native"
+
+COMPATIBLE_HOST = "(aarch64.*|arm.*|x86_64.*|i.86.*)-linux"
+
+do_configure[noexec] = "1"
+do_compile[noexec] = "1"
+
+do_install () {
+	install -d ${D}${bindir}
+	install -m 0755 ${S}/src/ukify/ukify.py ${D}${bindir}/ukify
+}
--- a/recipes-kernel/images/esp-qcom-image.bb
+++ b/recipes-kernel/images/esp-qcom-image.bb
@@ -1,0 +1,20 @@
+DESCRIPTION = "EFI System Partition Image to boot Qualcomm boards"
+
+COMPATIBLE_HOST = '(x86_64.*|arm.*|aarch64.*)-(linux.*)'
+
+PACKAGE_INSTALL = " \
+    kernel-devicetree \
+    kernel-uki \
+    systemd-boot \
+    systemd-bootconf \
+"
+
+KERNELDEPMODDEPEND = ""
+KERNEL_DEPLOY_DEPEND = ""
+
+inherit image
+
+IMAGE_FSTYPES = "vfat"
+IMAGE_FSTYPES_DEBUGFS = ""
+
+LINGUAS_INSTALL = ""
--- a/recipes-kernel/linux/linux-linaro-qcom.inc
+++ b/recipes-kernel/linux/linux-linaro-qcom.inc
@@ -25,6 +25,7 @@ KERNEL_DEFCONFIG:arm ?= "${S}/arch/arm/configs/qcom_defconfig"
 KERNEL_CONFIG_FRAGMENTS += "${S}/kernel/configs/distro.config"
 
 inherit linux-qcom-bootimg
+inherit ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'uki', '', d)}
 
 kernel_conf_variable() {
 	sed -e "/CONFIG_$1[ =]/d;" -i ${B}/.config
--- a/recipes-kernel/linux/linux-yocto_%.bbappend
+++ b/recipes-kernel/linux/linux-yocto_%.bbappend
@@ -8,6 +8,12 @@ SRC_URI:append:qcom = " \
     file://qcom.scc \
 "
 
+# For boot.img
 QCOM_BOOTIMG = ""
 QCOM_BOOTIMG:qcom = "linux-qcom-bootimg"
 inherit ${QCOM_BOOTIMG}
+
+# For UKI (linux.efi)
+QCOM_UKI = ""
+QCOM_UKI:qcom = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'uki', '', d)}"
+inherit ${QCOM_UKI}
--- a/conf/machine/include/qcom-common.inc
+++ b/conf/machine/include/qcom-common.inc
@@ -37,3 +37,19 @@ SERIAL_CONSOLES ?= "115200;ttyMSM0"
 # image.  All our boards (except db410c) have 2GiB and db410c has 1GiB of RAM,
 # so this image would fit.
 INITRAMFS_MAXSIZE = "393216"
+
+# Use systemd-boot as the EFI bootloader
+EFI_PROVIDER = "systemd-boot"
+
+# Install packages at root of ESP
+EFI_PREFIX = ""
+
+# Location of Kernel and dtb inside ESP
+EFILINUXDIR ?= "${EFI_PREFIX}EFI/Linux"
+EFIDTBDIR ?= "${EFI_PREFIX}dtb"
+
+# Unified Kernel Image (UKI) name
+EFILINUXIMG ?= "linux-${MACHINE}.efi"
+
+# Place dtb at EFIDTDIR to seamlessly package
+KERNEL_DTBDEST = "${EFIDTBDIR}"


classes/linux-qcom-efiimg.bbclass
```diff
@@ -0,0 +1,203 @@
...
+
+    # Determine the final size in blocks accounting for some padding
+    BLOCKS=$(expr $(expr $SECTORS / 2) + ${BOOTIMG_EXTRA_SPACE})
+
+    # mkfs.vfat will sometimes use FAT16 when it is not appropriate,
+    # resulting in a boot failure. Use FAT32 for images larger
+    # than 512MB, otherwise let mkfs.vfat decide.
+    if [ $(expr $BLOCKS / 1024) -gt 512 ]; then
+        FATSIZE="-F 32"
+    fi
```
lumag, line 147: If this is a mkfs.vfat bug, it should be handled as a patch against dosfstools.

classes/linux-qcom-efiimg.bbclass
```diff
@@ -0,0 +1,203 @@
...
+do_uki[depends] += "${@ '${INITRAMFS_IMAGE}:do_image_complete' if d.getVar('INITRAMFS_IMAGE') else ''}"
+
+do_uki[dirs] = "${B}"
+
+python do_uki() {
+    import glob
+    import subprocess
+
+    # Construct the ukify command
+    ukify_cmd = d.getVar('RECIPE_SYSROOT') +"/" + d.getVar('USRBINPATH') +"/ukify build"
```
lumag, line 39: You shouldn't need to generate a path to `ukify` if you are doing things right.

classes/linux-qcom-efiimg.bbclass
```diff
@@ -0,0 +1,203 @@
...
+#   - an UEFI stub, from systemd-boot
+#   - the kernel Image
+#   - an initramfs
+#   - other metadata like kernel cmdline
+#
+# Step 2: Generate esp.bin as a fat image by packaging
+#   - UKI generated in Step 1.
+#   - systemd-boot efi
+#   - systemd-boot config
+#   - UEFI default dtb
```
lumag, line 14: Two steps, two different classes (or two recipes).

lumag (changes requested):
Comments inline. Also please take care to fix CI breakage on 32-bit platforms.

lumag:
Also, did you consider using the wic's bootimg-efi plugin?

lumag:
Also see https://lore.kernel.org/all/20230901233231.1109712-1-michelle.linto91@gmail.com/T/.

recipes-core/systemd/systemd-boot_%.bbappend
```diff
@@ -0,0 +1,18 @@
...
+# and provide ukify tool, systemd-boot.efi and efi.stub as part of sysroot.
+#
+# Consumers can get these files by adding a dependency on systemd-boot:do_populate_sysroot task.
+
+SYSROOT_PREPROCESS_FUNCS += "ukify_sysroot_preprocess"
+
+# Copy ukify tool, efi.stub and efi bootimg to staging directory.
+ukify_sysroot_preprocess () {
+    mkdir -p ${SYSROOT_DESTDIR}${bindir}
+    install -m 0755 ${S}/src/ukify/ukify.py ${SYSROOT_DESTDIR}${bindir}/ukify
```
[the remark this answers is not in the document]
quic-vkraleti: First, I tired the same. Compiling systemd-boot native is not trivial. Recipe need to be tweaked a lot to fit in both -native & -target compilations. Moreover, to build UKI, don't need any of the compiled artifacts except Ukify tool. Which is a script used as is from sources of systemd-boot. Also, from the comments in https://lore.kernel.org/all/20230901233231.1109712-1-michelle.linto91@gmail.com/T/.  I could see community is not preferring to have a -native recipe for systemd.  With all these in mind, felt copying ukify tool to staging directory is much simpler way to consume.

recipes-core/systemd/systemd-boot_%.bbappend
```diff
@@ -0,0 +1,18 @@
...
+    install ${B}/src/boot/efi/linux*.efi.stub ${SYSROOT_DESTDIR}${bindir}
+    install ${B}/src/boot/efi/systemd-boot*.efi ${SYSROOT_DESTDIR}${bindir}/${SYSTEMD_BOOT_IMAGE}
```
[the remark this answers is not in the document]
quic-vkraleti: At present boot.img is generated as soon as "bitbake virtual/kernel" is complete. To achieve the same even with esp.bin generation,  if efi.stub & systemd-boot.efi are picked from DEPLOYDIR, kernel compilation has to wait till deploy directory is populated with these. Moreover as ukify script is picked from STAGING, felt coping even these files into staging would acceptable.

classes/linux-qcom-efiimg.bbclass
```diff
@@ -0,0 +1,203 @@
...
+    #systemd-boot efi
+    mkdir -p ${EFIIMGDIR}${EFIDIR}
+    cp ${RECIPE_SYSROOT}${bindir}/systemd-boot${EFI_ARCH}.efi ${EFIIMGDIR}${EFIDIR}/${EFI_BOOT_IMAGE}
+
+    # Uefi dtb
+    if [ -f ${B}/${KERNEL_OUTPUT_DIR}/dts/${UEFI_DEFAULT_DTB} ]; then
+        cp ${B}/${KERNEL_OUTPUT_DIR}/dts/${UEFI_DEFAULT_DTB} ${EFIIMGDIR}
+    fi
+    # Create a vfat image
+    build_fat_img ${EFIIMGDIR} ${EFIIMG_DEPLOYDIR}/${EFIIMG}
```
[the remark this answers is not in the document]
quic-vkraleti: Sure, I'll inherit image-live class to generate esp.bin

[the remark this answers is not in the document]
lumag: You even don't need image-live. You only need a separate image recipe, which pulls in the uki kernel and some additional data.

recipes-core/systemd/systemd-boot_%.bbappend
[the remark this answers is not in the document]
quic-vkraleti: Should I make this native package as part of a new recipe instead of systemd-boot-native, by using exactly same code base?

recipes-core/systemd/systemd-boot_%.bbappend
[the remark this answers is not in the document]
lumag: It is fine to generate uki image during `bitbake virtual/kernel`. However the esp.bin is a separate story. It might contain different images. I might contain other files, e.g. UEFI shell.

recipes-core/systemd/systemd-boot_%.bbappend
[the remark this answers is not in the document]
lumag: systemd-boot-native seems like a right solution to me

quic-vkraleti:
Build error on arm7 is due to a bug in systemd code base. Faced same issue even without any of the above changes.  https://github.com/systemd/systemd/commit/cecbb162a3134b43d2ca160e13198c73ff34c3ef fixes it. I'll pull this change a .patch for now to mitigate.

classes/linux-qcom-efiimg.bbclass
quic-vkraleti: Sure.

lumag:
> Also see https://lore.kernel.org/all/20230901233231.1109712-1-michelle.linto91@gmail.com/T/.

After internal discussion, I think the uki.bbclass is a way to go.

recipes-core/systemd/systemd-boot_%.bbappend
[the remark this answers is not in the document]
quic-vkraleti: Added uki.bbclass to build UKI as part of kernel compilation.

classes/linux-qcom-efiimg.bbclass
quic-vkraleti: Even image-live.bbclass is handling it in the same way.

lumag: It doesn't mean that we have to copy it.

quic-vkraleti: Noted, I'll work on it and submit a patch to mtools. But not as part of this pull request.

recipes-core/systemd/systemd-boot_%.bbappend
```diff
@@ -0,0 +1,6 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
+
+# Fixes: https://github.com/systemd/systemd/issues/29381
+SRC_URI += " \
```
lumag, line 4: :qcom

recipes-core/systemd/systemd-boot-native_254.4.bb
```diff
@@ -0,0 +1,33 @@
...
+elaborate transactional dependency-based service control logic. It can \
+work as a drop-in replacement for sysvinit."
+
+LICENSE = "GPL-2.0-only & LGPL-2.1-only"
+LIC_FILES_CHKSUM = "file://LICENSE.GPL2;md5=751419260aa954499f7abaabaa882bbe \
+                    file://LICENSE.LGPL2.1;md5=4fbd65380cdd255951079008b364516c"
+
+SRCREV = "2e7504449a51fb38db9cd2da391c6434f82def51"
+SRCBRANCH = "v254-stable"
+SRC_URI = "git://github.com/systemd/systemd-stable.git;protocol=https;branch=${SRCBRANCH}"
```
lumag, line 19: Can we please use BBCLASSEXTEND instead?

classes/image-vfat.bbclass
```diff
@@ -0,0 +1,68 @@
...
+	# new image, it is safe to delete any previous image.
+	if [ -e ${FATIMG} ]; then
+		rm ${FATIMG}
+	fi
+
+	mkdosfs ${FATSIZE} -n ${BOOTIMG_VOLUME_ID} ${MKDOSFS_EXTRAOPTS} -C ${FATIMG} ${BLOCKS}
+
+	# Copy FATSOURCEDIR recursively into the image file directly
+	mcopy -i ${FATIMG} -s ${FATSOURCEDIR}/* ::/
+}
```
lumag, line 68: Is there any reason for using this approach instead of installing all the files via usual packages?

recipes-core/systemd/systemd-boot_%.bbappend
quic-vkraleti: I'll change SRC_URI to SRC_URI:qcom. I don't find anyone submitting this to Oe-Core. 'll do that as well.

recipes-core/systemd/systemd-boot-native_254.4.bb
quic-vkraleti: I tired that but it is mandating changes for most of the major tasks like, do_configure, do_compile, do_install. So felt it is much simpler to have systemd-boot-native.bb instead. Take a look attachment to know the changes I need to make.
[systemd-boot_254.4.bb](https://github.com/Linaro/meta-qcom/files/13441149/systemd-boot_254.4.bb.txt)

classes/image-vfat.bbclass
quic-vkraleti: Creating FAT filesystem images is slightly different from other formats. Unlike ext2/3/4, one can't place all files under a directory and use mkfs to get the image. FAT requires below steps 
1. Preallocate required space (using tools like dd). 
2. Format it using mkdosfs utility. 
3. Mount it and place content. Or copy content using mcopy.

Above bbclass is just doing the same.

recipes-core/systemd/systemd-boot_%.bbappend
lumag: Please do that now. There is no need to have generic patches in the BSP layer.

classes/image-vfat.bbclass
lumag: No. Existing framework already has `IMAGE_FSTYPES = "vfat"`. Please stick to it unless there is a good reason to implement it by hand.

recipes-core/systemd/systemd-boot-native_254.4.bb
lumag: This is fine from my POV. Another option might be to make meson inheritance conditional on the native build.
I do not want to see systemd-boot and systemd-boot-native to deviate.
And anyway, this should go to OE-core instead of meta-qcom.

lumag:
And by the play, please take a look at resolving the build error reported by the CI (for nodistro builds):

> linux-yocto PROVIDES virtual/kernel but was skipped: missing required distro feature 'systemd' (not in DISTRO_FEATURES)

Thank you.

classes/image-vfat.bbclass
quic-vkraleti: I tired that but with the generated images, couldn't boot device. Probably 'll go ahead with "wic" implementation as  I see bootimg-efi plugin in wic. If you are fine, let this bbclass in for now. Later with "wic" implementaiton i'll drop it. If this is something that bothers, I'l drop this commit and will just have UKI support for now.

recipes-core/systemd/systemd-boot-native_254.4.bb
quic-vkraleti: I dropped -native.bb file and folded required changes into systemd-boot_%.bbappend. I'll submit a change to OE-Core also to get Ukify tool out on the host.

lumag: Pease send these patches to OE-core. I'l wait for their decision.

classes/image-vfat.bbclass
lumag: What was the issue? Coud you compare the generated images?

lumag:
@quic-vkraleti Please send the systemd patches to the oe-core list.

classes/image-vfat.bbclass
quic-vkraleti: Previously I tried on Kirkstone. There is no support for vfat in Krikstone's image-types.bbclass. This support is added in master branch with [440fa508d362b8a449beb1b82dd999e980b753b7] (https://patchwork.yoctoproject.org/project/oe-core/patch/20230818163244.1733763-1-JPEWhacker@gmail.com/) I guess we are good now. I'll drop the custom logic to pack vfat.

lumag (changes requested):

recipes-core/systemd/systemd-boot_%.bbappend
```diff
@@ -4,3 +4,30 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
 SRC_URI:append:qcom = " \
             file://0001-meson-Pass-all-static-pie-args-to-linker.patch \
             "
+
+DEPENDS:class-native = " python3-pefile-native"
```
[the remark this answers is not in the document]
lumag: Should we change this RDEPENDS to RRECOMMENDS instead?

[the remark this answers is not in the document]
quic-vkraleti: Changed to RDEPENDS.

recipes-kernel/linux/linux-yocto_%.bbappend
```diff
@@ -7,7 +7,12 @@ FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}:"
 SRC_URI:append:qcom = " \
     file://qcom.scc \
 "
-
```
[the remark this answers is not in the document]
quic-vkraleti: Done

recipes-kernel/linux/linux-linaro-qcom.inc
```diff
@@ -24,7 +24,8 @@ KERNEL_DEFCONFIG:aarch64 ?= "${S}/arch/arm64/configs/defconfig"
 KERNEL_DEFCONFIG:arm ?= "${S}/arch/arm/configs/qcom_defconfig"
 KERNEL_CONFIG_FRAGMENTS += "${S}/kernel/configs/distro.config"
 
-inherit linux-qcom-bootimg
```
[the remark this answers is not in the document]
quic-vkraleti: Done

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -0,0 +1,50 @@
...
+IMAGE_FSTYPES_DEBUGFS = ""
+
+LINGUAS_INSTALL = ""
+
+EFI_PROVIDER = "systemd-boot"
+EFIIMGDIR = "${IMAGE_ROOTFS}/boot"
+EFIIMG = "${IMGDEPLOYDIR}/efi-${MACHINE}.bin"
+require conf/image-uefi.conf
+
+UEFI_DEFAULT_DTB ?= ""
```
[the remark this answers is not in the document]
quic-vkraleti: changes made to populate both idp & rb3 dtb files.

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -0,0 +1,50 @@
...
+    mkdir -p ${EFIIMGDIR}/${EFIDIR}
+    install -m 0644 ${DEPLOY_DIR_IMAGE}/systemd-${EFI_BOOT_IMAGE} ${EFIIMGDIR}/${EFIDIR}/${EFI_BOOT_IMAGE}
+
+    # Uefi dtb
+    if [ -f ${DEPLOY_DIR_IMAGE}/${UEFI_DEFAULT_DTB} ]; then
+        install -m 0644 ${DEPLOY_DIR_IMAGE}/${UEFI_DEFAULT_DTB} ${EFIIMGDIR}
+    fi
+
+    # Create a vfat image
+    build_fat_img ${EFIIMGDIR} ${EFIIMG}
```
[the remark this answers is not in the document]
quic-vkraleti: Generating vaft image using do_image_vfat task image_fstypes.bblcass. I needed to tinker the rootfs a bit to get appropriate directory structure UEFI is looking for.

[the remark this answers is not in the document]
lumag: Yes, please.

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -0,0 +1,56 @@
...
+            install -m 0644 ${DEPLOY_DIR_IMAGE}/$dtbf boot/
+        fi
+    done
+
+    # Avoid installing unnecessary files
+    rm -rf `ls -d ${IMAGE_ROOTFS}/* | grep -v boot`
+
+    # Move content to root of the image
+    mv boot/* ${IMAGE_ROOTFS}
+    rm -rf boot
```
lumag, line 54: If you set `EFI_PREFIX` to the empty string, you will not need this hack.

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -0,0 +1,56 @@
...
+
+    # UEFI devicetree(s)
+    for dtbf in ${UEFI_DEFAULT_DTB}; do
+        if [ -f ${DEPLOY_DIR_IMAGE}/$dtbf ]; then
+            install -m 0644 ${DEPLOY_DIR_IMAGE}/$dtbf boot/
+        fi
+    done
+
+    # Avoid installing unnecessary files
+    rm -rf `ls -d ${IMAGE_ROOTFS}/* | grep -v boot`
```
lumag, line 50: No. The point of using existing mechanism is to allow adding more files to the ESP partition depending on the additional requirements.

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -0,0 +1,56 @@
...
+# As rootfs gets tainted, regenerate always
+do_rootfs[nostamp] = "1"
+
+# Copy UKI image and UEFI dtb into rootfs.
+fakeroot rootfs_copy_uki () {
+    cd ${IMAGE_ROOTFS}
+
+    #Kernel uki efi
+    mkdir -p boot/EFI/Linux
+    install -m 0644 ${DEPLOY_DIR_IMAGE}/linux-${MACHINE}.efi boot/EFI/Linux/
```
lumag, line 40: Override KERNEL_IMAGEDEST and use the kernel package instead. You have to change kernel.bbclass to allow overriding the KERNEL_IMAGEDEST for the particular image type. This way the UKI image gets installed to `/`,  all other files go to the proper `/boot` location.

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -0,0 +1,56 @@
...
+
+    # UEFI devicetree(s)
+    for dtbf in ${UEFI_DEFAULT_DTB}; do
+        if [ -f ${DEPLOY_DIR_IMAGE}/$dtbf ]; then
+            install -m 0644 ${DEPLOY_DIR_IMAGE}/$dtbf boot/
+        fi
+    done
```
lumag, line 47: Why do we need them in the first place? I think the goal is to pass DTB files from the bootloader, so that the ESP doesn't contain DTB files at all.

lumag (changes requested):

recipes-kernel/images/esp-qcom-image.bb
quic-vkraleti: Got it. Will do.

recipes-kernel/images/esp-qcom-image.bb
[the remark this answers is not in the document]
lumag: Adding more DTB doesn't disturb other boards. However I'm more interested in why do you need these DTB in the first place.

recipes-kernel/images/esp-qcom-image.bb
quic-vkraleti: systemd-boot completely relies on the EFI environment to find a good devicetree.  It can't mange dtb on it's own. Rather it can only replace an existing devicetree in the corresponding EFI configuration table with the one embedded in the ".dtb" section.  [1] 

DTB files at root of ESP serves two purposes
 1. Provides a default dtb in EFI environment for systemd-boot to override.
 2. They serve as a fallback in case dtb in UKI can't be loaded by systemd-boot.

[1] https://manpages.debian.org/experimental/systemd-boot-efi/linuxx64.efi.stub.7.en.html 
"If a DeviceTree is embedded in the ".dtb" section, it replaces an existing DeviceTree in the corresponding EFI configuration table. systemd-stub will ask the firmware via the "EFI_DT_FIXUP_PROTOCOL" for hardware specific fixups to the DeviceTree."

lumag: `However, EBBR does require the system description to be supplied by the platform, not the OS.`, this is a quote from EBBR 2.1.0 spec. If we want for this platform to be compliant with the SystemReady IR (and thus EBBR), the DTB must be provided by the firmware (XBL).

lumag: Thus I expect that the board must be able to boot without any additional DTB files in the ESP partition.

lumag: I can accept this chunk if:
- it comes in a separate commit
- it is clearly documented as a temporal measure to be reverted later, ideally with the estimated date.
- it uses `KERNEL_DEVICETREE` variable instead of adding new one for the DTB list.

hashim-shiraz: > Thus I expect that the board must be able to boot without any additional DTB files in the ESP partition.

When UKI doesn't have dtb, firmware (uefi) is picking valid dtb from ESP partition. Does that violate the spec ?
It is still compliant since uefi is the one that is selecting dtb and installing in this case, no ?

Or placing dtbs in ESP partition violates ? I don't know if spec enumerates how the dtb has to be provisioned in such cases (outside UKI).

lumag: @hashim-shiraz This way the DT is still supplied by the OS. It is the operating system, who provides the ESP partition contents. For example, consider the SystemReady ACS. It will not provide the DT in the ESP, so the UEFI will not be able to find one.

calebccff: sd-boot now supports loading the DTB all by itself: https://github.com/systemd/systemd/pull/29749

> Provides a default dtb in EFI environment for systemd-boot to override.
...
> When UKI doesn't have dtb, firmware (uefi) is picking valid dtb from ESP partition

This is wrong. The DTB should be embedded in the firmware or loaded from some other "hidden" partition. Loading the DTB from the ESP is fine, but it must be optional as Dmitry said.

> If a DeviceTree is embedded in the ".dtb" section

This refers to embedding the DTB directly inside the sd-boot EFI binary, this would make the UKI image board-specific (as it would override the DTB provided by the firmware). Is this really your intention?

recipes-kernel/images/esp-qcom-image.bb
lumag: ... Alternatively you can make UKI kernel go to the `${EFI_PREFIX}` instead of `KERNEL_IMAGEDEST`, as it is a special kind of the image. Probably it is up to the review at the oe-core ML.

recipes-kernel/images/esp-qcom-image.bb
lumag: If I remember correctly, one of the goals was to shift towards SystemReady IR and thus supporing generic kernel images.  Is that still the plan? Having a board-specific UKI mostly defeats the idea of using UEFI and systemd-boot.

lumag: We are really puzzled here, because with the board-specific UKI basically there is no need to go through all depths of UEFI and sd-boot support. You can continue using ABL. My opinion is that the DT should be provided firmware from one of partitions and the kernel image should be completely generic. Maybe I misunderstand you.

I noticed that uki.bbclass forces first dtb into the UKI. I have commented on the mailing list. We should drop that clause. First DTB is `apq8016-sbc.dtb`, which means nothing for other boards we support.

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -0,0 +1,40 @@
...
+
+# Copy UKI image and UEFI dtb into ESP
+fakeroot copy_uki_to_rootfs () {
+    #Kernel uki efi
+    mkdir -p ${IMAGE_ROOTFS}${EFILINUXDIR}
+    install -m 0644 ${DEPLOY_DIR_IMAGE}/linux-${MACHINE}.efi ${IMAGE_ROOTFS}${EFILINUXDIR}
+
+    # UEFI devicetree(s)
+    mkdir -p ${IMAGE_ROOTFS}/${EFIDTBDIR}
+    for DTB in ${KERNEL_DEVICETREE}; do
```
lumag, line 34: Set `KERNEL_DTBDEST` to `EFIDTBDIR` and install the `kernel-devicetree` instead?

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -0,0 +1,40 @@
...
+
+do_rootfs[depends] += " \
+    virtual/kernel:do_deploy \
+    "
+
+# Copy UKI image and UEFI dtb into ESP
+fakeroot copy_uki_to_rootfs () {
+    #Kernel uki efi
+    mkdir -p ${IMAGE_ROOTFS}${EFILINUXDIR}
+    install -m 0644 ${DEPLOY_DIR_IMAGE}/linux-${MACHINE}.efi ${IMAGE_ROOTFS}${EFILINUXDIR}
```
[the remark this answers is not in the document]
quic-vkraleti: ok, I'll update uki.bbclass to create a package instead of deploying into DEPLOY_DIR_IMAGE.

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -0,0 +1,40 @@
...
+    systemd-boot \
+    systemd-bootconf \
+"
+
+inherit image
+
+IMAGE_FSTYPES = "vfat"
+IMAGE_FSTYPES_DEBUGFS = ""
+
+LINGUAS_INSTALL = ""
```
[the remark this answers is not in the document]
quic-vkraleti: After converting devicetree & linux.efi into packages, 'll set KERNEL_DEPLOY_DEPEND to empty.

recipes-kernel/images/esp-qcom-image.bb
quic-vkraleti: I tired but found an issue in linux-qcom-boot.img. I'll address it first and then fix this.

lumag:
@quic-vkraleti after these refactorings the ESP recipe together with the EFI_* configuration starts to look more and more like a distro feature. Maybe we repurpose it to the distro layer? (Again, maybe later, not now, if you don't have a distro layer yet).

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -0,0 +1,40 @@
...
+
+require conf/image-uefi.conf
+# Location of Kernel and dtb inside ESP
+EFILINUXDIR ?= "${EFI_PREFIX}/EFI/Linux"
+EFIDTBDIR ?= "${EFI_PREFIX}/dtb"
```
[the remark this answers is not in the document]
lumag: Sure

classes/linux-qcom-bootimg.bbclass
```diff
@@ -132,7 +132,7 @@ python do_qcom_img_deploy() {
 do_qcom_img_deploy[depends] += "skales-native:do_populate_sysroot"
 do_qcom_img_deploy[vardeps] = "QCOM_BOOTIMG_PAGE_SIZE QCOM_BOOTIMG_KERNEL_BASE KERNEL_CMDLINE_EXTRA QCOM_BOOTIMG_ROOTFS"
 
-addtask qcom_img_deploy after do_populate_sysroot do_packagedata bundle_initramfs before do_deploy
+addtask qcom_img_deploy after do_populate_sysroot do_packagedata do_bundle_initramfs before do_deploy
```
lumag, line 135: Separate commit too.
Each logical set of changes should go to a separate commit. This is a rule of thumb.

lumag (changes requested):
The meson fix was merged to oe-core, please drop it.

I think that changes to qcom-common.inc and the esp-image belong to the distro layer. They describe distro-specific solutions, which might change from distro to distro. Consider e.g. Yoe or poky deciding to use grub. This decision will conflict with the qcom-common changes / setup. Please move these chunks to your distro instead.

With these changes done I will merge this PR.

lumag:
I'm going to merge this now.

- Please _work_ with upstream to get systemd-boot-native into the oe-core. I want to get it dropped from this layer before the end of the year. This might require discussions on the mailing list, not the send-and-forget approach.
- Please follow the upstream developments on the UKI support.
- I still have impression that ESP image should be moved into the distro layer. The whole set of `EFI_*` variables look like distro-level decisions. I have asked one of my colleagues for his opinion.

classes/uki.bbclass
```diff
@@ -0,0 +1,90 @@
...
+#   - the kernel Image
+#   - an initramfs
+#   - other metadata like kernel cmdline
+
+DEPENDS:append = " \
+            systemd-boot-native \
+            python3-native \
+            python3-pefile-native \
+            "
+
```
mikkorapeli-linaro, line 13: REQUIRED_DISTRO_FEATURES += "usrmerge systemd"
inherit features_check
?

lumag: systemd should not be required for systemd-boot (or for UKI).

classes/uki.bbclass
```diff
@@ -0,0 +1,90 @@
...
+require conf/image-uefi.conf
+
+inherit python3native
+
+do_install[depends] += " \
+    systemd-boot:do_deploy \
+    "
+do_install[depends] += "${@ '${INITRAMFS_IMAGE}:do_image_complete' if d.getVar('INITRAMFS_IMAGE') else ''}"
+do_uki[dirs] = "${B}"
+do_install[postfuncs] += "do_uki"
```
mikkorapeli-linaro, line 23: I'm leaning towareds adding the do_uki() task to
addtask uki after do_rootfs before do_image_complete do_image_wic
This way sstate cache has the files. Maybe they are cached also with your approach but I'm not sure. Worth a comment in commit message, and of course followup if upstream poky applys a similar class.

lumag: I hope that we get uki.bbclass in oe-core.

classes/uki.bbclass
```diff
@@ -0,0 +1,90 @@
...
+
+python do_uki() {
+    import glob
+    import subprocess
+
+    # Construct the ukify command
+    ukify_cmd = ("ukify build")
+
+    deploy_dir_image = d.getVar('DEPLOY_DIR_IMAGE')
```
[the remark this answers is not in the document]
quic-vkraleti: This is to read from DEPLOY_DIR not to write into it. I couldn't find any better place to get ramdisk from.

classes/uki.bbclass
mikkorapeli-linaro: Yep, I'm also using an uki.bbclass derived from the upstream submitted proposal with some fixes/changes on top for secure boot (sbsign-native) and dm-verity, but it's trickty to get all the pieces together. For example had to split wic image creation from uki and dm-verity generation recipes since the order of functions/tasks was not possible to set correctly. WIP still on my side too.

lumag:
@mikkorapeli-linaro thank you for your comments!

PR #559, opened 2023-12-07
uki: drop INITRAMFS_IMAGE support

lumag:
Enabling INITRAMFS_IMAGE and UKI results in the dependency loop. Initramfs depends on the kernel to be packaged, while the kernel's do_install gets dependency on the initramfs do_image_complete task.

Drop the whole INITRAMFS_IMAGE support from the uki.bbclass. One should use systemd-boot's initrd support natively.

### DIFF
--- a/classes/uki.bbclass
+++ b/classes/uki.bbclass
@@ -18,7 +18,6 @@ inherit python3native
 do_install[depends] += " \
     systemd-boot:do_deploy \
     "
-do_install[depends] += "${@ '${INITRAMFS_IMAGE}:do_image_complete' if d.getVar('INITRAMFS_IMAGE') else ''}"
 do_uki[dirs] = "${B}"
 do_install[postfuncs] += "do_uki"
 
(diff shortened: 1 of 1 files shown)


lumag:
cc @sbanerjee-quic

sbanerjee-quic:
> cc @sbanerjee-quic

@lumag we will review and feedback. We have a long weekend.

cc @quic-vkraleti

lumag:
@quic-vkraleti I'm not sure, what is Type1 / Type2 here. A reference would be a appreciated. Anyway, currently built-in initrd support is broken, try building such config on your own.

quic-vkraleti:
> @quic-vkraleti I'm not sure, what is Type1 / Type2 here. A reference would be a appreciated. Anyway, currently built-in initrd support is broken, try building such config on your own.

https://uapi-group.org/specifications/specs/boot_loader_specification/#the-boot-loader-specification describes details about Type#1 & Type#2 specifications. Goal is to support Type#2 and this mandates initrd to be part of UKI. 

Regd broken built-in initrd support,  can you share more details on how to reproduce? As I see even kernel.bblcass is fetching initrd from INITRAMFS_DEPLOY_DIR_IMAGE, which is a soft assignment to DEPLOY_DIR_IMAGE. uki.bbclass is also using same method to get initrd.

quic-vkraleti:
> For me it was simple, I enabled `INITRAMFS_IMAGE = "initramfs-test-image"` and got a errors regarding the task dependency loop.

From build POV, ramdisk image is also another image which by default it pulls in a dependency on do_packagedata task of kernel. Here uki.bbclass is trying to create kernel-uki package, that need ramdisk image to complete. This is the cause for circular dependency. To address this, need to stop creating kernel-uki package and place efi image in deploydir instead.

lumag:
Closed in favour of #562  which provides a better, more complex solution

PR #569, opened 2024-01-18
linux-yocto: enable ARM64_VA_BITS_48 for QCM6490

adhudase:
Enable CONFIG_ARM64_VA_BITS_48 for qcm6490. Currently, boot to shell is broken with systemd-boot  flow and it is observed that CONFIG_ARM64_VA_BITS_48 is needed.

CONFIG_ARM64_VA_BITS_39 is enabled by default on linux-yocto unlike upstream kernel which uses CONFIG_ARM64_VA_BITS_48.

Further debugging is in progress to figure out this dependency.


lumag:
Is broken ... how? What symtoms do you observe? What gets broken?

lumag:
@sbanerjee-quic
Currently there are 21 patches marked as Pending and 8 marked as Submitted. A quick check shows that some of the patches are now accepted. Could you please update Upstream-Status for the patches? Also in the longer-term goal I'd like to see the amount of 'Pending' patches going lower, not higher. Is there any progress on submitting them?

sbanerjee-quic:
> @sbanerjee-quic Currently there are 21 patches marked as Pending and 8 marked as Submitted. A quick check shows that some of the patches are now accepted. Could you please update Upstream-Status for the patches? Also in the longer-term goal I'd like to see the amount of 'Pending' patches going lower, not higher. Is there any progress on submitting them?

ACK, will respond with data

lumag:
@adhudase I'd still like to view what gets broken (because I have tested the linu-yocto on RB1, RB2, RB5 and dragonboards). My generic advise would be to follow https://github.com/torvalds/linux/commit/211102d85f71dc5a01dde771e3f833f60b494661 and to push this config option to linux-yocto itself to be enabled by default for arm64. This is enabled in defconfig, so 80% of builds use this setting. All other options receive much less testing.

adhudase:
Hi @lumag , I just realized that meta-qcom is not using any linux-yocto config (yocto-kernel-cache). It is set as **nocfg** as below.

```
linux-yocto\bsp\qcom-armv8a\qcom-armv8a-standard.scc
include ktypes/standard/standard.scc **nocfg**
```

So when we post our config change, it won't come as default to meta-qcom; we need to use that cfg/scc explicitly in our bsp. Is that okay?

Also, I do see CONFIG_ARM64_VA_BITS_48 enabled in other bsps in yocto-kernel-cache but not in common fragments.

Please correct me if I am wrong and guide us to resolve this issue.

Thanks

lumag:
Ack. sounds good to me. Please move it to a generic qcom.cfg file.

adhudase:
Hi @lumag, updated commit to move config to qcom.cfg. Please help review and merge it. Thanks

PR #574, opened 2024-01-19
linux-yocto: initial support for qcom-armv7a machine

lumag:
Add support for the qcom-armv7a machine. Tested on ifc6410 and (up to some extent) on db8074. The msm8974 needs additional patches to fix the interconnect driver.


adhudase:
LGTM

PR #590, opened 2024-04-30
Fix playback using WSA881x codecs

lumag:
Backport patch fixing playback issues on the RB5 board, which uses WSA881x codec.


lumag:
cc @adhudase

adhudase (approved):
LGTM. Thank you!

PR #596, opened 2024-05-24
gpsd: fix compilation with update OE layers

lumag:
GCC 14 and newer glibc are more strict regarding the headers. Add inlcude of the required header.


lumag:
cc @adhudase

PR #618, opened 2024-07-30
Enhance cmdline handling in linux-qcom-uki.bb

quic-vkraleti:
Construct kernel cmdline by reading QCOM_BOOTIMG_ROOTFS, SERIAL_CONSOLES and KERNEL_CMDLINE_EXTRA variables and update the same into uki cmdline.

### DIFF
--- a/recipes-kernel/images/linux-qcom-uki.bb
+++ b/recipes-kernel/images/linux-qcom-uki.bb
@@ -51,9 +51,27 @@ do_compile() {
     ukify_cmd="$ukify_cmd --uname ${KERNEL_VERSION}"
 
     # Kernel cmdline
-    if [ -n "${KERNEL_CMDLINE_EXTRA}" ]; then
-        ukify_cmd="$ukify_cmd --cmdline='${KERNEL_CMDLINE_EXTRA}'"
+    cmdline=""
+    if [ -n "${QCOM_BOOTIMG_ROOTFS}" ]; then
+        cmdline="$cmdline root=${QCOM_BOOTIMG_ROOTFS} rw rootwait"
     fi
+
+    if [ ! -z "${SERIAL_CONSOLES}" ]; then
+        tmp="${SERIAL_CONSOLES}"
+        for entry in $tmp ; do
+            baudrate=`echo $entry | sed 's/\;.*//'`
+            tty=`echo $entry | sed -e 's/^[0-9]*\;//' -e 's/\;.*//'`
+	    console="$tty","$baudrate"n8
+            cmdline="$cmdline console=$console"
+        done
+    fi
+
+    if [ -n "${KERNEL_CMDLINE_EXTRA}" ]; then
+        cmdline="$cmdline ${KERNEL_CMDLINE_EXTRA}"
+    fi
+
+    printf '%s' "$cmdline" > ${B}/cmdline
+    ukify_cmd="$ukify_cmd --cmdline @${B}/cmdline"
 
     # Architecture
     ukify_cmd="$ukify_cmd --efi-arch ${EFI_ARCH}"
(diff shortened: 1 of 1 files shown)


recipes-kernel/images/linux-qcom-uki.bb
```diff
@@ -51,10 +51,28 @@ do_compile() {
     ukify_cmd="$ukify_cmd --uname ${KERNEL_VERSION}"
 
     # Kernel cmdline
+    cmdline=""
+    if [ -n "${QCOM_BOOTIMG_ROOTFS}" ]; then
+        cmdline="$cmdline root=${QCOM_BOOTIMG_ROOTFS} rw rootwait"
+    fi
```
lumag, line 57: `QCOM_BOOTIMG_ROOTFS` varies per dtb. Please take that into account.

recipes-kernel/images/linux-qcom-uki.bb
```diff
@@ -51,10 +51,28 @@ do_compile() {
...
+
+    if [ ! -z "${SERIAL_CONSOLES}" ]; then
+        tmp="${SERIAL_CONSOLES}"
+        for entry in $tmp ; do
+            baudrate=`echo $entry | sed 's/\;.*//'`
+            tty=`echo $entry | sed -e 's/^[0-9]*\;//' -e 's/\;.*//'`
+	    console="$tty","$baudrate"n8
+            cmdline="$cmdline console=$console"
+        done
+    fi
```
lumag, line 67: We have similar code in `linux-qcom-bootimg.bbclass`, please consider extracting a function to OE-Core / `kernel.bbclass`.

For the reference:
```Python
        consoles = ' '.join(map(lambda c: "console=%(tty)s,%(rate)sn8" % dict(zip(("rate", "tty"), c.split(';'))), getVarDTB("SERIAL_CONSOLES").split()))
```

lumag (changes requested):

recipes-kernel/images/linux-qcom-uki.bb
quic-vkraleti: KERNEL_CMDLINE_EXTRA is a meta-qcom variable. OE-Core kernel.bbclass don't support updating cmdline. For now do consider kernel cmdline setting to be a meta-qcom specific implementation only. I'll try to submit a proposal to oe-core but not sure how far it'll fly.

recipes-kernel/images/linux-qcom-uki.bb
quic-vkraleti: @lumag if you recollect discussions in https://github.com/Linaro/meta-qcom/pull/548, we agreed to make UKI, dtb agnostic and keep a separate dtb section in ESP. We also discussed an approach to have a new partition to host dtb files.  So updating cmdline in UKI that varies per dtb won't be feasible.

quic-vkraleti: @lumag submitted PR: https://github.com/Linaro/meta-qcom/pull/620 to move DTB away from esp and create a separate dtb image. With this dtb and uki are completely isolated.

recipes-kernel/images/linux-qcom-uki.bb
lumag: Ack. Please cc me if you post such a patch.

recipes-kernel/images/linux-qcom-uki.bb
lumag: The issue is that all those variables wary from machine to machine. So you end up with an image that is expected to be generic, but in reality it isn't.

lumag: I don't have a good solution for the issue. maybe a setup-specific DT overlay to be applied by the EFI/bootloader. Maybe a bootloader should set them.

lumag: I'm going to merge this, but let's think for a longer-term plan.

PR #620, opened 2024-08-08
Generate vfat image by packing device-tree blob.

quic-vkraleti:
This PR introduces a bbclass, which upon inherited by kernel will produce a vfat image for each device-tree blob generated by kernel.

This helps to meet one of the ARM System Ready IR requirements of keeping dtb out of kernel binary.

### DIFF
--- a/classes/linux-qcom-dtbbin.bbclass
+++ b/classes/linux-qcom-dtbbin.bbclass
@@ -1,0 +1,33 @@
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+DTBBIN_DEPLOYDIR = "${WORKDIR}/qcom_dtbbin_deploy-${PN}"
+DTBBIN_SIZE ?= "4096"
+
+do_qcom_dtbbin_deploy[depends] += "dosfstools-native:do_populate_sysroot mtools-native:do_populate_sysroot"
+do_qcom_dtbbin_deploy[cleandirs] = "${DTBBIN_DEPLOYDIR}"
+do_qcom_dtbbin_deploy() {
+    for dtbf in ${KERNEL_DEVICETREE}; do
+        bbdebug 1 " combining: $dtbf"
+        dtb=`normalize_dtb "$dtbf"`
+        dtb_ext=${dtb##*.}
+        dtb_base_name=`basename $dtb .$dtb_ext`
+        mkdir -p ${DTBBIN_DEPLOYDIR}/$dtb_base_name
+        cp ${D}/${KERNEL_DTBDEST}/$dtb_base_name.dtb ${DTBBIN_DEPLOYDIR}/$dtb_base_name/combined-dtb.dtb
+        mkfs.vfat $@ -C ${DTBBIN_DEPLOYDIR}/dtb-${dtb_base_name}-image.vfat ${DTBBIN_SIZE}
+        mcopy -i "${DTBBIN_DEPLOYDIR}/dtb-${dtb_base_name}-image.vfat" -vsmpQ ${DTBBIN_DEPLOYDIR}/$dtb_base_name/* ::/
+        rm -rf ${DTBBIN_DEPLOYDIR}/$dtb_base_name
+    done
+}
+addtask qcom_dtbbin_deploy after do_populate_sysroot do_packagedata before do_deploy
+
+# Setup sstate, see deploy.bbclass
+SSTATETASKS += "do_qcom_dtbbin_deploy"
+do_qcom_dtbbin_deploy[sstate-inputdirs] = "${DTBBIN_DEPLOYDIR}"
+do_qcom_dtbbin_deploy[sstate-outputdirs] = "${DEPLOY_DIR_IMAGE}"
+
+python do_qcom_dtbbin_deploy_setscene () {
+    sstate_setscene(d)
+}
+addtask do_qcom_dtbbin_deploy_setscene
+
+do_qcom_dtbbin_deploy[stamp-extra-info] = "${MACHINE_ARCH}"
--- a/recipes-kernel/linux/linux-yocto_%.bbappend
+++ b/recipes-kernel/linux/linux-yocto_%.bbappend
@@ -12,3 +12,8 @@ SRC_URI:append:qcom = " \
 QCOM_BOOTIMG = ""
 QCOM_BOOTIMG:qcom = "linux-qcom-bootimg"
 inherit ${QCOM_BOOTIMG}
+
+# For dtb.bin
+QCOM_DTBBIN = ""
+QCOM_DTBBIN:qcom = "linux-qcom-dtbbin"
+inherit ${QCOM_DTBBIN}


recipes-kernel/images/dtb-qcom-image.bb
```diff
@@ -0,0 +1,21 @@
+DESCRIPTION = "DTB Partition Image to boot Qualcomm boards"
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/BSD-3-Clause-Clear;md5=7a434440b651f4a472ca93716d01033a"
+
+COMPATIBLE_HOST = '(x86_64.*|arm.*|aarch64.*)-(linux.*)'
```
ricardosalveti, line 5: Do we need to define compatible host? This is just creating a vfat image.

ricardosalveti:
I would also split this change in at least 2 commits, one adding the recipe and another the image.

recipes-kernel/images/linux-qcom-combinedtb.bb
```diff
@@ -0,0 +1,49 @@
...
+do_patch[noexec] = "1"
+
+SRC_URI = ""
+
+do_configure[depends] += " \
+    virtual/kernel:do_deploy \
+    "
+
+python do_compile() {
+    import os, shutil, subprocess
```
quaresmajose, line 23: the `subprocess` is not used so it can be dropped

recipes-kernel/images/linux-qcom-combinedtb.bb
```diff
@@ -0,0 +1,49 @@
...
+        if not file.endswith(".dtb"):
+            continue
+        kdtb = d.getVar('DEPLOY_DIR_IMAGE') + "/" + file
+        if not os.path.islink(kdtb):
+            #Append each dtb file to combined-dtb.dtb
+            combined_dtb = os.path.join(dtb_dir, "combined-dtb.dtb")
+            with open(combined_dtb, 'ab') as fout:
+                with open(kdtb, 'rb') as fin:
+                    bb.debug(1, "combining: %s" % kdtb)
+                    shutil.copyfileobj(fin, fout)
```
lumag, line 38: Isn't it just `cat ${DEPLOY_DIR_IMAGE} > ${B}/combined-dtb.dtb` ? Shell script will be shorter.

recipes-kernel/images/linux-qcom-combinedtb.bb
```diff
@@ -0,0 +1,49 @@
+SUMMARY = "Qualcomm linux kernel combined dtb creation"
+
+DESCRIPTION = "Generates a single dtb by combining multiple dtb files generated by kernel."
+
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/BSD-3-Clause-Clear;md5=7a434440b651f4a472ca93716d01033a"
+
+COMPATIBLE_HOST = '(arm.*|aarch64.*)-(linux.*)'
```
lumag, line 8: No need to.

recipes-kernel/images/linux-qcom-combinedtb.bb
```diff
@@ -0,0 +1,49 @@
+SUMMARY = "Qualcomm linux kernel combined dtb creation"
+
+DESCRIPTION = "Generates a single dtb by combining multiple dtb files generated by kernel."
+
+LICENSE = "BSD-3-Clause-Clear"
```
lumag, line 5: MIT, please

lumag (changes requested):
- please split into two commits
- please describe, how the dtb us going to be selected. Refer existing standards, etc. Note, that using qcom,msm-id and qcom,board-id is generally frowned upon by DT maintainers and are forbidden to be used for new devices.

As a general note, I dislike the idea of a single merged dtb image just as a concatenation of all DT files. It is not easily extensible, overriding it also is not easy. If not for the SecureBoot, I'd have suggested following DtbLoader way (dtb/CHID.dtb). However with SecureBoot being an important thing the DTB provisioning mechanism should leave place for the signature verification. I would suggest exploring either a set of PE files, each having one or several DTB sections and a signature, a FIT image with a signature or just external .pk7 signature being placed next to each dtb file.

recipes-kernel/images/linux-qcom-combinedtb.bb
quic-vkraleti: Removed COMPATIBLE_HOST.

recipes-kernel/images/linux-qcom-combinedtb.bb
quic-vkraleti: Done.

recipes-kernel/images/linux-qcom-combinedtb.bb
quic-vkraleti: No it is not a simple concatenation of all files under DEPLOY_DIR_IMAGE. There are symlinks of each dtb and those need to be omitted. There could be more files as well which other recipes may put under DEPLOY_DIR_IMAGE.

recipes-kernel/images/linux-qcom-combinedtb.bb
quic-vkraleti: Done.

recipes-kernel/images/dtb-qcom-image.bb
quic-vkraleti: Removed COMPATIBLE_HOST.

recipes-kernel/images/linux-qcom-combinedtb.bb
lumag: Still a shell script with a for loop will be shorter and easier to understand.

quic-vkraleti: Ok, I'll convert this into a shell script and update.

lumag: Other point: consider using`KERNEL_DEVICETREE` instead of looping over all dtb files

quic-vkraleti: Well I thought the same but couldn't do it because of the way dtbs are added in 
https://github.com/Linaro/meta-qcom/blob/master/conf/machine/qcom-armv8a.conf#L26

At preset DTBs are kernel recipe specific. So using KERNEL_DEVICETREE is not possible. Thus ended up looping over all generated files.

lumag: Ack could you please leave a Todo comment with an explanation.

quic-vkraleti: Added TODO in linux-qcom-combined.bb file. Also updated DESCRIPTION in dtb-qcom-image.bb

recipes-kernel/images/dtb-qcom-image.bb
```diff
@@ -0,0 +1,25 @@
+SUMMARY = "DTB Partition Image to boot Qualcomm boards"
+
+DESCRIPTION = " For DTB selection on QCOM boards with EFI boot support, all probable DTBs \
+                are concatenated as a VFAT image and kept in a separate partition. UEFI \
+                accesses this partition, loads a suitable DTB based on qcom,board-id. \
+                There is no DTB presence in ESP image. UKI have no .dtb section either."
```
ricardosalveti, line 6: This last line can be removed, as it is outside the scope of this image. An user could later have both the dtb partition (with concatenated dtbs) and UKIs with OS-provided dtbs.

quic-vkraleti: Removed.

quic-vkraleti:
> * please split into two commits
Done.

> * please describe, how the dtb us going to be selected. Refer existing standards, etc. Note, that using qcom,msm-id and qcom,board-id is generally frowned upon by DT maintainers and are forbidden to be used for new devices.
Updated description in dtb-qcom-image.bb describing how combined-dtb is put into use. 

> 
> As a general note, I dislike the idea of a single merged dtb image just as a concatenation of all DT files. It is not easily extensible, overriding it also is not easy. If not for the SecureBoot, I'd have suggested following DtbLoader way (dtb/CHID.dtb). However with SecureBoot being an important thing the DTB provisioning mechanism should leave place for the signature verification. I would suggest exploring either a set of PE files, each having one or several DTB sections and a signature, a FIT image with a signature or just external .pk7 signature being placed next to each dtb file.

Discussions in progress with other stake holders on alternative proposals like "keep a DTB in UKI and pass the same to Linux instead of what is provided by UEFI". Once there is a conclusion, will share a PR for the same.

> Also please make sure that your new recipes are included into the CI build.
Sure. What is the best place for adding a dependency on esp & dtb images to get built by default?

recipes-kernel/images/linux-qcom-combinedtb.bb
quic-vkraleti: Done.

lumag:
Yes, I saw that. However it is not acceptable. The `qcom,board-id` property isn't allowed to be used on the affected SoCs, see [arm/qcom.yaml](https://www.kernel.org/doc/Documentation/devicetree/bindings/arm/qcom.yaml)

lumag:
> keep a DTB in UKI and pass the same to Linux instead of what is provided by UEFI

This is good as an override, not I don't think it will be acceptable for System ready. Please review the old DtbLoader proposal of passing multiple dtb files via the <ESP>/dtb/<CHID>.dtb, where CHID is generated from the platform properties (like Windows and fwupdate do).

lumag:
> Sure. What is the best place for adding a dependency on esp & dtb images to get built by default?

Please extend the CI config in this repo. Be sure not to enable new images for the branches which don't have corresponding recipe.

ricardosalveti:
> This is good as an override, not I don't think it will be acceptable for System ready.

SR just requires the firmware to provide a DTB that is compliant with the spec, the issue here is that the DTB provided by the firmware is not going to be compliant, but the way it gets loaded is outside the SR scope. It is possible to have the DTB available via the firmware directory (ESP/FIRMWARE), but that is optional.

IIRC the idea is also to have the firmware providing a generic and upstream aligned device tree when the partition is not available (or empty), which can be a tradeoff until we're done with the the board-id / compatible string upstream discussion.

ricardosalveti:
> Discussions in progress with other stake holders on alternative proposals like "keep a DTB in UKI and pass the same to Linux instead of what is provided by UEFI". Once there is a conclusion, will share a PR for the same.

Supporting DTB loading via UKI is going to be needed anyway, even if the firmware is or not providing any DTB before UKI. DTB via UKI is basically allowing the OS to take control of the DTB content (which is also a valid use case).

lumag (changes requested):

recipes-kernel/images/dtb-qcom-image.bb
```diff
@@ -0,0 +1,25 @@
...
+
```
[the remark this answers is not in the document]
lumag: Or rather ..  isn't a proper/sensible package description

recipes-kernel/images/linux-qcom-combinedtb.bb
```diff
@@ -0,0 +1,38 @@
...
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+DEPENDS = "dtc-native"
+
+do_fetch[noexec] = "1"
+do_unpack[noexec] = "1"
+do_patch[noexec] = "1"
+
+SRC_URI = ""
```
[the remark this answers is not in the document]
quic-vkraleti: No need of? setting empty SRC_URI? marking do_fetch, do_unpack, do_patch as noexec? Or both? These are added to avoid potential setscene errors that can happen while eSDK generation.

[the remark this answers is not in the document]
lumag: Actually both. 
What kind of errors did you observe and how to reproduce them?

ricardosalveti:
> > the DTB provided by the firmware is not going to be compliant
> 
> Could you please be more specific, which DTB is it and why isn't it compliant? Can we replace it with the proper, compliant DTB?

Selection done by the boot firmware is still done with qcom,board-id, for example. Moving away from that is a larger discussion that is going in parallel.

lumag:
I think there are two distinct questions: 
- a DTB to be used by default (e.g. for SR tests)
- a DTB to be used for booting the feature full system

From the SR perspective it should be fine if the UEFI provides a not-so-perfect DTB, especially if this DTB can later be upgraded via different mechanism, like fwupd

quic-vkraleti:
> @ricardosalveti can we flash a single DTB and ignore selection completely?

AFAIK, even if a single DTB is flashed, board-id check would still happen. Only way to avoid this check is by placing dtb inside UKI.

lumag:
Oh, my. It would be really nice if such checks were skipped if there is just a single DTB. I think in the past we had to patch ABL to disable such checks.

ricardosalveti:
We had discussions to allow:
* dtb partition with a series of device tree files, selected via board-id (mostly used by downstream)
* firmware dtb that from linux upstream (that can be updated with the firmware)
* supporting OS-provided dtb via grub/systemd-boot/UKI

But that firmware is still to be release iirc.

ricardosalveti:
> We had discussions to allow:
> 
> * dtb partition with a series of device tree files, selected via board-id (mostly used by downstream)
> * firmware dtb that from linux upstream (that can be updated with the firmware, used when dtb partition is empty)
> * supporting OS-provided dtb via grub/systemd-boot/UKI
> 
> But that firmware is still to be release iirc.

lumag:
Well, from my POV the best option is a separate partition with a single DTB file, loaded by the firmware. Then we have a sane default, users can manually change it, fwupd can handle updates.
And if at some point there is a need to go to multiple files, UEFI can use compat strings to make a selection. The board-id will never fly and will never be accepted upstream as a generic solution. In the mobile world nearly every phone vendor mishandles those options, telling that the phone is MTP (or QRD).

Of course grub/systemdboot still can override the loaded DTB if it is specified in the config.

ricardosalveti:
> Well, from my POV the best option is a separate partition with a single DTB file, loaded by the firmware. Then we have a sane default, users can manually change it, fwupd can handle updates.

That is indeed a goal, but need to check on the details there as the name (and possibly board-id) needs to match what is expected by the firmware.

lumag:
Name -- agreed, but isn't firmware already loading a combined DTB file? We can use that file name.
Board-id -- make firmware ignore it if there is a single entry. Or maybe make a special case for something like `default.dtb`.

recipes-kernel/images/linux-qcom-combinedtb.bb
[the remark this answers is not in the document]
quic-vkraleti: Removed for now. Will share a separate PR with clear description of issue I was facing.

recipes-kernel/images/dtb-qcom-image.bb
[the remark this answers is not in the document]
quic-vkraleti: I tired my best in latest commit. Please feel free to reword as appropriate.

recipes-kernel/images/linux-qcom-combinedtb.bb
```diff
@@ -0,0 +1,30 @@
...
+DEPENDS = "dtc-native"
+
+do_configure[depends] += " \
+    virtual/kernel:do_deploy \
+    "
+
+# NOTE: At present dtbs are enabled based on the recipe provider of virtual/kernel
+# If dtb enablement happens in a generic way, below can be rewritten using KERNEL_DEVICETREE
+do_compile() {
+    rm -f ${B}/combined-dtb.dtb
```
ricardosalveti, line 15: This is probably not needed, and if you need a custom B folder you can extend do_compile[cleandirs] with it.

recipes-kernel/images/linux-qcom-combinedtb.bb
```diff
@@ -0,0 +1,30 @@
...
+    for kdtb in `find ${DEPLOY_DIR_IMAGE} -type f -name "*.dtb"`; do
+      echo "combining: $kdtb"
```
ricardosalveti, line 17: First line has 4 spaces and the additional lines just 2 for indentation, align everything with 4 spaces instead.

recipes-kernel/images/linux-qcom-uki.bb
```diff
@@ -67,22 +67,43 @@ do_compile() {
...
     ukify_cmd="$ukify_cmd --stub $stub"
 
-    # Output
+    # NOTE: At present dtbs are enabled based on the recipe provider of virtual/kernel
+    # If dtb enablement happens in a generic way, below can be rewritten using KERNEL_DEVICETREE
+    # Generate one uki per dtb
     mkdir -p "${B}${EFI_UKI_PATH}"
-    output="${B}${EFI_UKI_PATH}/${EFI_LINUX_IMG}"
+    for kdtb in `find ${DEPLOY_DIR_IMAGE} -type f -name "*.dtb"`; do
+        output="${B}${EFI_UKI_PATH}/${PN}-$(basename "$kdtb" .dtb)-${MACHINE}.efi"
```
ricardosalveti, line 75: While useful this might later explode in a large number of UKIs, based on the device trees available and set by the user. Can't we instead just create one with the defined dtb from the user (and empty if not set by the user)?

ricardosalveti: Mostly concerned about size issues, as UKI will contain kernel + initrd + dtb for every dtb available in the list.

recipes-kernel/images/dtb-qcom-image.bb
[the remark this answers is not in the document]
lumag: You have described UEFI behavior rather than the package itself. Please describe what the package does.

recipes-kernel/images/linux-qcom-uki.bb
```diff
@@ -44,7 +44,7 @@ do_compile() {
     # Kernel Image
     # Note: systemd-boot can't handle compressed kernel image.
     kernel_image="${DEPLOY_DIR_IMAGE}/Image"
-    [ -f $kernel_image ] && echo "Creating UKI with $kernel_image" || bbfatal "No valid kernel image to create UKI. Add 'Image' to KERNEL_IMAGETYPES."
+    [ -f $kernel_image ] && echo "Creating UKI with $kernel_image"
```
lumag, line 47: Why is this commit necessary at all? Why do you need to override the DTB? Commit message should answer such questions. In general, commit messages should respond to the ’why?’ question instead of simply describing what is being done (which is usually obvious from the commit itself).

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -1,10 +1,7 @@
 DESCRIPTION = "EFI System Partition Image to boot Qualcomm boards"
 
-COMPATIBLE_HOST = '(x86_64.*|arm.*|aarch64.*)-(linux.*)'
-
 PACKAGE_INSTALL = " \
-    kernel-devicetree \
-    linux-qcom-uki \
+    ${EFI_LINUX_IMG} \
```
lumag, line 4: Same comment. Commit message doesn't describe the reason for a change. Default DTB should be fine, otherwise there is little point in it.

recipes-kernel/images/linux-qcom-combinedtb.bb
lumag: I think it is fine, this can save us from possible issues if the task or the script gets executed manually.

quic-vkraleti:
@lumag @ricardosalveti I dropped two commits for enhancing uki to accept optional DTB from this PR. I'll submit another PR for those in some time.

lumag:
- package description is still incorrect
- DTB selection is based on the non-standard property 
- commit message describes `board-id`, while the actual property is `qcom,board-id`

ricardosalveti:
> * package description is still incorrect

The description was updated with "This recipe creates a vfat image with the 'combined-dtb.dtb' for UEFI consumption.", which is the intent.

> * DTB selection is based on the non-standard property

While not ideal, this is the current firmware behavior and will probably stay this way until an upstream compatible solution is found (discussion in progress). Not something that can be changed as part of this PR.

lumag:
Ack for the description. 
For the DTB, can we pretty please have a single DTB with no additional selection. There is no reason to install several different revisions if we know in advance, which one is going to be selected.

lumag:
I'm reluctant to merge a solution that depends on the patches that are not going to bed accepted upstream. And [upstream DT](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/boot/dts/qcom/qcs6490-rb3gen2.dts?h=v6.11-rc4) has neither `qcom, msm-id` nor `qcom,board-id` properties.

conf/machine/include/qcom-common.inc
```diff
@@ -50,6 +50,11 @@ EFI_DTB_DIR ?= "${EFI_PREFIX}dtb"
 # Unified Kernel Image (UKI) name
 EFI_LINUX_IMG ?= "linux-${MACHINE}.efi"
 
+# List of device trees to pack into combined-dtb
+COMBINED_DTB_DEVICETREES ?= " \
+    qcs6490-rb3gen2.dtb \
+    "
```
lumag, line 56: Why can't we pack all KERNEL_DEVICETREE ?

recipes-kernel/images/linux-qcom-combinedtb.bb
```diff
@@ -0,0 +1,43 @@
...
+            if [ -r "${DEPLOY_DIR_IMAGE}/$kdtb" ]; then
+                echo "combining: $kdtb"
+                cat ${DEPLOY_DIR_IMAGE}/$kdtb >> ${B}/combined-dtb.dtb
+            else
+                bbwarn "Couldn't find a valid $kdtb in deploydir"
+            fi
+        done
+    else
+        bbfatal "No valid device trees specified to prepare dtb image"
+    fi
```
lumag, line 32: The more I look at this recipe the more I think that it's better turn into a bbclass to be used by kernel recipes. Then we can use `KERNEL_DEVICETREE` even if it get overriden for the kernel recipe and/or version. And the combined-dtb will be automatically repopulated if the kernel image changes.

recipes-kernel/images/dtb-qcom-image.bb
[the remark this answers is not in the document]
lumag: Still not resolved.

lumag (changes requested):

recipes-kernel/images/dtb-qcom-image.bb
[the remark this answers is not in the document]
quic-vkraleti: "This recipe creates a vfat image with 'combined-dtb.dtb' for UEFI consumption." isn't it appropriate?

conf/machine/include/qcom-common.inc
quic-vkraleti: dtbs are enabled based on the recipe provider of virtual/kernel. If dtb enablement happens in a generic way, KERNEL_DEVICETREE can be parsed to get dtbs. Till then this list has to be maintained. As discussed over telcon UEFI firmware is being enhanced to stop reading board-id if there is a single dtb. Packing all of  KERNEL_DEVICETREE would mean the qcom,board-id based checks still need to be maintained.

recipes-kernel/images/linux-qcom-combinedtb.bb
quic-vkraleti: Sure, will consider for next PR.

quic-vkraleti (commented):
Qualcomm is exploring the suggestion to use FIT format based approach to pick a DTB so that it gets upstream aligned. Present dependency on qcom,board-id would be dropped with in an year time frame.

recipes-kernel/images/dtb-qcom-image.bb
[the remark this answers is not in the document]
lumag: Please describe the package and not the recipe.

recipes-kernel/images/linux-qcom-combinedtb.bb
```diff
@@ -0,0 +1,43 @@
+SUMMARY = "Generates a single dtb by combining multiple dtb files"
+
+DESCRIPTION = "QCOM UEFI implementation supports loading a matching device-tree \
+               by reading 'combined-dtb.dtb' from the dtb partition. This recipe \
+               combines user specified dtbs into a single 'combined-dtb.dtb' \
+               binary blob that can be provided to the UEFI."
```
lumag, line 6: Please describe the package a not the recipe.

```
DESCRIPTION[doc] = "The package description used by package managers. If not set, DESCRIPTION takes the value of the SUMMARY variable."
```

conf/machine/include/qcom-common.inc
lumag: In such a case you can just create one image per DTB. From the bbclass iterate over the `KERNEL_DEVICETREE` and call `mkfs.vfat` manually for each DTB.

quic-vkraleti: Sure, I'll drop linux-qcom-combineddtb recipe and add a kernel bbclass to generate one vfat image per DTB.

quic-vkraleti: Done, added a bbclass to create a dtb image for each dtb produced by kernel.

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -0,0 +1,33 @@
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+DTBBIN_DEPLOYDIR = "${WORKDIR}/qcom_dtbbin_deploy-${PN}"
+DTBBIN_SIZE ?= "65536"
```
lumag, line 4: I think 64M is a bit too big for a DTB partition. What about getting that down to 2M or 4M? If we need more, we can expand it later.

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -0,0 +1,33 @@
...
+do_qcom_dtbbin_deploy[cleandirs] = "${DTBBIN_DEPLOYDIR}"
+do_qcom_dtbbin_deploy() {
+    for dtbf in ${KERNEL_DEVICETREE}; do
+        bbdebug 1 " combining: $dtbf"
+        dtb=`normalize_dtb "$dtbf"`
+        dtb_ext=${dtb##*.}
+        dtb_base_name=`basename $dtb .$dtb_ext`
+        mkdir -p ${DTBBIN_DEPLOYDIR}/$dtb_base_name
+        cp ${D}/${KERNEL_DTBDEST}/$dtb_base_name.dtb ${DTBBIN_DEPLOYDIR}/$dtb_base_name/combined-dtb.dtb
+        mkfs.vfat $@ -C ${DTBBIN_DEPLOYDIR}/combined-dtb-$dtb_base_name.vfat ${DTBBIN_SIZE}
```
lumag, line 16: What about `dtb-${dtb_base_name}-image.vfat` ? This way it's not specific to the `combined/etc.` logic. And if the logic or partition internals change, the user won't have to look for a new name.

lumag (changes requested):
Looks great now, let's fix two minor issues.

classes/linux-qcom-dtbbin.bbclass
quic-vkraleti: Sure, I'll update.

classes/linux-qcom-dtbbin.bbclass
quic-vkraleti: reduced to 4MB.

lumag (approved):

lumag:
LGTM now

ricardosalveti (approved):
LGTM

PR #622, opened 2024-08-21
Support optional dtb loading via UKI

quic-vkraleti:
If a DTB is present in UKI, systemd-boot can override the default UEFI provided DTB. This provisioning of DTB via UKI allows the users to customize DTBs as per product needs without regenerating the firmware.

### DIFF
--- a/recipes-kernel/images/esp-qcom-image.bb
+++ b/recipes-kernel/images/esp-qcom-image.bb
@@ -1,9 +1,6 @@
 DESCRIPTION = "EFI System Partition Image to boot Qualcomm boards"
 
-COMPATIBLE_HOST = '(x86_64.*|arm.*|aarch64.*)-(linux.*)'
-
 PACKAGE_INSTALL = " \
-    kernel-devicetree \
     linux-qcom-uki \
     systemd-boot \
     systemd-bootconf \
--- a/recipes-kernel/images/linux-qcom-uki.bb
+++ b/recipes-kernel/images/linux-qcom-uki.bb
@@ -85,6 +85,13 @@ do_compile() {
     [ -f $stub ] && echo "Creating UKI with $stub" || bbfatal "$stub is not a valid stub to create UKI."
     ukify_cmd="$ukify_cmd --stub $stub"
 
+    # DTB
+    if [ -n "${EFI_LINUX_IMG_DTB}" ]; then
+        dtb="${DEPLOY_DIR_IMAGE}/${EFI_LINUX_IMG_DTB}"
+        [ -r "$dtb" ] && echo "Creating UKI with $dtb" || bbfatal "$dtb is not valid dtb to create UKI."
+        ukify_cmd="$ukify_cmd --devicetree=$dtb"
+    fi
+
     # Output
     mkdir -p "${B}${EFI_UKI_PATH}"
     output="${B}${EFI_UKI_PATH}/${EFI_LINUX_IMG}"
--- a/conf/machine/include/qcom-common.inc
+++ b/conf/machine/include/qcom-common.inc
@@ -50,6 +50,10 @@ EFI_DTB_DIR ?= "${EFI_PREFIX}dtb"
 # Unified Kernel Image (UKI) name
 EFI_LINUX_IMG ?= "linux-${MACHINE}.efi"
 
+# dtb file to pack into UKI.
+# empty value would skip packing dtb.
+EFI_LINUX_IMG_DTB ?= ""
+
 # Place dtb at EFIDTDIR to seamlessly package
 KERNEL_DTBDEST = "${EFI_DTB_DIR}"
 


recipes-kernel/images/linux-qcom-uki.bb
```diff
@@ -44,7 +44,7 @@ do_compile() {
     # Kernel Image
     # Note: systemd-boot can't handle compressed kernel image.
     kernel_image="${DEPLOY_DIR_IMAGE}/Image"
-    [ -f $kernel_image ] && echo "Creating UKI with $kernel_image" || bbfatal "No valid kernel image to create UKI. Add 'Image' to KERNEL_IMAGETYPES."
+    [ -f $kernel_image ] && echo "Creating UKI with $kernel_image"
```
lumag, line 47: Why? Commit message provides no reason to drop this bbfatal.

recipes-kernel/images/linux-qcom-uki.bb
```diff
@@ -85,6 +85,15 @@ do_compile() {
     [ -f $stub ] && echo "Creating UKI with $stub" || bbfatal "$stub is not a valid stub to create UKI."
     ukify_cmd="$ukify_cmd --stub $stub"
 
+    # optional DTB
+    dtb="${DEPLOY_DIR_IMAGE}/${EFI_LINUX_IMG_DTB}"
```
lumag, line 89: Please, drop extra empty line

recipes-kernel/images/linux-qcom-uki.bb
```diff
@@ -85,6 +85,15 @@ do_compile() {
...
+    if [ -f "$dtb" ]; then
+        echo "Creating UKI with $dtb"
+        ukify_cmd="$ukify_cmd --devicetree=$dtb"
+    else
+        bbdebug 1 "$dtb is not valid dtb to create UKI."
```
lumag, line 94: `[ -r $dtb ] || bbfatal dtbs not found`

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -1,9 +1,6 @@
 DESCRIPTION = "EFI System Partition Image to boot Qualcomm boards"
 
-COMPATIBLE_HOST = '(x86_64.*|arm.*|aarch64.*)-(linux.*)'
-
 PACKAGE_INSTALL = " \
-    kernel-devicetree \
```
lumag, line 6: If this commit depends on the boot flow changes, then it should be a part of #620

lumag (changes requested):

recipes-kernel/images/esp-qcom-image.bb
ricardosalveti: It doesn't depend on the combined dtb logic, since what is basically happening here is moving the dtb from esp to be part of UKI.

recipes-kernel/images/linux-qcom-uki.bb
ricardosalveti: Yeah, without kernel image we should still fail as before, right?

recipes-kernel/images/linux-qcom-uki.bb
ricardosalveti: For it to be fatal we need to always define a default for EFI_LINUX_IMG_DTB (which we should probably do anyway, like using the first dtb available in KERNEL_DEVICETREE or similar, with a weak assignment.

recipes-kernel/images/linux-qcom-uki.bb
quic-vkraleti: SKIP_RECIPE check is present at end of the file to check for 'Image' in kernel image types. This condition is not needed any more.

lumag: Unrelated change ergo separate commit with its own justification.

recipes-kernel/images/linux-qcom-uki.bb
lumag: Not quite. Empty EFI_LINUX_IMG_DTB should be a separate condition. And also there default should be empty string, i.e. use DTB provided by the firmware.

recipes-kernel/images/linux-qcom-uki.bb
lumag: Argh, it's just a GitHub on a cell phone, there is no empty line. Please ignore the comment.

recipes-kernel/images/esp-qcom-image.bb
lumag: No, it is not a move. This PR provides a way to override the DTB provided by the firmware.

ricardosalveti: The latest firmware is not loading from the esp partition anymore, but from combined-dtb. With latest firmware at least the dtb files from esp would only really be useful to the OS, and a similar behavior is applied with UKIs here.

recipes-kernel/images/linux-qcom-uki.bb
ricardosalveti: Right, my thinking is that we could have 2 variables, one to control if the dtb should be added to UKI and the other with the dtb string to be used, which can be set to the first one by default (with weak assignment).

lumag: No, weak assignment is really incorrect. The first DTB is the one for db410c if I'm not mistaken.

recipes-kernel/images/esp-qcom-image.bb
lumag: So... If anybody has DTB in ESP, they can get unbootable device by updating the firmware? That's ridiculous.

lumag: I'm currently OOO. Could you please schedule a telco on Monday or Tuesday, let's discuss what is going on and what we can do.

lumag: Please also invite @calebccff

recipes-kernel/images/linux-qcom-uki.bb
ricardosalveti: Right, for a generic target/config this won't really be useful due the amount of dtbs used.

lumag: Yep. So if for some stranger reason somebody has a single-machine config, it makes sense to have that variable assigned in the machine config. But it's not sensible to set it in generic config.

recipes-kernel/images/esp-qcom-image.bb
lumag: I'd suggest having a recipe that generates a set of DTB-vfat images with a single DTB in each. This is going to work with the current approach and it will continue to work once FW gets updated to ignore custom properties.

ricardosalveti: Will sync internally first to understand the previous boot firmware expectations here, I'm just familiar with what the latest version is currently doing.

lumag (changes requested):
Please make it a fatal error if the user specifies non-empty EFI_LINUX_IMG_DTB, but the `$dtb` doesn't exist. Other than that looks good.

recipes-kernel/images/linux-qcom-uki.bb
lumag: I'd really like to merge this. I think the `bbfatal` is the only comment that still needs attention.

lumag (approved):

PR #623, opened 2024-08-22
50-adbd-cmdline.conf: Change /var to /etc in ConditionPathExists

quic-raghuvar:
If android-tools-adbd.service service needs to be up upon boot, then the path assigned to ConditionPathExists must be present at boot time. This means that the path set to ConditionPathExists must be created at build time itself. /etc is a better place to keep files and directories that are created at build time rather than /var. /var is expected to house files that are created at run time.

Hence, change ConditionPathExists=|/var/usb-debugging-enabled to ConditionPathExists=|/etc/usb-debugging-enabled

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-adbd-cmdline/50-adbd-cmdline.conf
+++ b/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-adbd-cmdline/50-adbd-cmdline.conf
@@ -2,5 +2,5 @@
 # Clear all conditions
 ConditionPathExists=
 # And start if kernel adbd argument is provided or if the file exists
-ConditionPathExists=|/var/usb-debugging-enabled
+ConditionPathExists=|/etc/usb-debugging-enabled
 ConditionKernelCommandLine=|adbd


lumag:
Meta-oe still [uses](https://github.com/openembedded/meta-openembedded/blob/kirkstone/meta-oe/recipes-devtools/android-tools/android-tools/android-tools-adbd.service) /var for the kirkstone branch. Please get the corresponding patch backported first.

lumag:
BTW: with your meta-oe patch you have created internal inconsistency between selinux-enabled and -disabled configurations. See https://github.com/openembedded/meta-openembedded/blob/master/meta-oe/dynamic-layers/selinux/recipes-devtool/android-tools/android-tools/android-tools-adbd.service

Please send a patch fixing this inconsistency ASAP, add me to the Reported-by tag

lumag:
@quic-raghuvar I've pointed out at meta-openembedded, but these items apply here too:

- Please use your full name to create git commits
- Please add DCO trailer
- Please format commit message to fit into 72-77 chars.

lumag:
@quic-raghuvar any updates? meta-oe's kirstone branch still don't have these patches.

quic-raghuvar:
@lumag , I have requested a backport request to both kirkstone and scarthgap for this change

lumag:
@quic-raghuvar while we are waiting for the backports to be applied to the `kirkstone` branch, please provide a backport to the `scarthgap` branch. Corresponding changes were merged to meta-oe/scarthgap.

PR #624, opened 2024-08-23
mesa: Drop _git from recipe name

kraj:
This matches with OE-core, recently dropping _git from the mesa recipe name


lumag:
Truck

lumag:
Thank you!

PR #619, opened 2024-08-07
Add firmware package support for QCS6490 RB3Gen2

adhudase:
Add firmware package support for QCS6490 RB3Gen2 board.

This PR addresses the same purpose as the existing PR #551. However, due to limitations in writing to @kraparthy’s tree, we created this fresh PR. Please note that @kraparthy is no longer part of this project.

PR #627, opened 2024-08-26
qca-swiss-army-knife: fix ath11k/board-2.json generation

lumag:
ath11k tools have slighly changed the format of the board-2.json by adding a nested board / regdb arrays. Fix the script that generates board-2.json to accomodate for those changes.


lumag:
@adhudase fixes the issue from #619

PR #637, opened 2024-09-12
Fix build error if INITRAMFS_IMAGE undefined

quic-vkraleti:
Fixed below error in linux-qcom-uki.bb when INITRAMFS_IMAGE is undefined.

ERROR: Error for linux-qcom-uki.bb:do_configure[depends], dependency '${@ '${INITRAMFS_IMAGE}:do_image_complete' if d.getVar('INITRAMFS_IMAGE') else ''}' does not contain exactly one ':' character. Task 'depends' should be specified in the form 'packagename:task'

### DIFF
--- a/recipes-kernel/images/linux-qcom-uki.bb
+++ b/recipes-kernel/images/linux-qcom-uki.bb
@@ -18,6 +18,8 @@ require conf/image-uefi.conf
 
 KERNEL_VERSION = "${@get_kernelversion_file('${STAGING_KERNEL_BUILDDIR}')}"
 
+INITRAMFS_IMAGE ?= ''
+
 do_configure[depends] += " \
     systemd-boot:do_deploy \
     virtual/kernel:do_deploy \
--- a/recipes-kernel/images/linux-qcom-uki.bb
+++ b/recipes-kernel/images/linux-qcom-uki.bb
@@ -44,7 +44,7 @@ do_compile() {
     # Kernel Image
     # Note: systemd-boot can't handle compressed kernel image.
     kernel_image="${DEPLOY_DIR_IMAGE}/Image"
-    [ -f $kernel_image ] && echo "Creating UKI with $kernel_image" || bbfatal "No valid kernel image to create UKI. Add 'Image' to KERNEL_IMAGETYPES."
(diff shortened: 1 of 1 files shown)


lumag (changes requested):
Rule of thumb "Also do this and that" => separate commits.

Also please use imperative language to describe the second commit ('remove' instead of 'removed').

lumag:
Also could you please expand CI configuration so that such cases can be caught (please submit a PR with CI changes separately so that we can merge it first and then test this PR against the merged tree).

quic-vkraleti:
> and CI configuration so that such cases can be caught (please submit a PR with CI changes separately so that we can merge it first and then test this

Sure, I'll update CI scripts in a separate PR and get back.

lumag:
I don't see CI failing after the CI change in #639 .

quic-vkraleti:
> I don't see CI failing after the CI change in #639 . Was it enough to demonstrate the issue being fixed by this PR?

To reproduce need to set INITRAMFS_IMAGE = "". Is there any configuration in CI that tests kernel without initrd?

lumag:
> > I don't see CI failing after the CI change in #639 . Was it enough to demonstrate the issue being fixed by this PR?
> 
> To reproduce need to set INITRAMFS_IMAGE = "". Is there any configuration in CI that tests kernel without initrd?

I asked you to add necessary CI configuration.

quic-vkraleti:
> > > I don't see CI failing after the CI change in #639 . Was it enough to demonstrate the issue being fixed by this PR?
> > 
> > 
> > To reproduce need to set INITRAMFS_IMAGE = "". Is there any configuration in CI that tests kernel without initrd?
> 
> I asked you to add necessary CI configuration.

https://github.com/Linaro/meta-qcom/pull/644 updated build template to build with & without initramfs.

lumag (changes requested):
I've reviewed the error case. Please keep existing `do_configure[depends]` and provide default value (empty) for the `INITRAMFS_IMAGE`, like `kernel.bbclass` does.

lumag:
@quic-vkraleti any chance of getting this done? I'd like to merge the fix and then backport necessary changes to the `scarthgap` branch.

lumag (approved):

PR #641, opened 2024-09-12
Split WoA firmware recipes

lumag:
WoA firmware recipes depend on cabextract-native, which is a part of meta-oe. Move WoA recipes to `dynamic-layers/openembedded-layer`.
**cc** @ricardosalveti


ricardosalveti (approved):
LGTM

PR #649, opened 2024-09-13
python3-pefile: vendor in the recipe

lumag:
Vendor the recipe from meta-oe. This allows us to build the esp-qcom-image without adding dependency on the meta-oe. Python-pefile should be moved from meta-oe to OE-Core, then this recipe can be dropped.

Suggestd-by: Ricardo Salveti <ricardo@foundries.io>

### DIFF
--- a/recipes-vendor-in/python/python3-pefile/run-ptest
+++ b/recipes-vendor-in/python/python3-pefile/run-ptest
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+pytest --automake --ignore=./tests/pefile_test.py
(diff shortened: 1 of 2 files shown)


ricardosalveti:
We should probably vendor the latest version (updated 3 days ago) instead: https://git.openembedded.org/meta-openembedded/commit/?id=0e6ba2d70844843d057f41f884df676b9314c9d6

Probably also good to point to the git hash used as reference in the commit message.

Typo nit: s/Suggestd-by/Suggested-by/

lumag:
Thanks for the suggestions. I've updated the recipe and I'll merge it once build succeeds.

PR #650, opened 2024-09-13
Revert "CI: Enable building of the images with and without initramfs"

lumag:
Reverts Linaro/meta-qcom#644,it breaks CI by introducing duplicate labels .


lumag:
@quic-vkraleti see https://github.com/Linaro/meta-qcom/actions/runs/10852055637/job/30117205624

PR #655, opened 2024-09-17
CI: retain only licenses as build artifacts

lumag:
There is little point in retaining full set of build artifacts for the test builds, which is what the empty "artifacts" config does. Keep just the licences and throw away everything else.


ricardosalveti:
Is there any value on retaining the licenses?

lumag:
@ricardosalveti it's mostly a bandaid for TuxOE, which populates all artifacts if it finds empty or non-existing `artifacts` array. I think it's better than having `artifacts: [ "nonexisting"]`. But I can use that if you'd prefer it.

ricardosalveti:
Sure, works for me, was just curious why it was needed.

ricardosalveti (approved):

PR #562, opened 2023-12-29
Fix dependency loop between INITRAMFS_IMAGE and UKI

quic-vkraleti:
Ramdisk image by default pulls in a dependency on do_packagedata task of kernel.
Uki.bbclass is trying to create kernel-uki package, by packaging ramdisk image before packagedata task. 
This is the causing circular dependency of tasks. To address this, need to stop creating kernel-uki package and place efi image in deploydir instead.

PR #666, opened 2024-10-15
linux-qcom-uki: KERNEL_IMAGETYPES should be fully populated before use

ndechesne:
In response to https://github.com/quic-yocto/meta-qcom-hwe/pull/42.

KERNEL_IMAGETYPES is not enough to be checked, since kernel types can also be set using KERNEL_IMAGETYPE or KERNEL_ALT_IMAGETYPE. Like it's done in OE-core kernel class, we need to combine all these variables.

### DIFF
--- a/recipes-kernel/images/linux-qcom-uki.bb
+++ b/recipes-kernel/images/linux-qcom-uki.bb
@@ -20,6 +20,21 @@ KERNEL_VERSION = "${@get_kernelversion_file('${STAGING_KERNEL_BUILDDIR}')}"
 
 INITRAMFS_IMAGE ?= ''
 
+# This recipe is skipped based on KERNEL_IMAGETYPES. However the list of kernel type can also be set with
+# KERNEL_IMAGETYPE or KERNEL_ALT_IMAGETYPE.
+# We need to combine them all to be correct (like it's done in kernel.bbclass)
+python __anonymous () {
+    # Merge KERNEL_IMAGETYPE and KERNEL_ALT_IMAGETYPE into KERNEL_IMAGETYPES
+    type = d.getVar('KERNEL_IMAGETYPE') or ""
+    alttype = d.getVar('KERNEL_ALT_IMAGETYPE') or ""
+    types = d.getVar('KERNEL_IMAGETYPES') or ""
+    if type not in types.split():
+        types = (type + ' ' + types).strip()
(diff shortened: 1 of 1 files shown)


ndechesne:
@rsalveti @adhudase @quic-vkraleti . FWIW. It's untested, but that should be close to what we want for https://github.com/quic-yocto/meta-qcom-hwe/pull/42

ricardosalveti:
> @rsalveti @adhudase @quic-vkraleti . FWIW. It's untested, but that should be close to what we want for [quic-yocto/meta-qcom-hwe#42](https://github.com/quic-yocto/meta-qcom-hwe/pull/42)

This should work, but wonder if worthy maintaining this piece just to set SKIP_RECIPE, while the actual runtime check already exists and produces a fatal error.

We should also probably move to https://lists.openembedded.org/g/openembedded-core/message/205686 once merged (hopefully soon once the autobuilder issues are sorted out), but it will require changes as this layer is using Image.gz as the default target.

ndechesne:
Yes, we should backport the upstream uki class when it's merged, but it might take some time. I just tried to provide an alternate to the PR submitted on meta-qcom-hwe (which is wrong). so we can live with this change here until we remove this recipe.

lumag:
@ndechesne @ricardosalveti For your information, we were using `uki.bbclass` and then we had to stop using it, see https://github.com/Linaro/meta-qcom/pull/562 . Basically, we wanted to have the UKI image to be availabel as a package, not only as a file inside `${DEPLOY_DIR}`
I hope that @quic-vkraleti and other Qualcomm engineers are participating in the discussions on the ML regarding the UKI bbclass and provide additional information on Qualcomm requirements.

ndechesne:
hmm. thanks for pointing this out. I will review #562.. However we really need a very strong reason for not using the 'default' mechanism. Or we should indeed participate in the upstream discussions..

ricardosalveti:
> Yes, we should backport the upstream uki class when it's merged, but it might take some time. I just tried to provide an alternate to the PR submitted on meta-qcom-hwe (which is wrong). so we can live with this change here until we remove this recipe.

No need to backport at this stage since we're only using main/master for now, but I'm ok with your change now to solve the issue we have.

ricardosalveti (approved):

PR #667, opened 2024-10-16
linux-qcom-uki: use bb.parse.SkipRecipe for the kernel Image check

ricardosalveti:
Move the kernel Image check to the same python parsing logic created for merging KERNEL_IMAGETYPES, as SKIP_RECIPE does not take into account dynamic python logic, rejecting the updated KERNEL_IMAGETYPES value.

Since KERNEL_IMAGETYPES is not used anywhere else in the recipe, remove setVar and just use the merged 'types' variable instead.


ricardosalveti:
@ndechesne fyi, tested with meta-qcom-hwe master.

Issue #110, opened 2019-08-20
No 4k output on Qualcomm Robotics RB3 Platform(SDA845) (Thundercomm)


HowardDunn:
Using the thundercomm board for the 845c. Built an image using the warrior branch 5.2 kernel. The HDMI output maxes out at 1920x1080p @ 60Hz. The hardware manual I believe says the HDMI can output 4k @ 30hz. 

There is also an issue drawing to the display, when using gstreamer to draw multiple videos on the screen, the videos would decode but the screen has a hard time drawing the videos to the screen. I've tried both an X11 and Weston image but still see the weird screen drawing issue. Using the v4l2decoder with gstreamer with the xvimagesink on X11 and the waylandsink on Weston. The videos would flicker on the screen. I  also launched a GTK application and that flickered briefly as well.  It also crashes the device frequently whenever I run the gstreamer pipelines.

PR #670, opened 2024-11-01
Fix HDMI 4k output on Qualcomm Robotics RB3 Platform

vishwamartur:
Related to #110

Add support for 4k resolution output on Qualcomm Robotics RB3 Platform (SDA845).

* Update `recipes-bsp/packagegroups/packagegroup-firmware-dragonboard845c.bb` to include `linux-firmware-qcom-sdm845-hdmi` and `linux-firmware-qcom-sdm845-4k`.
* Update `recipes-bsp/firmware/firmware-qcom-dragonboard845c_20190529180356-v4.bb` to add `4k-support.bin` to the `do_install` function.
* Update `dynamic-layers/openembedded-layer/recipes-graphics/images/qcom-x11-image.bb` to add `xrandr --output HDMI-1 --mode 3840x2160 --rate 30` to `CORE_IMAGE_BASE_INSTALL`.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-graphics/images/qcom-x11-image.bb
+++ b/dynamic-layers/openembedded-layer/recipes-graphics/images/qcom-x11-image.bb
@@ -27,4 +27,5 @@ CORE_IMAGE_BASE_INSTALL += " \
     mesa-demos \
     openbox \
     rsync \
+    xrandr --output HDMI-1 --mode 3840x2160 --rate 30 \
 "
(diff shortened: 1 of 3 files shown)


lumag:
Unfortunately this more looks like a hoax. There is no `4k-support.bin` file in that archive. There is no `linux-firmware-qcom-sdm845-hdmi` package that you are trying to install. Calling xrandr for the fixed mode is also an error, what if the monitor doesn't support it.
No S-o-B, no real email in the commit. It's not the April Fool's day.

PR #673, opened 2024-11-04
firmware-qcom-rb3gen2: migrate to linux-firmware

lumag:
Stop providing the DSP and Adreno firmware as a part of the firmware-qcom-rb3gen2 recipe, they are now a provided by the linux-firmware recipe.


lumag:
cc @adhudase @rsalveti

PR #674, opened 2024-11-05
Use Hexagon DSP binaries recipe

lumag:
Make meta-qcom use new hexagon-dsp-binaries repository for the DSP binaries location. Bind-mounting the DSP location to `/usr/lib/rfsa` is a long-standing hack which should be dropped either when we switch the layer to using hexagonrpcd or when the fastrpc daemon gets support for specifying the location for the binaries.


lumag:
**cc** @adhudase @rsalveti

lumag:
@adhudase @ricardosalveti @ndechesne unless there are any comments, I'd like to merge this.

ricardosalveti (approved):
LGTM

PR #679, opened 2024-11-20
Update dsp binaries recipe

lumag:
Update hexagon-dspb-binaries to 20241110 and handle new RB3gen2 binaries.


ricardosalveti (approved):
LGTM

PR #681, opened 2024-11-27
Drop recipes migrating to meta-qcom-extras

lumag:
Drop recipes which should migrate to meta-qcom-extras during the process of meta-qcom handover.


ndechesne (commented):
LGTM, thanks!

PR #689, opened 2024-12-02
esp-qcom-image cleanup 

ricardosalveti:
Improve esp-qcom-image to make it aligned with the partition size and with the correct files for an ESP partition.

Fixes:
- Change default image size to 512MB, aligned with the default ESP partition size
- Change sector size to 4096 to allow kernel mounting ESP at runtime
- Drop systemd-bootconf, not required with UKIs
- Remove unused files and directories (e.g. /usr, /bin, /var, /etc, etc).


recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -14,4 +14,7 @@ inherit image
 IMAGE_FSTYPES = "vfat"
 IMAGE_FSTYPES_DEBUGFS = ""
 
+# Default vfat sector size (512) is too small for UFS
+EXTRA_IMAGECMD:vfat += " -S 4096"
```
lumag, line 17: Does it work for eMMC-based systems?

ricardosalveti: Linux handles it fine, u-boot will complain I believe (on eMMC).

I didn't want to have this config set by default at the machine conf as that would apply for all vfat runs, which is also not ideal as the user might write to different storage devices, which is why I moved to the image level.

We could have an override here per machine / soc, or define a custom image command for esp-vfat, which can then be configured at the machine level. Won't change anything to the current machines using it but it could later be used by eMMC devices booting out of u-boot.

lumag: could you please turn sector size into an overridable variable? Then I'm fine with the change.

ricardosalveti: Sure, done.

PR #699, opened 2024-12-11
hexagon-dsp-binaries: bump the version

lumag:
Bump the version of the Hexagon DSP binaries package to keep it 1:1 to the linux-firmware packaege version. This depends on [OE-Core](https://lore.kernel.org/openembedded-core/20241210234624.1650171-1-dmitry.baryshkov@linaro.org/) patches.


lumag:
cc @ndechesne @ricardosalveti

ndechesne:
Why do you need this tight coupling? are you going to update the filename each time linux-firmware in OE changes? does that really make sense?

lumag:
Unfortunately the DSP binaries are tied to the particular DSP firmware. So, we have to make sure that their versions match. The easiest way to handle that is by versioned RDEPENDS from DSP package onto the corresponding linux-firmware package.
So, basically yes. Another option might be to make linux-firmware-qcom-foo provide crafted virtual packages like `sm8250-qualcomm-rb5-ADSP.HT.5.3.c2-00082-SM8250-1` and then make Hexagon DSP binaries package depend on such virtual name. However such a dependency quickly becomes distro-specific and can not be easily automated. So, it seems, just forcing 1:1 version dependency is the easiest way to handle the issue in a generic way.

lumag:
@ndechesne gracious ping.

ricardosalveti (approved):

PR #700, opened 2024-12-11
Enable linux-yocto-6.10 support

lumag:


lumag:
cc @ricardosalveti @quic-vkraleti

lumag:
Closing, we should be switching to 6.12 instead. 6.10 is gone from OE-Core

PR #723, opened 2025-01-28
Cdt qcs9100

ndechesne:


recipes-bsp/partition/qcom-partition-confs_1.0.bb
```diff
@@ -36,6 +36,10 @@ do_compile:prepend:qcs6490-rb3gen2-core-kit() {
     fixup_cdt
 }
 
+do_compile:prepend:sa8775p-ride-sx() {
+    fixup_cdt
+}
+
```
lumag, line 42: I think QDL skips non-existing binaries. Is that the case for other Qualcomm tools? It would be nice to enable fixup_cdt by default and let QDL skip non-existing cdt.bin

ndechesne: I know.. I don't like that at all either. I would like to merge it for now, and see how we can improve things at some point. QDL will error out if the file is missing, unless if invoked with -f in which case it will ignore missing files..

lumag (approved):

ricardosalveti (approved):

PR #727, opened 2025-01-30
Yocto 6.12

lumag:
Add recipe for linux-yocto 6.12. Drop most of the backported patches, keep only workarounds and pending dma-heap patches.


ndechesne:
I think I understand why you sent this PR.. but it's not very clear from the commit messages. Can you please add more information. 

We have many machines and kernel.. so perhaps we also need a dedicate document (or in the README) that explains how we maintain machines and kernel here.. maybe not blocking for this PR, but we should start something, as I think it will get even worse over time ..

lumag:
Well... For now I really wanted a fresher kernel, being able to test it too (and getting rid of 6.6 in the long term). But probably I can move that to the commit message. But in the longer term we need to discuss the kernel strategy.

lumag:
We can wait for a few days for https://lore.kernel.org/openembedded-core/cover.1739145061.git.bruce.ashfield@gmail.com/ to be merged and then change this PR to move from 6.6 to 6.12 instead of adding 6.12 next to 6.6.

ndechesne:
The changes look good to me. thanks for the cleanup / additional information in the commits. We should merge it now, before 6.6. is deprecated, otherwise, the build will break when it's removed.

ndechesne (approved):

ricardosalveti (approved):
LGTM

PR #739, opened 2025-02-07
Test action v6

mwasilew:


mwasilew:
This is the existing summary that is already added to the PR
gh_test_summary

This comment with the job links will now be added. There will be a new comment for each version of the patch, unlike the summary, which is updated.
gh_test_links

.github/workflows/test.yml
```diff
@@ -59,14 +59,15 @@ jobs:
 
       - name: Submit ${{ matrix.target }}
         timeout-minutes: 20
-        uses: foundriesio/lava-action@v5
+        uses: foundriesio/lava-action@v6
```
lumag, line 62: Missing commit message.

.github/workflows/test.yml
```diff
@@ -59,14 +59,15 @@ jobs:
...
           lava_token: ${{ secrets.LAVATOKEN }}
           lava_url: 'lava.infra.foundries.io'
           job_definition: ${{ matrix.target }}
           wait_for_job: true
           fail_action_on_failure: false
           save_result_as_artifact: true
+          save_job_details: true
 
   publish-test-results:
     name: "Publish Tests Results"
```
lumag, line 73: Is it a part of the upgrade or a separate change on its own?

.github/workflows/test.yml
```diff
@@ -92,3 +92,14 @@ jobs:
...
+      - name: Publish Test Job Details
+        run: |
+          for json_file in $(find ${{ github.workspace }} -name "test-job-*.json")
+          do
+              DEVICE_TYPE=$(cat "$json_file" | jq -r ".requested_device_type")
+              URL=$(cat "$json_file" | jq -r ".url")
+              JOB_ID=$(cat "$json_file" | jq -r ".id")
+              echo " * [Job $JOB_ID on $DEVICE_TYPE]($URL)"
+              echo " * [Job $JOB_ID on $DEVICE_TYPE]($URL)" >> $GITHUB_STEP_SUMMARY
+          done
```
lumag, line 105: Again, missing commit message.

lumag (changes requested):
Is there a way to attach test results to a workflow/pipeline run instead of posting them as a comment?

mwasilew:
> Is there a way to attach test results to a workflow/pipeline run instead of posting them as a comment?

It's already there. The problem is it's not easy to find the proper run from the PR. For example this is a test job for PR #744 https://github.com/qualcomm-linux/meta-qcom/actions/runs/13249889524. You can see the results posted as a comment, there is a check added to the commit, but getting to the actual test job log requires finding proper workflow on the list.

.github/workflows/test.yml
mwasilew: fixed.

.github/workflows/test.yml
mwasilew: Yes, part of the upgrade. There is a default, but it's set to false and the job details need to be saved to allow for updates in PR/workflow status.

lumag: This should be a part of the commit message (it is not).

lumag: Well, usual thing: rationale for a change. Either new default doesn't match old behaviour (and so this change is legit here) or you are just enabling new functionality (and as such it should go to a separate or  to a next commit, which makes use of job details).

lumag (approved):

ricardosalveti (approved):

lumag:
@mwasilew could you please force-push a new head to your branch in order to trigger CI build against a fixed repo?

PR #740, opened 2025-02-07
CI: rework to enable optional meta-qcom-distro configs

koenkooi:
Consume the upstream repos directly, not through the combo repo.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -12,10 +12,19 @@ defaults:
 repos:
   meta-qcom:
 
-  poky:
-    url: https://git.yoctoproject.org/git/poky
+  oe-core:
+    url: https://git.openembedded.org/openembedded-core
     layers:
       meta:
+
+  bitbake:
+    url: https://git.openembedded.org/bitbake
+    layers:
+      .: excluded
+
+  meta-yocto:
+    url: https://git.yoctoproject.org/git/meta-yocto
+    layers:
       meta-poky:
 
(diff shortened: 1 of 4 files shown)


lumag:
Why?
CI needs to be fixed too.

koenkooi:
> Why? 

Because we need to test the layer properly, which means against non-poky configurations as well. With this PR poky becomes optional instead of mandatory.

ndechesne:
the yocto-pacthreview test fails, can you check why?

koenkooi:
> 
> the yocto-pacthreview test fails, can you check why?

Yes, I updated the patch from poky to meta-yocto, but they are in OE-core

krzk (changes requested):
Missing commit msg.

ricardosalveti (approved):
LGTM

ndechesne:
I am ok with this change. But I would like @sbanerjee-quic and/or @quic-vkraleti to review/approve before we merge.

ricardosalveti:
There is no functional change here, just using the repositories directly instead of using poky. Once merged we can add a new fragment for meta-qcom-distro here (since there is no dependency on poky / meta-yocto) and update https://github.com/qualcomm-linux/meta-qcom-distro/pull/8 to only include the github workflows.

lumag (approved):

sbanerjee-quic (approved):

PR #741, opened 2025-02-08
Follow QCA firmware packages split

lumag:
OE-Core has split single `linux-firmware-qca` package into smaller packages. Follow the split and select only required firmware packages for the Qualcomm boards.


ndechesne (approved):

ricardosalveti (approved):
LGTM

PR #744, opened 2025-02-10
Add BSP selection support to machine conf

quic-vkraleti:
For better configuration of QCOM BSP, introduce BSP selection support with valid selections being ‘base’ and ‘custom’. While neither of these is the default offering, MACHINEs or DISTROs may change the BSP selection as needed. 

For example add following in machine.conf to choose 'base' BSP support. 
QCOM_SELECTED_BSP = "base"

### DIFF
--- a/conf/machine/include/qcom-base.inc
+++ b/conf/machine/include/qcom-base.inc
@@ -1,5 +1,11 @@
 # Common configurations and variables for Qualcomm platforms.
 
+# Supported BSP selections are: base, custom.
+# Specific MACHINEs or DISTROs may change this
+# selection as needed in their configuration.
+QCOM_SELECTED_BSP ??= "none"
+MACHINEOVERRIDES =. "${@bb.utils.contains_any('QCOM_SELECTED_BSP', 'base custom', 'qcom-${QCOM_SELECTED_BSP}-bsp:', '', d)}"
+
 # Provider for linux kernel
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-yocto-dev"
 


ndechesne (commented):
NACK for now. 

I want to understand and see how it will be used before we can merge something like that. I do not want to clutter meta-qcom master branch with un-needed, complicated, not-clean metadata.

ricardosalveti:
And as we discussed, we should use 'base' as the default option and then see how we can add the 'custom' providers as we start integrating them (hopefully as overlays).

lumag (changes requested):
A total NAK from my side too. This doesn't explain, why this is necessary, there is no description of the problem that you are trying to solve. It doesn't provide any kind of an actual use for these overrides, etc. (no API without users is a typical requirement). There is no explanation what is the difference between 'none', 'base' and 'custom' values, etc.

quic-vkraleti:
> And as we discussed, we should use 'base' as the default option and then see how we can add the 'custom' providers as we start integrating them (hopefully as overlays).

If 'base' is default, for boards supporting linux-yocto-dev/linux-qcom kernels, need to find out a way to choose between these & linux-qcom kernel. For now, assumed 'none' variant selection will use linux-yocto-dev & base/custom variant selection will pick corresponding linux-qcom kernel. I am open for suggestions.

quic-vkraleti:
> A total NAK from my side too. This doesn't explain, why this is necessary, there is no description of the problem that you are trying to solve. It doesn't provide any kind of an actual use for these overrides, etc. (no API without users is a typical requirement). There is no explanation what is the difference between 'none', 'base' and 'custom' values, etc.

These BSP variants are introduced to allow users to experience and experiment with the value-added software provided by QCOM, which is either not present in the upstream or is present but not feature-rich.
This variant selection will also provide the ability to quickly switch among multiple competing frameworks in userspace, (Eg, freedreno vs Adreno., Pulseaudio vs pipewire) and find out what best suits the user needs.
The ‘none’ BSP variant simply fallback to the existing MACHINEOVERRIDE selection without disturbing anything in the composition. 

An example usage of these variant selections is part of https://github.com/qualcomm-linux/meta-qcom/pull/745 Where the kernel recipe is decided based on the variant.

lumag:
The API goes with the user, so this commit should have been a part of #745. But that PR already got several NAKs on its own.

quic-vkraleti:
Please consider this PR as closed. Moved the BSP variant selection support change also into https://github.com/qualcomm-linux/meta-qcom/pull/745. So that every thing can be reviewed at one go.

PR #745, opened 2025-02-11
Add BSP selection support, linux-qcom 6.12 recipe, and use them in qcs6490-rb3gen2-core-kit compilation

quic-vkraleti:
For better configuration of QCOM BSP, This PR introduces BSP selection support with valid selections being ‘base’ and ‘custom’. While neither of these is the default offering, MACHINEs or DISTROs may change the BSP selection as needed.
For example add following in machine.conf to choose 'base' BSP support.
QCOM_SELECTED_BSP = "base"

This PR also add a recipe to compile linux-qcom 6.12 kernel and demonstrates how base and custom BSP variant overrides can be setup on a same machine to choose device tree and defconfig fragments. When none of the BSP overrides are set effectively linux-yocto-dev is built.

This PR also updated firmware selection loigc based on 'base' and 'custom' BSP variant in use.

### DIFF
--- a/conf/machine/include/qcom-base.inc
+++ b/conf/machine/include/qcom-base.inc
@@ -9,6 +9,7 @@ MACHINEOVERRIDES =. "${@bb.utils.contains('QCOM_SELECTED_BSP', 'custom', 'qcom-b
 
 # Provider for linux kernel
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-yocto-dev"
+PREFERRED_PROVIDER_virtual/kernel:qcom-base-bsp ?= "linux-qcom"
 
 KERNEL_IMAGETYPE ?= "Image"
 KERNEL_IMAGETYPES ?= "Image.gz"
--- a/conf/machine/include/qcom-base.inc
+++ b/conf/machine/include/qcom-base.inc
@@ -1,5 +1,12 @@
 # Common configurations and variables for Qualcomm platforms.
 
+# Supported BSP selections are: 'base', 'custom'. Where 'custom' selection is an overlay on top of 
+# 'base'.i.e. with 'custom' BSP selection both base & custom overrides come into effect.
+# Specific MACHINEs or DISTROs may change this selection as needed in their configuration.
+QCOM_SELECTED_BSP ??= "none"
+MACHINEOVERRIDES =. "${@bb.utils.contains('QCOM_SELECTED_BSP', 'base', 'qcom-base-bsp:', '', d)}"
+MACHINEOVERRIDES =. "${@bb.utils.contains('QCOM_SELECTED_BSP', 'custom', 'qcom-base-bsp:qcom-custom-bsp:', '', d)}"
+
 # Provider for linux kernel
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-yocto-dev"
 
--- a/conf/machine/qcs6490-rb3gen2-core-kit.conf
+++ b/conf/machine/qcs6490-rb3gen2-core-kit.conf
@@ -10,6 +10,10 @@ KERNEL_DEVICETREE ?= " \
                       qcom/qcs6490-rb3gen2.dtb \
                       "
 
+KERNEL_DEVICETREE:qcom-custom-bsp ?= " \
+                      qcom/qcs6490-addons-rb3gen2.dtb \
+                      "
+
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     linux-firmware-qcom-adreno-a660 \
     linux-firmware-qcom-qcm6490-adreno \
--- a/conf/machine/qcs6490-rb3gen2-core-kit.conf
+++ b/conf/machine/qcs6490-rb3gen2-core-kit.conf
@@ -14,9 +14,15 @@ KERNEL_DEVICETREE:qcom-custom-bsp ?= " \
                       qcom/qcs6490-addons-rb3gen2.dtb \
                       "
 
-MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
+MACHINE_FIRMWARE_PKGS ?= " \
     linux-firmware-qcom-adreno-a660 \
     linux-firmware-qcom-qcm6490-adreno \
     linux-firmware-qcom-qcm6490-audio \
     linux-firmware-qcom-qcm6490-compute \
 "
+
+MACHINE_FIRMWARE_PKGS:qcom-custom-bsp ?= " \
+    packagegroup-firmware-rb3gen2 \
+"
+
+MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += "${MACHINE_FIRMWARE_PKGS}"
--- a/recipes-kernel/linux/linux-qcom-headers_6.12.bb
+++ b/recipes-kernel/linux/linux-qcom-headers_6.12.bb
@@ -1,0 +1,68 @@
+SECTION = "kernel"
+
+DESCRIPTION = "Linux ${PV} kernel headers required to build userspace."
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
+
+inherit kernel-arch
+
+COMPATIBLE_MACHINE = "(qcom)"
+
+LINUX_QCOM_GIT ?= "git://git.codelinaro.org/clo/la/kernel/qcom.git;protocol=https"
+SRCBRANCH ?= "qclinux.6.12.y"
+SRC_URI = "${LINUX_QCOM_GIT};branch=${SRCBRANCH}"
+
+SRCREV = "608736233891ff0d6575f440d94cc2c821a7b87f"
+PV = "6.12+git${SRCPV}"
+
+S = "${WORKDIR}/git"
+
+DEPENDS += "flex-native bison-native rsync-native"
+
+do_configure[noexrec] = "1"
+do_compile[noexec] = "1"
+
+do_install () {
+    cd ${B}
+    headerdir=${B}/headers
+    kerneldir=${D}${includedir}/linux-kernel-qcom
+    install -d $kerneldir
+
+    # Install all headers inside B and copy only required ones to D
+    oe_runmake_call -C ${B} ARCH=${ARCH} headers_install O=$headerdir
+
+    if [ -d $headerdir/include/generated ]; then
+        mkdir -p $kerneldir/include/generated/
+        cp -fR $headerdir/include/generated/* $kerneldir/include/generated/
+    fi
+
+    if [ -d $headerdir/arch/${ARCH}/include/generated ]; then
+        mkdir -p $kerneldir/arch/${ARCH}/include/generated/
+        cp -fR $headerdir/arch/${ARCH}/include/generated/* $kerneldir/arch/${ARCH}/include/generated/
+    fi
+
+    if [ -d $headerdir/${includedir} ]; then
+        mkdir -p $kerneldir/${includedir}
+        cp -fR $headerdir/${includedir}/* $kerneldir/${includedir}
+    fi
+
+    # Remove .install and .cmd files
+    find $kerneldir/ -type f -name .install | xargs rm -f
+    find $kerneldir/ -type f -name "*.cmd" | xargs rm -f
+}
+
+# kernel headers are generally machine specific
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+# Allow to build empty main package, to include -dev package into the SDK
+ALLOW_EMPTY_${PN} = "1"
+
+FILES_${PN}-dev += "linux-qcom/*"
+
+INHIBIT_DEFAULT_DEPS = "1"
+
+python () {
+    mach_overrides = d.getVar('MACHINEOVERRIDES').split(":")
+    if ('qcom-custom-bsp' not in mach_overrides):
+        raise bb.parse.SkipRecipe("linux-qcom-headers are compatible only with qcom-custom-bsp")
+}
--- a/recipes-kernel/linux/linux-qcom_6.12.bb
+++ b/recipes-kernel/linux/linux-qcom_6.12.bb
@@ -1,0 +1,73 @@
+SECTION = "kernel"
+
+DESCRIPTION = "Linux ${PV} kernel for QCOM devices"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
+
+inherit kernel
+
+COMPATIBLE_MACHINE = "(qcom)"
+
+LINUX_QCOM_GIT ?= "git://git.codelinaro.org/clo/la/kernel/qcom.git;protocol=https"
+SRCBRANCH ?= "qclinux.6.12.y"
+SRC_URI = "${LINUX_QCOM_GIT};branch=${SRCBRANCH}"
+
+SRCREV = "608736233891ff0d6575f440d94cc2c821a7b87f"
+PV = "6.12+git${SRCPV}"
+
+S = "${WORKDIR}/git"
+
+KERNEL_CONFIG ??= "qcom_defconfig"
+
+# Additional configs for qcom custom variant kernel.
+KERNEL_CONFIG_FRAGMENTS:append:qcom-custom-bsp = " ${S}/arch/arm64/configs/qcom_addons.config"
+
+kernel_conf_variable() {
+    sed -e "/CONFIG_$1[ =]/d;" -i ${B}/.config
+    if test "$2" = "n"
+    then
+        echo "# CONFIG_$1 is not set" >> ${B}/.config
+    else
+        echo "CONFIG_$1=$2" >> ${B}/.config
+    fi
+}
+
+do_configure:prepend() {
+    if [ ! -f "${S}/arch/${ARCH}/configs/${KERNEL_CONFIG}" ]; then
+        bbfatal "KERNEL_CONFIG '${KERNEL_CONFIG}' was specified, but not present in the source tree"
+    else
+        cp '${S}/arch/${ARCH}/configs/${KERNEL_CONFIG}' '${B}/.config'
+    fi
+
+    # Check for kernel config fragments.  The assumption is that the config
+    # fragment will be specified with the absolute path.  For example:
+    #   * ${WORKDIR}/config1.cfg
+    #   * ${S}/config2.cfg
+    # Iterate through the list of configs and make sure that you can find
+    # each one.  If not then error out.
+    # NOTE: If you want to override a configuration that is kept in the kernel
+    #       with one from the OE meta data then you should make sure that the
+    #       OE meta data version (i.e. ${WORKDIR}/config1.cfg) is listed
+    #       after the in kernel configuration fragment.
+    # Check if any config fragments are specified.
+    if [ ! -z "${KERNEL_CONFIG_FRAGMENTS}" ]
+    then
+        for f in ${KERNEL_CONFIG_FRAGMENTS}
+        do
+            # Check if the config fragment was copied into the WORKDIR from
+            # the OE meta data
+            if [ ! -e "$f" ]
+            then
+                echo "Could not find kernel config fragment $f"
+                exit 1
+            fi
+        done
+
+        # Now that all the fragments are located merge them.
+        ( cd ${WORKDIR} && ${S}/scripts/kconfig/merge_config.sh -m -r -O ${B} ${B}/.config ${KERNEL_CONFIG_FRAGMENTS} 1>&2 )
+    fi
+}
+
+do_configure:append() {
+    oe_runmake -C ${S} O=${B} savedefconfig && cp ${B}/defconfig ${WORKDIR}/defconfig.saved
+}
--- a/recipes-kernel/linux/linux-qcom_6.12.bb
+++ b/recipes-kernel/linux/linux-qcom_6.12.bb
@@ -12,6 +12,9 @@ LINUX_QCOM_GIT ?= "git://git.codelinaro.org/clo/la/kernel/qcom.git;protocol=http
 SRCBRANCH ?= "qclinux.6.12.y"
 SRC_URI = "${LINUX_QCOM_GIT};branch=${SRCBRANCH}"
 
+# Patch to address a compilation issue in ice driver with GCC 14.2
+SRC_URI:append = " file://0001-QCLINUX-soc-qcom-ice-Type-cast-return-error-code-as-.patch"
+
 SRCREV = "608736233891ff0d6575f440d94cc2c821a7b87f"
 PV = "6.12+git${SRCPV}"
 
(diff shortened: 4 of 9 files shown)


ndechesne:
The v6.12 is a regression since we are currently using 6.13 (linux-yocto-dev). Going backward do not seem appropriate for our main development branch. How soon are we going to have 6.13 or 6.14 dev branch for the kernel? Perhaps we should wait for it.

ndechesne:
> NAK, use linux-yocto instead of linux-qcom. Add necessary patches on top of that kernel. Using external kernel trees reduces visibility of the patch status and also makes it harder to integrate other layers which frequently .bbappend linux-yocto, but know nothing about linux-qcom.

Not quite. Well, there is a little bit of context to add to have a fully open conversation here. 

There are 2 main purpose for meta-qcom, maybe 3. 

1. We want meta-qcom to be used for the Qualcomm Linux 'mainline' development. e.g. the place where all dev teams can get the latest / greatest version of everything and to be used for development. This is what we want for the 'master' branch. Up until now we've used linux-yocto-dev for that reason. Our kernel team wants to provide a continuously up-to-date kernel that we can consume in the distros that we support/build, and we are starting to support Debian builds as well. So the idea is to decouple from linux-yocto-dev in favor of a 'rolling' kernel development tree to be used in Yocto 'master' and in Debian builds.
2. We want meta-qcom LTS branches to be the foundation for future Qualcomm commercial Linux offering, starting with the next LTS in 2026. So the development which we are doing on the master branch today, should become the baseline for commercial support in 2026. And we will need to decide how to deal with the kernel, which will then be aligned with linux-yocto (LTS).
3. Now you are suggesting support for linux-yocto in master branch, which is not a use case I had in minds.. so we need to discuss the value of that.

recipes-kernel/linux/files/disable_crypto.config
```diff
@@ -0,0 +1,3 @@
+# CONFIG_MMC_CRYPTO is not set
+# CONFIG_SCSI_UFS_CRYPTO is not set
+# CONFIG_QCOM_INLINE_CRYPTO_ENGINE is not set
```
lumag, line 3: Commit message is useless. What kind of issues? Describe them so that other developers can reproduce them (and then decide when they are solved). Please add a Link: tag pointing to the bug report to the upstream.

lumag:
My proposal is very simple: use linux-yocto+patches for OE, use Debian kernel+patches for Debian builds. Using vendor kernel results in all different kinds of patches and strange workarounds in it with no clear background, no reviews, etc.

recipes-kernel/linux/linux-qcom-custom_6.12.bb
```diff
@@ -0,0 +1,6 @@
+
```
ricardosalveti, line 1: We don't need two kernel recipes in order to have a behavior defined by the override, we could as well just have one single linux-qcom recipe and additions/removals be done based on the override as part of the recipe (e.g. in this case adding one extra fragment).

conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -10,6 +10,14 @@ KERNEL_DEVICETREE ?= " \
                       qcom/qcs6490-rb3gen2.dtb \
                       "
 
+KERNEL_DEVICETREE:qcom-base-bsp ?= " \
+                      qcom/qcs6490-rb3gen2.dtb \
+                      "
+
+KERNEL_DEVICETREE:qcom-custom-bsp ?= " \
+                      qcom/qcs6490-addons-rb3gen2.dtb \
```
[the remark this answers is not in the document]
ricardosalveti: I believe overlays are indeed being used to change the behavior of the kernel (e.g. changing compats to make it use other drivers), which I believe it is also how rpi is doing in a way (agree it is not using as it should).

Besides changing how the kernel is built and the rootfs behavior (e.g. loading x or y modules), is there any other suggested approach the teams could take?

recipes-kernel/linux/files/disable_crypto.config
ricardosalveti: Yeah, please describe the issue, as this is an interesting one and we should probably find how to get this fixed instead.

ricardosalveti:
> > My proposal is very simple: use linux-yocto+patches for OE, use Debian kernel+patches for Debian builds. Using vendor kernel results in all different kinds of patches and strange workarounds in it with no clear background, no reviews, etc.
> 
> As you can imagine, it's not quite practical given the rather large number of outstanding patches we need to manage (pending upstream merge). I know it's not ideal, but it's a compromise to allow us a faster kernel dev/integration pace. For distro/commercial support, the situation might be different. We do want to have 'official' distro support using the official distro kernel.

From the yocto side having a different kernel provider as a recipe is fine, but I also agree we should first discuss how to best manage the vendor integration tree and changes.

My preference would be to have an integration tree available under qualcomm-linux (a bit similar to the qcom lt integration tree), with a well defined set of rules and update process (as we don't want to merge strange workarounds, etc), and then have a recipe pointing to such tree. Maintaining 200+ patches via this layer is quite a bit of work, and hard to review and maintain, and won't necessarily help when integrating the same changes on debian, as the same work will be duplicated.

So here I believe we need to:

1. Decide how and where to maintain the vendor changes / tree;
2. Change this pr to simply add the kernel recipe as a provider (dealing with base only first, we can discuss how to integrate 'custom' as a follow up);

lumag:
Ack. A well defined process is fine with me. What about havng a yocto-based git tree (+patches on top)? Then we can basically follow yocto process _and_ package the same kernel tree for Debian? And, BTW, if that works we can do yet another step: define Debian kernel as linux-yocto basline + set of patches from this tree. Then at any point we can reapply them, build a source tree, package it as a set of Debian packages and use it.

conf/machine/qcs6490-rb3gen2-core-kit.conf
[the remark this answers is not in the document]
lumag: The DT must be describing the hardware, no matter what modules are to be used. Then you can use blacklist unnecessary modules or use Kconfig to select between required modules. DT does not describe drivers and as such it _must_ be the same.

quic-vkraleti:
> > The v6.12 is a regression since we are currently using 6.13 (linux-yocto-dev). Going backward do not seem appropriate for our main development branch. How soon are we going to have 6.13 or 6.14 dev branch for the kernel? Perhaps we should wait for it.
> 
> I don't understand what regression mean here as different recipe selecting different kernel variants.
> 
> BAse and Custom would always use QCOM kernel. You can consider pristine linux-yocto-dev as third offering that remains aligned to community.

@ndechesne started linux-qcom with 6.12 as it is currently available in house. However we'll keep catching up with linux-yocto-dev version at a set cadence. May be we can discuss what that period could be and have plan in place for  6.13, 6.14 dev branches.

recipes-kernel/linux/linux-qcom-custom_6.12.bb
quic-vkraleti: Discussed a bit around this internally and felt having two recipes will give a clear view to user what kernel is in use. As this is what 1.0 is also doing (I ack, on 1.0 as kernel trees are different, so can't use single recipe) felt it'd be easy to corelate. If doing a single recipe is what meta-qcom maintainers recommend, I'll update accordingly.

recipes-kernel/linux/files/disable_crypto.config
quic-vkraleti: Apologies for not being explicit in commit message. It is  a type casting error from 
https://git.codelinaro.org/clo/la/kernel/qcom/-/blob/torvalds/master/drivers/soc/qcom/ice.c?ref_type=heads#L335

> build/tmp/work-shared/qcs6490-rb3gen2-core-kit/kernel-source/drivers/soc/qcom/ice.c:635:24: error: returning 'long int' from a function with return type 'struct qcom_ice *' makes pointer from integer without a cast [-Wint-conversion]
> 2050   635 |                 return PTR_ERR(base);

Instead of disabling the driver will check if the code can be fixed.

conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -10,6 +10,14 @@ KERNEL_DEVICETREE ?= " \
                       qcom/qcs6490-rb3gen2.dtb \
                       "
 
+KERNEL_DEVICETREE:qcom-base-bsp ?= " \
+                      qcom/qcs6490-rb3gen2.dtb \
+                      "
```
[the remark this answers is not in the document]
quic-vkraleti: Typo in commit message. Will address in next patchset.

recipes-kernel/linux/linux-qcom-custom_6.12.bb
lumag: The main issue is that I don't see a good explanation of the difference between 'base' and 'custom'.

conf/machine/include/qcom-base.inc
```diff
@@ -1,5 +1,12 @@
 # Common configurations and variables for Qualcomm platforms.
 
+# Supported BSP selections are: 'base', 'custom'. Where 'custom' selection is an overlay on top of 
+# 'base'.i.e. with 'custom' BSP selection both base & custom overrides come into effect.
+# Specific MACHINEs or DISTROs may change this selection as needed in their configuration.
+QCOM_SELECTED_BSP ??= "none"
```
lumag, line 6: Updated commit message still doesn't explain, _why_ do you need two BSP variants. What is going to be guarded by the 'custom' override. In the end, everything is custom.

recipes-kernel/linux/linux-qcom-headers_6.12.bb
```diff
@@ -0,0 +1,68 @@
+SECTION = "kernel"
+
+DESCRIPTION = "Linux ${PV} kernel headers required to build userspace."
```
lumag, line 3: No. That's the standard answer regarding kernel uAPI headers. Please copy corresponding headers to your userspace applications instead of depending on the kernel headers. There is a very limited set of usecases for using kernel-provided headers, which is mostly limited to building of libc.

recipes-kernel/linux/linux-qcom-headers_6.12.bb
```diff
@@ -0,0 +1,68 @@
...
+    find $kerneldir/ -type f -name "*.cmd" | xargs rm -f
+}
+
+# kernel headers are generally machine specific
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+# Allow to build empty main package, to include -dev package into the SDK
+ALLOW_EMPTY_${PN} = "1"
+
+FILES_${PN}-dev += "linux-qcom/*"
```
lumag, line 60: It doesn't look like this was actually tested. It is not a dir where the headers are being installed.

conf/machine/qcs6490-rb3gen2-core-kit.conf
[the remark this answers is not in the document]
lumag: It is fine to use DT overlays to describe additional hardware (mezzanines, externally conected devices, etc). But DT must nost be used to select drivers.

lumag (changes requested):

conf/machine/qcs6490-rb3gen2-core-kit.conf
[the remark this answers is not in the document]
ricardosalveti: Right, I believe the thinking here was to try to use the same kernel build (with an extended config option) and find an automatic way to disable/enable drivers at runtime, and overlays is an easy way as it can be managed at a higher level (probably also why this is heavily used on rpi).

One other way would indeed just have a denylist of some sort, but that moves the configuration to the distro (which should probably be fine, a single package / config can be created which would change the behavior). 

We might just need to review and revisit all the custom overlays as they might be including other downstream changes which might not be describing the hardware and instead the behavior.

recipes-kernel/linux/linux-qcom-custom_6.12.bb
ricardosalveti: I think it is better to start with one and later split if we see the need.

lumag: Agree. However if there is a single recipe, there is no need to have `QCOM_SELECTED_BSP` at all.

ricardosalveti: > The main issue is that I don't see a good explanation of the difference between 'base' and 'custom'.

'custom' would include downstream / proprietary extras, in this case it could be a kernel config fragment that enables / disables certain drivers for example.

There are other BSP layers handling this a bit differently, but most end up creating overrides. Meta-freescale uses generic-bsp for sourcing upstream components and nxp-bsp for sourcing downstream / vendor BSP components, and that is how they manage using the open source gpu driver x proprietary one at build time, for example.

ricardosalveti:
> What about havng a yocto-based git tree (+patches on top)? Then we can basically follow yocto process _and_ package the same kernel tree for Debian? And, BTW, if that works we can do yet another step: define Debian kernel as linux-yocto basline + set of patches from this tree. Then at any point we can reapply them, build a source tree, package it as a set of Debian packages and use it.

I believe that is how it was done with this tree (based on linux-yocto/linux-yocto-dev), but to avoid yocto specifics in other distros (since this would be the reference kernel for any distro, for example), creating a generic based on linus tree seems to be easier to manage and expose to other users.

This is currently under discussion, and @ndechesne is taking the lead from our side.

recipes-kernel/linux/linux-qcom-custom_6.12.bb
lumag: This should be a distro-level decision, pulling required set of modules through packagegroups. In other words instead of having 'custom' BSP, it is much easier if it is possible to build both sets of drivers / modules / etc and just let the user select correct packages via dependencies. For examples, this would allow users to select MSM GPU driver _and_ downstream VIDC or WiFi drivers.

ricardosalveti: Sure, the final decision is done by the distro, the override is only useful to have in the BSP when that creates dependencies with other components (e.g. vendor bsp requiring vendor kernel + vendor userspace + vendor boot component).

But we can create the override when we see the need.

recipes-kernel/linux/linux-qcom_6.12.bb
```diff
@@ -0,0 +1,73 @@
...
+
+inherit kernel
+
+COMPATIBLE_MACHINE = "(qcom)"
+
+LINUX_QCOM_GIT ?= "git://git.codelinaro.org/clo/la/kernel/qcom.git;protocol=https"
+SRCBRANCH ?= "qclinux.6.12.y"
+SRC_URI = "${LINUX_QCOM_GIT};branch=${SRCBRANCH}"
+
+SRCREV = "608736233891ff0d6575f440d94cc2c821a7b87f"
```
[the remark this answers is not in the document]
ricardosalveti: I might be wrong, but I believe this was done as a stop-gap until the team is ready to transition to the latest in-dev rev (e.g. 6.14-rcX).

recipes-kernel/linux/linux-qcom-custom_6.12.bb
lumag: This still can be expressed as package dependencies instead.

recipes-kernel/linux/linux-qcom_6.12.bb
```diff
@@ -0,0 +1,73 @@
...
+            fi
+        done
+
+        # Now that all the fragments are located merge them.
+        ( cd ${WORKDIR} && ${S}/scripts/kconfig/merge_config.sh -m -r -O ${B} ${B}/.config ${KERNEL_CONFIG_FRAGMENTS} 1>&2 )
+    fi
+}
+
+do_configure:append() {
+    oe_runmake -C ${S} O=${B} savedefconfig && cp ${B}/defconfig ${WORKDIR}/defconfig.saved
```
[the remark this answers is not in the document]
ricardosalveti: Probably a copy & paste error from linux-linaro-qcom.inc

recipes-kernel/linux/linux-qcom-headers_6.12.bb
```diff
@@ -0,0 +1,68 @@
...
+PV = "6.12+git${SRCPV}"
+
+S = "${WORKDIR}/git"
+
+DEPENDS += "flex-native bison-native rsync-native"
+
+do_configure[noexrec] = "1"
+do_compile[noexec] = "1"
+
+do_install () {
```
ricardosalveti, line 25: I would prefer something more explicit as done by https://github.com/Freescale/meta-freescale/blob/master/recipes-kernel/linux/linux-imx-headers_6.6.bb#L21, if a common headers recipe is indeed required.

ricardosalveti:
I would prefer just having the new kernel recipe proposed first and skip any 'custom' configuration out of the initial changes, as we should review those separately.

With the kernel recipe we can already create multiple builds based on the preferred provider settings done at the distro / local.conf level, and start validating it.

Once we see the need of a custom override, we can review and see what would be the best way to merge those, as in the end we want custom to just be a clean set of overlays, which hopefully should all be able to be built together, and installed based on what the distro decides.

recipes-kernel/linux/linux-qcom-custom_6.12.bb
ricardosalveti: Right, which is why I want to see the need first in order to justify it. In meta-freescale it makes sense as there are quite a few configure options which are based on the bsp selected, not just providers, but not sure if we will have those here yet.

recipes-kernel/linux/linux-qcom-headers_6.12.bb
lumag: If userspace depends on particular kernel headers, it means there is no uAPI between user- and kernel- space. Userspace tools must vendor in required headers or just constants. Not to mention that using kernel headers in the userspace is generally impossible. They use kernel-based types (u32, u64, etc), while userspace should use stdint-types (uint32_t, etc).

quic-vkraleti:
> I would prefer just having the new kernel recipe proposed first and skip any 'custom' configuration out of the initial changes, as we should review those separately.
> 
> With the kernel recipe we can already create multiple builds based on the preferred provider settings done at the distro / local.conf level, and start validating it.
> 
> Once we see the need of a custom override, we can review and see what would be the best way to merge those, as in the end we want custom to just be a clean set of overlays, which hopefully should all be able to be built together, and installed based on what the distro decides.

@ndechesne @shashim-quic @sbanerjee-quic If you also agree for this, I am more than happy to submit only kernel recipe for now and add CI variant to build it along with linux-yocto-dev.

@lumag, as @ricardosalveti mentioned there are discussions happening around the best possible way to host qcom kernel tree. Till they conclude please do accept this approach of hosting a tree in CLO and refer it in linux-qcom recipe. Many others are waiting for this recipe to land to start submitting their usersapce recipes to meta-qcom master branch.

lumag:
@quic-vkraleti the headers patch is NAKed. I'm fine with getting the qclinux package in. Please make it non-preferred by default.

ndechesne:
> > I would prefer just having the new kernel recipe proposed first and skip any 'custom' configuration out of the initial changes, as we should review those separately.
> > With the kernel recipe we can already create multiple builds based on the preferred provider settings done at the distro / local.conf level, and start validating it.
> > Once we see the need of a custom override, we can review and see what would be the best way to merge those, as in the end we want custom to just be a clean set of overlays, which hopefully should all be able to be built together, and installed based on what the distro decides.
> 
> @ndechesne @shashim-quic @sbanerjee-quic If you also agree for this, I am more than happy to submit only kernel recipe for now and add CI variant to build it along with linux-yocto-dev.
> 
> @lumag, as @ricardosalveti mentioned there are discussions happening around the best possible way to host qcom kernel tree. Till they conclude please do accept this approach of hosting a tree in CLO and refer it in linux-qcom recipe. Many others are waiting for this recipe to land to start submitting their usersapce recipes to meta-qcom master branch.

Unfortunately, we need to wait to have an agreed dev process for the kernel before we can merge the kernel recipe. Whoever needs it now, can pull in this unmerged PR locally for development.

lumag:
@quic-vkraleti to unblock userspace developers you can start submitting userspace apps after vendoring in necessary kernel headers.

idlethread:
> > What about havng a yocto-based git tree (+patches on top)? Then we can basically follow yocto process _and_ package the same kernel tree for Debian? And, BTW, if that works we can do yet another step: define Debian kernel as linux-yocto basline + set of patches from this tree. Then at any point we can reapply them, build a source tree, package it as a set of Debian packages and use it.
> 
> I believe that is how it was done with this tree (based on linux-yocto/linux-yocto-dev), but to avoid yocto specifics in other distros (since this would be the reference kernel for any distro, for example), creating a generic based on linus tree seems to be easier to manage and expose to other users.
> 
> This is currently under discussion, and @ndechesne is taking the lead from our side.

We really shouldn't build a git tree specific to Yocto and then reuse for Debian. And then RHEL. And so on...

IMO, it is also wrong to expect all teams to have Yocto or distro-specific expertise and track their changes as individual patches.

As this point, we're leaning towards maintaining a distro-agnostic single kernel integration tree with the build configuration staying in the distro recipes/packaging.

lumag:
I'd really suggest to shift the focus from the kernel team integrating patches into the single tree into kernel team providing the patches upstream and then pinging meta-qcom maintainers to integrate the series once it has been at least  reviewed by the upstream community. Doing it in any other way results in variable quality of patches being integrated into the tree, no way for anybody from outside to track those changes, etc.

Likewise having an distro-agnostic tree means that there is no easy way for the outside developers (including community) to contribute to the kernel. E.g. one of my colleagues is interested in using meta-qcom for SM8550 devkits. If the patches are handled withing meta-qcom, he can directly work on enabling missing features by updating the recipe. If the kernel is hosted as an external CLO repo, he has no sensible way of collaborating.

The only viable alternative to linux-yocto+patches (from my point of view) is to move the in-development kernel repo to github, enable pull requests and work on developing the kernel by following the open process, allowing external contributions, etc.

lumag:
As a second comment, I stopped believing into distro-agnostic kernels. That works quite nice if you just need to enable the platform. However in a longer term that approach has a lot of deficiencies. The resulting kernel isn't well-integrated into the distribution. For example, making a Debian installation image using an external kernel is a pain.

ndechesne:
> The only viable alternative to linux-yocto+patches (from my point of view) is to move the in-development kernel repo to github, enable pull requests and work on developing the kernel by following the open process, allowing external contributions, etc.

This is exactly where we are heading to. We aim to have a kernel branch on github.com/qualcomm-linux, which will be continuously updated (tracking kernel.org master) where we integrate ongoing patches being under development (and being upstreamed). The 6.12 branch on CLO is just a 'first' iteration of what it might look like, and something we use to see what needs to be improved in the process. 

Note that meta-qcom can still take patches as metadata on top of the distro agnostic kernel. 

I also agree that distro agnostic kernel is not appropriate for product. The goal of this kernel and its recipe in meta-qcom is to accelerate kernel development, and our upstream pace. For commercially support product, the kernel will be a distro kernel. E.g. a QCOM customer shipping a Yocto based distro will get a yocto based kernel, if they use Ubuntu, they will use the Ubuntu kernel, .. the goal is to have more content upstream (and then available by default in each distro kernel), and a clear split of what is not upstream (or not upstreamable) so that the distro know how and what to integrate with the non upstream content.

lumag:
Well... this means that the product kernel will not be tested during development. I don't think it is a good idea.

lumag:
Closing, it has been replaced by #784

PR #748, opened 2025-02-13
Fix default vfat sector size for the UFS devices and partitions

ricardosalveti:
Set default vfat sector size for UFS devices and partitions.

### DIFF
--- a/classes/linux-qcom-dtbbin.bbclass
+++ b/classes/linux-qcom-dtbbin.bbclass
@@ -17,7 +17,7 @@ do_qcom_dtbbin_deploy() {
         dtb_base_name=`basename $dtb .$dtb_ext`
         mkdir -p ${DTBBIN_DEPLOYDIR}/$dtb_base_name
         cp ${D}/${KERNEL_DTBDEST}/$dtb_base_name.dtb ${DTBBIN_DEPLOYDIR}/$dtb_base_name/combined-dtb.dtb
-        mkfs.vfat $@ -C ${DTBBIN_DEPLOYDIR}/dtb-${dtb_base_name}-image.vfat ${DTBBIN_SIZE}
+        mkfs.vfat -C ${DTBBIN_DEPLOYDIR}/dtb-${dtb_base_name}-image.vfat ${DTBBIN_SIZE}
         mcopy -i "${DTBBIN_DEPLOYDIR}/dtb-${dtb_base_name}-image.vfat" -vsmpQ ${DTBBIN_DEPLOYDIR}/$dtb_base_name/* ::/
         rm -rf ${DTBBIN_DEPLOYDIR}/$dtb_base_name
     done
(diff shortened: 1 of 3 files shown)


lumag (approved):

quic-vkraleti (approved):
LGTM

koenkooi:
LGTM

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -7,6 +7,10 @@
 DTBBIN_DEPLOYDIR = "${WORKDIR}/qcom_dtbbin_deploy-${PN}"
 DTBBIN_SIZE ?= "4096"
 
+# UFS requires vfat sector size of 4096 (default is 512)
+VFAT_SECTOR_SIZE ?= "4096"
```
ndechesne, line 11: while the change is needed, it's buried down here. how are you planning to support eMMC devices? Wouldn't it better to have a parameter in the $MACHINE conf file instead, so that it's not hidden here?

lumag: Isn't it backwards compatible? In othere words, you can have 4k FS on the storage with 512-byte sectors, but you can't have 512-byte FS on 4k storage.

ricardosalveti: It is generally supported, which is why I added it to be the default but with an easy way to override. We could move to machine conf with ?=, but users can still build the exact same image and boot over sdcard, for example, so more up to the user configuration than machine really.

ricardosalveti: Yup, that would work, let me update it.

ricardosalveti: Updated

lumag:
I think that there are more platforms that use UFS than eMMC. Wouldn't it be better to default to 4096? In the end, eMMC will work with 4k clusters, while UFS will not work with 512-byte clusters.

conf/machine/include/qcom-common.inc
```diff
@@ -32,6 +32,9 @@ IMAGE_ROOTFS_ALIGNMENT ?= "4096"
 QCOM_BOOTIMG_KERNEL_BASE ?= "0x80000000"
 QCOM_BOOTIMG_PAGE_SIZE ?= "4096"
 
+# Default vfat sector size, customized per platform when UFS is used
+VFAT_SECTOR_SIZE ??= "512"
```
ndechesne, line 36: Should we prefix this variable (now that it is a global var) with QCOM_?

ricardosalveti: Updated.

ricardosalveti:
Made 4096 the default value, changed to QCOM_VFAT_SECTOR_SIZE.

lumag (approved):

PR #749, opened 2025-02-13
recipes-bsp/firmware: package QAR2130P firmware

lumag:
Package firmware for the Qualcomm QAR2130P development platform.

### DIFF
--- a/recipes-bsp/firmware/firmware-qcom-sar2130p.bb
+++ b/recipes-bsp/firmware/firmware-qcom-sar2130p.bb
@@ -1,0 +1,44 @@
+# Specify location of the corresponding NON-HLOS.bin file by adding
+# NHLOS_URI:pn-firmware-qcom-sar2130p = "..."  to local.conf. Use "file://"
+# if the file is provided locally.
+
+DESCRIPTION = "QCOM Firmware for SAR2130P boards"
+
+LICENSE = "CLOSED"
+
+FW_QCOM_NAME = "sar2130p"
+
+FW_QCOM_LIST = "\
+    a620_zap.mbn \
+    adsp.mbn adsp_dtb.mbn adspr.jsn adsps.jsn adspua.jsn \
+    cdsp.mbn cdsp_dtb.mbn cdspr.jsn \
+    ipa_fws.mbn \
+"
+
+S = "${UNPACKDIR}"
+
+require recipes-bsp/firmware/firmware-qcom.inc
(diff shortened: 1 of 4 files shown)


recipes-bsp/firmware/firmware-qcom-sar2130p.bb
```diff
@@ -0,0 +1,44 @@
+# Specify location of the corresponding NON-HLOS.bin file by adding
+# NHLOS_URI:pn-firmware-qcom-sar2130p = "..."  to local.conf. Use "file://"
+# if the file is provided locally.
+
+DESCRIPTION = "QCOM Firmware for SAR2130P boards"
+
+LICENSE = "CLOSED"
+
+FW_QCOM_NAME = "sar2130p"
```
ndechesne, line 9: you are using QAR2130p and SAR2130p here. Can you explain why in the commit message?

recipes-bsp/images/initramfs-firmware-qar2130p-image.bb
```diff
@@ -0,0 +1,10 @@
+DESCRIPTION = "Tiny ramdisk image with QAR2130P devices firmware files"
```
ndechesne, line 1: can you remind me why we have all these tiny initramfs? What are the use case for that?

recipes-bsp/firmware/firmware-qcom-sar2130p.bb
lumag: done

recipes-bsp/firmware/firmware-qcom-sar2130p.bb
```diff
@@ -0,0 +1,44 @@
+# Specify location of the corresponding NON-HLOS.bin file by adding
+# NHLOS_URI:pn-firmware-qcom-sar2130p = "..."  to local.conf. Use "file://"
+# if the file is provided locally.
```
ricardosalveti, line 3: Is this because this file is not available in a public location?

lumag: Yes. It is handled in the same way as all HDKs: the owner of the board is required to provide NHLOS.bin and an archive with Adreno firmware. Several days ago I raised a question regarding Adreno bits, but got no response up to now.

ricardosalveti (approved):

ndechesne (approved):

PR #750, opened 2025-02-13
initramfs-test-image: move Qualcomm packages to -tiny-

lumag:
In order to be able to utilize initramfs-tiny-image on small targets (RB1, RB2, IFC6640, etc), move Qualcomm-related packages from test initramfs to the tiny one.


recipes-test/images/initramfs-tiny-image.bb
```diff
@@ -8,6 +8,15 @@ PACKAGE_INSTALL = " \
     udev \
 "
 
+PACKAGE_INSTALL:append:qcom = " \
```
ndechesne, line 11: why append:qcom? is that really needed? everything in there is fairly specific to qcom. If we are concerned about that we could prefix our images with qcom in there, instead?

lumag: Funny enough, the rest of the image is pretty generic. One can easily use those initramfs images on any machine.

lumag: Replaced `:append:qcom =` with `+=`

ricardosalveti (approved):

PR #753, opened 2025-02-17
RB1 support

ndechesne:
This is WIP, just to share what I have so far. It is not boot tested yet.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -76,6 +76,7 @@ jobs:
           - qcm6490-idp
           - qcs6490-rb3gen2-core-kit
           - qcs9100-ride-sx
+          - qrb2210-rb1-core-kit
           - qcom-armv8a
           - qcom-armv7a
     runs-on: [self-hosted, x86]
--- a/ci/qcm2290.yml
+++ b/ci/qcm2290.yml
@@ -1,0 +1,6 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+  includes:
+  - ci/base.yml
--- a/ci/qrb2210-rb1-core-kit.yml
+++ b/ci/qrb2210-rb1-core-kit.yml
@@ -1,0 +1,8 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+  includes:
+  - ci/qcm2290.yml
+
+machine: qrb2210-rb1-core-kit
--- a/conf/machine/include/qcom-qcm2290.inc
+++ b/conf/machine/include/qcom-qcm2290.inc
@@ -1,0 +1,11 @@
+# Configurations and variables for QCM2290 (aka QRB2210) SoC family.
+
+SOC_FAMILY = "qcm2290"
+require conf/machine/include/qcom-base.inc
+require conf/machine/include/qcom-common.inc
+
+require conf/machine/include/arm/armv8a/tune-cortexa53.inc
+
+KERNEL_CMDLINE_EXTRA ?= "earlycon clk_ignore_unused pd_ignore_unused"
+
+EXTRA_IMAGEDEPENDS += "firmware-qcom-boot-qrb2210-rb1 qcom-gen-partition-bins"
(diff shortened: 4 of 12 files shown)


lumag (changes requested):

recipes-kernel/linux/linux-yocto-dev.bbappend
```diff
@@ -1,6 +1,9 @@
 # do not override KBRANCH and SRCREV_machine, use default ones.
 COMPATIBLE_MACHINE:qcom = "(qcom)"
 
+# HACK: needed on RB1 until we fix the EFI boot
+inherit linux-qcom-bootimg
```
lumag, line 5: If this is c&p'ed from linux-linaro-qcomlt it becomes less of a hack. Let's merge it maybe?

ndechesne: I am still hoping we can get a new bootloader release that supports EFI. so for now I don't want to merge this change.

ndechesne: If we want to merge it, I will make it RB1 specific, not in this current form then.

lumag: Why? There are other RB boards and dragonboard which can also benefit from this recipe. Should we force-update all of them to the UEFI environment?

lumag: e.g. with this change one can boot linux-yocto-dev on db410c or db820c, identifying possible issues.

conf/machine/include/qcom-qrb2210.inc
```diff
@@ -0,0 +1,11 @@
+# Configurations and variables for QCS6490 SoC family.
+
+SOC_FAMILY = "qrb2210"
```
loicpoulain, line 3: Shouldn't it be `qcm2290` family to align with Linux kernel devicetree (`qrb2210-rb1.dts` including `qcm2290.dtsi` SoC)

ndechesne: this is a good point. qcm2290 is a better name. I will change it. 
that said we are using qcm6490 for the RB3G2 while the kernel platform is sc7280 ;-(

recipes-kernel/linux/linux-yocto-dev.bbappend
ndechesne: I changed the hack and moved this config in the machine file instead. I don't think we should inherit linux-qcom-bootimg by default, since it creates a bunch of boot images which might be un-expected for some users.

lumag: Nice! We might adopt the same idea for qcom-armv8a, dropping it from the recipes.

conf/machine/include/qcom-qrb2210.inc
ndechesne: I decided to go with qcm2290. just pushed a new branch.

recipes-graphics/mesa/mesa/freedreno-gpu-check.patch
```diff
@@ -0,0 +1,71 @@
...
++#include <stddef.h>
+ #include <stdint.h>
+ 
+ #ifdef __cplusplus
+@@ -379,6 +380,14 @@ fd_dev_gpu_id(const struct fd_dev_id *id)
+ /* Unmodified dev info as defined in freedreno_devices.py */
+ const struct fd_dev_info *fd_dev_info_raw(const struct fd_dev_id *id);
+ 
++/* Helper to check if GPU is known before going any further */
++static uint8_t
```
lumag, line 41: Please change this to `sstatic inline uint8_t`

koenkooi:
It's missing qbootctl from MACHINE_ESSENTIAL*

lumag:
i think it was better to have `KERNEL_CLASSES` and `MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS` in machine conf (rather than in the include file). In the end we can have different qcm2290-based machines, some of them using ABL, others using U-Boot, etc.

recipes-graphics/mesa/mesa/freedreno-gpu-check.patch
lumag: And the patch is now `Backport`, not just `Submitted`

ndechesne:
@lumag @ricardosalveti , I think it's ready for merge. I removed the hack to include all kernel modules. @koenkooi is working on a packagegroup with all the kernel modules, which is the right solution. But this branch is ready to merge, as far as I know..

lumag (approved):
I like how it has shaped in the end. I will take a glance at porting A702 Mesa patches later this week.

ricardosalveti (approved):

PR #758, opened 2025-02-18
ci: base.yaml: add ssh-server-openssh

ldts:
No authentication will be required when connecting as the root user.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -40,7 +40,7 @@ local_conf_header:
     IMAGE_FSTYPES += "qcomflash"
   extra: |
     DISTRO_FEATURES:append = " efi"
-    EXTRA_IMAGE_FEATURES = "allow-empty-password empty-root-password allow-root-login"
+    EXTRA_IMAGE_FEATURES = "allow-empty-password empty-root-password allow-root-login ssh-server-openssh"
     IMAGE_ROOTFS_EXTRA_SPACE = "307200"
 
 machine: unset


ricardosalveti:
I would prefer integrating a proper user, removing root login and forcing user passwd to be changed on first boot as part of the meta-qcom-distro files instead, as this can quite easily can cause problems at someones network.

If needed locally you can create a custom yml file and pass that over to kas.

ldts:
> I would prefer integrating a proper user, removing root login and forcing user passwd to be changed on first boot as part of the meta-qcom-distro files instead, as this can quite easily can cause problems at someones network.

I don't have the full context—was the recent commit https://github.com/qualcomm-linux/meta-qcom/commit/bdb293260d5ccae850364d1f38aa77509940b61f only temporary? I had assumed it was setting a direction.

So we want to drop these extras before enabling ssh? I remember mentioning this to @ndechesne when I first booted the board.

ndechesne:
hmm. I didn't know we wanted to build such images, with proper user, and password policy.. I also think we need 'developers' images. so perhaps we can indeed create a devel.yml config file to include devel features.

ricardosalveti:
> I don't have the full context—was the recent commit [bdb2932](https://github.com/qualcomm-linux/meta-qcom/commit/bdb293260d5ccae850364d1f38aa77509940b61f) only temporary? I had assumed it was setting a direction.
> 
> So we want to drop password less before enabling ssh? I remember mentioning this extras (no password) to @ndechesne when I first booted the board.
> 
> Makes sense to me - just not aware of the sequence in which this should be delivered (I assume more features need to be enabled and verified first and at some point later users will be enabled? - ie I just did some simple venus encondig/decoding verification for one of the suported codecs)

Right, the concern is because of ssh / network.

We can define a devel.yml or similar with an extended set of configs, or just extend meta-qcom-distro instead, I just want to make sure that the default image is not enabling ssh with root and without passwd :-)

koenkooi:
> > I don't have the full context—was the recent commit [bdb2932](https://github.com/qualcomm-linux/meta-qcom/commit/bdb293260d5ccae850364d1f38aa77509940b61f) only temporary? I had assumed it was setting a direction.
> > So we want to drop password less before enabling ssh? I remember mentioning this extras (no password) to @ndechesne when I first booted the board.
> > Makes sense to me - just not aware of the sequence in which this should be delivered (I assume more features need to be enabled and verified first and at some point later users will be enabled? - ie I just did some simple venus encondig/decoding verification for one of the suported codecs)
> 
> Right, the concern is because of ssh / network.
> 
> We can define a devel.yml or similar with an extended set of configs, or just extend meta-qcom-distro instead, I just want to make sure that the default image is not enabling ssh with root and without passwd :-)

What's stopping us from adding a custom image that does 'require core-image-base.bb' and adds the things we need?

ricardosalveti:
> What's stopping us from adding a custom image that does 'require core-image-base.bb' and adds the things we need?

My preference would be to have image related recipes in the distro layer instead.

ldts:
Many of us working on various projects and architectures appreciate the simplicity of development when testing and implementing complex OS/firmware features—cloning a meta, building a minimal image, mounting, and working.

As a contributor and evaluator, I’d find it useful to have a devel.yaml (which is what I thought we had!) in meta-qcom that provides full control while saving both time and disk space. Of course, this assumes that developers understand the associated benefits and risks—though we could explicitly state this if we want to be extra cautious.

Given that there's already a layer and a recipe allowing password-less root access via the console, it seems like a natural candidate for this. In that sense, the concerns raised in this PR seem unrelated.

So, to bring the focus back—can we just make an __explicit__ devel.yaml in meta-qcom?

ricardosalveti:
> So, to bring the focus back—can we just make an **explicit** devel.yaml in meta-qcom?

Sure, I think we all gave +1 to it.

PR #761, opened 2025-02-19
RB3 Gen2: switch to linux-firmware

lumag:
Switch RB3 Gen2 support to the firmware from linux-firmware.


ricardosalveti (approved):

ricardosalveti:
> For the stable branches it would be possible to reinstantiate the `firmware-com-rb3gen2` recipe and make the distro use `PREFERRED_PROVIDER` or `PREFERRED_VERSION` in order to select the correct package.

Even yocto stable releases quite often brings updates to linux-firmware, so backporting in oe-core could be an option.

ricardosalveti (approved):

ndechesne (approved):

PR #756, opened 2025-02-18
RFC: Populate MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS with packagegroups

koenkooi:
To have more hardware features working on the supported machines out of the box we should add, at the very least, add drivers and firmware to` MACHINE_ESSENTIAL_*.`

This PR adds _fastrpc, msm, qcom-stats, rtc-pm8xxx and venus-enc/dec_ kernel modules to qcom-common.inc. This should be a big improvement over the status quo, but should be evaluated each time a machine is added or removed.

The commit to the RB3 machine adds virtually all kernel modules that were autoloaded, minus coresight.

PR #764, opened 2025-02-20
Rework MACHINE_FEATURES

koenkooi:
Pulled out from #756, this PR puts all common `MACHINE_FEATURES` in qcom-common.inc and adds `pci` to RB3.

### DIFF
--- a/conf/machine/include/qcom-common.inc
+++ b/conf/machine/include/qcom-common.inc
@@ -1,6 +1,9 @@
 SOC_FAMILY:prepend = "qcom:"
 require conf/machine/include/soc-family.inc
 
+# Set a default set of MACHINE_FEATURES, indivudual MACHINEs can append it
+MACHINE_FEATURES = "alsa bluetooth usbgadget ushbost wifi"
+
 XSERVER_OPENGL ?= " \
     xf86-video-modesetting \
     xserver-xorg-extension-glx \
--- a/conf/machine/qcm6490-idp.conf
+++ b/conf/machine/qcm6490-idp.conf
@@ -4,7 +4,7 @@
 
 require conf/machine/include/qcom-qcs6490.inc
 
-MACHINE_FEATURES = "efi usbhost usbgadget alsa wifi bluetooth"
+MACHINE_FEATURES += "efi pci"
 
 KERNEL_DEVICETREE ?= " \
                       qcom/qcm6490-idp.dtb \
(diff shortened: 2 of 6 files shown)


koenkooi:
I strongly suspect we can move more features to qcom-common.inc, like `alsa`, but that would introduce a functional change.

conf/machine/qcom-armv7a-modem.conf
```diff
@@ -6,8 +6,6 @@ require conf/machine/include/qcom-common.inc
 # MDM9615 is Cortex-A5 + VFP4, so it should be compatible
 require conf/machine/include/arm/armv7a/tune-cortexa7.inc
 
-MACHINE_FEATURES = "usbhost usbgadget"
-
```
lumag, line 10: I think -modem is pretty unique kind of machines, it might be better to have a more common data in `qcom-common.inc` and provide an override here.

ricardosalveti:
As you are adding pci to rb3gen2 here, mind also adding at qcm6490-idp.conf and qcs9100-ride-sx.conf?

koenkooi:
> As you are adding pci to rb3gen2 here, mind also adding at qcm6490-idp.conf and qcs9100-ride-sx.conf?

Done. Should this move into the qcom-<soc>.inc or stay in the MACHINE.conf?

ricardosalveti:
> Done. Should this move into the qcom-.inc or stay in the MACHINE.conf?

I think it is fine to have it at board level for now.

ricardosalveti (approved):

lumag (approved):

lumag:
I've opened #765 , hopefully it can help with the config unification.

ricardosalveti:
@koenkooi please rebase

koenkooi:
> @koenkooi please rebase

Done

conf/machine/include/qcom-common.inc
```diff
@@ -1,6 +1,9 @@
 SOC_FAMILY:prepend = "qcom:"
 require conf/machine/include/soc-family.inc
 
+# Set a default set of MACHINE_FEATURES, indivudual MACHINEs can append it
+MACHINE_FEATURES = "usbhost usbgadget"
```
lumag, line 5: Don't we have now more of the common features? I think "alsa", "screen", "wifi" and "bluetooth" also should be added to the list (if any of the remaining platforms doesn't have it, it is an omission).

koenkooi: Qcs-9100-ride-sx doesn't list bluetooth, but the specs for the board state that both wifi and bluetooth are an add-on board. So only having one of the 2 listed is wrong either way :) I'll add both to the .inc.

Issue #713, opened 2024-11-15
Please move bootloaders to a subdir


lumag:
Having bootloader binaries just under `${DEPLOYDIR}` unnecessary pollutes the directory and disallows multi-machine configurations. Please move bootloader binaries under board-specific dir under `${DEPLOYDIR}`.

PR #766, opened 2025-02-21
recipes-bsp: libdmabufheap: Add recipe to compile libdmabufheap

quic-vjitta:
Add recipe to compile libdmabuf which provices apis to access dmabuf heap framework.

### DIFF
--- a/recipes-bsp/packagegroups/packagegroup-qcom.bb
+++ b/recipes-bsp/packagegroups/packagegroup-qcom.bb
@@ -24,4 +24,5 @@ RDEPENDS:${PN}-miscellaneous = " \
     hexagonrpc \
     ${@bb.utils.contains("BBLAYERS", "openembedded-layer", "libssc","", d)} \
     libvmmem-dev \
+    libdmabufheap-dev \
 "
--- a/recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
+++ b/recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
@@ -1,0 +1,16 @@
+SUMMARY = "Qualcomm Libdmabufheap library"
+HOMEPAGE = "https://git.codelinaro.org/clo/le/platform/system/memory/libdmabufheap"
+DESCRIPTION = "Libdmabufheap library to access DMA-BUF heap framework with some custom changes \
+removing dependency on libion and libbase"
+
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://BufferAllocator.cpp;beginline=1;endline=15;md5=f0d8beef101110303cf3728fcf3e73a9"
+
+PV = "0.0+git"
+SRCREV = "5ead341efb667cbd2cf519bfaf2837ee8e97b747"
+SRC_URI = " \
+git://git.codelinaro.org//clo/le/platform/system/memory/libdmabufheap.git;branch=memory-le-apps.lnx.3.0;protocol=https \
+"
+S = "${WORKDIR}/git"
+
+inherit autotools pkgconfig


recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
+SUMMARY = "Qualcomm Libdmabufheap"
+HOMEPAGE = "https://git.codelinaro.org/clo/le/platform/system/memory/libdmabufheap"
+
+DESCRIPTION = "Qualcomm Libdmabufheap to use DMABUF heap framework"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/${LICENSE};md5=89aea4e17d99a7cacdbeed46a0096b10"
```
lumag, line 6: Please reference a file from the lib sources rather than the common licence file. The goal of this variable is to catch the licence changes. If you reference a common licence file, you can not catch such issues.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+
+DEPENDS += "linux-kernel-qcom-headers"
```
lumag, line 8: This won't build. Please import necessary headers into the source tree. If you really can not change original source, please import headers via SRC_URI. But depending on the non-uAPI headers is a NAK. Let me quote [kernel docs](https://docs.kernel.org/kbuild/headers_install.html):

> These kernel header files are used by the system’s C library (such as glibc or uClibc) to define available system calls, as well as constants and structures to be used with these system calls.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+
+SRC_URI   = " \
+             git://git.codelinaro.org//clo/le/platform/system/memory/libdmabufheap.git;branch=memory-le-apps.lnx.3.0;protocol=https \
```
lumag, line 11: Missing SRCREV.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+             "
+EXTRA_OECONF:append = " \
+    --disable-static \
+    --with-sanitized-headers=${STAGING_INCDIR}/linux/usr/include \
```
lumag, line 15: Should be dropped.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+"
+
+inherit autotools-brokensep pkgconfig
```
lumag, line 18: Why does it use `-brokenstep`? Qualcomm controls original source files. Can you fix it instead?

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+
+PACKAGES +="${PN}-test-bin"
+
+PACKAGECONFIG[ion] = "--with-ion, --without-ion, libion"
+
+FILES:${PN}     = "${libdir}/pkgconfig/* ${libdir}/* ${sysconfdir}/*"
```
lumag, line 24: - Drop extra whitespaces
- `${pkgconfig}` and `${libdir}/*` files should go to correspnding packages (mostly -dev). Don't pull them into `${PN}`.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+FILES:${PN}-test-bin = "${base_bindir}/*"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
```
lumag, line 27: why?

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+
```
lumag, line 28: Drop extra empty line

lumag (changes requested):

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
ricardosalveti: You can use something similar to:

`LIC_FILES_CHKSUM = "file://stack_chk.c;beginline=1;endline=30;md5=97e59d9deee678a9332c9ddb2ab6360d"`

`BufferAllocator.cpp` would be a good sample to use, since there is no LICENSE file there.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
ricardosalveti: We also need to find a way to extend our CI coverage to build such recipe as well, as it won't even build as is.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+DESCRIPTION = "Qualcomm Libdmabufheap to use DMABUF heap framework"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/${LICENSE};md5=89aea4e17d99a7cacdbeed46a0096b10"
+
+DEPENDS += "linux-kernel-qcom-headers"
+
+SRC_URI   = " \
+             git://git.codelinaro.org//clo/le/platform/system/memory/libdmabufheap.git;branch=memory-le-apps.lnx.3.0;protocol=https \
+             "
+EXTRA_OECONF:append = " \
```
ricardosalveti, line 13: Should either set directly or use += instead, append makes harder to override.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+    --disable-static \
```
lumag, line 14: BTW: what is the reason for --disable-static?

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+    --with-sanitized-headers=${STAGING_INCDIR}/linux/usr/include \
+"
+
+inherit autotools-brokensep pkgconfig
+
+PACKAGES +="${PN}-test-bin"
```
lumag, line 20: Use PACKAGES_BEFORE_PN, then you can drop FILES assignments.

lumag (commented):
What is the reason for the recipe being named `qcom-libdmabufheap` instead of just `libdmabufheap` ?

lumag:
There doesn't seem to be a versioned release tag. Please rename the recipe to foo_git.bb and set `PV = "0.0+git${SRCPV}"`

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
```diff
@@ -0,0 +1,28 @@
...
+
+PACKAGECONFIG[ion] = "--with-ion, --without-ion, libion"
+
+FILES:${PN}     = "${libdir}/pkgconfig/* ${libdir}/* ${sysconfdir}/*"
+FILES:${PN}-test-bin = "${base_bindir}/*"
```
lumag, line 25: There are no binaries, so it is not even required.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
lumag: I'm not sure about bb internals, but something like this might work:
```
bitbake `find ../layers/meta-qcom -name "*.bb" -o -name "*.bbappend" | sed -e 's!.*/!!g' -e 's!\(_.*\)\?\.bb\(append\)\?!!g' | grep -v 'qcom-partition-confs\|firmware-qcom-boot\|qcom-gen-partition-bins\|linux-yocto\|linux-linaro-qcomlt'`
```
Fixing #713 would allow us to drop qcom-* and firmware-qcom-boot from the list of exclusions.

lumag: ... and u-boot fails to fetch. I pinged Caleb for a stable commit.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
quic-vjitta: Thanks, updated this to use BufferAllocator.cpp.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
quic-vjitta: Added

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
quic-vjitta: Done

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
quic-vjitta: updated to use autotools

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
quic-vjitta: done

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
quic-vjitta: removed

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
quic-vjitta: Done

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
quic-vjitta: removed this.

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
quic-vjitta: Looks like its not need , removed

recipes-bsp/libdmabufheap/qcom-libdmabufheap_1.0.bb
quic-vjitta: removed

quic-vjitta:
> There doesn't seem to be a versioned release tag. Please rename the recipe to foo_git.bb and set `PV = "0.0+git"`
Added it.

quic-vjitta:
> What is the reason for the recipe being named `qcom-libdmabufheap` instead of just `libdmabufheap` ?

This source code has some changes from qualcomm and is diverged from android open source libdmabufheap code, So it's being named with qcom-libdmabufheap.

lumag (changes requested):
Other than the unnecessary `FILES`, what is the reason for the recipe being named qcom-libdmabufheap instead of just libdmabufheap ?

koenkooi:
> Other than the unnecessary `FILES`, what is the reason for the recipe being named qcom-libdmabufheap instead of just libdmabufheap ?

And if it's a variant of an existing upstream project, there need to the appropriate (R)CONFLICTS /(R)REPLACES/(R)PROVIDES stanzas to alert the parser to the conflicts.

quic-vjitta:
> Also please drop the gerrit tag from the commit message

done

recipes-bsp/libdmabufheap/qcom-libdmabufheap_git.bb
```diff
@@ -0,0 +1,22 @@
...
+SRC_URI = " \
+    git://git.codelinaro.org//clo/le/platform/system/memory/libdmabufheap.git;branch=memory-le-apps.lnx.3.0;protocol=https \
+"
+
+inherit autotools pkgconfig
+
+PACKAGECONFIG[ion] = "--with-ion, --without-ion, libion"
+
+FILES:${PN} = "${sysconfdir}/*"
+FILES:${PN}-dev = "${includedir}/* ${libdir}/pkgconfig/* ${libdir}/*"
```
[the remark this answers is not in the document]
quic-vjitta: removed them

lumag (changes requested):
- Drop the `qcom-` prefix from the recipe name, everything in this layer is qcom-related.
- Make it tested in the CI

lumag:
> > What is the reason for the recipe being named `qcom-libdmabufheap` instead of just `libdmabufheap` ?
> 
> This source code has some changes from qualcomm and is diverged from android open source libdmabufheap code, So it's being named with qcom-libdmabufheap.

In such a case, it might be better package libdmabufheap _and_ add your patches on top of it. I strongly dislike the idea of qcom-specific forks of the existing projects.

lumag:
Okay. After looking at the AOSP source... Don't package it. But the commit message should document that it's a fork of the AOSP library. Inside the recipe please add a brief comment describing Qualcomm changes, so that a random user browsing the repo tree would understand the difference and the reason. Please follow @koenkooi's suggestion and add 'PROVIDES' notes.

lumag:
Also, while I noticed it. Neither libdmabufheap nor libvmmem are `BSP`-related packages. Please move them to a more sensible location. Maybe `recipes-support`.

lumag:
I think it should be fine to skip `recipes-foo`. Most of the layers don't require that in the commit message.

lumag:
@quic-vjitta @quic-vkraleti @sbanerjee-quic gracious ping. This PR wasn't updated for last three weeks. Should it be updated and merged or rather we should close it for now?

quic-vjitta:
> * Drop the `qcom-` prefix from the recipe name, everything in this layer is qcom-related.
> * Make it tested in the CI

Removed qcom prefix and have added it to package group.

quic-vjitta:
> @quic-vjitta @quic-vkraleti @sbanerjee-quic gracious ping. This PR wasn't updated for last three weeks. Should it be updated and merged or rather we should close it for now?

sorry for the delay , have updated the change now.

quic-vjitta:
> nit:
> 
> `recipes-bsp: libdmabufheap: Add recipe to build libdmabufheap`
> 
> Please change to just `recipes-bsp: libdmabufheap: add recipe`, since the main purpose of this commit is to add a new recipe into this repo.
> 
> Also agree it should be moved out of recipes-bsp.

Done , moved it to recipes-support

quic-vjitta:
> Also, while I noticed it. Neither libdmabufheap nor libvmmem are `BSP`-related packages. Please move them to a more sensible location. Maybe `recipes-support`.

done

quic-vjitta:
> Okay. After looking at the AOSP source... Don't package it. But the commit message should document that it's a fork of the AOSP library. Inside the recipe please add a brief comment describing Qualcomm changes, so that a random user browsing the repo tree would understand the difference and the reason. Please follow @koenkooi's suggestion and add 'PROVIDES' notes.

Have added Provides/conflicts/replaces as both the libraray names are the same.

recipes-support/libdmabufheap/libdmabufheap_git.bb
```diff
@@ -0,0 +1,22 @@
...
+
+SRC_URI = " \
+    git://git.codelinaro.org//clo/le/platform/system/memory/libdmabufheap.git;branch=memory-le-apps.lnx.3.0;protocol=https \
+    "
+
+inherit autotools pkgconfig
+
+RPROVIDES_${PN} = "libdmabufheap"
+RREPLACES_${PN} = "libdmabufheap"
+RCONFLICTS_${PN} = "libdmabufheap"
```
lumag, line 22: Why?

recipes-support/libdmabufheap/libdmabufheap_git.bb
```diff
@@ -0,0 +1,22 @@
+SUMMARY = "Qualcomm Libdmabufheap library"
```
lumag, line 1: commit message: AOSP, API

recipes-bsp/packagegroups/packagegroup-qcom.bb
```diff
@@ -5,6 +5,7 @@ inherit packagegroup
 PACKAGES = " \
     ${PN}-boot-essential \
     ${PN}-boot-additional \
+    ${PN}-qcom-miscellaneous \
```
lumag, line 8: This ends up being `packagegroup-qcom-qcom-miscellaneous`. Why do you need a double `qcom`?

lumag (changes requested):
s

recipes-support/libdmabufheap/libdmabufheap_git.bb
quic-vjitta: As aosp library also has the same name , so added above fields.

recipes-bsp/packagegroups/packagegroup-qcom.bb
quic-vjitta: missed it , will update it.

recipes-support/libdmabufheap/libdmabufheap_git.bb
lumag: So, you end up with the package `libdmabufheap` RREPLACES `libdmabufheap`. What is the point?

lumag: You have dropped `RREPLACES`. Do you actually need my comment in order to drop `RPROVIDES_foo = "foo"` and `RCONFLICTS_foo = "foo"`. What is the purpose of those two clauses?

recipes-support/libdmabufheap/libdmabufheap_git.bb
```diff
@@ -0,0 +1,21 @@
...
+
+DESCRIPTION = "Builds Libdmabufheap library to access DMA-BUF heap framework with some custom changes \
+               removing dependency on libion and libbase"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://BufferAllocator.cpp;beginline=1;endline=15;md5=f0d8beef101110303cf3728fcf3e73a9"
+
+SRCREV = "5ead341efb667cbd2cf519bfaf2837ee8e97b747"
+S = "${WORKDIR}/git"
+
+PV = "0.0+git"
```
sbanerjee-quic, line 12: @quic-vjitta 
In configure.ac, why 0.0 here? 
`AC_INIT([libdmabufheap],1.0.0)`

It is better to name the recipe as libdmabufheap_<version>.bb

quic-vjitta:
> It really doesn't look as this was properly tested:
> 
> ```
> ERROR: packagegroup-qcom-1.0-r0 do_package_write_ipk: An allarch packagegroup shouldn't depend on packages which are dynamically renamed (libdmabufheap to libdmabufheap0)
> ```

resolved this error.

recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
```diff
@@ -0,0 +1,21 @@
+SUMMARY = "Qualcomm Libdmabufheap library"
+HOMEPAGE = "https://git.codelinaro.org/clo/le/platform/system/memory/libdmabufheap"
+
+DESCRIPTION = "Builds Libdmabufheap library to access DMA-BUF heap framework with some custom changes \
```
lumag, line 4: "Builds foo bar" is not a package description.

recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
```diff
@@ -0,0 +1,21 @@
...
+               removing dependency on libion and libbase"
```
lumag, line 5: Drop extra indentation, we don't want it to sneak into the generated packages.

recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
```diff
@@ -0,0 +1,21 @@
...
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://BufferAllocator.cpp;beginline=1;endline=15;md5=f0d8beef101110303cf3728fcf3e73a9"
+
+SRCREV = "5ead341efb667cbd2cf519bfaf2837ee8e97b747"
+S = "${WORKDIR}/git"
+
+PV = "1.0.0+git"
```
lumag, line 12: Is this a real version? Was there a tag with `1.0.0`? Will there be a `1.1` or `1.0.1` at some point?

lumag (changes requested):

recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
ricardosalveti: Yeah, there is no meaningful tag there (hundreds of tags but nothing specific to the project itself), no 1.0.0 (version seems to be just the default used when creating it - https://git.codelinaro.org/clo/le/platform/system/memory/libdmabufheap/-/commit/904441744b5c22e991ee4d722dbe9ac67139be63).

ricardosalveti:
> > Okay. After looking at the AOSP source... Don't package it. But the commit message should document that it's a fork of the AOSP library. Inside the recipe please add a brief comment describing Qualcomm changes, so that a random user browsing the repo tree would understand the difference and the reason. Please follow @koenkooi's suggestion and add 'PROVIDES' notes.
> 
> Have added Provides/conflicts/replaces as both the libraray names are the same.

We should probably avoid using the upstream name for this recipe as it is a fork, so we need to find the best mechanism to show that this is a fork that was done by qualcomm. If named differently (e.g. libdmabufheap-qcom_1.0.0.bb) then I agree we need a PROVIDES, since if someone creates a recipe for the pure upstream version, it would show as an option.

Another way would be to make it qcom version specific, like libdmabufheap_1.0.0.qcom.bb (as commonly done in some BSPs), and have a preferred version to it, but this is mostly used when there is already a recipe from the same project in oe-core or meta-oe.

quic-vjitta:
> > > Okay. After looking at the AOSP source... Don't package it. But the commit message should document that it's a fork of the AOSP library. Inside the recipe please add a brief comment describing Qualcomm changes, so that a random user browsing the repo tree would understand the difference and the reason. Please follow @koenkooi's suggestion and add 'PROVIDES' notes.
> > 
> > 
> > Have added Provides/conflicts/replaces as both the libraray names are the same.
> 
> We should probably avoid using the upstream name for this recipe as it is a fork, so we need to find the best mechanism to show that this is a fork that was done by qualcomm. If named differently (e.g. libdmabufheap-qcom_1.0.0.bb) then I agree we need a PROVIDES, since if someone creates a recipe for the pure upstream version, it would show as an option.
> 
> Another way would be to make it qcom version specific, like libdmabufheap_1.0.0.qcom.bb (as commonly done in some BSPs), and have a preferred version to it, but this is mostly used when there is already a recipe from the same project in oe-core or meta-oe.

My initial change was naming the recipe as qcom-libdmabufheap , but suggestion from @lumag was to remove qcom from name as library itself is named as libdmabufheap. So, i renamed it.

So , is it better to just remove PROVIDES/REPLACES and keep name as is as I am already mentioning in recipe description that it has custom changes ?  or is it better to rename ? Lets me know your suggestion

lumag:
> > > Okay. After looking at the AOSP source... Don't package it. But the commit message should document that it's a fork of the AOSP library. Inside the recipe please add a brief comment describing Qualcomm changes, so that a random user browsing the repo tree would understand the difference and the reason. Please follow @koenkooi's suggestion and add 'PROVIDES' notes.
> > 
> > 
> > Have added Provides/conflicts/replaces as both the libraray names are the same.
> 
> We should probably avoid using the upstream name for this recipe as it is a fork, so we need to find the best mechanism to show that this is a fork that was done by qualcomm. If named differently (e.g. libdmabufheap-qcom_1.0.0.bb) then I agree we need a PROVIDES, since if someone creates a recipe for the pure upstream version, it would show as an option.

There is no 'upstream' per se. It is a library inside AOSP which is unusable without fixes similar to those implemented in this fork. Probably the best course of action would be to work with Google / AOSP people to _become_ an upstream for this component.

> 
> Another way would be to make it qcom version specific, like libdmabufheap_1.0.0.qcom.bb (as commonly done in some BSPs), and have a preferred version to it, but this is mostly used when there is already a recipe from the same project in oe-core or meta-oe.

recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
lumag: I'd then suggest reverting back to 0.0+git (and maybe changing the library's build script to stop confusing people).

ricardosalveti:
> > > > Okay. After looking at the AOSP source... Don't package it. But the commit message should document that it's a fork of the AOSP library. Inside the recipe please add a brief comment describing Qualcomm changes, so that a random user browsing the repo tree would understand the difference and the reason. Please follow @koenkooi's suggestion and add 'PROVIDES' notes.
> > > 
> > > 
> > > Have added Provides/conflicts/replaces as both the libraray names are the same.
> > 
> > 
> > We should probably avoid using the upstream name for this recipe as it is a fork, so we need to find the best mechanism to show that this is a fork that was done by qualcomm. If named differently (e.g. libdmabufheap-qcom_1.0.0.bb) then I agree we need a PROVIDES, since if someone creates a recipe for the pure upstream version, it would show as an option.
> 
> There is no 'upstream' per se. It is a library inside AOSP which is unusable without fixes similar to those implemented in this fork. Probably the best course of action would be to work with Google / AOSP people to _become_ an upstream for this component.

Sure, if there is no separated 'upstream' I'm fine having it named libdmabufheap_1.0.0.bb then.

lumag:
@quic-vjitta please rebase.

quic-vjitta:
> @quic-vjitta please rebase.
Have rebased it.

lumag (dismissed):

quic-vjitta:
> Please rebase it properly, there is now a merge commit here.

done.

lumag (dismissed):

recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
```diff
@@ -0,0 +1,18 @@
+SUMMARY = "Qualcomm Libdmabufheap library"
+HOMEPAGE = "https://git.codelinaro.org/clo/le/platform/system/memory/libdmabufheap"
+
+DESCRIPTION = "Libdmabufheap library to access DMA-BUF heap framework with some custom changes \
+removing dependency on libion and libbase"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://BufferAllocator.cpp;beginline=1;endline=15;md5=f0d8beef101110303cf3728fcf3e73a9"
+
+SRCREV = "5ead341efb667cbd2cf519bfaf2837ee8e97b747"
```
koenkooi: Nitpick: if you're using whitespace to separate out variables, please keep related variables together. In this case:

- SUMMARY, HOMEPAGE and DESCRIPTION belong together
- LICENSE and LINC_FILES_CHKSUM belong togethers
- PV, SRCREV, SRC_URI and S belong together, in that specific order.

lumag: Is oe-stylize.py still recommended to be used or is it outdated?

recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
```diff
@@ -0,0 +1,16 @@
...
+DESCRIPTION = "Libdmabufheap library to access DMA-BUF heap framework with some custom changes \
+removing dependency on libion and libbase"
+
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://BufferAllocator.cpp;beginline=1;endline=15;md5=f0d8beef101110303cf3728fcf3e73a9"
+
+PV = "0.0+git"
+SRCREV = "5ead341efb667cbd2cf519bfaf2837ee8e97b747"
+SRC_URI = " \
+git://git.codelinaro.org//clo/le/platform/system/memory/libdmabufheap.git;branch=memory-le-apps.lnx.3.0;protocol=https \
```
sbanerjeeqc, line 12: @quic-vjitta, @shashim-quic inline with mainline development principles should this repo be moved to qualcomm-linux github.

lumag: Why? It it not being developed as a part of Qualcomm Linux.

lumag (approved):

recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
ricardosalveti: If moved we can update the recipe once done.

ricardosalveti (approved):

recipes-support/libdmabufheap/libdmabufheap_1.0.0.bb
koenkooi: I haven't used it in a while, but it should catch the things I mentioned.

PR #774, opened 2025-02-22
Disable seccomp for dhcpcd

bpothuno-quic:
SECCOMP is enabled by default, causing SIGSYS errors for dhcpcd wlan0.  Due to the dependency of Linux SECCOMP on the libc and kernel versions, privilege separation is being disabled entirely to accommodate changes in libc/kernel compatibility.

### DIFF
--- a/recipes-support/dhcpcd/dhcpcd_10.1.0.bbappend
+++ b/recipes-support/dhcpcd/dhcpcd_10.1.0.bbappend
@@ -1,0 +1,3 @@
+# SECCOMP is very dependant on libc vs kernel,
+# Disable seccomp to prevent dhcpcd break
+EXTRA_OECONF:append = " --disable-seccomp"


recipes-support/dhcpcd/dhcpcd_10.1.0.bbappend
```diff
@@ -0,0 +1,3 @@
+# SECCOMP is very dependant on libc vs kernel,
+# Disable seccomp to prevent dhcpcd break
+EXTRA_OECONF:append = " --disable-seccomp"
```
quic-vkraleti, line 3: Use ‘EXTRA_OECONF:append:qcom’ to ensure this change doesn’t break other vendors’ dhcpcd when meta-qcom layer sits alongside.

recipes-support/dhcpcd/dhcpcd_10.1.0.bbappend
```diff
@@ -0,0 +1,3 @@
+# SECCOMP is very dependant on libc vs kernel,
+# Disable seccomp to prevent dhcpcd break
```
quic-vkraleti, line 2: Is it possible to mention why enabling SECCOMP is causing SIGSYS errors on QCOM platforms and how disabling this config helping to address those?

lumag:
What kind of incompatibility is it? Is it an upstream bug? Is it reported anywhere?

recipes-support/dhcpcd/dhcpcd_10.1.0.bbappend
bpothuno-quic: DHCP module restricted system call via SECCOMP method.
DHCPD service run as part of boot-up and as soon as wifi is connected, dhcpcd receiving __NR_sysfs call which is invalid to this module as per seccomp filter, because of this SIGSYS signal is received and it got terminated.
So when SECCOMP disabled then dhcpd don't have any restrcition to system call, Hence its working as expected.

bpothuno-quic:
> Please describe what the actual issue is, since this is unexpected and we shouldn't change the behavior of the upstream project unless we can justify it.

DHCP module restricted system call via SECCOMP  method.
DHCPD service run as part of boot-up and as soon as wifi is connected, dhcpcd receiving __NR_sysfs call which is invalid to this module as per seccomp filter, because of this SIGSYS signal is received and it got terminated.
So when SECCOMP disabled then dhcpd don't have any restrcition to system call, Hence its working as expected.`

bpothuno-quic:
> What kind of incompatibility is it? Is it an upstream bug? Is it reported anywhere?

As per the following commit: https://github.com/NetworkConfiguration/dhcpcd/commit/8b65e2d6c7da1688df00753f872d8805631e07cb  SECCOM is dependent on libc or kernel where dhcpcd may face imcompatibility and Introduced option to disable using "disable-seccomp",

As we faced following behavior hence disabled:

DHCP module restricted system call via SECCOMP method.
DHCPD service run as part of boot-up and as soon as wifi is connected, dhcpcd receiving __NR_sysfs call which is invalid to this module as per seccomp filter, because of this SIGSYS signal is received and it got terminated.
So when SECCOMP disabled then dhcpd don't have any restrcition to system call, Hence its working as expected.`

lumag:
I don't see this issue being reported (maybe I missed it, I am using my phone). Please report it to the upstream so that we know when we can disable the workaround. Also be sure to include a comment with the link to the upstream bug report.

lumag:
Okay. Now I wonder, what kind of analysys you've performed. git grep __NR_sysfs points only the ARC architecture. Your log message points out the `syscall=135`. On arm64 this is `__NR_rt_sigprocmask. This systcall is pretty old and should be supported by both libc and kernel. Quick glance through the issue tracker points out relevant fix that must be pushed to the upstream repo, https://github.com/NetworkConfiguration/dhcpcd/commit/e9e40400003db2e4f12dba85acabbaf2212a520f. Either that patch should be backported to OE-Core or the dhcpcd recipe should be upgraded to 10.2.0.

bpothuno-quic:
> __NR_rt_sigprocmask

Thanks for suggestion, following change is working as expected : https://github.com/NetworkConfiguration/dhcpcd/commit/e9e40400003db2e4f12dba85acabbaf2212a520f

lumag:
@bpothuno-quic I don't see a patch being sent to OE-Core. This issue will not solve itself.

PR #776, opened 2025-02-24
CI base.yml: add `wifi` to `DISTRO_FEATURES`

koenkooi:
The magic in recipes-bsp/packagegroups demands 'wifi' to be present before it includes the wifi firmware.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -39,7 +39,7 @@ local_conf_header:
     IMAGE_CLASSES += "image_types_qcom"
     IMAGE_FSTYPES += "qcomflash"
   extra: |
-    DISTRO_FEATURES:append = " efi"
+    DISTRO_FEATURES:append = " efi wifi"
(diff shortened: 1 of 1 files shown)


ci/base.yml
```diff
@@ -39,7 +39,7 @@ local_conf_header:
     IMAGE_CLASSES += "image_types_qcom"
     IMAGE_FSTYPES += "qcomflash"
   extra: |
-    DISTRO_FEATURES:append = " efi"
+    DISTRO_FEATURES:append = " efi wifi"
```
lumag, line 42: Isn't `wifi` a part of the `DISTRO_FEATURES_DEFAULT`?

ricardosalveti: Yeah, seems to be enabled by default because of that already.

koenkooi: Indeed it is, I was thrown off by it being in the middle of that variable :/

PR #779, opened 2025-02-25
qcs6490-rb3gen2-core-kit: add `packagegroup-rb3gen2-firmware` to `MACHINE_ESSENTIAL_RECOMMENDS`

koenkooi:
This machine config doesn't include qcom-armv8.inc, on purpose, so include the firmwares here manually.

### DIFF
--- a/conf/machine/qcs6490-rb3gen2-core-kit.conf
+++ b/conf/machine/qcs6490-rb3gen2-core-kit.conf
@@ -11,8 +11,5 @@ KERNEL_DEVICETREE ?= " \
                       "
 
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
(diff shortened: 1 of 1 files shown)


conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -15,4 +15,5 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     linux-firmware-qcom-qcm6490-adreno \
     linux-firmware-qcom-qcm6490-audio \
     linux-firmware-qcom-qcm6490-compute \
+    packagegroup-rb3gen2-firmware \
```
lumag, line 18: I think we should also drop linux-firmware-qcom-foo names from this variable, packagegroup does that for us.

PR #781, opened 2025-02-26
firmware-qcom-boot-qcs9100: use correct ride-sx v3 cdt

ricardosalveti:
Fix incorrect CDT for qcs9100-ride-sx, which is based on ride-sx v3 instead.


quic-vkraleti (approved):
LGTM

ndechesne (approved):

lumag:
What is the relationship between 'v3' in this name and the 'r3' in the devicetrees?

ricardosalveti:
v3 cdt causes the r3 device-tree to be loaded, in this case.

lumag:
Ack, thanks.

PR #785, opened 2025-03-03
Add bbappend for wpa-supplicant and hostapd

miaoqing-quic:
Switch the code source from w1.fi/releases to w1.fi/hostap.git so that we can specify the latest code version. Additionally, enable features such as 11AX, 11BE, 11R, ACS and OWE.

### DIFF
--- a/conf/machine/include/qcom-common.inc
+++ b/conf/machine/include/qcom-common.inc
@@ -2,7 +2,7 @@ SOC_FAMILY:prepend = "qcom:"
 require conf/machine/include/soc-family.inc
 
 # Set a default set of MACHINE_FEATURES, indivudual MACHINEs can append it
-MACHINE_FEATURES = "alsa bluetooth usbgadget ushbost wifi"
+MACHINE_FEATURES = "alsa bluetooth usbgadget usbhost wifi"
 
 XSERVER_OPENGL ?= " \
     xf86-video-modesetting \


koenkooi:
Generally we follow the OE commit message guidelines, so start with the recipe/class/conf and then say what it does and explain the how in the message body. So for this PR, the commit message should look like this:

> hostapd: switch from tarball releases to git, add custom qcom config
> 
> Switch the code source from w1.fi/releases to w1.fi/hostap.git so that
> we can specify the latest code version. Additionally, enable features
> such as 11AX, 11BE, 11R, ACS and OWE.
> 
> Signed-off-by: Miaoqing Pan <quic_miaoqing@quicinc.com>

and

> wpa-supplicant: switch from tarball releases to git, add custom qcom config
> 
> Switch the code source from w1.fi/releases to w1.fi/hostap.git so that
> we can specify the latest code version. Additionally, enable features
> such as 11BE and OWE.
> 
> Signed-off-by: Miaoqing Pan <quic_miaoqing@quicinc.com>

recipes-connectivity/hostapd/hostapd_%.bbappend
```diff
@@ -0,0 +1,23 @@
+LIC_FILES_CHKSUM:qcom = "file://hostapd/README;md5=0e430ef1be3d6eebf257cf493fc7661d"
+
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${BPN}:"
+
+SRC_URI:qcom = "git://w1.fi/hostap.git;protocol=git;branch=main \
+                file://defconfig.qcom \
+                "
+
+SRCREV:qcom = "ad8d30e0a3601804c7464f53e54b1f4087364fd2"
```
koenkooi, line 9: This needs a comment why this specific revision was selected. It also needs a PV="x.y+git", since the version is different from the tarball based recipe.

recipes-connectivity/hostapd/hostapd_%.bbappend
```diff
@@ -0,0 +1,23 @@
...
+
+S:qcom = "${UNPACKDIR}/git"
+
+SYSTEMD_SERVICE:${PN}:qcom = ""
```
koenkooi, line 13: This change isn't listed in the commit message and I'm fairly sure we shouldn't sabotage startup scripts like that. If we do, there needs to be a expansive comment in the recipe.

recipes-connectivity/hostapd/hostapd_%.bbappend
```diff
@@ -0,0 +1,23 @@
...
+
+do_configure:prepend:qcom() {
+    install -m 0644 ${UNPACKDIR}/defconfig.qcom ${UNPACKDIR}/defconfig
+}
+
+do_install:qcom() {
```
koenkooi, line 19: Same as above: this change isn't listed in the commit message and I'm fairly sure we shouldn't sabotage startup scripts like that. If we do, there needs to be a expansive comment in the recipe.

koenkooi (changes requested):

recipes-connectivity/wpa-supplicant/wpa-supplicant_%.bbappend
```diff
@@ -0,0 +1,19 @@
...
+
+SRCREV:qcom = "ad8d30e0a3601804c7464f53e54b1f4087364fd2"
+
+S:qcom = "${UNPACKDIR}/git"
+
+do_configure:prepend:qcom() {
+	install -m 0644 ${UNPACKDIR}/defconfig.qcom ${UNPACKDIR}/defconfig
+}
+
+do_install:qcom() {
```
koenkooi, line 17: Same comment as the other commit: This change isn't listed in the commit message and I'm fairly sure we shouldn't sabotage startup scripts like that. If we do, there needs to be a expansive comment in the recipe.

And this also removes the dbus RPC configs.

koenkooi (changes requested):

koenkooi:
Side note: hostapd is in meta-oe, which isn't enabled in our default build configs, so this will never get tested by the CI.

PR #786, opened 2025-03-03
ci/distro.yml: enable meta-xfce and its dependencies

koenkooi:
This enables meta-xfce and friends for DISTRO='qcom-distro'

### DIFF
--- a/ci/distro.yml
+++ b/ci/distro.yml
@@ -13,6 +13,9 @@ repos:
   meta-openembedded:
     url: https://github.com/openembedded/meta-openembedded
     layers:
-      meta-oe:
+      meta-gnome:
+      meta-multimedia: 
       meta-networking:
+      meta-oe:
       meta-python:
+      meta-xfce:


lumag:
What is the expected usecase? If you are building an image. wouid it make sense to add corresponding changes to your kas file as well?

koenkooi:
The use case is indeed building an xfce image, but the specific image(s) can be a separate yaml.

ndechesne:
It is getting 'complicated', no? Why are we not keeping everything in distro.yml? distro.yml is our reference distro config file, no? or at least include distro.yml from xfce.yml, and set the image name in there? 

our CI/yml scripts are supposed to define our reference builds, they are not generic hooks to add OE layers/features, no?

lumag:
> It is getting 'complicated', no? Why are we not keeping everything in distro.yml? distro.yml is our reference distro config file, no? or at least include distro.yml from xfce.yml, and set the image name in there?
> 
> our CI/yml scripts are supposed to define our reference builds, they are not generic hooks to add OE layers/features, no?

Another option might be to push 'useful, but not completely BSP nor distro' files to `meta-qcom-extras`.

koenkooi:
> It is getting 'complicated', no? Why are we not keeping everything in distro.yml? distro.yml is our reference distro config file, no? or at least include distro.yml from xfce.yml, and set the image name in there?
> 
> our CI/yml scripts are supposed to define our reference builds, they are not generic hooks to add OE layers/features, no?

I wasn't sure and wanted to avoid a combinatorial explosion :) We did set a precedent with the distro.yml that we merged recently.

AIUI KAS will de-dupe entries, so we could have xfce.yml include base.yml, that would shorten the commandline a bit.

ricardosalveti:
Shouldn't this image be one of the standard distro builds we want to make? I assume we should do a graphics distro variant and console-based one, then we could simplify the options.

lumag:
well... I see a benefit in having extra layers configuration without requiring `meta-qcom-distro` layer. It makes it easier to test the BSP layer separately from the distro. At the same time, I don't feel like adding those to `meta-qcom` itself. That's why I proposed adding all the 'extra' files to `meta-qcom-extras`.

koenkooi:
I'm fine putting the snippets in meta-qcom-distro as well. I have no strong opinion, just that I'd like to have a way of testing meta-qcom-distro against a minimal set of layers.

ricardosalveti:
> I'm fine putting the snippets in meta-qcom-distro as well. I have no strong opinion, just that I'd like to have a way of testing meta-qcom-distro against a minimal set of layers.

I think having our standard distro yml here makes CI and testing a bit easier, as we can build everything we need out of one single repository, but we can probably simplify things a bit.

I would imagine we want 2 types of distro images, one that is console based and another that is graphics based (which could be based out of xfce).

koenkooi:
So which changes are needed before this can get merged?

ndechesne:
thanks. this looks good. this is what I had in mind somehow. we still need to build the desktop images somehow, right?

ndechesne:
I mean, we should fix the build failures too :)

koenkooi:
> thanks. this looks good. this is what I had in mind somehow. we still need to build the desktop images somehow, right?

Yes, but that requires a few more PRs on the meta-qcom-distro side.

.github/workflows/build-yocto.yml
```diff
@@ -79,8 +79,13 @@ jobs:
           - qrb2210-rb1-core-kit
           - qcom-armv8a
           - qcom-armv7a
+        distro:
+          - name: poky/systemd
+            yamlfile: ""
```
ricardosalveti, line 84: Should we move some of the local_conf_header changes from base.yml to avoid being used with distro.yml as well?

We can do the clean up after as well if needed, but I was thinking that we should try to avoid changes to local.conf when using the distro, as those options should be part of meta-qcom-distro instead (e.g. systemd, distro features, etc).

ndechesne: Are you suggesting that we test with vanilla poky distro without any of our 'local_conf' changes? e.g. without systemd, efi, pni-names?

I agree that when using qcom-distro we should have 'less' noise in local_conf, but I am not sure we want to test a pure poky distro.

koenkooi: Ideally we'd test these options:

1) a fully working meta-qcom-distro DISTRO
2) nodistro
3) poky

And if we really need to:

4) poky modified to be actually working, which wouldn't be poky anymore.

We need 3) for getting the Yocto Compatible thing, no? And if we test meta-qcom-distro, what's the point of having poky-but-not-poky in the test matrix?

ndechesne: well, right now, we have none of the 1-3 list ;-) 

What does that mean to test just 'poky' or 'nodistro'. Can we still have that in local.conf:
    IMAGE_FSTYPES += "qcomflash"

 Poky/nodistro does not have 'efi' by default neither.

The delta between poky and nodistro for distro/packages is

# Override these in poky based distros                                                                                                                      POKY_DEFAULT_DISTRO_FEATURES = "opengl ptest multiarch wayland vulkan"
POKY_DEFAULT_EXTRA_RDEPENDS = "packagegroup-core-boot"
POKY_DEFAULT_EXTRA_RRECOMMENDS = "kernel-module-af-packet"

Also poky uses sysvinit. So we need to decide if we really want to support/test for sysvinit and systemd

ricardosalveti: > Are you suggesting that we test with vanilla poky distro without any of our 'local_conf' changes? e.g. without systemd, efi, pni-names?

No, just make sure the distro itself is not using the 'local_conf' changes, since it is not needed.

For meta-qcom poky we should still keep using 'local_conf' for now, for simplicity.

koenkooi: I'll split out the local conf changes from base.yml into its own yml and rework the others

lumag (approved):

ci/base.yml
```diff
@@ -28,18 +28,10 @@ repos:
       meta-poky:
 
 local_conf_header:
-  base: |
-    CONF_VERSION = "2"
-    INIT_MANAGER = "systemd"
-    PACKAGE_CLASSES = "package_ipk"
-    INHERIT += "buildstats buildstats-summary"
-    INHERIT += "buildhistory"
-    INHERIT += "rm_work"
```
quaresmajose, line 37: This will break the local builds because all the machines includes ci/base.yml and with that change we will move away from systemd init manager.

koenkooi: Which is exactly the point, we shouldn't be hiding fixups in the CI!

quaresmajose: My point is that we should have a simple way to replicate locally what happens in the CI, this will make life easier for us by replicating issue and fixing bug.
After this change if you do a local build using: `kas build ci/machine.yml` you will produce a completely different result than what is produced in the ci.

koenkooi: Yes, because you're now doing something different than what the CI does! so add ```:ci/distro.yml``` or ```:ci/localconfmod.yml``` and you'll get what the CI does.

koenkooi:
I've removed the localconfmod commit, that change is to invasive to be in this PR, I'll create a separate PR for it.

ci/base.yml
koenkooi: I'm moving this commit to a separate PR

ricardosalveti:
We can merge https://github.com/qualcomm-linux/meta-qcom/pull/894 and rebase this one to just add meta-xfce.

ricardosalveti:
Commit message is still saying DISTRO='qcom-wayland'.

ricardosalveti (approved):

lumag (approved):

PR #787, opened 2025-03-03
hostapd/wpa-supplicant: switch from tarball releases to git, add custom qcom config

miaoqing-quic:
Switch the code source from w1.fi/releases to w1.fi/hostap.git so that we can specify the latest code version. Additionally, enable features such as 11AX, 11BE, 11R, ACS and OWE.

### DIFF
--- a/recipes-connectivity/hostapd/hostapd/defconfig.qcom
+++ b/recipes-connectivity/hostapd/hostapd/defconfig.qcom
@@ -1,0 +1,437 @@
+# Example hostapd build time configuration
+#
+# This file lists the configuration options that are used when building the
+# hostapd binary. All lines starting with # are ignored. Configuration option
+# lines must be commented out complete, if they are not to be included, i.e.,
+# just setting VARIABLE=n is not disabling that variable.
+#
+# This file is included in Makefile, so variables like CFLAGS and LIBS can also
+# be modified from here. In most cass, these lines should use += in order not
+# to override previous values of the variables.
+
+# Driver interface for Host AP driver
+#CONFIG_DRIVER_HOSTAP=y
+
+# Driver interface for wired authenticator
+#CONFIG_DRIVER_WIRED=y
+
+# Driver interface for drivers using the nl80211 kernel interface
+CONFIG_DRIVER_NL80211=y
+
+# QCA vendor extensions to nl80211
+#CONFIG_DRIVER_NL80211_QCA=y
+
+# driver_nl80211.c requires libnl. If you are compiling it yourself
+# you may need to point hostapd to your version of libnl.
+#
+#CFLAGS += -I$<path to libnl include files>
+#LIBS += -L$<path to libnl library files>
+
+# Use libnl v2.0 (or 3.0) libraries.
+#CONFIG_LIBNL20=y
+
+# Use libnl 3.2 libraries (if this is selected, CONFIG_LIBNL20 is ignored)
+CONFIG_LIBNL32=y
+
+
+# Driver interface for FreeBSD net80211 layer (e.g., Atheros driver)
+#CONFIG_DRIVER_BSD=y
+#CFLAGS += -I/usr/local/include
+#LIBS += -L/usr/local/lib
+#LIBS_p += -L/usr/local/lib
+#LIBS_c += -L/usr/local/lib
+
+# Driver interface for no driver (e.g., RADIUS server only)
+#CONFIG_DRIVER_NONE=y
+
+# WPA2/IEEE 802.11i RSN pre-authentication
+CONFIG_RSN_PREAUTH=y
+
+# Support Operating Channel Validation
+CONFIG_OCV=y
+
+# Integrated EAP server
+CONFIG_EAP=y
+
+# EAP Re-authentication Protocol (ERP) in integrated EAP server
+CONFIG_ERP=y
+
+# EAP-MD5 for the integrated EAP server
+CONFIG_EAP_MD5=y
+
+# EAP-TLS for the integrated EAP server
+CONFIG_EAP_TLS=y
+
+# EAP-MSCHAPv2 for the integrated EAP server
+CONFIG_EAP_MSCHAPV2=y
+
+# EAP-PEAP for the integrated EAP server
+CONFIG_EAP_PEAP=y
+
+# EAP-GTC for the integrated EAP server
+CONFIG_EAP_GTC=y
+
+# EAP-TTLS for the integrated EAP server
+CONFIG_EAP_TTLS=y
+
+# EAP-SIM for the integrated EAP server
+#CONFIG_EAP_SIM=y
+
+# EAP-AKA for the integrated EAP server
+#CONFIG_EAP_AKA=y
+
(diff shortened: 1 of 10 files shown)


recipes-connectivity/wpa-supplicant/wpa-supplicant_%.bbappend
```diff
@@ -0,0 +1,22 @@
+LIC_FILES_CHKSUM:qcom = "file://${COMMON_LICENSE_DIR}/${LICENSE};md5=550794465ba0ec5312d6919e203a55f9"
+
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${BPN}:"
+
+PV:qcom="2.12+git"
```
koenkooi, line 5: Please keep PV and SRCREV close together in a recipe, they need to be updated in lockstep, so having them apart makes it too easy to forget updating one.

koenkooi (changes requested):

recipes-connectivity/hostapd/hostapd/hostapd.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
```
koenkooi, line 1: This looks identical to the upstream one, why is it duplicated here?

recipes-connectivity/hostapd/hostapd/init
```diff
@@ -0,0 +1,58 @@
+#!/bin/sh
```
koenkooi, line 1: This looks identical to the upstream one, why is it duplicated here?

recipes-connectivity/hostapd/hostapd_%.bbappend
```diff
@@ -0,0 +1,20 @@
+LIC_FILES_CHKSUM:qcom = "file://hostapd/README;md5=0e430ef1be3d6eebf257cf493fc7661d"
+
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${BPN}:"
+
+PV:qcom = "2.12+git"
```
koenkooi, line 5: As in the other recipe, please keep PV and SRCREV together.

koenkooi (changes requested):

recipes-connectivity/wpa-supplicant/wpa-supplicant_%.bbappend
```diff
@@ -0,0 +1,22 @@
...
+
+SRC_URI:qcom = "git://w1.fi/hostap.git;protocol=git;branch=main \
```
lumag, line 7: The commit message doesn't answer the main question: why. Which features are present only in the Git version, so that they prompt switching to the Git version?

My preference would be to select necessary patches and use them here instead of switching versions. This way it is easier to drop them once they are applied.

recipes-connectivity/hostapd/hostapd/defconfig.qcom
```diff
@@ -0,0 +1,437 @@
+# Example hostapd build time configuration
```
lumag, line 1: Patch the defconfig instead of replacing it completely. It is impossible to track the changes once upstream starts to change the file.

recipes-connectivity/hostapd/hostapd_%.bbappend
```diff
@@ -0,0 +1,20 @@
...
+
+SRC_URI:qcom = "git://w1.fi/hostap.git;protocol=git;branch=main \
+                file://defconfig.qcom \
+		file://init \
+		file://hostapd.service \
```
lumag, line 10: The commit message doesn't describe, why do you need to switch to the Git repo. When can we switch back. As I wrote earlier, I'd prefer to see a set of patches instead of having a semi-random commit ID being selected for the build.

lumag (changes requested):

lumag:
This wasn't even properly tested on the developer's side.

ndechesne:
Please note that this is not appropriate to close a PR and open another one right away with the same content. We are loosing the history of the review doing such. please avoid doing that in the future. thanks!

lumag:
> Please note that this is not appropriate to close a PR and open another one right away with the same content. We are loosing the history of the review doing such. please avoid doing that in the future. thanks!

I think your internal docs must be updated to mention that one shouldn't use the default ('master') branch for PRs. Otherwise it's easy to close those by default by refreshing the repo / branch.

recipes-connectivity/wpa-supplicant/wpa-supplicant_%.bbappend
miaoqing-quic: The tarball version is often far behind the Git version, I wanted to sync to the latest version, but I think what you said makes sense.

recipes-connectivity/wpa-supplicant/wpa-supplicant/wpa_supplicant.conf-sane
```diff
@@ -0,0 +1,7 @@
+ctrl_interface=/var/run/wpa_supplicant
```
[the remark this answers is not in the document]
miaoqing-quic: I have no idea to reuse upstream one.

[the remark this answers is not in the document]
lumag: @miaoqing-quic I can't understand your comment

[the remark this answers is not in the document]
miaoqing-quic: Anyway, please ignore those comments. I will discard this patch， thanks for review.

ricardosalveti:
I don't think we should be forking upstream related recipes just to move them to a newer revision, we should instead work with oe-core and find a common solution here (or even the upstream project when a new release is desired). This will cause a component to be based on a revision that is not managed by OE, and won't receive stable / security updates after a release is made, which is bad to us as we will have to maintain the fork (even if purely based on upstream).

Backporting the patches are also one option, as suggested by @lumag, otherwise I would rather have a dev specific recipe with AUTOREV or similar that is only used when selected by the user, and still, I'm not sure this would be material for a BSP layer.

Also, for the config changes, please make those dynamic based on what gets selected by the recipe (via options which the user can control or that can be extended by the BSP layer).

miaoqing-quic:
> I don't think we should be forking upstream related recipes just to move them to a newer revision, we should instead work with oe-core and find a common solution here (or even the upstream project when a new release is desired). This will cause a component to be based on a revision that is not managed by OE, and won't receive stable / security updates after a release is made, which is bad to us as we will have to maintain the fork (even if purely based on upstream).
> 
> Backporting the patches are also one option, as suggested by @lumag, otherwise I would rather have a dev specific recipe with AUTOREV or similar that is only used when selected by the user, and still, I'm not sure this would be material for a BSP layer.
> 
> Also, for the config changes, please make those dynamic based on what gets selected by the recipe (via options which the user can control or that can be extended by the BSP layer).

Thanks for your suggestion, will discard this patch.

lumag:
Closing this for now. Please reopen if you'd like to return to this topic.

PR #792, opened 2025-03-10
sc7280: camera subsystem, rb3gen2-vision-mezzanine

ldts:
Frame Capture Test

```
media-ctl --reset
media-ctl -v -V '"imx577 '19-001a'":0[fmt:SRGGB10/4056x3040 field:none]'
media-ctl -V '"msm_csiphy3":0[fmt:SRGGB10/4056x3040]'
media-ctl -V '"msm_csid0":0[fmt:SRGGB10/4056x3040]'
media-ctl -V '"msm_vfe0_rdi0":0[fmt:SRGGB10/4056x3040]'
media-ctl -l '"msm_csiphy3":1->"msm_csid0":0[1]'
media-ctl -l '"msm_csid0":1->"msm_vfe0_rdi0":0[1]'

read -p "Start capturing " foo
yavta -B capture-mplane -c -I -n 5 -f SRGGB10P -s 4056x3040 -F /dev/video0

```

Videosink test (gstreamer)

`gst-launch libcamerasrc ! videoconvert ! autovideosink`

**Notes**
Note that the 12MP camera sensor (`imx577`) is multiplanar and therefore[ not supported natively](https://github.com/PipeWire/pipewire/blob/master/spa/plugins/v4l2/v4l2-utils.c#L80) by Pipewire. The libcamera spa (`libspa-0.2-libcamera`) is required for Pipewire to be able to detect the camera as a source.

Also, the camera generates `SRGGB10P` (10 bit packed format) and since there is no ISP converting Bayer to RGB, this must be done via software (libcamera takes care of that as well: run gst-launch in a verbose way to check the cpu involvement).

Convenient to enable CONFIG_UDMABUF for the Gstreamer pipeline.

### DIFF
--- a/conf/machine/qcs6490-rb3gen2-core-kit.conf
+++ b/conf/machine/qcs6490-rb3gen2-core-kit.conf
@@ -6,8 +6,11 @@ require conf/machine/include/qcom-qcs6490.inc
 
 MACHINE_FEATURES += "efi pci"
 
+QCOM_DTB_DEFAULT ?= "qcs6490-rb3gen2"
+
 KERNEL_DEVICETREE ?= " \
                       qcom/qcs6490-rb3gen2.dtb \
+                      qcom/qcs6490-rb3gen2-vision-mezzanine.dtb \
                       "
 
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
--- a/recipes-kernel/linux/linux-yocto-dev.bbappend
+++ b/recipes-kernel/linux/linux-yocto-dev.bbappend
@@ -9,6 +9,9 @@ SRC_URI:append:qcom = " \
     file://qcm6490-board-dts/0002-arm64-dts-qcom-sc7280-Add-global-PCIe-interrupt.patch \
     file://qcm6490-board-dts/0001-dt-bindings-PCI-Add-binding-for-Toshiba-TC956x-PCIe-.patch \
     file://qcm6490-board-dts/0002-arm64-dts-qcom-qcs6490-rb3gen2-Add-TC956x-PCIe-switc.patch \
+    file://qcm6490-board-dts/0001-arm64-dts-qcom-sc7280-Add-support-for-camss.patch \
+    file://qcm6490-board-dts/0002-arm64-dts-qcom-qcs6490-rb3gen2-vision-mezzanine-Add-.patch \
+    file://qcm6490-board-dts/0002-media-dt-bindings-update-clocks-for-sc7280-camss.patch \
     file://workarounds/0001-QCLINUX-arm64-dts-qcom-qcm6490-disable-sdhc1-for-ufs.patch \
     file://workarounds/0001-PENDING-arm64-dts-qcom-Remove-voltage-vote-support-f.patch \
     file://drivers/0003-PCI-Add-new-start_link-stop_link-function-ops.patch \
@@ -17,6 +20,7 @@ SRC_URI:append:qcom = " \
     file://drivers/0006-PCI-qcom-Add-support-for-host_stop_link-host_start_l.patch \
     file://drivers/0007-PCI-PCI-Add-pcie_is_link_active-to-determine-if-the-.patch \
     file://drivers/0008-PCI-pwrctrl-Add-power-control-driver-for-tc956x.patch \
+    file://drivers/0001-media-qcom-camss-update-clock-names-for-sc7280.patch \
 "
 
 # Include additional kernel configs.
--- a/recipes-kernel/linux/linux-yocto-dev/drivers/0001-media-qcom-camss-update-clock-names-for-sc7280.patch
+++ b/recipes-kernel/linux/linux-yocto-dev/drivers/0001-media-qcom-camss-update-clock-names-for-sc7280.patch
@@ -1,0 +1,94 @@
+From 306753e5d8a48b29ee3f73837efea50b421bff5a Mon Sep 17 00:00:00 2001
+From: Vikram Sharma <quic_vikramsa@quicinc.com>
+Date: Tue, 21 Jan 2025 23:37:46 +0530
+Subject: [PATCH 1/2] media: qcom: camss: update clock names for sc7280
+
+Update clock names to make them consistent with existing platform i.e
+sc8280xp. Rename gcc_cam_hf_axi to gcc_axi_hf and add gcc_axi_sf.
+
+Signed-off-by: Vikram Sharma <quic_vikramsa@quicinc.com>
+Upstream-Status: Submitted [https://lore.kernel.org/linux-arm-msm/20250121180746.1989996-1-quic_vikramsa@quicinc.com/]
+---
+ drivers/media/platform/qcom/camss/camss.c | 15 ++++++++++-----
+ 1 file changed, 10 insertions(+), 5 deletions(-)
+
+diff --git a/drivers/media/platform/qcom/camss/camss.c b/drivers/media/platform/qcom/camss/camss.c
+index a85e9df0f301..b1358457c66e 100644
+--- a/drivers/media/platform/qcom/camss/camss.c
++++ b/drivers/media/platform/qcom/camss/camss.c
+@@ -1443,12 +1443,13 @@ static const struct camss_subdev_resources vfe_res_7280[] = {
+ 		.regulators = {},
+ 
+ 		.clock = { "camnoc_axi", "cpas_ahb", "icp_ahb", "vfe0",
+-			   "vfe0_axi", "gcc_cam_hf_axi" },
++			   "vfe0_axi", "gcc_axi_hf", "gcc_axi_sf" },
+ 		.clock_rate = { { 150000000, 240000000, 320000000, 400000000, 480000000 },
+ 				{ 80000000 },
+ 				{ 0 },
+ 				{ 380000000, 510000000, 637000000, 760000000 },
+ 				{ 0 },
++				{ 0 },
+ 				{ 0 } },
+ 
+ 		.reg = { "vfe0" },
+@@ -1468,12 +1469,13 @@ static const struct camss_subdev_resources vfe_res_7280[] = {
+ 		.regulators = {},
+ 
+ 		.clock = { "camnoc_axi", "cpas_ahb", "icp_ahb", "vfe1",
+-			   "vfe1_axi", "gcc_cam_hf_axi" },
++			   "vfe1_axi", "gcc_axi_hf", "gcc_axi_sf" },
+ 		.clock_rate = { { 150000000, 240000000, 320000000, 400000000, 480000000 },
+ 				{ 80000000 },
+ 				{ 0 },
+ 				{ 380000000, 510000000, 637000000, 760000000 },
+ 				{ 0 },
++				{ 0 },
+ 				{ 0 } },
+ 
+ 		.reg = { "vfe1" },
+@@ -1493,12 +1495,13 @@ static const struct camss_subdev_resources vfe_res_7280[] = {
+ 		.regulators = {},
+ 
+ 		.clock = { "camnoc_axi", "cpas_ahb", "icp_ahb", "vfe2",
+-			   "vfe2_axi", "gcc_cam_hf_axi" },
++			   "vfe2_axi", "gcc_axi_hf", "gcc_axi_sf" },
+ 		.clock_rate = { { 150000000, 240000000, 320000000, 400000000, 480000000 },
+ 				{ 80000000 },
+ 				{ 0 },
+ 				{ 380000000, 510000000, 637000000, 760000000 },
+ 				{ 0 },
++				{ 0 },
+ 				{ 0 } },
+ 
+ 		.reg = { "vfe2" },
+@@ -1516,11 +1519,12 @@ static const struct camss_subdev_resources vfe_res_7280[] = {
+ 	/* VFE3 (lite) */
+ 	{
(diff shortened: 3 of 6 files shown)


ldts:
to validate we still need the kernel-modules 
tested on rb3gen2 6.14-rc3 (yocto-dev)

recipes-kernel/linux/linux-yocto-dev.bbappend
```diff
@@ -5,8 +5,12 @@ FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}:"
 
 SRC_URI:append:qcom = " \
     file://qcm6490-board-dts/0001-FROMLIST-arm64-dts-qcom-qcm6490-idp-Update-protected.patch \
+    file://qcm6490-board-dts/0001-arm64-dts-qcom-sc7280-Add-support-for-camss.patch \
```
lumag, line 8: Both the commit message and the subject needs to be improved. You are not enabling 12MP camera. You are picking up patches to enable vision mezzanine for the RB3 Gen2 board.

conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -7,7 +7,7 @@ require conf/machine/include/qcom-qcs6490.inc
 MACHINE_FEATURES += "efi pci"
 
 KERNEL_DEVICETREE ?= " \
-                      qcom/qcs6490-rb3gen2.dtb \
+                      qcom/qcs6490-rb3gen2-vision-mezzanine.dtb \
```
lumag, line 10: No. The core kit is a kit without the mezzanine. And as with the previous patch, both the subject and the message needs to be improved.

lumag (changes requested):

recipes-kernel/linux/linux-yocto-dev.bbappend
ldts: huh?? this is not an either or: this is doing both things, just different semantic layers in case you didn't notice (one targets the user trying to validate some functionality, the other the developer).
but sure, I understand the target audience, will amend.

lumag: I didn't get a message regarding the 'target audioence'. The changes needs to be decribed properly. Your commit message isn't specific enough.

ldts: right, this PR, at high functional level, tells someone wanting to test the 12MP camera what they need to pick up from the layer (unequivocally). The target audience is usually a tester or a just user.

At a lower level it does what is described in the commits. I am happy to adjust to that level. I wasn't sure which one to chose.

lumag: I think the kernel has the best principles and the best possible description for that: https://docs.kernel.org/process/submitting-patches.html#describe-your-changes

conf/machine/qcs6490-rb3gen2-core-kit.conf
lumag: I'd rather not, but it's up to the corresponding maintainers.

recipes-kernel/linux/linux-yocto-dev.bbappend
ldts: it does and I am aware but this is not the kernel. Please no need to insist. seriously. I get it.

conf/machine/qcs6490-rb3gen2-core-kit.conf
ldts: so something like this instead? trying to agree before proposing (rebuilding and retesting is a bit of a pain)

```
--- a/conf/machine/qcs6490-rb3gen2-core-kit.conf
+++ b/conf/machine/qcs6490-rb3gen2-core-kit.conf
@@ -6,9 +6,12 @@ require conf/machine/include/qcom-qcs6490.inc
 
 MACHINE_FEATURES += "efi pci"
 
-KERNEL_DEVICETREE ?= " \
-                      qcom/qcs6490-rb3gen2-vision-mezzanine.dtb \
-                      "
+USE_VISION_MEZZANINE ?= "0"
+
+KERNEL_DEVICETREE_qcs6490-rb3gen2 = "qcom/qcs6490-rb3gen2.dtb"
+KERNEL_DEVICETREE_qcs6490-rb3gen2_vision = "qcom/qcs6490-rb3gen2-vision-mezzanine.dtb"
+
+KERNEL_DEVICETREE = "${@bb.utils.contains('USE_VISION_MEZZANINE', '1', '${KERNEL_DEVICETREE_qcs6490-rb3gen2_vision}', '${KERNEL_DEVICETREE_qcs6490-rb3gen2}', d)}"
 
```

ndechesne: Unfortunately there is no 'good' solution for now.. 
The current firmware can select the 'best' DTB if we have more than one DTBs in the dtb.bin image on the dtb partition. However to make this choice the firmware uses the msm-id , board-id. However these properties are not available in the upstream kernel (rejected). 

We are working on a improved DTB image format that would allow the selection of the DTB, but it's not ready yet.. 

So right now, we only support 1 DTB in the DTB image. I agree with @lumag that duplicating the machine .conf should be avoided.. 

So, @ricardosalveti @lumag , how about we include a kernel patch to include the msm-id for now? It's not great, but perhaps it's still the best option we have?

ldts: I'll propose with the above and I can then modify if needed. just to have a consistent story in the thread

lumag: I'm fine with msm-id, provided it is documented as a temporary measure that will be dropped.

recipes-kernel/linux/linux-yocto-dev.bbappend
lumag, line 8: Nit: `linux-yocto-dev`

Non-nit: please drop the Tested-by tag. I mean, you are expected to test your own PRs, there is no need to document anything.

conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -6,9 +6,12 @@ require conf/machine/include/qcom-qcs6490.inc
 
 MACHINE_FEATURES += "efi pci"
 
-KERNEL_DEVICETREE ?= " \
-                      qcom/qcs6490-rb3gen2.dtb \
-                      "
+USE_VISION_MEZZANINE ?= "0"
```
lumag, line 9: This doesn't scale, there might be other mezzanines. If you don't want to introduce `qcom-msm-id` patch, I'd suggest something like `KERNEL_DEVICETREE_VARIANT`, being empty, `-vision-mezzanine` or  selecting some other mezz.

lumag (changes requested):

conf/machine/qcs6490-rb3gen2-core-kit.conf
ldts: dont mind using the msm-id patch, was just waiting for a decision and advancing something simple for review.

sounds good, could do the KERNEL_DEVICETREE_VARIANT instead if that is what we want (will wait for the msm-id decision though)

conf/machine/qcs6490-rb3gen2-core-kit.conf
ldts: thinking about it, wont using msm-id make it harder to revert? we will have to carry kernel patches whereas with your KERNEL_DEVICETREE_VARIANT we only need to revert one config change. Seems a better temporary choice.

also, do you know where is the board-id/msm-id for the vision mezzanine? I cant find it.

ldts: > thinking about it, wont using msm-id make it harder to revert? we will have to carry kernel patches whereas with your KERNEL_DEVICETREE_VARIANT we only need to revert one config change. Seems a better temporary choice.
> 
> also, do you know where is the board-id/msm-id for the vision mezzanine? I cant find it.

let me test the msm-id/board-id. will report back.

ricardosalveti: I'm -1 on merging msm-id / board-id as it will be harder to revert later on (and maintain), I would prefer having the team focusing on the current fit proposal (which will be upstream aligned).

I'm fine extending the list of device tree files used by this board configuration, the user can select the default one to be loaded via QCOM_DTB_DEFAULT, or by having all of them available in the dtb partition and having the user renaming what should be the default dtb used.

ldts: I'll propose anyway as I did with the previous one and we can nack it ..

conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -6,9 +6,9 @@ require conf/machine/include/qcom-qcs6490.inc
 
 MACHINE_FEATURES += "efi pci"
 
-KERNEL_DEVICETREE ?= " \
-                      qcom/qcs6490-rb3gen2.dtb \
-                      "
+KERNEL_DEVICETREE_VARIANT ?= ""
+KERNEL_DEVICETREE_BASE ?= "qcom/qcs6490-rb3gen2"
+KERNEL_DEVICETREE ?= "${KERNEL_DEVICETREE_BASE}${@bb.utils.contains('KERNEL_DEVICETREE_VARIANT', '', '-${KERNEL_DEVICETREE_VARIANT}', '', d)}.dtb"
```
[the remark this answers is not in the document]
lumag: I completely forgot about the `QCOM_DTB_DEFAULT`. I'd support using it.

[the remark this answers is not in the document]
ldts: I don't think it's a better solution. If `QCOM_DTB_DEFAULT` wasn’t already in place, we wouldn’t implement it. Why generate all the VFAT images when we already know at compile time which one we’re going to use? It adds no value—just unnecessary waste.  

Is it useful to restrict the list of valid device trees at build time for a platform when we don’t perform DTB detection at runtime, when we dont do a compound dtb? Maybe, maybe not—I’m not sure. But honestly, I fail to see the benefit.  

As a user, I find `KERNEL_DEVICETREE_VARIANT` more flexible and resource-efficient and logical.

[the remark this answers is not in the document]
ndechesne: generating multiple artifact per board/device is something we do in a few other places already. Until we can manage all DTBs in a single image, I don't think it's a concern to generate 1 VFAT image per DTB, and let the user flash the appropriate one. QCOM_DTB_DEFAULT will just select which one of these images is called 'dtb.bin'. But having all of them available is convenient for a developer.

conf/machine/qcs6490-rb3gen2-core-kit.conf
ldts: I prototyped and validated the qcom,board-id approach. It requires maintaining the DTB patch and making some infrastructure changes to create the compound DTB and provide the new CDT. It’s manageable, but after internal discussions, we’ve decided to drop it for now. We’re waiting for a proper runtime DTB selection feature.

conf/machine/qcs6490-rb3gen2-core-kit.conf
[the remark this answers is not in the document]
ldts: So QCOM_DTB_DEFAULT is a better choice because 1) it restricts and 2) it builds all the possible device tree images?
This logic and naming is clearly wrong: seems to me that QCOM_DTB_DEFAULT should be renamed as QCOM_DTB_ALL and DTB_DEFAULT should just be a parameter to the flash command.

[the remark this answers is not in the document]
ndechesne: QCOM_DTB_DEFAULT is the 'default' dtb flashed if the developer does not do anything. If the user wants to flash another DTB, then they need to run an extra command (either override the dtb.bin file or rename the file in the rawprogram XML file). 

In the default use case, which is 1) git clone, 2) build, 3) run qdl on the generated folder, we want the default DTB to be used.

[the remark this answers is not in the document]
ldts: right and I do understand how it works: what I question is that QCOM_DTB_DEFAULT has the side effect of building and deploying **_all_** the images when the user asked for just **one**.

If you say this is what makes it better then fair enough. But I usually trust the user to know what it wants to build.

anyway, I'll update, I get the picture.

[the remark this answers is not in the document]
ndechesne: I like the idea of generic machines too. But it has plenty of implications on the QCOM 'internal' releases as well, that we should discuss..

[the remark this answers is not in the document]
lumag: Sure!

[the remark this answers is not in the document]
ricardosalveti: Yeah, we should raise a separated issue to discuss how to better use CDT here, and how it will match with the fit proposal, with the goal to better support a generic image and runtime detection of some sort instead.

ldts:
> Please rebase.

yep, also moving patches to `drivers`/ now and retesting

ldts:
@ricardosalveti done

lumag (approved):

ricardosalveti (approved):

ndechesne (approved):

PR #793, opened 2025-03-11
adb serial number: fall back option to populate serial number

vishwasudupa:
when androidboot.serialno is not available in cmdline then populate adb serial number from firmware platform-parts-info dt entry.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf-configfs/qcom/android-gadget-setup.machine
+++ b/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf-configfs/qcom/android-gadget-setup.machine
@@ -1,5 +1,7 @@
 manufacturer=Qualcomm
 model=`hostname`
 androidserial="$(sed -n -e '/androidboot.serialno/  s/.*androidboot.serialno=\([^ ]*\).*/\1/gp ' /proc/cmdline)"
+[ -z "$androidserial" ] && [ -e /dev/sda ] && scsiserialline=$(udevadm info /dev/sda | grep ID_SCSI_SERIAL) && scsiserialvalue=$(echo "$scsiserialline" | awk -F= '{print $2}') && androidserial=$(echo -n "$scsiserialvalue" | crc32)
+[ -z "$androidserial" ] && [ -e /sys/class/mmc_host/mmc0/mmc0:0001/serial ] && androidserial=$(sed 's/0x//' /sys/class/mmc_host/mmc0/mmc0:0001/serial)
 [ -n "$androidserial" ] && serial="$androidserial"
 true


lumag:
Please provide additional details regarding this node. It is not described in bindings, so it is not clear, which party populates it, on which platforms, etc.
Also please take a look at #547.

vishwasudupa:
> Please provide additional details regarding this node. It is not described in bindings, so it is not clear, which party populates it, on which platforms, etc. Also please take a look at #547.

this node is populated by uefi during runtime and the serial number is fetched from 
CardInfoData. Same place from where abl used to fetch and populate in cmdline.
Since for non abl targets this is not available, populating it from this node.

lumag:
This node is not described anywhere. Also it is specific to a particular UEFI implementation, I don't have it on either of my Snapdragon laptops. It is not supported by other UEFI bootloaders (U-Bot). it doesn't follow DT standards, etc. As such, no this is not acceptable from my point of view.

quic-vkraleti:
> This node is not described anywhere. Also it is specific to a particular UEFI implementation, I don't have it on either of my Snapdragon laptops. It is not supported by other UEFI bootloaders (U-Bot). it doesn't follow DT standards, etc. As such, no this is not acceptable from my point of view.

Yes, this implementation is specific to UEFI used in Qualcomm Linux and is not generic or applicable to other boot loaders. Unlike ABL, systemd-boot cannot handle kernel command line fixup, so passing androidboot.serialno will no longer work.

A possible solution would be to read /sys/class/mmc_host/mmc0/mmc0:0001/serial for EMMC devices and to read ID_SCSI_SERIAL using udevadm info /dev/sda for UFS devices.

ndechesne:
> A possible solution would be to read /sys/class/mmc_host/mmc0/mmc0:0001/serial for EMMC devices and to read ID_SCSI_SERIAL using udevadm info /dev/sda for UFS devices.

this looks better to me, since this is the exact implementation of the serial id, which we are exposing very clearly.

lumag (approved):
I don't particularly like hardcoded `sda` and `mmc0`, but I don't see  a better way.

Could you please confirm that it works on a eMMC board which had USB thumbdrive installed during boot and that there is no serial change in this config.

ricardosalveti:
> I don't particularly like hardcoded `sda` and `mmc0`, but I don't see a better way.
> 
> Could you please confirm that it works on a eMMC board which had USB thumbdrive installed during boot and that there is no serial change in this config.

Should we swap the order to prioritize mmc0?

ndechesne:
I suppose this code runs after we mounted the root fs, right? so , we could find the root block device from '/' without hardcoding sda or mmc0 here, no?

lumag:
> I suppose this code runs after we mounted the root fs, right? so , we could find the root block device from '/' without hardcoding sda or mmc0 here, no?

It can be as well be executed from the initramfs, so no, we cannot depend on / being the actual rootfs.

lumag:
> > I don't particularly like hardcoded `sda` and `mmc0`, but I don't see a better way.
> > Could you please confirm that it works on a eMMC board which had USB thumbdrive installed during boot and that there is no serial change in this config.
> 
> Should we swap the order to prioritize mmc0?

But why? It will cause (almost) the same question of the MMC card being wired to the uSD slot (MMC to make sure the address is the same, SD cards use slightly different addressing).

ricardosalveti:
> > Should we swap the order to prioritize mmc0?
> 
> But why? It will cause (almost) the same question of the MMC card being wired to the uSD slot (MMC to make sure the address is the same, SD cards use slightly different addressing).

Just thinking on the most common case, but even mmc0 is not ideal here, since the slot might quite easily change.

It looks a bit too fragile to me still.

lumag:
> > > Should we swap the order to prioritize mmc0?
> > 
> > 
> > But why? It will cause (almost) the same question of the MMC card being wired to the uSD slot (MMC to make sure the address is the same, SD cards use slightly different addressing).
> 
> Just thinking on the most common case, but even mmc0 is not ideal here, since the slot might quite easily change.

Well, mmc0 is usually fixed through aliases (unlike sda)

vishwasudupa:
> Could you please confirm that it works on a eMMC board which had USB thumbdrive installed during boot and that there is no serial change in this config.

the serial value does not change, even if thumbdrive is installed during the boot

lumag:
> > Could you please confirm that it works on a eMMC board which had USB thumbdrive installed during boot and that there is no serial change in this config.
> 
> the serial value does not change, even if thumbdrive is installed during the boot

Will it change if the board uses UFS and the thumbdrive gets probed before the UFS?

quic-vkraleti:
> > > Could you please confirm that it works on a eMMC board which had USB thumbdrive installed during boot and that there is no serial change in this config.
> > 
> > 
> > the serial value does not change, even if thumbdrive is installed during the boot
> 
> Will it change if the board uses UFS and the thumbdrive gets probed before the UFS?

We couldn't get a 6490/9100 based board that supports UFS and has thumb drive support. We'll check and get back once device is handy. For now as remaining cases are fine, approving this PR to let the changes merge.

quic-vkraleti (approved):

lumag:
> > > > Could you please confirm that it works on a eMMC board which had USB thumbdrive installed during boot and that there is no serial change in this config.
> > > 
> > > 
> > > the serial value does not change, even if thumbdrive is installed during the boot
> > 
> > Will it change if the board uses UFS and the thumbdrive gets probed before the UFS?
> 
> We couldn't get a 6490/9100 based board that supports UFS and has thumb drive support. We'll check and get back once device is handy. For now as remaining cases are fine, approving this PR to let the changes merge.

Can't you use RB5 to test it?

PR #795, opened 2025-03-12
Bump SRCREV of linux-qcom-staging to latest.

quic-vkraleti:
Update linux-qcom-staging recipe to fetch the latest git revision to pull in several fixes.

### DIFF
--- a/recipes-kernel/linux/linux-qcom-staging_6.12.bb
+++ b/recipes-kernel/linux/linux-qcom-staging_6.12.bb
@@ -18,7 +18,8 @@ BBCLASSEXTEND = "devupstream:target"
 PN:class-devupstream = "linux-qcom-staging-tip"
 SRCREV:class-devupstream = "${AUTOREV}"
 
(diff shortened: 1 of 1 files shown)


ricardosalveti:
Would you mind extending the commit message giving a bit more details about the changes? I know there are several changes between both revs, but just a small overall would already be useful.

Might also be good to say this is pointing to the LKP.QCLINUX.0.0-11000-kernel.0 tag (seems to be the case). That should help later if we are also expecting more tags to be created.

quic-vkraleti:
> Would you mind extending the commit message giving a bit more details about the changes? I know there are several changes between both revs, but just a small overall would already be useful.
> 
> Might also be good to say this is pointing to the LKP.QCLINUX.0.0-11000-kernel.0 tag (seems to be the case). That should help later if we are also expecting more tags to be created.

Added tag information in commit as well as in recipe for later reference.

ricardosalveti (approved):

lumag (approved):

sbanerjee-quic (approved):

PR #796, opened 2025-03-12
Add custom BSP selection support

quic-vkraleti:
To enable additional value adds of the QCOM BSP, introduce custom BSP selection support. Selecting this override would bring in add-on device-tree overlays, additional kernel modules, and the option to choose user space frameworks specific to QCOM value adds.
    
The 'custom BSP' can be selected by adding `QCOM_SELECTED_BSP = "custom"` in a configuration file like local.conf. This results in `MACHINEOVERRIDES = "xxx:qcom-custom-bsp:yyy"`.

### DIFF
--- a/conf/machine/qcm6490-idp.conf
+++ b/conf/machine/qcm6490-idp.conf
@@ -10,6 +10,12 @@ KERNEL_DEVICETREE ?= " \
                       qcom/qcm6490-idp.dtb \
                       "
 
+# The addon DTs produced by the linux-qcom-staging kernel are not yet upstreamed.
+KERNEL_DEVICETREE:append:pn-linux-qcom-staging = " \
+                      qcom/qcm6490-addons-idp.dtb \
+                      qcom/qcm6490-addons-idp-amoled.dtb \
+                      "
+
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-rb3gen2-firmware \
     packagegroup-rb3gen2-hexagon-dsp-binaries \
--- a/recipes-kernel/linux/linux-qcom-staging_6.12.bb
+++ b/recipes-kernel/linux/linux-qcom-staging_6.12.bb
@@ -26,7 +26,7 @@ S = "${WORKDIR}/git"
 KERNEL_CONFIG ?= "qcom_defconfig"
 
 # Additional fragment for qcom value add features
-KERNEL_CONFIG_FRAGMENTS += " ${S}/arch/arm64/configs/qcom_addons.config"
+KERNEL_CONFIG_FRAGMENTS:append:qcom-custom-bsp = " ${S}/arch/arm64/configs/qcom_addons.config"
 
 do_configure:prepend() {
     if [ ! -f "${S}/arch/${ARCH}/configs/${KERNEL_CONFIG}" ]; then
--- a/conf/machine/include/qcom-base.inc
+++ b/conf/machine/include/qcom-base.inc
@@ -1,5 +1,11 @@
 # Common configurations and variables for Qualcomm platforms.
 
+# 'custom' BSP selection on top of the default BSP, provides additional functionality
+# such as addon DT overlays, downstream DLKMs, and QCOM-specific user space frameworks.
+# Specific MACHINEs or DISTROs may choose this configuration to utilize addon features.
+QCOM_SELECTED_BSP ??= ""
+MACHINEOVERRIDES =. "${@bb.utils.contains('QCOM_SELECTED_BSP', 'custom', 'qcom-custom-bsp:', '', d)}"
+
 # Provider for linux kernel
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-yocto-dev"
 
--- a/conf/machine/qcs6490-rb3gen2-core-kit.conf
+++ b/conf/machine/qcs6490-rb3gen2-core-kit.conf
@@ -10,6 +10,11 @@ KERNEL_DEVICETREE ?= " \
                       qcom/qcs6490-rb3gen2.dtb \
                       "
 
+# The addon DTs produced by the linux-qcom-staging kernel are not yet upstreamed.
+KERNEL_DEVICETREE:append:pn-linux-qcom-staging = " \
+                      qcom/qcs6490-addons-rb3gen2.dtb \
+                      "
+
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-rb3gen2-firmware \
     packagegroup-rb3gen2-hexagon-dsp-binaries \
--- a/conf/machine/qcs9100-ride-sx.conf
+++ b/conf/machine/qcs9100-ride-sx.conf
@@ -13,6 +13,14 @@ KERNEL_DEVICETREE ?= " \
                       qcom/sa8775p-ride-r3.dtb \
                       "
 
+# The addon DTs produced by the linux-qcom-staging kernel are not yet upstreamed.
+KERNEL_DEVICETREE:append:pn-linux-qcom-staging = " \
+                      qcom/qcs9100-addons-ride.dtb \
+                      qcom/qcs9100-addons-ride-r3.dtb \
+                      qcom/sa8775p-addons-ride.dtb \
+                      qcom/sa8775p-addons-ride-r3.dtb \
+                      "
+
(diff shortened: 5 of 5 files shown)


recipes-kernel/linux/linux-qcom-staging_6.12.bb
```diff
@@ -26,7 +26,7 @@ S = "${WORKDIR}/git"
 KERNEL_CONFIG ?= "qcom_defconfig"
 
 # Additional fragment for qcom value add features
-KERNEL_CONFIG_FRAGMENTS += " ${S}/arch/arm64/configs/qcom_addons.config"
+KERNEL_CONFIG_FRAGMENTS:append:qcom-custom-bsp = " ${S}/arch/arm64/configs/qcom_addons.config"
```
lumag, line 29: Why? Which features inside that config make it applicable to the value-added build?
I'd prefer to be able to use package selection and/or kmod blacklists instead of adding extra overrides at the build time.

quic-vkraleti: Agreed. Runtime switching to custom BSP using kmod blacklists, DT-based driver selection from the bootloader menu, etc., are works in progress. Teams are working on a war footing to achieve this. Until then, this build-time selection of value adds is a stopgap to make progress with the rest of the feature development.

conf/machine/qcm6490-idp.conf
```diff
@@ -10,6 +10,14 @@ KERNEL_DEVICETREE ?= " \
                       qcom/qcm6490-idp.dtb \
                       "
 
+# The addon DTs produced by the linux-qcom-staging kernel are not yet upstreamed.
+# A separate KERNEL_DEVICETREE would avoid conflicts with other kernel providers like linux-yocto-dev.
+KERNEL_DEVICETREE:pn-linux-qcom-staging:qcom-custom-bsp ?= " \
+                      qcom/qcm6490-idp.dtb \
+                      qcom/qcm6490-addons-idp.dtb \
+                      qcom/qcm6490-addons-idp-amoled.dtb \
```
lumag, line 18: This is absolute and definitive NAK. Device Tree defines hardware. As such, it can not depend on whether you enable 'normal' BSP or a 'custom' one. Hardware description doesn't depend on the driver. So, no, this can not go in.
Note: it is fine to have extra dtbs for the `linux-qcom-staging` recipe. But DTB selection can not depend on drivers.

conf/machine/qcm6490-idp.conf
```diff
@@ -10,6 +10,14 @@ KERNEL_DEVICETREE ?= " \
...
+                      "
```
ricardosalveti, line 19: There is a bit of duplication here (same for other machines), not sure if we should add recipe specific kernel devicetree files as part of the machine configuration.

Maybe define a .inc that gets included when QCOM_SELECTED_BSP is set to custom? Still not ideal, but at least we can have all the temporary settings in a single location.

recipes-kernel/linux/linux-qcom-staging_6.12.bb
lumag: It doesn't look like a stop-gap. It looks more like a virus, which we will never be able to get rid of if we allow it into the repo. We have a stop-gap of `QCOM_DTB_DEFAULT`, which worksaround the DT selection (as it is hanled outside of Yocto layers). Everything else can be fixed within this layer.

quic-vkraleti: Not all viruses are bad. Some of them have helped develop the human brain in the past. 

Setting QCOM_DTB_DEFAULT is only helping to pack one of the produced DTBs into dtb.bin. custom BSP support is beyond DT selection.  As of today, there are custom drivers  like camx, kgsl etc. that are using dt bindings not aligned to upstream. These dt bindings have been segregated into addons dtbs and all downstream configs are clubbed together in qcom_addons.config.  Ask is to support these only on 'custom BSP' so that there is a clean separation for users. 

Nevertheless, I'll convey to kernel team that the selection of addons.config via an override is not desirable and will check if there is a better solution.

conf/machine/qcm6490-idp.conf
quic-vkraleti: Ok, will inform team the same. For now I tested and all addon dtbs are compiling even without an override. So removed override part and converted KERNEL_DEVICETREE to a linux-qcom-staging recipe specific append.

conf/machine/qcm6490-idp.conf
quic-vkraleti: Converted KERNEL_DEVICETREE  to an linux-qcom-staging append for now. Hope that addresses this concern.

quic-vkraleti: As this is no longer tied to custom BSP support, If required I can move this commit to a new PR.

recipes-kernel/linux/linux-qcom-staging_6.12.bb
lumag: Let me explain, why I oppose to this option (and always will).
Having `QCOM_SELECTED_BSP` makes everybody believe that there are two and only two options: either an open-source BSP or a 'custom' BSP with proprietary drivers. This is not true. There is a definite request from hardware and software vendors to mix-and-match different components. A typical example would be using the Mesa/Freedreno driver _and_ proprietary camera stack. Any other proprietary driver adds another axis to this combo list. And yes, this comes from the actual vendors we have been working with.

That's why I said that `QCOM_SELECTED_BSP` is a virus: it collapses all variety of customer-required options to simple two usecases with no ability to untangle them later. Supporting Qualcomm camera stack _and_ Mesa would require restructuring the layer (again).

Having said that, I really hope that we can drop this idea and work on feature-based approach rather than having a single switch.

quic-vkraleti: I am completely aligned with this. Not only me, but everyone in the organization is. Our collective goal is to achieve all valid permutations and combinations of different components to serve the needs of the end users, not to limit to one or two.

However, today's reality is that we are not fully ready to realize this. There are conflicting software pieces that can't coexist in a single image. There are still downstream implementations that are slowly moving away. While things are evolving, from the distribution side, there is a need to validate and integrate the features that we have on hand. Postponing the distribution to consume downstream software would mean there is very little time in the future to meet end user asks.

Now, to be able to use downstream software, there is a need to have selections via QCOM_SELECTED_BSP. If this selection can't be defined in places like machine.confs or distro.confs, is it okay to bring this via CI YAML files to local.conf?

As the software matures, the appends and overrides can be dropped in a staggered manner, and QCOM_SELECTED_BSP support can be finally removed.

lumag: I'd prefer to have explicit `RCONFLICTS` for pieces that can't coexist in a single image.

For example, the full featured `vidc` external module should `RCONFLICT` with `kernel-module-iris`.
Adreno OpenGL libraries should be mutually exclusive with Mesa libraries (but it should be possible to co-install Adreno and Mesa CL and Vulkan libraries - they already provide coexistence mechanisms in a form of ICD). Once Adreno libraries are updated to use GLVND, there should be no more coexistence issue at all. 
Likewise, CamX drivers should conflict with existing camss modules.

As you can see, it's all expressed in the form of package management and properly describing dependencies and conflicts. On top of these packages you can create packagegroups, which pull in drivers and libraries. And on top of everything, a KAS YAML file which builds two rootfs images. One with the open-source BSP and another one with the proprietary bits. There is no need for `QCOM_SELECTED_BSP`.

ricardosalveti: I didn't check all the downstream kernel changes, but are we at least able to build all the standard + custom modules in a single build? Or do we require to disable certain standard modules in order to have the custom configs enabled? Just trying to understand if we have a build or runtime/package conflict at this stage.

quic-vkraleti: At this time, all the standard + custom modules can't be in a single build. Both build & runtime conflicts do exist. This clean up is going happen eventually. 

As there is a difference of opinion, for now I'll stop pursuing this PR further and introduce a new KAS YML to pick staging kernel and all custom modules into one build.

lumag: What is the build conflict? Please always point to the log or provide a necessary excerpt when reporting such issues.

PR #797, opened 2025-03-12
ci: force build of new/updates recipes

ndechesne:
In a PR test build, it's possible to have a new recipe (or updates with a bbappend) which is not included in any image we build. We could do a 'world' build but that would be really expensive. Instead this change tries to catch any updated .bb or .bbappend file, and force a bitbake build.

If the recipe was already built as part of the image, it should be a quick no-op. If the recipe was not built, then it's a good thing to check that it builds!


.github/workflows/build-yocto.yml
```diff
@@ -136,6 +136,18 @@ jobs:
...
+      # For PR, we might have new recipes not included in any image we build
+      # let's try to build any new or modified recipes, just in case.
+      # In most case it should be a quick no-op.
+      - name: Poor man's meta-qcom world build
+        if: github.event_name == 'pull_request'
+        run: |
+          export DL_DIR=${CACHE_DIR}/downloads
+          export SSTATE_DIR=${CACHE_DIR}/sstate-cache
+          export KAS_WORK_DIR=$PWD/../kas
+          mkdir $KAS_WORK_DIR
```
ricardosalveti, line 148: mkdir: cannot create directory ‘/opt/actions-runner/_work/meta-qcom/meta-qcom/../kas’: File exists

quic-vkraleti: mkdir -p $KAS_WORK_DIR

.github/workflows/build-yocto.yml
```diff
@@ -136,6 +136,17 @@ jobs:
...
+      # For PR, we might have new recipes not included in any image we build
+      # let's try to build any new or modified recipes, just in case.
+      # In most case it should be a quick no-op.
+      - name: Poor man's meta-qcom world build
+        if: github.event_name == 'pull_request'
+        run: |
+          export DL_DIR=${CACHE_DIR}/downloads
+          export SSTATE_DIR=${CACHE_DIR}/sstate-cache
+          export KAS_WORK_DIR=$PWD/../kas
+          kas shell ci/mirror.yml:ci/${{ matrix.machine }}.yml --command "bitbake $(echo $(for f in $(git diff --name-only ${{ github.event.pull_request.base.sha  }} -- *.{bb,bbappend}); do echo $(basename $f) | cut -d_ -f1 | cut -d. -f1; done))"
```
lumag, line 148: `echo $f | sed -e 's!.*\/\(.*\)\(_.*\)\?\.bb\(append\)\?!\2!g`? But as I wrote that I doubt that it is more readable...
Maybe something like `sed -e 's!.*/!!g' -e 's!\(_.*\)\.bb\(append\)\?!!g'` ?

lumag: `\(_.*\)\?` ?

ndechesne: yes. i used that. just pushed a new version.

lumag (dismissed):

ricardosalveti:
> wait wait.. this needs more testing/thoughts before we merge! e.g. if we try to build linux-qcom-staging (assuming a PR changes that file), we get:
> 
> ERROR: Nothing PROVIDES 'linux-qcom-staging' linux-qcom-staging was skipped: PREFERRED_PROVIDER_virtual/kernel set to linux-yocto-dev, not linux-qcom-staging

Yeah, recipes depending on provides won't work that easily. How is this managed with world?

ndechesne:
> How is this managed with world?

world is not a recipe/target. it's intercepted by bitbake and it uses the internal recipe/data store. it would be nice to have a 'world' target that only builds everything from a layer instead.. i am not sure how difficult that would be..

koenkooi:
> wait wait.. this needs more testing/thoughts before we merge! e.g. if we try to build linux-qcom-staging (assuming a PR changes that file), we get:
> 
> ERROR: Nothing PROVIDES 'linux-qcom-staging' linux-qcom-staging was skipped: PREFERRED_PROVIDER_virtual/kernel set to linux-yocto-dev, not linux-qcom-staging

If this was purely for populating sstate-cache, I'd suggest using `-k`, but for actually catching errors, I'm going to suggest filtering out `recipes-kernel/linux`.

lumag:
> > wait wait.. this needs more testing/thoughts before we merge! e.g. if we try to build linux-qcom-staging (assuming a PR changes that file), we get:
> > ERROR: Nothing PROVIDES 'linux-qcom-staging' linux-qcom-staging was skipped: PREFERRED_PROVIDER_virtual/kernel set to linux-yocto-dev, not linux-qcom-staging
> 
> If this was purely for populating sstate-cache, I'd suggest using `-k`, but for actually catching errors, I'm going to suggest filtering out `recipes-kernel/linux`.

Currently we also need to filter boot images and some other packages. If I have time, I will take a look at reworking those recipes to remove possible conflicts.

vkraleti:
> the qcom world build fails currently with
> 
> ```
> ERROR: qcom-ptool-0.0+git-r0 do_package_qa: QA Issue: /usr/bin/msp.py contained in package qcom-ptool requires /usr/bin/python, but no providers found in RDEPENDS:qcom-ptool? [file-rdeps]
> ERROR: qcom-ptool-0.0+git-r0 do_package_qa: Fatal QA errors were found, failing task.
> ```

https://github.com/qualcomm-linux/qcom-ptool/pull/30 should address this. Once merged I'll update qcom-ptool SHA.

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -37,3 +37,7 @@ remove_unused_files() {
 IMAGE_PREPROCESS_COMMAND:append = " remove_unused_files"
 
 do_uki[vardeps] += "KERNEL_CMDLINE_EXTRA"
+
+# ESP image is currently only used and tested on ARMv8 (aarch64) machines.
+COMPATIBLE_MACHINE = "^$"
+COMPATIBLE_MACHINE:aarch64 = "(.*)"
```
lumag: Would it be more idiomatic to use REQUIRED_MACHINE_FEATURES= "efi"?

ndechesne: that is a really good suggestion. just pushed this change.

ndechesne:
> For this we also need to inherits the features_check class

ah, thanks. so the current build should fail. I will push an update shortly

recipes-devtools/qca-swiss-army-knife/qca-swiss-army-knife_git.bb
```diff
@@ -12,6 +12,7 @@ SRC_URI = " \
 	file://ath10k-generate-pci-board-2_json.sh \
 	file://ath11k-generate-pci-board-2_json.sh \
 	file://ath11k-generate-ahb-board-2_json.sh \
+        file://0001-scripts-use-usr-bin-env-python3.patch \
```
ricardosalveti: nit, different indentation, tab x spaces.

ndechesne: fixed and updated

conf/machine/qcom-armv8a.conf
```diff
@@ -1,7 +1,7 @@
 require conf/machine/include/qcom-common.inc
 require conf/machine/include/arm/arch-armv8a.inc
 
-MACHINE_FEATURES += "screen ext2"
+MACHINE_FEATURES += "screen ext2 efi"
```
lumag, line 4: I had mixed feelings towards this, but in the end I think this is correct: some of machines on which we are running the qcom-armv8a might have UEFI.

lumag (approved):

ricardosalveti (approved):

PR #802, opened 2025-03-13
ci: base: enable pni-names in DISTRO_FEATURES

ricardosalveti:
Enable Predictable Network Interface Names by default in order to have a stable interface naming in our testing infrastructure.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -39,7 +39,7 @@ local_conf_header:
     IMAGE_CLASSES += "image_types_qcom"
     IMAGE_FSTYPES += "qcomflash"
   extra: |
-    DISTRO_FEATURES:append = " efi"
+    DISTRO_FEATURES:append = " efi pni-names"
     EXTRA_IMAGE_FEATURES = "allow-empty-password empty-root-password allow-root-login"
     IMAGE_ROOTFS_EXTRA_SPACE = "307200"
 


ricardosalveti:
@mwasilew this should enable predictable network interface naming in our ci runs and allow us to have a stable interface naming in our ci/cd.

ricardosalveti:
I proposed a similar change to meta-qcom-distro (https://github.com/qualcomm-linux/meta-qcom-distro/pull/21) but it is also important here as we need it enabled by default here as well, otherwise it will cause testings issues as we enable networking validation.

ci/base.yml
```diff
@@ -39,7 +39,7 @@ local_conf_header:
     IMAGE_CLASSES += "image_types_qcom"
     IMAGE_FSTYPES += "qcomflash"
   extra: |
-    DISTRO_FEATURES:append = " efi"
+    DISTRO_FEATURES:append = " efi pni-names"
```
ndechesne, line 42: hmm. are we not trying to hide a real problem here? Why is the kernel not giving a predictable interface?

mwasilew: I think the interface names in kernel depend on the order they're enumerated. This is why systemd does the renaming.

ndechesne: on my laptop, 'mac' policy is not added , and I have predictable names.. 
on which boards do you see non predictable names ?

mwasilew: RIDE SX is one example. Ethernet interface name jump between eth0 and eth1 after reboot. I mean interface with the same mac address can be named either eth0 or eth1 and there is no way to predict it.

mwasilew: I take this back. I think the problem is different. After reflashing, even with the same build, MAC address changes. Example:
https://lava.infra.foundries.io/scheduler/job/65791#L2161
https://lava.infra.foundries.io/scheduler/job/66111#L2154

mwasilew: Here is example with the same build on a different physical board. Reflashed between the jobs:
https://lava.infra.foundries.io/scheduler/job/66112#L2153
https://lava.infra.foundries.io/scheduler/job/65798#L2158

```
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether ee:f8:45:71:21:0c brd ff:ff:ff:ff:ff:ff
    inet 10.95.134.41/24 brd 10.95.134.255 scope global dynamic noprefixroute eth0
       valid_lft 1765sec preferred_lft 1765sec
    inet6 fe80::2a93:f5e6:46bc:9418/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever
3: eth1: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc mq state DOWN group default qlen 1000
    link/ether 32:87:d6:d5:59:b5 brd ff:ff:ff:ff:ff:ff
```

and

```
2: eth0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc mq state DOWN group default qlen 1000
    link/ether 8e:44:2b:28:70:35 brd ff:ff:ff:ff:ff:ff
3: eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether b6:6f:94:9a:12:ec brd ff:ff:ff:ff:ff:ff
    inet 10.95.134.43/24 brd 10.95.134.255 scope global dynamic noprefixroute eth1
       valid_lft 1765sec preferred_lft 1765sec
    inet6 fe80::da90:b63d:c593:2284/64 scope link noprefixroute 
       valid_lft forever preferred_lft forever
```

On this board only one interface is connected to the switch.

ndechesne: they are "locally administrated mac address" (bit 1 of the first byte is 1). That means they are set by the kernel, because there is no MAC address associated with them.

ricardosalveti: Not hiding any issue, lack of mac is a different problem, without pni-names everything gets disabled in systemd (https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-core/systemd/systemd_257.3.bb#L408), and the probe order from the kernel side can be done at random.

With pni-names it will rely on the device tree aliases when renaming the interfaces, making it stable between reboots.

ndechesne: hmm. i missed the 'else' statement that disables it completely.. so regardless of the MAC issue discussed above, we should merge this change..

ndechesne (approved):

quic-vkraleti (approved):

PR #803, opened 2025-03-16
Rework boot firmware handling

lumag:
Having boot firmware as machine-spefic recipes complicates testing (as one would have to filter those from CI builds lke #797) and prevents building rescue images with generic machine configs. Rework boot firmware, partiion configuration and qcomflash recipes in order to make that code work with any machine.


conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -17,3 +17,5 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-rb3gen2-firmware \
     packagegroup-rb3gen2-hexagon-dsp-binaries \
 "
+
+CDT_FILE = "cdt_core_kit"
```
koenkooi, line 21: Nitpick: should we call it QCOM_CDT_FILE to make it even more clear it is a Qualcomm specific thing?

conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -19,3 +19,4 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
 "
 
 CDT_FILE = "cdt_core_kit"
+QCOM_BOOT_FILES_SUBDIR = "qcm6490"
```
koenkooi, line 22: Would setting these variables to SOC_FAMILY work as well? That would save us from adding more and more stanzas for each new board/soc-family combination.

recipes-bsp/partition/qcom-partition-confs_1.0.bb
```diff
@@ -23,24 +21,28 @@ inherit python3native deploy
...
+    install -m 0644 ${B}/patch*.xml -D ${DEPLOYDIR}
+    install -m 0644 ${B}/rawprogram*.xml -D ${DEPLOYDIR}
+    install -m 0644 ${B}/zeros_*.bin -D ${DEPLOYDIR}
+    install -m 0644 ${B}/wipe_rawprogram_PHY*.xml -D ${DEPLOYDIR}
 }
-addtask deploy after do_install
+addtask deploy before do_build after do_compile
 
 COMPATIBLE_MACHINE = "(qcm6490|qcs9100|qcm2290)"
-PACKAGE_ARCH = "${MACHINE_ARCH}"
```
ndechesne, line 59: why? it is really machine (or soc) specific, no?

lumag: There are no packages anymore, so they are not machine-specific

conf/machine/qcs6490-rb3gen2-core-kit.conf
lumag: Well... Yes and no. Currently it works this way. However consider adding support for [Tachyon](https://www.kickstarter.com/projects/particle-iot/tachyon-powerful-5g-single-board-computer-w-ai-accelerator/description). It might share some bits with existing qcs6490/qcm6490 products, but at the same time it will likely use different partition schema or different bootloaders.
So, I'd prefer to keep those out of SoC configs.
It might be a good idea to switch to macg\hine names completely rather than using SoC-named subdirs. WDYT?

lumag:
> I am not sure about this PR.. I would like to better understand the goal here. I find there is too much hardcoded string everywhere.

Well, there were two main goals:
- to be able to test recipes in a generic way (otherwise all those -boot recipes will stand out in your #797, as they can not be compiled on any machine)
- to be able to generate and observe data files for multiple targets without having to recompile everything (or without having a pile of almost-equivalent build dirs)

Longer term goals:
- to remove a need to create a sparate machine for each board variant. E.g. currently there is no way to hadle core vs vision kits because `CDT_FILE` is handled as a machine-speific file way down the pipeline
- to be able to build machine-speific images with a generic machine config.

> I believe we want to move the partition data, XML, .. into https://github.com/qualcomm-linux/qcom-ptool eventually. I would like to check how this would impact all of that first.

I don't see significant difference here: I'd still deploy all partition data. anoter option might be to create per-board recipes deploying data in a board-specific location. Maybe I should already rework `qcom-partitions-conf` in this way to prevent us from getting extra/unwanted data in deploy dir

lumag:
@ndechesne reworked partition configuration

lumag:
@ndechesne I'd really like to be able to get rid of machine-specific packages and `COMPATIBLE_MACHINES`  here. Would you suggest any way forward?

recipes-bsp/partition/qcom-partition-confs_1.0.bb
ndechesne: allarch then?

recipes-bsp/partition/qcom-partition-confs_1.0.bb
```diff
@@ -23,6 +23,7 @@ do_configure[noexec] = "1"
 do_install[noexec] = "1"
 
 QCOM_PARTCONF_SUBDIR ?= ""
+QCOM_PARTCONF_SUBDIR:qcm6490 ?= "qcm6490"
```
ndechesne: QCOM_PARTCONF_SUBDIR, QCOM_BOOT_IMG_SUBDIR and QCOM_BOOT_FILES_SUBDIR have to match, right? Should we not define a single variable in a single place?

ndechesne: I take that back..

conf/machine/qrb2210-rb1-core-kit.conf
```diff
@@ -23,3 +23,4 @@ KERNEL_CLASSES += "linux-qcom-bootimg"
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += "qbootctl"
 
 QCOM_BOOT_FILES_SUBDIR = "qrb2210"
+QCOM_PARTITION_CONF = "qcom-partition-confs"
```
ndechesne: should we use a virtual/qcom-partition-confs instead? and let BSP layer set it accordingly?

ndechesne (commented):
I don't have any strong objection (anymore). I can see how it improves things. And we can clean up more things after this one..

conf/machine/qrb2210-rb1-core-kit.conf
lumag: Using `virtual/foo` would make other providers for the same name to be skipped once one is selected. Also it's not like they are providing exactly the same configuration.

recipes-bsp/partition/qcom-partition-confs_1.0.bb
lumag: The split recipes become allarch, yes. Maybe I should drop this chunk.

lumag: dropped

conf/machine/qrb2210-rb1-core-kit.conf
ndechesne: indeed. i realized that after i made the comment.

ndechesne (approved):

koenkooi (approved):
Looks good to me, the only change I want to suggest is a nitpick: reword "qcom-parition-confs: split to a per-machine recipes" to fix the typo in 'par**t**ition' and drop either the 'a' or make 'recipes' singular.

PR #804, opened 2025-03-17
Build add-on DTBs with linux-qcom-staging kernel 

vkraleti:
The linux-qcom-staging kernel can produce an additional set of DTB files called 'addons DTB' files with downstream additions. Build them along with upstream DTB files to allow users to use downstream features.

### DIFF
--- a/conf/machine/qcm6490-idp.conf
+++ b/conf/machine/qcm6490-idp.conf
@@ -14,3 +14,14 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-rb3gen2-firmware \
     packagegroup-rb3gen2-hexagon-dsp-binaries \
 "
+
+#
+# Configuration variables for linux-qcom-staging kernel
+#
+
+# These addon DTs are not yet upstreamed. Build them only in staging kernel
+# to avoid conflicts with other kernel providers like linux-yocto-dev.
+KERNEL_DEVICETREE:append:pn-linux-qcom-staging = " \
+                      qcom/qcm6490-addons-idp.dtb \
+                      qcom/qcm6490-addons-idp-amoled.dtb \
+                      "
--- a/conf/machine/qcs6490-rb3gen2-core-kit.conf
+++ b/conf/machine/qcs6490-rb3gen2-core-kit.conf
@@ -21,3 +21,14 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
 QCOM_CDT_FILE = "cdt_core_kit"
 QCOM_BOOT_FILES_SUBDIR = "qcm6490"
 QCOM_PARTITION_CONF = "qcom-partition-conf-qcs6490-rb3gen2"
+
+#
+# Configuration variables for linux-qcom-staging kernel
+#
+
+# These addon DTs are not yet upstreamed. Build them only in staging kernel
+# to avoid conflicts with other kernel providers like linux-yocto-dev.
+KERNEL_DEVICETREE:append:pn-linux-qcom-staging = " \
+                      qcom/qcs6490-addons-rb3gen2.dtb \
+                      qcom/qcs6490-addons-rb3gen2-vision-mezz.dtb \
+                      "
--- a/conf/machine/qcs9100-ride-sx.conf
+++ b/conf/machine/qcs9100-ride-sx.conf
@@ -23,3 +23,16 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
 QCOM_CDT_FILE = "cdt_ride_sx"
 QCOM_BOOT_FILES_SUBDIR = "qcs9100"
 QCOM_PARTITION_CONF = "qcom-partition-conf-qcs9100-ride-sx"
+
+#
+# Configuration variables for linux-qcom-staging kernel
+#
+
+# These addon DTs are not yet upstreamed. Build them only in staging kernel
+# to avoid conflicts with other kernel providers like linux-yocto-dev.
+KERNEL_DEVICETREE:append:pn-linux-qcom-staging = " \
+                      qcom/qcs9100-addons-ride.dtb \
+                      qcom/qcs9100-addons-ride-r3.dtb \
+                      qcom/sa8775p-addons-ride.dtb \
+                      qcom/sa8775p-addons-ride-r3.dtb \
+                      "


recipes-multimedia/video/qcom-videodtb_1.0.bb
```diff
@@ -0,0 +1,37 @@
+inherit module
+
+DESCRIPTION = "QCOM Video device-tree overlays"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/${LICENSE};md5=550794465ba0ec5312d6919e203a55f9"
+
+SRCREV  = "52d2fbdaf53ebc13d559f85bcb15df4bdc7a06c5"
+SRC_URI = "git://git.codelinaro.org/clo/le/platform/vendor/opensource/video-devicetree.git;protocol=https;branch=video.qclinux.0.0"
```
lumag, line 9: What is the timeline for merging those into the main repo?
Judging by `qcm6490-video.dtsi` there are two kinds of changes:
- use of interconnect-names. This is unnecessary and must be fixed in the driver
- use of additional / non-standard IOMMU declartions. This needs to be aligned with the upstream schema.

recipes-multimedia/video/qcom-videodtb_1.0.bb
```diff
@@ -0,0 +1,37 @@
...
+
+S = "${WORKDIR}/git"
+
+DTBO_TARGET = ""
+DTBO_TARGET:qcm6490 = "qcm6490-video"
+DTBO_TARGET:qcs9100 = "sa8775p-video"
```
lumag, line 15: No. Please don't add extra machine-specific defines. See #803

recipes-kernel/linux/linux-qcom-staging_6.12.bb
```diff
@@ -29,6 +29,9 @@ KERNEL_CONFIG ?= "qcom_defconfig"
 # Additional fragment for qcom value add features
 KERNEL_CONFIG_FRAGMENTS += " ${S}/arch/arm64/configs/qcom_addons.config"
 
+# Add '-@' option to compile DTB with symbols.
+KERNEL_DTC_FLAGS += "-@"
```
lumag, line 33: How does overlaying of in-kernel DTBOs work without this flag?

lumag (changes requested):

recipes-multimedia/video/qcom-videodtb_1.0.bb
Vikash-Garodia: >use of interconnect-names. This is unnecessary and must be fixed in the driver

Thats right. Those should be fixed in driver.

>use of additional / non-standard IOMMU declarations. This needs to be aligned with the upstream schema.

Yes, Schema should be updated with additional (optional) sub-node

Vikash-Garodia: >What is the timeline for merging those into the main repo?
Timelines wise, the plan is to enable the remaining codecs and then add the binding + driver change to support additional subnodes having IOMMUs. This would come around Nov,2025

lumag: Could you possibly shift priorities and work on schema first? I'd say, we don't even need driver backing, just the ability to describe the hardware properly. Then we should be able to drop video-devicetree from the addons, etc.

Vikash-Garodia: For qcm6490, trying an experiment to absorb the IOMMU sub node in driver, while for SA8775p and QCS8300, need to define the sub node in bindings. Let me come back with an earlier timelines and result for qcm6490 experiment.

lumag: No other devices use sub-nodes for IOMMU if I remember correctly. At least the NS IOMMU should be a part of the node itself (otherwise current Iris driver will not work with it).

lumag: Anyway, this is a separate topic, not related to this PR.

conf/machine/qcm6490-idp.conf
```diff
@@ -14,3 +14,14 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-rb3gen2-firmware \
     packagegroup-rb3gen2-hexagon-dsp-binaries \
 "
+
```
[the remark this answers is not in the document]
quic-vkraleti: Kernel team agreed to drop addon DTBs when hosting a kernel branch on github.com/qualcomm-linux, which is expected to happen around the end of May 2025 or early June 2025.

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -4,9 +4,80 @@
...
+
+        # Copy latest overlayed dtb file with original name
+        output = dtb + "." + str(dtbos_found)
+        shutil.copy2(os.path.join(dtbo_dir, output), os.path.join(dtbo_dir, dtb))
+}
+
 DTBBIN_DEPLOYDIR = "${WORKDIR}/qcom_dtbbin_deploy-${PN}"
 DTBBIN_SIZE ?= "4096"
 
+do_qcom_dtbbin_deploy[prefuncs] += "qcom_dtbmerge"
```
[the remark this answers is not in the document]
quic-vkraleti: Separating out-of-tree (OOT) DTBO merging and DTB.bin generation tasks using 'prefuncs' helps make this class more readable. This also allows for conditional bypassing of DTBO merging, adding flexibility to the workflow. If this doesn't look convincing, I can make these changes into do_qcom_dtbbin_deploy itself.

recipes-multimedia/video/qcom-videodtb_1.0.bb
quic-vkraleti: Sure, will generate all DTBOs without any conditions.

recipes-kernel/linux/linux-qcom-staging_6.12.bb
quic-vkraleti: As I understand, this flag is needed even for in-tree DTBOs to work. A few vendors submitted patches upstream to add this flag into makefiles.
https://patchwork.kernel.org/project/linux-arm-kernel/patch/20230530175624.2360218-2-u.kleine-koenig@pengutronix.de/ 

However, QCOM doesn't have any in-tree DTBOs to date, so this wasn't added. This is probably a potential upstream change to pursue.

conf/machine/qcm6490-idp.conf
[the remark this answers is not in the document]
lumag: Ack, please add that to the commit message. But maybe it's also possible to merge those -addons DTB parts in to the main DTB files inside the current Linux kernel repo too?

classes/linux-qcom-dtbbin.bbclass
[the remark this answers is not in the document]
lumag: do_deploy should be limited to deploying. If you are to combine the FDTs, please do so during do_compile state.

recipes-kernel/linux/linux-qcom-staging_6.12.bb
lumag: > However, QCOM doesn't have any in-tree DTBOs to date

This is not true.

```
$ git grep dtbo arch/arm64/boot/dts/qcom/Makefile 
arch/arm64/boot/dts/qcom/Makefile:apq8016-sbc-usb-host-dtbs     := apq8016-sbc.dtb apq8016-sbc-usb-host.dtbo
arch/arm64/boot/dts/qcom/Makefile:qcs6490-rb3gen2-vision-mezzanine-dtbs := qcs6490-rb3gen2.dtb qcs6490-rb3gen2-vision-mezzanine.dtbo
arch/arm64/boot/dts/qcom/Makefile:qrb5165-rb5-vision-mezzanine-dtbs     := qrb5165-rb5.dtb qrb5165-rb5-vision-mezzanine.dtbo
arch/arm64/boot/dts/qcom/Makefile:sdm845-db845c-navigation-mezzanine-dtbs       := sdm845-db845c.dtb sdm845-db845c-navigation-mezzanine.dtbo
arch/arm64/boot/dts/qcom/Makefile:sm8650-hdk-display-card-dtbs  := sm8650-hdk.dtb sm8650-hdk-display-card.dtbo
```

classes/linux-qcom-dtbbin.bbclass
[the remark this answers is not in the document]
quic-vkraleti: `do_qcom_dtbbin_deploy` is not a regular deploy task. This custom task is introduced in the kernel build workflow to convert the generated DTBs into VFAT format for flashing into the DTB partition. With this PR it is further being extended to merge DTBOs and then convert them to VFAT.

Merging OOT DTBOs can't happen in kernel `do_compile`, as generation of DTBOs in itself need dtc tool and headers from kernel work-shared directory which will be populated in `do_shared_workdir` task that runs after `do_compile`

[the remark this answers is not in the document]
lumag: So, it should be a task executed before do_deploy, but after do_shared_workdir.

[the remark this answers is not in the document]
quic-vkraleti: Yes, this is the current task order.
`addtask qcom_dtbbin_deploy after do_populate_sysroot do_packagedata before do_deploy`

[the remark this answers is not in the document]
lumag: This is correct for the qcom_dtbbin_deploy task. However I'm asking you to add new task that generates merged images instead of using prefunc. I want to be able to run `bitbake -c apply-dtbos virtual/kernel` and inspect generated DTB images. Also with your approach generated files do not end up in the `${KERNEL_PACKAGE_NAME}-devicetree`, which is incorrect.

TL;DR. Don't use `prefunc`, build the appended DTBs closer to the `do_compile` task and make sure that they get into the `kernel-devicetree` package.

conf/machine/qcm6490-idp.conf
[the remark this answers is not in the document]
ricardosalveti: Probably only when they start using a compatible dt schema.

@quic-vkraleti even when they drop from the kernel branch, having downstream using an upstream schema and overlay will probably take longer, no?

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -4,9 +4,81 @@
...
+# That is, the DTBO file that depends on another DTBO file should be listed
+# later in the list. For example, if my-dtbo1.dtbo depends on my-dtbo2.dtbo,
+# they should be mentioned as:
+# KERNEL_DTB_OVERLAYS[kernel-dt] = "my-dtbo2.dtbo my-dtbo1.dtbo"
+#
+# Recipes generating out-of-tree DTBOs must be listed in KERNEL_DTBO_PROVIDERS
+# for BitBake to set the correct task dependency order and ensure DTBOs are built
+# and available for overlaying. For example, if the 'my-dtb.bb' recipe generates
+# 'my-dt-overlay.dtbo', it should be mentioned as:
+# KERNEL_DTBO_PROVIDERS = "my-dtb"
```
ricardosalveti, line 22: I wonder if we really want to create such specifics here to handle something that should really be generic, in the end what we want is an external device-tree provider (full dtb or overlays), and oe-core defines PREFERRED_PROVIDER_virtual/dtb for that.

The main difference here is that we expect multiple recipes providing multiple overlays in the end, which kind of just complicates everything, since the final solution will be to have all provided by the kernel (single recipe). We can always have out of tree dtbs, but I wonder if it makes sense to have one repo containing all the external dtbos we have.

classes/linux-qcom-dtbbin.bbclass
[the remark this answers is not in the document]
ricardosalveti: We do have https://github.com/openembedded/openembedded-core/blob/master/meta/classes-recipe/devicetree.bbclass for building out of tree device tree files (also used by https://github.com/Xilinx/meta-xilinx/blob/master/meta-xilinx-core/recipes-bsp/device-tree/device-tree.bb) which will do the right thing at the right order, can't use use it here?

recipes-multimedia/video/qcom-videodtb_1.0.bb
```diff
@@ -0,0 +1,37 @@
+inherit module
```
ricardosalveti, line 1: Why is this needed since we are just installing the overlay?

ricardosalveti: This would align better with devicetree.bbclass.

recipes-kernel/linux/linux-qcom-staging_6.12.bb
ricardosalveti: +1, might not be needed with current mainline.

classes/linux-qcom-dtbbin.bbclass
sbanerjee-quic: @ricardosalveti useful suggestion, I think we should take this as a future improvement?

lumag: I'd prefer to have all extra DTBOs to be a part of the linux-staging repo.

recipes-multimedia/video/qcom-videodtb_1.0.bb
```diff
@@ -0,0 +1,37 @@
...
+
+DESCRIPTION = "QCOM Video device-tree overlays"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/${LICENSE};md5=550794465ba0ec5312d6919e203a55f9"
```
lumag, line 6: This should be referencing a file in the sources, not the external licence.

recipes-multimedia/video/qcom-videodtb_1.0.bb
```diff
@@ -0,0 +1,37 @@
...
+    install -m 0644 ${S}/*.dtbo -D ${D}/devicetree
+}
+
+PACKAGES = "${PN}"
+FILES:${PN} = "/devicetree/*"
+
+inherit deploy
+do_deploy() {
+    install -d ${DEPLOYDIR}/devicetree
+    install -m 0644 ${D}/devicetree/*.dtbo -D ${DEPLOYDIR}/devicetree
```
lumag, line 35: Do we need dtbo's in DEPLOYDIR? Can't the fdtoverlay pick those from stagingdir?

lumag: OTOH, it's probably fine.

classes/linux-qcom-dtbbin.bbclass
ricardosalveti: Right, if we plan on using the extra DTBOs with multiple kernel trees than I would prefer to keep them out of the tree (and built via a recipe using devicetree.bbclass), but if we are really only expecting to use them with one specific kernel tree (which seems to be the current scenario), then it is indeed better to just make them part of the kernel repository.

lumag: I'd say, this is close to the main 'DTB are a part of the kernel' vs ' DTB should be separated' question. I think, that even if we were to serve multiple kernels, the DTBOs are better to stay together with the rest of the DTBs, even if that creates extra work. In several months (hopefully) there will be no extra DTBOs, everything will be served by the single DTB. So we might as well start getting used to that earlier than later.

lumag:
@quic-vkraleti please `git rebase` rather than `git merge`.

quic-vkraleti:
> @quic-vkraleti please `git rebase` rather than `git merge`.

Done.

lumag (approved):

PR #807, opened 2025-03-18
Fix qcomflash dir for RB1 board

### LINKED ISSUE #806: missing files in rawprogram0.xml for RB1
When trying to use generated qcomflash tarball for RB1, the following files are not included:
 - not built: 
```
boot-erase.img
persist.img
logfs_ufs_8mb.bin
```
 - missing from tarball:
```
NON-HLOS.bin
dspso.bin
BTFM.bin
```

lumag:
As noted in #806 the qcomflash dir for the RB1 board misses several files, which can cause errors during flashing.
- Install boot image into the qcomflash dir and use it for `boot_a`/`boot_b` partitions
- Don't flash `persist` and `logfs`, they are unused by Linux
- Don't flash firmware and DSP partitions, they are unused with meta-qcom

Closes #806

### DIFF
--- a/classes-recipe/image_types_qcom.bbclass
+++ b/classes-recipe/image_types_qcom.bbclass
@@ -43,6 +43,19 @@ create_qcomflash_pkg() {
         done
     fi
 
+    # Legacy boot images
+    if [ -n "${QCOM_DTB_DEFAULT}" ]; then
+        [ -e "${DEPLOY_DIR_IMAGE}/boot-initramfs-${QCOM_DTB_DEFAULT}-${MACHINE}.img" -a \
+            ! -e "boot.img" ] && \
+            install -m 0644 "${DEPLOY_DIR_IMAGE}/boot-initramfs-${QCOM_DTB_DEFAULT}-${MACHINE}.img" boot.img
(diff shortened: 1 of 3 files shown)


ricardosalveti (approved):

ndechesne (dismissed):
Looks pretty good, thanks! I was hoping we wouldn't need to do and get some EFI support soon.. but I don't think we will get it soon!

Can you confirm that you've tested the PR? Or should we ask someone to test before we merge?

lumag:
> Can you confirm that you've tested the PR? Or should we ask someone to test before we merge?

Only build-tested for now (still setting up the hardware).
@mwasilew could you possibly test it?

mwasilew:
Still only boots to fastboot (I think this is expected) but at least there are no files that qdl reports as missing. LGTM

ndechesne:
No, it should boot a kernel, and fails to mount the root file system. @lumag and @koenkooi have an RB1, perhaps it's better to test locally with them first.

lumag:
Just for the reference: this has been boot-tested on the actual RB1 device.

lumag:
@mwasilew I've pushed the fix for the fastboot problem.

mwasilew:
I'm able to boot the board after flashing with QDL.

ndechesne (approved):

ricardosalveti (approved):

PR #809, opened 2025-03-19
RB1: add platform-specific modules

lumag:
This is a temporal solution until @koenkooi finishes module packagegroups to let other developers test recent kernels on this platform. I do not intend for this to be merged, so marking it as draft.


ndechesne:
typo in $commit of #1 (packagegrup-rb1)

ndechesne:
we have this problem on all boards. can we instead start a qcom-modules which is slighly more generic and at least works on RB3 as well, since our focus is on RB1/3 right now?

ndechesne:
ah, ok, i saw it was temporary.. i thought you wanted to merge it and revert later. adding "kernel-modules" is what I am doing..

lumag:
Maybe I should have done that too.

lumag:
Closing in favour of new package groups.

PR #810, opened 2025-03-20
mesa: build patched 25.0 version for QCM2290

lumag:
Mesa in OE-Core is stuck with the version 24.0.x, since newer Mesa versions require libclc in order to build Intel drivers, and libclc in not available in OE-Core.

The QRB2210 / QCM2290 SoC uses A702, which is not currently supported by Mesa, support is provided via an external patchset. As such it doesn't make sense to backport patches to the 24.0.x branch.

In order to be able to use GPU acceleration on RB1, upgrade mesa to 25.x only for this platform.


lumag:
Yes, that's why it's marked as draft.

sbanerjee-quic:
> Well, NACK. as discussed privately, we should either backport A702 for 24.x or bring 25.x in oe-core. Our BSP layer is not the right place to manage such large version upgrade..

+1

PR #812, opened 2025-03-25
ci: add qrb2210-rb1-core-kit boot test job

mwasilew:
Add LAVA test job template for qrb2210-rb1-core-kit. The jobs is the same as for other existing boards.


mwasilew:
Example job from one of the previous builds: https://lava.infra.foundries.io/scheduler/job/67351

ricardosalveti (approved):

Issue #712, opened 2024-11-21
CI: Update KAS mirrors


ndechesne:
We added support for KAS mirrors in f916c17, and thought that the mirrors would be updated with new objects (when using the --update command). As it happens this is not correct. KAS will effectively create the mirror repo when/if it does not exist, however it will never update the mirrors.

When running a checkout command with --update, KAS will first clone the repo from the mirror (which contains the git objects from the time the mirror was created), and then it fetches new objects from the remote repo. However the new objects are fetched in the kas workspace, but not in the mirror. Next time we run checkout with --update again, the same new objects are fetched again. 

At first, I thought it was a bug, but it happens to be done 'by design' as discussed on the kas mailing list here:
https://groups.google.com/g/kas-devel/c/-TtiBdDlhJ0

We probably want to update the kas mirrors regardless, e.g. running git fetch command from time to time. That's what meta-arm is doing at https://git.yoctoproject.org/meta-arm/tree/ci/update-repos.

PR #819, opened 2025-03-31
ci: pre-fetch KAS_REPO_REF_DIR with just git

quaresmajose:
We initially used KAS_REPO_REF_DIR to keep a local copy of the repos mirrors,
to avoid fetching projects from scratch in CI. However, it is implemented such
that the mirrors, once created by kas the first time, are never updated by kas,
they are 'read only'.
    
Sometimes the fetch from upstream fails and the KAS ended up falling back
to the old content of the mirrors, silently. To fix this we will pre-fetch
all ref front the origin the repos in KAS_REPO_REF_DIR.
    
We will for each repo retry 3 times sleeping 10 seconds between attempts.

Fixes https://github.com/qualcomm-linux/meta-qcom/issues/712

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -13,7 +13,19 @@ env:
   BASE_ARTIFACT_URL: "https://quic-yocto-fileserver-1029608027416.us-central1.run.app/${{ github.run_id }}"
 
 jobs:
+  kas-mirror:
+    if: github.repository == 'qualcomm-linux/meta-qcom'
+    runs-on: [self-hosted, x86]
+    steps:
+      - name: Update kas mirrors
+        run: |
+          for r in $(find ${KAS_REPO_REF_DIR}/* -maxdepth 0 -type d); do
+            echo "pre-fetch: $r"
+            git -C $r fetch --prune origin '+refs/*:refs/*'
+          done
+
   kas-lock:
+    needs:  kas-mirror
     if: github.repository == 'qualcomm-linux/meta-qcom'
     runs-on: [self-hosted, x86]
     steps:


.github/workflows/build-yocto.yml
```diff
@@ -23,6 +23,13 @@ jobs:
 
       - name: Run kas lock
         run: |
+          for r in $(find ${KAS_REPO_REF_DIR}/* -maxdepth 0 -type d); do
+            echo "pre-fetch: $r"
+            # transform the repo in mirror
+            git -C $r config remote.origin.fetch '+refs/*:refs/*'
+            git -C $r config remote.origin.mirror true
```
ndechesne, line 30: kas does not set mirror=true when it builds the tree. why is that needed?

.github/workflows/build-yocto.yml
```diff
@@ -23,6 +23,13 @@ jobs:
...
+            git -C $r fetch --prune
```
ndechesne: running locally, I can see that 
`git fetch --prune origin '+refs/*:refs/*'
`
is enough to update all refs. e.g. a one-liner.

.github/workflows/build-yocto.yml
```diff
@@ -23,6 +23,19 @@ jobs:
 
       - name: Run kas lock
         run: |
+          for r in $(find ${KAS_REPO_REF_DIR}/* -maxdepth 0 -type d); do
```
ndechesne, line 26: also, we should move the mirrors update in its own job, not under kas lock job.

.github/workflows/build-yocto.yml
quaresmajose: I tried that because get it configured as a mirror seems more appropriate for what we have

.github/workflows/build-yocto.yml
quaresmajose: If only this line serves us, even better

.github/workflows/build-yocto.yml
quaresmajose: but if we move to the other job we will continue to have the inconsistencies we had before:
the kas fails in some of the repo and starts using mirror which may not be aligned with all the others repos.

ndechesne: move as a dependency. adds a needs statement in kas-lock

quaresmajose: yes that makes sense

.github/workflows/build-yocto.yml
ndechesne: git clone --mirror does quite a few things. I don't think setting mirror to true is really equivalent to doing git clone with --mirror. 

In fact, if kas was doing a git clone with --mirror, we would not have these issues. since we could run git remote update --prune and it would work. With --mirror, +refs/*:refs/* would be set automatically.. So perhaps we should propose this change to kas (or understand why it's not done like that already)

quaresmajose: Calling just `git fetch --prune origin '+refs/*:refs/*'  should do mostly the same and we can drop the "mirror true" and be more aligned with kas.
I think propose the kas clone with --mirror it has to be somehow conditional because most users don't need it.

.github/workflows/build-yocto.yml
quaresmajose: done

.github/workflows/build-yocto.yml
quaresmajose: done

.github/workflows/build-yocto.yml
```diff
@@ -13,7 +13,25 @@ env:
...
+  kas-mirror:
+    if: github.repository == 'qualcomm-linux/meta-qcom'
+    runs-on: [self-hosted, x86]
+    steps:
+      - name: Update kas mirrors
+        run: |
+          for r in $(find ${KAS_REPO_REF_DIR}/* -maxdepth 0 -type d); do
+            echo "pre-fetch: $r"
+            for i in $(seq 3); do
+              if git -C $r fetch --prune origin '+refs/*:refs/*'; then
```
ndechesne, line 25: if it fails 3 times, the for loop will still exit with a success exit code, and the job won't fail, no? I think it's ok to run git just once and let the job fail, if it ever happens.

quaresmajose: Nice catch! yeah I can remove the loop and keep it simple.  When the git fails we can manually relaunch only that job.

ndechesne (approved):

PR #820, opened 2025-03-31
ci: build-yocto: cleanup sstate-cache after the build

### LINKED ISSUE #715: ci: add job to cleanup the unused local sstate-cache artifacts
We need to clean up the ci local sstate-cache that is not used, otherwise it will grow infinitely.
This will take up significant space that will not be used by anyone since invalid sstate-cache artefacts will no longer be needed.

This can be done by adding a new job that runs after all builds are successful and we can do this using the upstream script in oe-core or call the rsync directly.

quaresmajose:
We need to clean up the ci local sstate-cache that is not used, otherwise it will grow infinitely.
This will take up significant space that will not be used by anyone since invalid sstate-cache
artefacts will no longer be needed.

Fixes https://github.com/qualcomm-linux/meta-qcom/issues/715

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -123,41 +123,6 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Download kas lockfile
-        uses: actions/download-artifact@v7
-        with:
-          name: kas-lockfile
-          path: ci
-
-      - name: Download kas-container
-        uses: actions/download-artifact@v7
-        with:
-          name: kas-container
-          path: ${{ runner.temp }}
-
-      - name: Setting up kas-container
-        run: |
-          KAS_CONTAINER=$RUNNER_TEMP/kas-container
-          echo "KAS_CONTAINER=$KAS_CONTAINER" >> $GITHUB_ENV
-          chmod +x $KAS_CONTAINER
-
-      - name: Setup cache variables (used by oe-selftest)
-        run: |
-          echo "DL_DIR=${CACHE_DIR}/downloads" >> $GITHUB_ENV
-          echo "SSTATE_DIR=${CACHE_DIR}/sstate-cache-$(date '+%Y-%m')" >> $GITHUB_ENV
-
-      - name: Run Yocto patchreview
-        run: |
-          ci/kas-container-shell-helper.sh ci/yocto-patchreview.sh
-
-      - name: Run yocto-check-layer
-        run: |
-          ci/kas-container-shell-helper.sh ci/yocto-check-layer.sh
-
-      - name: Run oe-selftest
-        run: |
-          ci/kas-container-shell-helper.sh ci/oe-selftest.sh
-
   compile-env:
     if: github.repository_owner == 'qualcomm-linux'
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


0001-sstate-cache-management-dry-run.patch
```diff
@@ -0,0 +1,34 @@
+From f28a08f27ca26579f2f5fa56ae52365fb16b8a32 Mon Sep 17 00:00:00 2001
+From: Jose Quaresma <jose.quaresma@foundries.io>
+Date: Mon, 31 Mar 2025 11:15:12 +0100
+Subject: [PATCH] sstate-cache-management: dry-run
```
lumag, line 4: It might be a good idea to add an actual `--dry-run` to the script.

.github/workflows/build-yocto.yml
```diff
@@ -117,6 +117,13 @@ jobs:
           mkdir $KAS_WORK_DIR
           kas build ci/mirror.yml:ci/${{ matrix.machine }}.yml
 
+      - name: Sstate-cache cleanup
+        run: |
+          export SSTATE_DIR=${CACHE_DIR}/sstate-cache
+          export KAS_WORK_DIR=$PWD/../kas
+          SSTATE_CACHE_MANAGEMENT="${KAS_WORK_DIR}/poky/scripts/sstate-cache-management.py --cache-dir ${SSTATE_DIR} --remove-duplicated --remove-orphans --debug --yes"
+          kas shell ci/${{ matrix.machine }}.yml --command "$SSTATE_CACHE_MANAGEMENT"
```
doanac, line 125: not sure how long this script will take. My past recollection makes me think this will be pretty slow. Have you considered doing this on a cron schedule instead so that it doesn't affect day-to-day stuff as much?

ndechesne: Yocto docs about sstate-cache-management says 

> It has the ability to purge all but the newest cache files on each architecture, and also to remove files that it considers unreachable by exploring a set of build configurations. However, this command requires a full build environment to be available and doesn’t work well covering multiple releases.

Which is not very encouraging to be honest..

quaresmajose: for this first run it takes about 15 minutes for each machine but note that we have a very big sstate folder since it has never been cleaned yet. I expect to get much less time after this first interaction

ndechesne:
We should run it on push event only, not PR. First it would run less often, and then we really care about sstate for metadata which is merged, not what's in a PR which is temporary.

quaresmajose:
> We should run it on push event only, not PR. First it would run less often, and then we really care about sstate for metadata which is merged, not what's in a PR which is temporary.

God point! I have changed it to just run on git push events.

0001-sstate-cache-management-dry-run.patch
quaresmajose: Patch sent https://lore.kernel.org/openembedded-core/20250401165912.1603-1-jose.quaresma@foundries.io/T/#u

quaresmajose:
> When I looked at clean-sstate-cache a few years ago, it would remove everything unrelated to your existing build, so it would delete sstate for all other machines. Does it still do this or has it become smarter?

I believe the script is smart enough to support builds with more than one machine but I'm not sure it will be like that. I will need to test this locally on my machine to actually see what it deletes.

lumag:
> We should run it on push event only, not PR. First it would run less often, and then we really care about sstate for metadata which is merged, not what's in a PR which is temporary.

I'd go for a scheduled cleanup, e.g. once a week or so.

ndechesne:
that would indeed work better. even once a month.. but that also means we need to duplicate the build config/machines in that cleanup job, since we need to inherit the same matrix config, right?

lumag:
Can't we define machines in a common config file which gets included by both workflows?

PR #822, opened 2025-04-01
ci/base.yml: switch to github.com/openembedded mirrors where possible

koenkooi:
The Yocto Project infrastructure has been mostly unreachable for the past week, so switch to more reliable mirrors where possible.

Since meta-yocto is, well, a Yocto Project project, this isn't mirrored by OE.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -13,12 +13,12 @@ repos:
   meta-qcom:
 
   oe-core:
-    url: https://git.openembedded.org/openembedded-core
+    url: https://github.com/openembedded/openembedded-core
     layers:
       meta:
 
   bitbake:
-    url: https://git.openembedded.org/bitbake
+    url: https://github.com/openembedded/bitbake
     layers:
       .: excluded
 


ndechesne:
hmm. but if meta-yocto is not mirrored, then we aren't fixing our issues, then? we will still suffer from infra issues. should we create our own mirrors perhaps?

lumag:
It's not a meta-yocto, of course, but... poky is [mirrored](https://github.com/yoctoproject/poky)

koenkooi:
> hmm. but if meta-yocto is not mirrored, then we aren't fixing our issues, then? we will still suffer from infra issues. should we create our own mirrors perhaps?

Or, you know, stop using it by switching to `nodistro`?

lumag:
Well... Found a good enough [mirror](https://github.com/distro-core-curated-mirrors/meta-yocto/commits/master) for meta-yocto.

quaresmajose:
for let all know we have a foundries mirror at https://github.com/lmp-mirrors

quaresmajose:
> for let all know we have a foundries mirror at https://github.com/lmp-mirrors

but it is good to keep in mind that synchronization is scheduled and done hourly.

ci/base.yml
```diff
@@ -13,17 +13,17 @@ repos:
...
 
   bitbake:
-    url: https://git.openembedded.org/bitbake
+    url: https://github.com/openembedded/bitbake
     layers:
       .: excluded
 
   meta-yocto:
-    url: https://git.yoctoproject.org/meta-yocto
+    url: https://github.com/distro-core-curated-mirrors/meta-yocto
```
ndechesne, line 26: I am not in favor of using random/untrusted mirrors in the default QCOM config. 

I think I would be ok with any of these 3 options:
1. if we want to use GH YP official repo, we use poky.git and only configure meta-poky from it. I think it 'should' work.
2. we keep the upstream repo in base.yml and we setup our CI to use mirrors with git insteadof. 
3. we create/maintain our own mirrors. I noticed that we have lmp-mirrors. But again the link/association with meta-qcom is unclear.. so we could keep the upstream repo and use lmp-mirrors in our GH runners..

quaresmajose:
I think we can achieve mostly the same using the KAS_PREMIRRORS. This way we can use of the mirrors (github or another) on our CI and and continue to use the upstream projects url references on the ci/base.yml

For example with env we will start using the oe-core github mirror
```
KAS_PREMIRRORS = "https://git\\.openembedded\\.org/ https://github.com/openembedded/\n"
```

references:
https://github.com/siemens/kas/commit/2f7650bb05b9e06f1530353032260a9bd2c8d642
https://github.com/siemens/kas/commit/be8307b586a491636fa744c68f0117f17de671b0

koenkooi:
> I think we can achieve mostly the same using the KAS_PREMIRRORS. This way we can use the mirrors (github or another) on our CI and and continue to use the upstream projects url references on the ci/base.yml
> 
> For example, with this env we will start using the oe-core github mirror.
> 
> ```
> KAS_PREMIRRORS = "https://git\\.openembedded\\.org/ https://github.com/openembedded/\n"
> ```
> 
> references: [siemens/kas@2f7650b](https://github.com/siemens/kas/commit/2f7650bb05b9e06f1530353032260a9bd2c8d642) [siemens/kas@be8307b](https://github.com/siemens/kas/commit/be8307b586a491636fa744c68f0117f17de671b0)

It does mean we abstract our actual config behind yet another layer, making it harder for downstream users to see what's going on. In this specific case we're just adding more KAS logic, but in general meta-qcom has waaaaaay too much logic going on in ci/*.yml that downstream users miss out on when just including the layer itself (e.g. like meta-lpm-manifest would do).

ndechesne:
@koenkooi as discussed privately, I am ok to merge a change that at least switches oe-core and bitbake to their github mirrors. We can keep meta-yocto as is for now. Can you please refresh the patch?

koenkooi:
> @koenkooi as discussed privately, I am ok to merge a change that at least switches oe-core and bitbake to their github mirrors. We can keep meta-yocto as is for now. Can you please refresh the patch?

Done

quaresmajose (approved):
LGTM

ndechesne (approved):

ricardosalveti:
> > for let all know we have a foundries mirror at https://github.com/lmp-mirrors
> 
> but it is good to keep in mind that synchronization is scheduled and done hourly.

If we need to mirror additional layers and content we can extend lmp-mirrors (and even rename the org if needed), as we control this entirely.

ndechesne:
> If we need to mirror additional layers and content we can extend lmp-mirrors (and even rename the org if needed), as we control this entirely.

We discussed this in a call, and realized we have mirrors here https://git.codelinaro.org/clo/ype/external. The 'branches' on the mirrors have prefixes.. not sure why, but @quic-vkraleti is trying to get that cleaned up.

PR #823, opened 2025-04-03
workflows: Use unique build ids

lool:
Build ids on the fileserver simply used the GITHUB_RUN_ID, but it is not
guaranteed to be unique, and other repositories and workflows share the
same tree. Change build ids to be prefixed with the GITHUB_REPOSITORY
name, and suffixed with the GITHUB_RUN_ATTEMPT (in case the run has been
given back).

This is still safe from the perspective of directory names and URLs as
GitHub prevents org and repo names to start with a dash, and use only
alphanumeric chars plus single dashes.


lool:
build-yocto.yml is the only workflow where the fileserver local path and URL are defined/used directly.

The LAVA test workflow inject the GITHUB_RUN_ID in their files, and also submit jobs using the GITHUB_RUN_ID, but these workflow runs are entirely separate from the build workflow runs.  For instance, this [recent LAVA job](https://lava.infra.foundries.io/scheduler/job/67670) is named after the test.yml workflow run 14129155762 (`boot test (qrb2210-rb1-core-kit) 14129155762`), but is testing artifact URL from build-yocto workflow run 14127965204 (`https://quic-yocto-fileserver-1029608027416.us-central1.run.app/14127965204/qrb2210-rb1-core-kit/core-image-base-qrb2210-rb1-core-kit.rootfs.qcomflash.tar.gz`).

koenkooi (approved):
Looks good to me!

ndechesne:
Don't you need to update BASE_ARTIFACT_URL as well? That's the variable we use to define where the LAVA job will download the image from, and it's currently using the old path (RUN ID only).

lool:
Indeed, I saw this BASE_URL in my first read, but I forgot to update it. I've pushed an updated version, and grep-ed for github.run_id across workflows, this was the only use.

lumag (approved):

.github/workflows/build-yocto.yml
```diff
@@ -8,9 +8,10 @@ on:
         value: ${{ jobs.create-output.outputs.url }}
 
 env:
+  BUILD_ID: ${{ github.repository }}-${{ github.run_id }}-${{ github.run_attempt }}
   CACHE_DIR: /srv/gh-runners/quic-yocto
   KAS_REPO_REF_DIR: /srv/gh-runners/quic-yocto/kas-mirrors
-  BASE_ARTIFACT_URL: "https://quic-yocto-fileserver-1029608027416.us-central1.run.app/${{ github.run_id }}"
+  BASE_ARTIFACT_URL: "https://quic-yocto-fileserver-1029608027416.us-central1.run.app/"
```
ndechesne, line 14: So the CI worked, and the image was booted on the devices. LAVA picked up the image from:

"https://quic-yocto-fileserver-1029608027416.us-central1.run.app//qualcomm-linux/meta-qcom-14239561725-1/qcs6490-rb3gen2-core-kit/core-image-base-qcs6490-rb3gen2-core-kit.rootfs.qcomflash.tar.gz"

as shown here https://lava.infra.foundries.io/scheduler/job/68427#L3

As you can see there is a double // in the URL because you left a trailing / here. Do you want to fix that before we merge?

lool: Good catch, I've pushed an updated commit, also removing the quotes in the URL definition

ndechesne (approved):

lumag (approved):

PR #824, opened 2025-04-03
ci: use git force in pre-fetch

quaresmajose:
If the upstream server get corruped and fixed in some way we need to replicate evething in our mirror so keep exavtly from the server.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -21,7 +21,7 @@ jobs:
         run: |
           for r in $(find ${KAS_REPO_REF_DIR}/* -maxdepth 0 -type d); do
             echo "pre-fetch: $r"
-            git -C $r fetch --prune origin '+refs/*:refs/*'
+            git -C $r fetch --force --prune origin '+refs/*:refs/*'
           done
 
   kas-lock:


lumag (approved):

.github/workflows/build-yocto.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         run: |
           for r in $(find ${KAS_REPO_REF_DIR}/* -maxdepth 0 -type d); do
             echo "pre-fetch: $r"
-            git -C $r fetch --prune origin '+refs/*:refs/*'
+            git -C $r fetch --force --prune origin '+refs/*:refs/*'
```
ndechesne, line 24: I believe --force and the option '+' sign in the refspec are strictly equivalent, which would make this patch a no-op. Am i wrong or missing anything?

quaresmajose: According to the [documentation](https://git-scm.com/docs/git-fetch) the '+'  should be enough but then, how can the following happen?
```
pre-fetch: /srv/gh-runners/quic-yocto/kas-mirrors/git.openembedded.org.bitbake
error: cannot lock ref 'refs/heads/master': is at [13](https://github.com/qualcomm-linux/meta-qcom/actions/runs/14244733333/job/39922716348#step:2:14)6c06e251de68ed64355ec6b47a522ff3a372e3 but expected 5b4e20377eea8d428edf1aeb2187c18f82ca6757
From https://git.openembedded.org/bitbake
 ! 5b4e20377..136c06e25  master      -> master  (unable to update local ref)
error: cannot lock ref 'refs/heads/master-next': is at b827978a7f4b21c349b377bd884a345de6442890 but expected 47b7b1294[14](https://github.com/qualcomm-linux/meta-qcom/actions/runs/14244733333/job/39922716348#step:2:15)b69729e57c7b8a9003d123f05fb0f
 ! 47b7b1294...b827978a7 master-next -> master-next  (unable to update local ref)
Error: Process completed with exit code 1.
```
https://github.com/qualcomm-linux/meta-qcom/actions/runs/14244733333/job/39922716348

After this PR was executed this mentioned issue disappeared.

ndechesne: I don't know.. but if --force and the '+' are equivalent, then it cannot make such a difference.. we know that the git servers have been flaky. so perhaps that was a transient error?

quaresmajose: and it is very likely that it could have been something transient. I'll close it then

PR #825, opened 2025-04-04
partition-utils: use qcom-ptool

ndechesne:
qcom-ptool is a new project which is part of Qualcomm Linux, where we intend to provide all partition tools to generate GPT images, flashing artefact, ... These tools were maintained in various places up until now, and there was a copy of them in meta-qcom for some of them. With this patch, we removed the local copy from meta-qcom, and fetch all the tools we need from the new 'upstream' qcom-ptool.

Having these tools in a single, maintained repo (upstream) will make it easier for other distro to use them (such as our Qualcomm Linux Debian flavor).

Since qcom-ptool provides both ptool.py and gen_partition.py, the qcom-gen-partitions-tool recipe is simply removed from here.

This patch is effectively a no-op, and the tools we fetch are (at the time of this commit) exactly the same.

qcom-ptool also has a copy of each platform partition configuration file. For now, we are not using them in meta-qcom, and instead we are keeping our own copy. It's not clear yet, how we will manage the partition config files over time, over multiple LTS releases.. and it's not clear that we can have a single partition config file that will work everywhere. This change can be done later if needed.

The LICENSE md5 change is because of cosmetic changes in the LICENSE file, there is no change in the license itself.


recipes-devtools/partition-utils/qcom-ptool_git.bb
```diff
@@ -1,18 +1,15 @@
...
-SRC_URI = "git://git.codelinaro.org/linaro/qcomlt/partioning_tool.git;branch=master;protocol=https \
-	   file://0001-ptool.py-Generate-zero-files-in-output-folder-when-s.patch \
-	   file://0002-ptool.py-Python-3-support.patch \
-"
+SRC_URI = "git://github.com/qualcomm-linux/qcom-ptool.git;branch=main;protocol=https"
 
-SRCREV = "3484fc0a88088dea00397774fc93f9acd3a23ce0"
+SRCREV = "34d6580574bafd38751f526e1674eec3167546dc"
 
 PV = "0.0+git"
```
koenkooi, line 14: Is the version still 0.0?

ndechesne: there is no release done yet. so i didn't think about anything better than 0.0.. let me know if you have a better idea!

recipes-devtools/partition-utils/qcom-ptool_git.bb
```diff
@@ -27,6 +24,7 @@ do_compile[noexec] = "1"
 do_install() {
    install -d ${D}${bindir}
    install -m 755 ${S}/ptool.py ${D}${bindir}/ptool.py
+   install -m 755 ${S}/gen_partition.py ${D}${bindir}/gen_partition.py
```
lumag, line 27: I think this should be handled by the 'make install' step. If we are importing the repo under our GH org, we can add that targt too.

recipes-devtools/partition-utils/qcom-ptool_git.bb
```diff
@@ -1,18 +1,15 @@
 SUMMARY = "Qualcomm partitioning tool"
 DESCRIPTION = "Partitioning tool, generates the GPT and/or Partition MBN"
-HOMEPAGE = "https://git.codelinaro.org/linaro/qcomlt/partioning_tool"
+HOMEPAGE = "https://github.com/qualcomm-linux/qcom-ptool"
```
lumag, line 3: Should it be `/qualcomm-linux` or `/linux-msm`? I'd like to push all platforms from [this repo](https://git.codelinaro.org/linaro/qcomlt/db-boot-tools) too, so that we have a single place for all of them.

lumag (changes requested):

recipes-devtools/partition-utils/qcom-ptool_git.bb
sbanerjee-quic: IMO it should be qualcomm-linux

recipes-devtools/partition-utils/qcom-ptool_git.bb
quic-vkraleti: Agreed, will update ptool make file.

recipes-devtools/partition-utils/qcom-ptool_git.bb
lumag: Why?

ndechesne: it does not really matter.. if the repo had been in linux-msm initially we could have used it from there. now it's here, and maintained by some QC people. so let's use it. We can host any platforms config file here, I already merged your PRs earlier.

lumag: Ack

recipes-devtools/partition-utils/qcom-ptool_git.bb
ndechesne: done.

lumag (approved):

quaresmajose (approved):

ricardosalveti (approved):

PR #826, opened 2025-04-04
RB1: switch to chainloaded u-boot + ESP booting

koenkooi:
The RB1 currently uses an Android boot image to boot, this PR changes that to use chainload u-boot, which in turns loads an EFI bootloader from ESP.

To accomplish this:

- Add a recipe to build u-boot mainline + RB1 patches
- Add u-boot configs to the RB1 machine config
- Remove Android boot image from the RB1 machine config
- Make u-boot a dependency for the image_types_qcom class when UBOOT_CONFIG is non-empty

The current OE-core u-boot is too old (2025.01), when upstream u-boot releases 2025.04 we can update u-boot in OE-core and convert the recipe to a bbappend.


recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -0,0 +1,63 @@
+# Limit to qrb2210
+COMPATIBLE_MACHINE = "(qrb2210-rb1-core-kit)"
```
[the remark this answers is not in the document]
koenkooi: done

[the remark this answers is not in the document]
lumag: The official name for the SoC on the RB1 is QRB2210. I think we should use the kernel policy and use the base name for generic code, but exact SoC name for machine names.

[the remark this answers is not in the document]
lumag: After looking at the change... I'm fine either way.

ndechesne (approved):

b49020 (approved):
Glad to see U-Boot working in the CI.

recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -0,0 +1,63 @@
...
+            i=$(expr $i + 1);
+            for type in ${UBOOT_CONFIG}; do
+                j=$(expr $j + 1);
+                if [ $j -eq $i ]
+                then
+	            cd ${B}/${config}
+                    rm -f u-boot-nodtb.bin.gz
+	            gzip -k u-boot-nodtb.bin
+	            cat u-boot-nodtb.bin.gz dts/upstream/src/arm64/$UBOOT_DTB > u-boot-nodtb.bin.gz-dtb
+	            ${STAGING_BINDIR_NATIVE}/skales/mkbootimg --base 0x80000000 --pagesize 4096 --kernel u-boot-nodtb.bin.gz-dtb  --cmdline "root=/dev/notreal" --ramdisk u-boot.bin --output ${MACHINE}-u-boot-${PV}-boot.img
```
lumag, line 37: Would it be better to use `uboot_compile_config:append`?

recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -0,0 +1,63 @@
...
+    if [ -n "${UBOOT_CONFIG}" ]
+    then
+        for config in ${UBOOT_MACHINE}; do
+            i=$(expr $i + 1);
+            for type in ${UBOOT_CONFIG}; do
+                j=$(expr $j + 1);
+                if [ $j -eq $i ]
+                then
+	            cd ${B}/${config} && install -m 0644 ${MACHINE}-u-boot-${PV}-boot.img ${DEPLOYDIR}
+	            cd ${DEPLOYDIR} && ln -sf ${MACHINE}-u-boot-${PV}-boot.img boot-${MACHINE}.img && ln -sf ${MACHINE}-u-boot-${PV}-boot.img boot.img
```
lumag, line 56: Likewise `uboot_deploy_config:append`

lumag (changes requested):

quaresmajose:
Maybe we also need to bring in the u-boot-tools

recipes-bsp/u-boot/u-boot_git.bb
koenkooi: Good point, done!

recipes-bsp/u-boot/u-boot_git.bb
koenkooi: Good point, done!

recipes-bsp/u-boot/u-boot_git.bb
lumag: Not quite. This function gets all necessary arguments, so you don't need loops. The other one can be simplified too.

koenkooi: That tidies up the methods quite neatly, thanks for the suggestion!

lumag (dismissed):

recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -0,0 +1,35 @@
...
+    UBOOT_DTB=""
+    for dtb in ${KERNEL_DEVICETREE} ; do
+	UBOOT_DTB="$dtb"
+    done
+
+    cd ${B}/${config}
+    rm -f u-boot-nodtb.bin.gz
+    gzip -k u-boot-nodtb.bin
+    cat u-boot-nodtb.bin.gz dts/upstream/src/arm64/$UBOOT_DTB > u-boot-nodtb.bin.gz-dtb
+    ${STAGING_BINDIR_NATIVE}/skales/mkbootimg --base 0x80000000 --pagesize 4096 --kernel u-boot-nodtb.bin.gz-dtb  --cmdline "root=/dev/notreal" --ramdisk u-boot.bin --output ${MACHINE}-u-boot-${PV}-boot.img
```
lumag, line 29: As a second thought... Shouldn't we create an image per DTB in `KERNEL_DEVICETREE`?

koenkooi: That's a good idea, but it does raise the question of which DTB we pick to use as boot.img in qcomflash.

ndechesne: Ideally the one that comes from the kernel build we are doing in meta-qcom.

koenkooi:
I pushed 2 changes:

1. Update Upstream-status of the patches (shouldn't be controversial)
2. Use kernel dtbs, all of them and pick the one listed last to use as boot.img 

I haven't squashed those yet because 2. is a bit controversial and annoying to revert after squashing.

recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -15,21 +15,39 @@ SRC_URI:append:qcom = " \
            file://0004-qcom_defconfig-Disable-MMC-HS200-mode-support.patch \
 "
 
+do_compile[depends] += " \
+    virtual/kernel:do_deploy \
```
ndechesne, line 19: one line?

recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -15,21 +15,39 @@ SRC_URI:append:qcom = " \
...
-    cd ${DEPLOYDIR} && ln -sf ${MACHINE}-u-boot-${PV}-boot.img boot-${MACHINE}.img && ln -sf ${MACHINE}-u-boot-${PV}-boot.img boot.img
+    cd ${B}/${config} && install -m 0644 ${MACHINE}-u-boot-*-${PV}-boot.img ${DEPLOYDIR}
+    for dtb in ${KERNEL_DEVICETREE} ; do
+        UBOOT_DTB="$dtb"
+        # Most machines in meta-qcom put their DTBs strait in the deploy
+        # folder instead of using their vendor subfolder
+        if ! [ -e ${DEPLOY_DIR_IMAGE}/$UBOOT_DTB ] ; then
+            UBOOT_DTB="$(basename $dtb)"
+        fi
+        cd ${DEPLOYDIR} && ln -sf ${MACHINE}-u-boot-$(basename $dtb .dtb)-${PV}-boot.img boot-${MACHINE}.img && ln -sf ${MACHINE}-u-boot-${PV}-boot.img boot.img
```
ndechesne, line 51: Up until now, we picked the 'first' one as the default, see https://github.com/qualcomm-linux/meta-qcom/blob/master/classes-recipe/image_types_qcom.bbclass#L15

ndechesne:
The publishing is failing with 
PROGRESS=3/79 OBJ=qualcomm-linux/meta-qcom-14352972664-1/qrb2210-rb1-core-kit/boot.img STATUS=ERROR;item does not exist: qualcomm-linux/meta-qcom-14352972664-1/qrb2210-rb1-core-kit/boot.img

Also there are a bunch of unnecessary files published , such as u-boot-*.bin.

recipes-bsp/u-boot/u-boot_git.bb
koenkooi: I've copied that logic

recipes-bsp/u-boot/u-boot_git.bb
koenkooi: Done

recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -0,0 +1,44 @@
...
+           file://0003-arm-mach-snapdragon-Enable-OF_UPSTREAM_BUILD_VENDOR.patch \
+           file://0004-qcom_defconfig-Disable-MMC-HS200-mode-support.patch \
+"
+
+do_compile[depends] += "virtual/kernel:do_deploy"
+
+uboot_compile_config:append:qcom() {
+    cd ${B}/${config}
+
+    # Strip whitespace from ${KERNEL_DEVICETREE}
```
lumag, line 23: I think the comment is misleading a bit. It's not just about stripping whitespace, it's about creating a binary for each DTB

recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -0,0 +1,44 @@
...
+    UBOOT_DTB=""
+    for dtb in ${KERNEL_DEVICETREE} ; do
+        UBOOT_DTB="$dtb"
+        # Most machines in meta-qcom put their DTBs strait in the deploy
+        # folder instead of using their vendor subfolder
```
lumag, line 28: I think that's standard, not to have the subdir, isn't it?

recipes-bsp/u-boot/u-boot_git.bb
lumag: ... and the platforms were able to override that variable.

recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -0,0 +1,44 @@
...
+        if ! [ -e ${DEPLOY_DIR_IMAGE}/$UBOOT_DTB ] ; then
+            UBOOT_DTB="$(basename $dtb)"
+        fi
+        rm -f u-boot-nodtb.bin.gz
+        gzip -k u-boot-nodtb.bin
```
lumag, line 33: I think rm + gzip can be moved outside of the loop.

recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -0,0 +1,44 @@
...
+        cat u-boot-nodtb.bin.gz ${DEPLOY_DIR_IMAGE}/$UBOOT_DTB > u-boot-nodtb.bin.gz-dtb
+        ${STAGING_BINDIR_NATIVE}/skales/mkbootimg --base 0x80000000 --pagesize 4096 --kernel u-boot-nodtb.bin.gz-dtb  --cmdline "root=/dev/notreal" --ramdisk u-boot.bin --output ${MACHINE}-u-boot-$(basename $dtb .dtb)-${PV}-boot.img
```
lumag, line 35: Do we need `${MACHINE}` in the file name? it already goes to machine-specific deploy dir.

recipes-bsp/u-boot/u-boot_git.bb
```diff
@@ -0,0 +1,44 @@
...
+           file://0004-qcom_defconfig-Disable-MMC-HS200-mode-support.patch \
+"
+
+do_compile[depends] += "virtual/kernel:do_deploy"
+
+uboot_compile_config:append:qcom() {
+    cd ${B}/${config}
+
+    # Strip whitespace from ${KERNEL_DEVICETREE}
+    UBOOT_DTB=""
```
lumag, line 24: Unused, you set it inside the loop

lumag:
> > Also there are a bunch of unnecessary files published , such as u-boot-*.bin.
> 
> Unless we want to override do_deploy from u-boot.bb instead of appending it, we're stuck with those. Or add a cleanup step.

I played a little bit. Please consider the following suggestion:

```diff
diff --git a/conf/machine/qrb2210-rb1-core-kit.conf b/conf/machine/qrb2210-rb1-core-kit.conf
index 8c507ab44647..1b8819599bf3 100644
--- a/conf/machine/qrb2210-rb1-core-kit.conf
+++ b/conf/machine/qrb2210-rb1-core-kit.conf
@@ -9,6 +9,7 @@ MACHINE_FEATURES = "efi usbhost usbgadget alsa wifi bluetooth"
 PREFERRED_PROVIDER_virtual/bootloader = "u-boot"
 UBOOT_CONFIG = "qrb2210-rb1"
 UBOOT_CONFIG[qrb2210-rb1] = "qcom_defconfig"
+UBOOT_INITIAL_ENV = ""
 
 QCOM_DTB_DEFAULT ?= "qrb2210-rb1"
 
diff --git a/recipes-bsp/u-boot/u-boot_git.bb b/recipes-bsp/u-boot/u-boot_git.bb
index ef9c92e97d62..5fd3cd65bd08 100644
--- a/recipes-bsp/u-boot/u-boot_git.bb
+++ b/recipes-bsp/u-boot/u-boot_git.bb
@@ -1,5 +1,5 @@
 # Limit to qrb2210
-COMPATIBLE_MACHINE = "(qcm2290)"
+COMPATIBLE_MACHINE = "(qcom)"
 
 require recipes-bsp/u-boot/u-boot-common.inc
 require recipes-bsp/u-boot/u-boot.inc
@@ -20,25 +20,8 @@ do_compile[depends] += "virtual/kernel:do_deploy"
 uboot_compile_config:append:qcom() {
     cd ${B}/${config}
 
-    # Strip whitespace from ${KERNEL_DEVICETREE}
-    UBOOT_DTB=""
-    for dtb in ${KERNEL_DEVICETREE} ; do
-        UBOOT_DTB="$dtb"
-        # Most machines in meta-qcom put their DTBs strait in the deploy
-        # folder instead of using their vendor subfolder
-        if ! [ -e ${DEPLOY_DIR_IMAGE}/$UBOOT_DTB ] ; then
-            UBOOT_DTB="$(basename $dtb)"
-        fi
-        rm -f u-boot-nodtb.bin.gz
-        gzip -k u-boot-nodtb.bin
-        cat u-boot-nodtb.bin.gz ${DEPLOY_DIR_IMAGE}/$UBOOT_DTB > u-boot-nodtb.bin.gz-dtb
-        ${STAGING_BINDIR_NATIVE}/skales/mkbootimg --base 0x80000000 --pagesize 4096 --kernel u-boot-nodtb.bin.gz-dtb  --cmdline "root=/dev/notreal" --ramdisk u-boot.bin --output ${MACHINE}-u-boot-$(basename $dtb .dtb)-${PV}-boot.img
-    done
-
-}
-
-uboot_deploy_config:append:qcom() {
-    cd ${B}/${config} && install -m 0644 ${MACHINE}-u-boot-*-${PV}-boot.img ${DEPLOYDIR}
-    UBOOT_DTB="${@os.path.basename(d.getVar('KERNEL_DEVICETREE').split()[0][:-4]) if d.getVar('KERNEL_DEVICETREE') else ''}"
-    cd ${DEPLOYDIR} && ln -sf ${MACHINE}-u-boot-$(basename $UBOOT_DTB .dtb)-${PV}-boot.img boot-${MACHINE}.img && ln -sf  ${MACHINE}-u-boot-$(basename $UBOOT_DTB .dtb)-${PV}-boot.img boot.img
+    rm -f u-boot-nodtb.bin.gz
+    gzip -k u-boot-nodtb.bin
+    cat u-boot-nodtb.bin.gz ${DEPLOY_DIR_IMAGE}/${type}.dtb > u-boot-nodtb.bin.gz-${type}
+    ${STAGING_BINDIR_NATIVE}/skales/mkbootimg --base 0x80000000 --pagesize 4096 --kernel u-boot-nodtb.bin.gz-${type} --cmdline "root=/dev/notreal" --ramdisk u-boot.bin --output u-boot-${type}.bin
 }
```

lumag:
Ideally it would be nice to be able to insert a hook after the `oe_runmake` step in `uboot_compile_config()`. Then we would be able to declare necessary binary types via the `UBOOT_CONFIG` and merge `qtestsign` with our `mkbootimage` hooks.

recipes-bsp/u-boot/u-boot_git.bb
lumag, line 35: Hmm. What is the reason for using u-boot.bin as ramdisk?

recipes-bsp/u-boot/u-boot_git.bb
koenkooi: It was originally about stripping whitespace, but you are correct that it doesn't match the current code :)

recipes-bsp/u-boot/u-boot_git.bb
koenkooi: Yes, I'll move it out.

recipes-bsp/u-boot/u-boot_git.bb
koenkooi: I think so, it's a very specific file that might get copied out of deploy to be flashed separately. For rb1 the DTB filename might  be descriptive enough, but that is likely not the case for other MACHINEs based on this SoC.

recipes-bsp/u-boot/u-boot_git.bb
koenkooi: I've dropped it, the scoping issues I was worried about are not relevant anymore in this iteration :)

recipes-bsp/u-boot/u-boot_git.bb
koenkooi: I followed the instructions from @b49020 for that, I don't know the reason for it.

b49020: That was just to suffice `ramdisk` requirement for `mkbootimg` tooling from `skales` repo, it doesn't play any role during boot. It's optional though for `mkbootimg` available from distro packages.

lumag: Heh. We can use an empty file for ramdisk.

lumag: Please use an empty file. Then we can use `UBOOT_MAKE_TARGET = "u-boot-nodtb.bin"`

koenkooi: Using an empty file works, but the follow up `UBOOT_MAKE_TARGET:qcom = "u-boot-nodtb.bin"` breaks the u-boot build.

koenkooi:
> > > Also there are a bunch of unnecessary files published , such as u-boot-*.bin.
> > 
> > 
> > Unless we want to override do_deploy from u-boot.bb instead of appending it, we're stuck with those. Or add a cleanup step.
> 
> I played a little bit. Please consider the following suggestion. This way there are no extra binaries and the image is deployed in a native way.
> 
> ```diff
> diff --git a/conf/machine/qrb2210-rb1-core-kit.conf b/conf/machine/qrb2210-rb1-core-kit.conf
> index 8c507ab44647..1b8819599bf3 100644
> --- a/conf/machine/qrb2210-rb1-core-kit.conf
> +++ b/conf/machine/qrb2210-rb1-core-kit.conf
> @@ -9,6 +9,7 @@ MACHINE_FEATURES = "efi usbhost usbgadget alsa wifi bluetooth"
>  PREFERRED_PROVIDER_virtual/bootloader = "u-boot"
>  UBOOT_CONFIG = "qrb2210-rb1"
>  UBOOT_CONFIG[qrb2210-rb1] = "qcom_defconfig"
> +UBOOT_INITIAL_ENV = ""
>  
>  QCOM_DTB_DEFAULT ?= "qrb2210-rb1"
>  
> diff --git a/recipes-bsp/u-boot/u-boot_git.bb b/recipes-bsp/u-boot/u-boot_git.bb
> index ef9c92e97d62..5fd3cd65bd08 100644
> --- a/recipes-bsp/u-boot/u-boot_git.bb
> +++ b/recipes-bsp/u-boot/u-boot_git.bb
> @@ -1,5 +1,5 @@
>  # Limit to qrb2210
> -COMPATIBLE_MACHINE = "(qcm2290)"
> +COMPATIBLE_MACHINE = "(qcom)"
>  
>  require recipes-bsp/u-boot/u-boot-common.inc
>  require recipes-bsp/u-boot/u-boot.inc
> @@ -20,25 +20,8 @@ do_compile[depends] += "virtual/kernel:do_deploy"
>  uboot_compile_config:append:qcom() {
>      cd ${B}/${config}
>  
> -    # Strip whitespace from ${KERNEL_DEVICETREE}
> -    UBOOT_DTB=""
> -    for dtb in ${KERNEL_DEVICETREE} ; do
> -        UBOOT_DTB="$dtb"
> -        # Most machines in meta-qcom put their DTBs strait in the deploy
> -        # folder instead of using their vendor subfolder
> -        if ! [ -e ${DEPLOY_DIR_IMAGE}/$UBOOT_DTB ] ; then
> -            UBOOT_DTB="$(basename $dtb)"
> -        fi
> -        rm -f u-boot-nodtb.bin.gz
> -        gzip -k u-boot-nodtb.bin
> -        cat u-boot-nodtb.bin.gz ${DEPLOY_DIR_IMAGE}/$UBOOT_DTB > u-boot-nodtb.bin.gz-dtb
> -        ${STAGING_BINDIR_NATIVE}/skales/mkbootimg --base 0x80000000 --pagesize 4096 --kernel u-boot-nodtb.bin.gz-dtb  --cmdline "root=/dev/notreal" --ramdisk u-boot.bin --output ${MACHINE}-u-boot-$(basename $dtb .dtb)-${PV}-boot.img
> -    done
> -
> -}
> -
> -uboot_deploy_config:append:qcom() {
> -    cd ${B}/${config} && install -m 0644 ${MACHINE}-u-boot-*-${PV}-boot.img ${DEPLOYDIR}
> -    UBOOT_DTB="${@os.path.basename(d.getVar('KERNEL_DEVICETREE').split()[0][:-4]) if d.getVar('KERNEL_DEVICETREE') else ''}"
> -    cd ${DEPLOYDIR} && ln -sf ${MACHINE}-u-boot-$(basename $UBOOT_DTB .dtb)-${PV}-boot.img boot-${MACHINE}.img && ln -sf  ${MACHINE}-u-boot-$(basename $UBOOT_DTB .dtb)-${PV}-boot.img boot.img
> +    rm -f u-boot-nodtb.bin.gz
> +    gzip -k u-boot-nodtb.bin
> +    cat u-boot-nodtb.bin.gz ${DEPLOY_DIR_IMAGE}/${type}.dtb > u-boot-nodtb.bin.gz-${type}
> +    ${STAGING_BINDIR_NATIVE}/skales/mkbootimg --base 0x80000000 --pagesize 4096 --kernel u-boot-nodtb.bin.gz-${type} --cmdline "root=/dev/notreal" --ramdisk u-boot.bin --output u-boot-${type}.bin
>  }
> ```

Giving this a try now!

recipes-bsp/u-boot/u-boot_git.bb
lumag: Why? The DT is usually very descriptive. The kernel uses it :-)

lumag:
> I get this:
> 
> ```
> kkooi@hu-kkooi-ams:/local/mnt/workspace/koen/kas-github$ ls build/tmp/deploy/images/qrb2210-rb1-core-kit/ | grep "boot"
> systemd-bootaa64.efi
> u-boot.bin
> u-boot.bin-qrb2210-rb1
> u-boot-qrb2210-rb1-2025.03+2025.04-rc6-r0.bin
> u-boot-qrb2210-rb1-core-kit.bin
> u-boot-qrb2210-rb1-core-kit.bin-qrb2210-rb1
> ```

```
$ file tmp/deploy/images/qrb2210-rb1-core-kit/u-boot*
tmp/deploy/images/qrb2210-rb1-core-kit/u-boot.bin:                                    symbolic link to u-boot-qrb2210-rb1-2025.03+2025.04-rc6-r0.bin
tmp/deploy/images/qrb2210-rb1-core-kit/u-boot.bin-qrb2210-rb1:                        symbolic link to u-boot-qrb2210-rb1-2025.03+2025.04-rc6-r0.bin
tmp/deploy/images/qrb2210-rb1-core-kit/u-boot-qrb2210-rb1-2025.03+2025.04-rc6-r0.bin: Android bootimg, kernel, page size: 4096, cmdline (root=/dev/notreal)
tmp/deploy/images/qrb2210-rb1-core-kit/u-boot-qrb2210-rb1-core-kit.bin:               symbolic link to u-boot-qrb2210-rb1-2025.03+2025.04-rc6-r0.bin
tmp/deploy/images/qrb2210-rb1-core-kit/u-boot-qrb2210-rb1-core-kit.bin-qrb2210-rb1:   symbolic link to u-boot-qrb2210-rb1-2025.03+2025.04-rc6-r0.bin
```

So those are the bits that we actually need.

koenkooi:
> This needs to be rebased on top of u-boot 2025.04 which was merged upstream. and some squash/clean up in the PR

Yes, I have a commit locally to change it to a bbappend on top of the oe-core 2025.04 recipe.

koenkooi:
I've rebased on `master` and squashed everything, except for the magic @lumag showed us. The naming of the artefacts is opaque, but it results in the least amount of litter in `DEPLOY_DIR_IMAGE`, so I'd like to keep it. I'll squash it into the main u-boot commit if there are no objections.

I've boot tested the result on my RB1, it works as expected.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -28,6 +28,7 @@ do_image_qcomflash[dirs] = "${QCOMFLASH_DIR}"
 do_image_qcomflash[cleandirs] = "${QCOMFLASH_DIR}"
 do_image_qcomflash[depends] += "${@ ['', '${QCOM_PARTITION_CONF}:do_deploy'][d.getVar('QCOM_PARTITION_CONF') != '']} \
                                 virtual/kernel:do_deploy \
+				${@'virtual/bootloader:do_deploy' if 'qcm2290' in d.getVar('SOC_FAMILY') else  ''} \
```
quaresmajose, line 31: The virtual/bootloader:do_deploy comes from u-boot so maybe is better checking if it is the prefered provider
```
${@'virtual/bootloader:do_deploy' if 'u-boot' in d.getVar('PREFERRED_PROVIDER_virtual/bootloader') else  ''} \
```

koenkooi: That blows up when you aren't using u-boot:

```
bb.data_smart.ExpansionError: Failure expanding variable do_image_qcomflash[depends], expression was  ${@ ['', ':do_deploy'][d.getVar('QCOM_PARTITION_CONF') != '']}                                 virtual/kernel:do_deploy 				${@'virtual/bootloader:do_deploy' if 'u-boot' in d.getVar('PREFERRED_PROVIDER_virtual/bootloader') else  ''} 				${@'esp-qcom-image:do_image_complete' if d.getVar('QCOM_ESP_IMAGE') != '' else  ''} which triggered exception TypeError: argument of type 'NoneType' is not iterable
The variable dependency chain for the failure is: do_image_qcomflash[depends]
```

quaresmajose: According to the build error [argument of type 'NoneType' is not iterable](https://github.com/qualcomm-linux/meta-qcom/actions/runs/14463561784/job/40560817262#step:4:880) we can replace the `in` by `==` .
The PREFERRED_PROVIDER_virtual/bootloader is not a list and only include a single element.

I tested it on the python interface:
```bash
$ python
Python 3.13.2 (main, Feb  5 2025, 08:05:21) [GCC 14.2.1 20250128] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 'virtual/bootloader:do_deploy' if 'u-boot' in None else ''
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    'virtual/bootloader:do_deploy' if 'u-boot' in None else ''
                                      ^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable
>>> 'virtual/bootloader:do_deploy' if 'u-boot' == None else ''
''
```
So the fix should be:
```patch
-${@'virtual/bootloader:do_deploy' if 'u-boot' in d.getVar('PREFERRED_PROVIDER_virtual/bootloader') else  ''} \
+${@'virtual/bootloader:do_deploy' if 'u-boot' == d.getVar('PREFERRED_PROVIDER_virtual/bootloader') else  ''} \
```

lumag: What about more generic:
```
				${@'virtual/bootloader:do_deploy' if d.getVar('PREFERRED_PROVIDER_virtual/bootloader') else  ''} \
```

Note, suggestion not tested at all ;-)

lumag: @quaresmajose what if the platform uses othe bootloader? We should still depend on `virtual/bootloader` even it's not a u-boot.

quaresmajose: yeah, if the `PREFERRED_PROVIDER_virtual/bootloader` variable is defined we will have the task 'virtual/bootloader:do_deploy' in place. So I agree with you @lumag, keep it simple!

koenkooi: > What about more generic:
> 
> ```
> 				${@'virtual/bootloader:do_deploy' if d.getVar('PREFERRED_PROVIDER_virtual/bootloader') else  ''} \
> ```
> 
> Note, suggestion not tested at all ;-)

That one works locally, let's see what the CI thinks of it for non-RB1 boards.

koenkooi:
I've rebased, squashed and reworded the commits, I think all review comments have been addressed.

quaresmajose (approved):

lumag (dismissed):

recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -0,0 +1,28 @@
+# Limit to qcom machines
+COMPATIBLE_MACHINE = "(qcom)"
+
+DEPENDS += "skales-native xxd-native"
```
ricardosalveti, line 4: Won't hurt but will also affect any other layer that might be added together with this, maybe depends:append:qcom?

conf/machine/qrb2210-rb1-core-kit.conf
```diff
@@ -6,6 +6,7 @@ require conf/machine/include/qcom-qcm2290.inc
 
 MACHINE_FEATURES = "efi usbhost usbgadget alsa wifi bluetooth"
 
+PREFERRED_PROVIDER_virtual/bootloader = "u-boot"
```
ricardosalveti, line 9: At machine level it is usually better to set with ?= instead.

koenkooi: I strongly disagree with that, the machine knows best. If you want to override this, you already have to do `PREFERRED_PROVIDER_virtual/bootloader:<override>`, which will completely ignore the `=`. Unless you want a DISTRO wide, indiscriminate bootloader change, which is practically speaking impossible.

recipes-bsp/u-boot/u-boot_%.bbappend
koenkooi: That makes more sense, yes

recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -0,0 +1,28 @@
...
+
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/files:"
+SRC_URI:append:qcom = " \
+           file://0001-clk-stub-generic-rpm-requests-compatible.patch \
+           file://0002-arm-dts-Add-override-for-RB1.patch \
+           file://0003-arm-mach-snapdragon-Enable-OF_UPSTREAM_BUILD_VENDOR.patch \
+           file://0004-qcom_defconfig-Disable-MMC-HS200-mode-support.patch \
+"
+
+do_compile[depends] += "virtual/kernel:do_deploy"
```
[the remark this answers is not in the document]
koenkooi: done

recipes-bsp/u-boot/u-boot_%.bbappend
koenkooi: done

lumag (changes requested):

recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -0,0 +1,31 @@
+# Limit to qcom machines
+COMPATIBLE_MACHINE = "(qcom)"
```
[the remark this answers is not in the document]
koenkooi: Yes, putting COMPATIBLE_MACHINE in a bbappend is generally a very bad idea!

lumag (approved):

quaresmajose (approved):

conf/machine/qrb2210-rb1-core-kit.conf
ricardosalveti: It is actually not that common for the machine conf to set a strong override, and this also causes an issue when the distro specific layer wants to override the settings via a include file (without :override), as done with meta-updater.

E.g.:

`classes/sota.bbclass:inherit sota_sanity sota_${SOTA_MACHINE}`

This includes a machine specific file that could then override `PREFERRED_PROVIDER_virtual/bootloader`.

And to compare with other BSPs:

```
meta-arm/meta-arm-bsp/conf/machine/juno.conf:PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot"
meta-freescale/conf/machine/imx6sllevk.conf:PREFERRED_PROVIDER_virtual/bootloader = "u-boot-fslc"
meta-freescale/conf/machine/include/imx-base.inc:PREFERRED_PROVIDER_virtual/bootloader ??= "${IMX_DEFAULT_BOOTLOADER}"
meta-freescale/conf/machine/include/qoriq-base.inc:PREFERRED_PROVIDER_virtual/bootloader ??= "${PREFERRED_PROVIDER_u-boot}"
meta-freescale-3rdparty/conf/machine/nitrogen8mn.conf:PREFERRED_PROVIDER_virtual/bootloader ??= "u-boot-boundary"
meta-freescale-3rdparty/conf/machine/nitrogen6x-lite.conf:PREFERRED_PROVIDER_virtual/bootloader ??= "u-boot-boundary"
meta-freescale-3rdparty/conf/machine/nitrogen8mm.conf:PREFERRED_PROVIDER_virtual/bootloader ??= "u-boot-boundary"
meta-freescale-3rdparty/conf/machine/nitrogen7.conf:PREFERRED_PROVIDER_virtual/bootloader ??= "u-boot-boundary"
meta-freescale-3rdparty/conf/machine/nitrogen8m.conf:PREFERRED_PROVIDER_virtual/bootloader ??= "u-boot-boundary"
meta-freescale-3rdparty/conf/machine/imx6qdl-variscite-som.conf:PREFERRED_PROVIDER_virtual/bootloader = "u-boot-variscite"
meta-freescale-3rdparty/conf/machine/imx6ul-kontron.conf:PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot-kontron"
meta-freescale-3rdparty/conf/machine/nitrogen8mp.conf:PREFERRED_PROVIDER_virtual/bootloader ??= "u-boot-boundary"
meta-tegra/conf/machine/include/tegra-common.inc:PREFERRED_PROVIDER_virtual/bootloader ?= "edk2-firmware-tegra"
meta-ti/meta-ti-bsp/conf/machine/include/ti33x.inc:PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot-ti-staging"
meta-ti/meta-ti-bsp/conf/machine/include/omap-a15.inc:PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot-ti-staging"
meta-ti/meta-ti-bsp/conf/machine/include/k3.inc:PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot-ti-staging"
meta-ti/meta-ti-bsp/conf/machine/include/ti43x.inc:PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot-ti-staging"
meta-ti/meta-ti-bsp/conf/machine/include/k3r5.inc:PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot-ti-staging"
meta-xilinx/meta-xilinx-core/conf/machine/include/machine-xilinx-default.inc:PREFERRED_PROVIDER_virtual/bootloader ??= "u-boot-xlnx"
meta-yocto/meta-yocto-bsp/conf/machine/beaglebone-yocto.conf:PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot"
openembedded-core/meta/conf/machine/include/riscv/qemuriscv.inc:PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot"
```

ricardosalveti: When setting with '=' only this will force the user to also add a machine/soc override in order to replace it.

ricardosalveti (approved):
LGTM!

lumag (approved):

PR #827, opened 2025-04-07
initramfs-firmware-rb3gen2-image: add firmware image

lumag:
Add initramfs image with the firmware for the RB3 Gen2 board.


quic-vkraleti:
> Add initramfs image with the firmware for the RB3 Gen2 board.

packagegroup-rb3gen2-firmware is already part of rootfs image. Any specific need to move it to ramdisk?

ricardosalveti:
What is the reason for this image, just to facilitate kernel dev / debugging without having to flash rootfs?

lumag:
> What is the reason for this image, just to facilitate kernel dev / debugging without having to flash rootfs?

Yes, pretty much like the rest of initramfs images. I'm using them for testing the kernels in the CDBA lab without flashing anything.

PR #830, opened 2025-04-09
partition: fix rootfs GUID in the RB1 partition conf file

ndechesne:
As reported on https://github.com/qualcomm-linux/qcom-ptool/issues/9, the GUID used for rootfs is incorrectly. It was wrongly inherited from the original partition XML file of an Android build, and the GUID which was used corresponds to what Android uses for userdata partition.

This changes sets the correct GUID
(b921b045-1df0-41c3-af44-4c6f280d3fae) for Root Partition (64-bit ARM/AArch64).

Reported-by: Loic Minier <loic.minier@oss.qualcomm.com>


quic-vkraleti (approved):
LGTM

lumag (approved):

PR #836, opened 2025-04-10
mesa: add support for QCM2290

lumag:
The QRB2210 / QCM2290 SoC uses A702, which is not currently supported by Mesa. Import patches from the MR ([0]) and add a patch on top that fixes build issues.

[0] https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/27665


koenkooi (approved):
Looks good to me!

ndechesne (approved):

PR #838, opened 2025-04-11
Use KERNEL_CLASSES to pull in image-related classes

lumag:
Rather than clobbering linux-yocto (and linux-yocto-dev) with machine-specific inherits, specify required classes using the KERNEL_CLASSES variable.


koenkooi:
Nice

koenkooi (approved):

vkraleti:
This change is relevant for remaining machines as well. You may update them as well.

lumag:
> This change is relevant for remaining machines as well. You may update them as well.

I think they are already using `KERNEL_CLASSES`.

ndechesne (approved):

PR #839, opened 2025-04-11
WIP: CI: move local.conf modifications into their own yaml to make them optional

koenkooi:
Depends on #786

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -92,8 +92,13 @@ jobs:
           - qrb2210-rb1-core-kit
           - qcom-armv8a
           - qcom-armv7a
+        distro:
+          - name: poky/systemd
+            yamlfile: ""
+          - name: qcom
+            yamlfile: 'ci/distro.yml:'
     runs-on: [self-hosted, x86]
-    name: ${{ matrix.machine }}/poky/systemd
+    name: ${{ matrix.machine }}/${{ matrix.distro.name }}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-yocto.yml
```diff
@@ -92,8 +92,13 @@ jobs:
           - qrb2210-rb1-core-kit
           - qcom-armv8a
           - qcom-armv7a
+        distro:
+          - name: poky/systemd
+            yamlfile: ""
```
ndechesne: it is really a kas config file. so 'kas' would be a better name.

ndechesne: btw, it's empty, you can simply skip it.

ci/base.yml
```diff
@@ -28,18 +28,10 @@ repos:
       meta-poky:
 
 local_conf_header:
-  base: |
-    CONF_VERSION = "2"
-    INIT_MANAGER = "systemd"
-    PACKAGE_CLASSES = "package_ipk"
-    INHERIT += "buildstats buildstats-summary"
```
ndechesne, line 35: the 3 INHERIT in there, I see them as 'CI' configs that we want to enforce. I think we should keep them here. 
First buildstats now is required since f6cfffdd2234a7f357fbe2896c115182098a4e57, and buildhistory I would like eventually to keep a copy of all our builds, so it would be required too.

rm_work while not needed in CI with throw away instances, is really a nice user feature, which should be kept by default too.

.github/workflows/build-yocto.yml
```diff
@@ -94,7 +94,7 @@ jobs:
           - qcom-armv7a
         distro:
           - name: poky/systemd
-            yamlfile: ""
+            yamlfile: 'ci/localconfmods.yml:'
```
ndechesne, line 97: localconf.yml?

PR #845, opened 2025-04-20
Email update for vkraleti

vkraleti:
As an organizational guideline, it is preferred to use the oss.qualcomm.com email ID instead of the quicinc email ID.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,12 +1,12 @@
 # Default reviewers for all changes in entire repository,
 # unless a later match takes precedence
-*  @quic-vkraleti @ricardosalveti @sbanerjee-quic @ndechesne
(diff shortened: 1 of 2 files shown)


sbanerjee-quic (approved):

.github/CODEOWNERS
```diff
@@ -1,12 +1,12 @@
 # Default reviewers for all changes in entire repository,
 # unless a later match takes precedence
-*  @quic-vkraleti @ricardosalveti @sbanerjee-quic @ndechesne
+*  @vkraleti @ricardosalveti @sbanerjee-quic @ndechesne
```
lumag, line 3: There should be no need to change accounts just for the email change...

lumag (approved):

PR #846, opened 2025-04-22
ci: switch to kas-container 

quaresmajose:


lumag:
> _ No description provided. _

Could you please describe briefly, why are you performing those changes?

lumag:
None of the commits have commit messages. Please fix that.

ci/yocto-patchreview.sh
```diff
@@ -2,15 +2,13 @@
 # Copyright (c) 2024 Qualcomm Innovation Center, Inc. All rights reserved.
 # SPDX-License-Identifier: MIT
 
-TOPDIR=$(realpath $(dirname $(readlink -f $0))/..)
+if [ ! -f /work/oe-core/scripts/contrib/patchreview.py ]; then
```
lumag, line 5: I really don't like the idea of hard-coding the `/work/oe-core/` all over the place.

ci/yocto-check-layer.sh
```diff
@@ -1,25 +1,30 @@
-#!/bin/sh -e
+#!/bin/bash -e
```
lumag, line 1: Why?

quaresmajose:
This is a prototype to see if I can us the kas-container and it seems to work. I'm now going to put everything right and write the commits as they should be. Should have put this in draft to not draw attention.

ci/yocto-patchreview.sh
quaresmajose: I have changed this where we can check if we are within the container

quaresmajose:
> > _ No description provided. _
> 
> Could you please describe briefly, why are you performing those changes?

I added a description on PR 

> None of the commits have commit messages. Please fix that.

done

quaresmajose:
Currently this is blocked because I need to use the container from kas next branch until a new version is released.

ci/yocto-check-layer.sh
lumag: can you use `.` instead?

.github/workflows/build-yocto.yml
```diff
@@ -110,6 +110,9 @@ jobs:
 
       - name: Kas build
         run: |
+          # WIP: try container from kas next branch
```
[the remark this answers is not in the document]
quaresmajose: this is because we need the python3-cairo inside the container that is only available on the next branch
https://groups.google.com/g/kas-devel/c/h0DrRouDIOc

ci/yocto-check-layer.sh
quaresmajose: no, it fails!
```
Error: /build/oe-init-build-env doesn't exist!
Please run this script in oe-init-build-env's directory.
2025-05-07 14:57:09 - ERROR    - Shell returned non-zero exit status
2025-05-07 14:57:09 - ERROR    - Command "/bin/bash -c /repo/ci/yocto-check-layer.sh" failed with error 1
```

quaresmajose: the yocto-check-layer-wrapper also uses the source so I prefer doing the same
https://git.openembedded.org/openembedded-core/tree/scripts/yocto-check-layer-wrapper#n37

lumag: ack

.github/workflows/build-yocto.yml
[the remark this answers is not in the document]
lumag: thx

sbanerjee-quic:
> Currently this is blocked because I need to use the container from kas next branch until a new version is released.

HI @quaresmajose I see a 'latest' tag created. Is that usable for use? Or you do want to wait till 4.8? 
Is there a known date for 4.8 release?

ci/yocto-patchreview.sh
```diff
@@ -4,13 +4,13 @@
 
 TOPDIR=$(realpath $(dirname $(readlink -f $0))/..)
 
-# Ensure KAS workspace is outside of the checked out repo
-# Allows the caller to specify KAS_WORK_DIR, otherwise make temp one
-export KAS_WORK_DIR=$(realpath ${KAS_WORK_DIR:-$(mktemp -d)})
+if [ ! -d /build ] || [ ! -d /work/build ] || [ "$TOPDIR" != "/repo" ] ; then
```
ndechesne, line 7: I am not entirely sure we should do that.. this script (and check-layer too) should work with or without kas, and with or without kas-container.. I need to think about that..

ci/yocto-patchreview.sh
```diff
@@ -4,13 +4,13 @@
...
+    echo "ERROR: This script is designed to run inside kas-container"
```
ndechesne, line 8: another issue with this PR is that it breaks after this commit, it's not an atomic change.

ci/yocto-check-layer.sh
ndechesne: well, this is a really too bad.. i would prefer to avoid forcing bash ;-(

ci/yocto-check-layer.sh
```diff
@@ -1,25 +1,31 @@
...
 
-# Ensure KAS workspace is outside of the checked out repo
-# Allows the caller to specify KAS_WORK_DIR, otherwise make temp one
-export KAS_WORK_DIR=$(realpath ${KAS_WORK_DIR:-$(mktemp -d)})
+if [ ! -d /build ] || [ ! -d /work/build ] || [ "$TOPDIR" != "/repo" ] ; then
+    echo "ERROR: This script is designed to run inside kas-container"
+    exit 1
+fi
+
+# Creates a temporary build directory to run the yocto-check-layer
```
ndechesne, line 12: *if* we really need to stop using the wrapper, it should be done in 2 patches. 1 patch to remove the wrapper, 1 patch to switch to kas-container.

ci/yocto-check-layer.sh
```diff
@@ -1,25 +1,31 @@
...
+# script to avoid a contaminated environment.
+BUILDDIR="$(mktemp -p /build -d -t yocto-check-layer-XXXX)"
+source /work/oe-core/oe-init-build-env $BUILDDIR
+git -c advice.detachedHead=false clone --quiet --shared /repo meta-qcom
```
ndechesne, line 16: why? we might want to run the script on the repo with uncommitted changes. this looks complicated to me..

ci/yocto-patchreview.sh
quaresmajose: But it currenlty already runs using kas so it depends on kas. We can change that and add some environment variables plus some helpers scripts to abrtach the use of this scripts with/without kas

ci/yocto-patchreview.sh
quaresmajose: We need to merge the full PR to have it in proper shape. Squasing all commits did't make sense in my opinion.

ci/yocto-check-layer.sh
quaresmajose: I can do that

ci/yocto-check-layer.sh
quaresmajose: This is the same repo but it is copied to another palce. If you run the kas-container locally without setting KAS_WORK_DIR you will get the build folder inside th meta-qcom repo and this will break yocto-check-layer.
The above git clone fix it.

quaresmajose:
> > Currently this is blocked because I need to use the container from kas next branch until a new version is released.
> 
> HI @quaresmajose I see a 'latest' tag created. Is that usable for use? Or you do want to wait till 4.8? Is there a known date for 4.8 release?

I have changed to use the 'latest' tag one one addiction commit, I think we can use it forevever as I explined on tthe commit. I don't know the schedule and so I have no idea when the next version will be available.

ci/yocto-patchreview.sh
ndechesne, line 7: This script used to just run before this PR without running kas. e.g. 

git clone meta-qcom
./ci/yocto-patchreview.sh

and same for yocto-check-layer. 

Somehow I think we should not break that. Using kas-container is our own CI, we should not force it to our users.

quaresmajose: The scrit already use _kas_ but it runs on the host machine and not in the contianer
https://github.com/qualcomm-linux/meta-qcom/blob/f0d28c43c01b52ea382bd2e1dc2d91bb50703ad2/ci/yocto-patchreview.sh#L12

The dfirence is that:

- before: we call it directly from the host with `./ci/yocto-patchreview.sh` and _kas_ is called inside the script.

- now: we call it with the _kas-container_ helper from the host `kas-container shell ci/base.yml -c /repo/ci/yocto-patchreview.sh` and inside the script the _kas_ no longer used

.github/workflows/build-yocto.yml
```diff
@@ -31,7 +31,7 @@ jobs:
 
       - name: Run kas lock
         run: |
-          kas dump --update --lock --inplace ci/base.yml
+          kas-container dump --update --lock --inplace ci/base.yml
```
ricardosalveti, line 35: Don't we need to make the kas-container script available by downloading it first?

ci/yocto-patchreview.sh
ricardosalveti: Can't we make both cases work here? Maybe check if kas is available and use it directly instead of the container?

While this is CI specific, I agree that allowing it to be executed manually by the user, without requiring a container, is a good think to keep.

quaresmajose: I can try that aprouch to make the sripts works in both cases but will not be easy because the filesytem stucture is diferent inside the container.

.github/workflows/build-yocto.yml
```diff
@@ -10,6 +10,7 @@ on:
 env:
```
ricardosalveti, line 10: Typo on the commit message, should be 'container' instead of 'continer'.

.github/workflows/build-yocto.yml
```diff
@@ -17,6 +17,11 @@ jobs:
     if: github.repository == 'qualcomm-linux/meta-qcom'
     runs-on: [self-hosted, x86]
     steps:
+      - name: Update kas-container
+        run: |
+          wget -qO ${KAS_REPO_REF_DIR}/kas-container https://github.com/siemens/kas/raw/refs/heads/master/kas-container
```
ricardosalveti, line 22: At some point we should probably use a fixed revision here, but fine to use latest for now.

ricardosalveti: @quaresmajose can you reorder the commits here to first add the script and then switch to it?

quaresmajose: > At some point we should probably use a fixed revision here, but fine to use latest for now.

This is a requirement for now to use the latest helper and the latest container because we require a patch in there yet to be released. When the next 4.8 release comes out we can pin it.

> @quaresmajose can you reorder the commits here to first add the script and then switch to it?

Done

ci/yocto-patchreview.sh
quaresmajose: I will try this approach

lumag:
Are there any remaining issues? Or can this now be merged?

quaresmajose:
> Are there any remaining issues? Or can this now be merged?

I need to make the ci sripts works in both cases, with and without the container https://github.com/qualcomm-linux/meta-qcom/pull/846#discussion_r2113987574

quaresmajose:
This ready for reviwe and depends now on this https://github.com/qualcomm-linux/meta-qcom/pull/934

ci/yocto-patchreview.sh
quaresmajose: Intead if/else logic I added two new kas-shell helper scripts, one for the container and another for the host

ricardosalveti:
Please rebase.

quaresmajose:
> Please rebase.

done

lumag (approved):

PR #849, opened 2025-04-25
RFC: Add layer to make images 'Axiom ready'

koenkooi:
This adds the 'meta-axiom-ci' layer as well as a KAS fragment to enable it:

`kas build meta-qcom/ci/qcs6490-rb3gen2-core-kit.yml:meta-qcom/ci/distro.yml:meta-qcom/ci/axiom.yml`

Will build _core-image-base_, with adb, nmcli and all kernel modules installed. Images can be 'converted' by adding a bbappend that inherits the `axiom-image` bbclass.

A bbappend for `linux-yocto-dev` disables USB mass storage support to make adb USB gadget activation more robust, additional bbappends might be needed for platforms that use a different kernel.


lumag (commented):
I think, the commit message (and likely a small README) should mention, what is Axiom. Otherwise LGTM.

koenkooi:
> I think, the commit message (and likely a small README) should mention, what is Axiom. Otherwise LGTM.

Good point, I've added a placeholder README.md that I need to fill in some more.

ricardosalveti (dismissed):
Fine to merge as is and migrate in parallel to a different repo. @ndechesne we basically need to decide where we want to land the axiom related changes, separated repo might work better, knowing we need the axiom stuff on builds from meta-qcom and from meta-qcom-distro.

ndechesne:
sorry to join the discussion late, but that's because you did it during my vacation :) 

I am not much in favor of this PR unfortunately.. I don't know what led us to want to do it this way. I don't see anything here that cannot be done as a simple 'test image' in meta-qcom directly. Using a layer seems very expensive, and using a layer in its own repo even more. 

while adbd and nmcli needs came from internal discussions to generate an image for our internal axiom thingy, i don't think these features should be done in an 'axiom' layer. What's wrong with a test image that would enable adbd and nmcli?

lumag:
> while adbd and nmcli needs came from internal discussions to generate an image for our internal axiom thingy, i don't think these features should be done in an 'axiom' layer. What's wrong with a test image that would enable adbd and nmcli?

adbd and nmcli require meta-oe and meta-networking, so we can't enable it straight away

ndechesne:
Then back to my initial comment, why are we not adding that in meta-qcom-distro instead?

ricardosalveti:
> sorry to join the discussion late, but that's because you did it during my vacation :)
> 
> I am not much in favor of this PR unfortunately.. I don't know what led us to want to do it this way. I don't see anything here that cannot be done as a simple 'test image' in meta-qcom directly. Using a layer seems very expensive, and using a layer in its own repo even more.
> 
> while adbd and nmcli needs came from internal discussions to generate an image for our internal axiom thingy, i don't think these features should be done in an 'axiom' layer. What's wrong with a test image that would enable adbd and nmcli?

There are some additional overrides associated with the changes that are not image specific (e.g. additional kernel config fragment).

And while we could land in distro, in the end we want to test meta-qcom and meta-qcom-distro images on axiom.

lumag:
> Then back to my initial comment, why are we not adding that in meta-qcom-distro instead?

Being able to test meta-qcom without depending on meta-qcom-distro.

lumag:
@koenkooi I assume it is no longer needed and can be closed. Correct?

PR #850, opened 2025-04-25
packagegroup-qcom: use RDEPENDS instead of RRECOMMENDS

lumag:
All initramfs-\*-image recipes use NO_RECOMMENDS to keep the size minimal. The commit ebf3168 ("packagegroups: Define QCOM userspace helper packagegroups") made initramfs-tiny-image to pull     packagegroup-qcom-* packages, which RRECOMMENDs platform packages     instead of RDEPENDing on them, thus all testing initramfs images have     lost corresponding Qualcomm tools.

Make packagegroup-qcom use RDEPENDS instead of RRECOMMENDS in order to     brign tool packages back.

### DIFF
--- a/recipes-bsp/packagegroups/packagegroup-qcom.bb
+++ b/recipes-bsp/packagegroups/packagegroup-qcom.bb
@@ -7,13 +7,13 @@ PACKAGES = " \
     ${PN}-boot-additional \
 "
 
-RRECOMMENDS:${PN}-boot-essential = " \
+RDEPENDS:${PN}-boot-essential = " \
     pd-mapper \
     qrtr \
     rmtfs \
     tqftpserv \
 "
 
-RRECOMMENDS:${PN}-boot-additional = " \
+RDEPENDS:${PN}-boot-additional = " \
     fastrpc \
 "


recipes-test/images/initramfs-tiny-image.bb
```diff
@@ -10,8 +10,11 @@ PACKAGE_INSTALL = " \
 
 PACKAGE_INSTALL += " \
     debugcc \
-    packagegroup-qcom-boot-additional \
-    packagegroup-qcom-boot-essential \
+    fastrpc \
+    pd-mapper \
+    qrtr \
+    rmtfs \
+    tqftpserv \
```
vkraleti, line 17: How about updating packagegroup-qcom-* to use RDEPENDS instead?"

lumag: I thought about it, but it would prevent using`BAD_RECOMMENDATIONS` in order to skip recommendations. But... Maybe it's still a good idea.

lumag:
> The change is to use RDEPENDS instead of RRECOMMENDS, but the commit message indicates the opposite.

fixed

vkraleti (approved):

ricardosalveti (approved):

PR #843, opened 2025-04-17
Switch fastrpc recipe to GitHub version

quic-ekangupt:
Switched the fastrpc recipe from the CodeLinaro version to the GitHub version. This update includes:

1. Transition to the GitHub version of fastrpc, ensuring compatibility with all features supported by the CodeLinaro version.
2. Exposing header files to be used by developers for easier integration and development.
3. Addition of new features and improvements to enhance stability and performance.

The GitHub version of fastrpc has been tested on multiple platforms, confirming its reliability and robustness.

PR #851, opened 2025-04-25
fastrpc: simplify DSP library location handling

lumag:
The /usr/lib/rfsa subdir is supposed to be used for DSP libraries provided by external components. Instead of reusing it for mounting the DSP libraries there, pass `DSP_LIBRARY_PATH` and `ADSP_LIBRARY_PATH` variables to the daemons, pointing out the correct DSP library location.


lumag:
cc @quic-ekangupt @quic-bkumar

recipes-support/fastrpc/fastrpc/guess-dsp.sh
```diff
@@ -0,0 +1,54 @@
...
+
+if [ -z "$DSP_LIBRARY_PATH" -o ! -d "$DSP_LIBRARY_PATH" ] ; then
+	if [ -n "`ls /usr/share/qcom`" ] ; then
+		echo "DSP directory not detected, bailing out" 1>&2
+		exit 1
+	fi
+
+	echo "DSP directory not present, trying DSP partitions"
+
+	mkdir -p /usr/share/qcom/dsp
```
ricardosalveti, line 41: This will fail when the rootfs (or /usr) is mounted as ro. Can't we mount at a runtime specific rw directory instead, since DSP_LIBRARY_PATH is set later anyway?

lumag: I think it's better to provide this dir as a part of the recipe.

ricardosalveti: That works as well.

lumag: Done. Mounting under /usr/share is more of the band-aid for the platforms which provide neither dspso.bin (to be handled by the repackaging recipes) nor provide files in the hexagon-dsp-binaries repo.

ricardosalveti (approved):

recipes-support/fastrpc/fastrpc/0001-apps_std_fopen_with_env-account-for-domain-kinds-whe.patch
```diff
@@ -1,129 +0,0 @@
...
- 
-       // Account for slash char
--      absNameLen = std_strlen(dirName) + std_strlen(name) + 2;
-+      dirNameLen = std_strlen(dirName);
-+      absNameLen = dirNameLen + std_strlen(name) + 2;
-       VERIFYC(NULL != (absName = (char*)malloc(sizeof(char) * absNameLen)), AEE_ENOMEMORY);
-       if ('\0' != *dirName) {
-          std_strlcpy(absName, dirName, absNameLen);
-+         if (!std_strcmp(absName + dirNameLen - 4, "adsp"))
-+            std_memscpy(absName + dirNameLen - 4, 4, dspName, 4);
```
Chennak-quic, line 67: As guess-dsp.sh is setting DSP_LIBRARY_PATH to "/usr/share/qcom/sm8250/Thundercomm/RB5/dsp" and DSP shared objects are present in adsp and cdsp sub folders under "/usr/share/qcom/sm8250/Thundercomm/RB5/dsp", I think FastRPC framework will not be able to load if we drop this change of appending domain information (adsp/cdsp) when using "git://git.codelinaro.org/linaro/qcomlt/fastrpc.git".

lumag: It should be able to load those from under the ADSP_LIBRARY_PATH.

recipes-support/fastrpc/fastrpc/guess-dsp.sh
```diff
@@ -0,0 +1,53 @@
...
+			DSP_LIBRARY_PATH=/usr/share/qcom/sdm845/Thundercomm/db845c/dsp
+			;;
+		*"Robotics RB1"*)
+			DSP_LIBRARY_PATH=/usr/share/qcom/qcm2290/Thundercomm/RB1/dsp
+			;;
+		*"QRB4210 RB2"*)
+			DSP_LIBRARY_PATH=/usr/share/qcom/qrb4210/Thundercomm/RB2/dsp
+			;;
+		*"Robotics RB5"*)
+			DSP_LIBRARY_PATH=/usr/share/qcom/sm8250/Thundercomm/RB5/dsp
```
quic-ekangupt, line 28: can the paths for qcm6490, sa8775p and qcs8300 machines be added here? Or should those be added separately?

quic-ekangupt:
As per discussion here: (https://github.com/qualcomm-linux/meta-qcom/pull/843/files#r2052657232)
I'm testing changes to add service files to (https://github.com/quic/fastrpc).

Do you see any problem with this?

recipes-support/fastrpc/fastrpc/guess-dsp.sh
lumag: Separate PR, please

lumag:
> As per discussion here: (https://github.com/qualcomm-linux/meta-qcom/pull/843/files#r2052657232) I'm testing changes to add service files to (https://github.com/quic/fastrpc).
> 
> Do you see any problem with this?

Switching to that version of the fastrpc is a separate topic. So, no, I don't see a problem with that, but we'd need to work out how to handle the path detection.

quic-ekangupt (approved):

PR #847, opened 2025-04-23
Rb8 enablement 

vkraleti:
Added machine configuration, CI support, and kernel patches required to build RB8 DTB. The LAVA test job template is not included for now; it will be added once the hardware is available for testing.

Currently, the RB8 DT naming is not finalized. This PR will remain in draft state until this is concluded.

PR #852, opened 2025-04-25
Add a way to test upcoming firmware

lumag:
The main meta-qcom layer is settled on using linux-firmware and hexagon-dsp-binaries. However it might make sense to be able to test firmware (and boot firmware) before it gets released. Add a sublayer which provides support for testing the pending firmware changes from QArtifactory.


ricardosalveti:
I understand that having this as an additional meta layer makes sense, but also not a fan of having extra layers mixed up with meta-qcom, at least based on how the files are currently structured, would be nice to find a way for this to be part of meta-qcom and enabled with a custom override.

Also, for boot firmware we are basically using this to test what is already released downstream, would probably just make more sense to propose the updates once they are available directly, as in the end we don't have other providers and don't expect this to be updated outside qualcomm linux releases (at least not for now). Once we have weekly development snapshots this would make more sense to have (to use for continuous testing).

ricardosalveti:
Updated boot firmware for qcs9100 is also part of https://github.com/qualcomm-linux/meta-qcom/pull/847, and I'm about to propose a similar update for 6490 (won't conflict much in the end).

vkraleti:
If intention here is to test latest drop of firmware before accepting, can dev-upstream be used to pull latest version?

lumag:
> If intention here is to test latest drop of firmware before accepting, can dev-upstream be used to pull latest version?

I can give it a thought, but do you have any idea on your mind how to implement it?

vkraleti:
> > If intention here is to test latest drop of firmware before accepting, can dev-upstream be used to pull latest version?
> 
> I can give it a thought, but do you have any idea on your mind how to implement it?

Unfortunately, there is no predictable way to know what the latest firmware will be. Generally, the one with the highest number is the latest. May be firmware recipes can scan all the directory names and pick the one with the highest number as dev-upstream variant.

lumag:
> > > If intention here is to test latest drop of firmware before accepting, can dev-upstream be used to pull latest version?
> > 
> > I can give it a thought, but do you have any idea on your mind how to implement it?
> 
> Unfortunately, there is no predictable way to know what the latest firmware will be. Generally, the one with the highest number is the latest. May be firmware recipes can scan all the directory names and pick the one with the highest number as dev-upstream variant. 

This is AUTOREV, not devupstream. And I'd rather not implement such scanning. Traversing directories on the web server can open a can of worms.

sbanerjee-quic (commented):
> firmware-qcom-qcs6490: package latest firmware for QCS9100

The changes are for qcs9100, but subject says qcs6490 @lumag

ricardosalveti:
I would prefer not mixing one layer into another here (as might also have additional layers as we saw with the axiom related changes), should we move everything under meta-qcom into a meta-qcom folder then?

Would be a bit similar to https://github.com/jonmason/meta-arm and https://git.yoctoproject.org/meta-ti/tree/.

lumag:
@ricardosalveti moved meta-qcom to the subdir

ricardosalveti:
> @ricardosalveti moved meta-qcom to the subdir

Sure, this would work for me, might just need to update ci/kas/.github and README to point to the new directory.

@ndechesne @vkraleti @sbanerjee-quic this would allow for us to host additional meta layers that are relevant to meta-qcom (e.g. tip testing, axiom integration, extras, etc) as part of this repo, a common behavior that is found in several other bsp layers.

mwasilew:
I think we're missing test jobs using the new firmware. Or is the new firmware already included in the default builds?

koenkooi:
> > @ricardosalveti moved meta-qcom to the subdir
> 
> Sure, this would work for me, might just need to update ci/kas/.github and README to point to the new directory.
> 
> @ndechesne @vkraleti @sbanerjee-quic this would allow for us to host additional meta layers that are relevant to meta-qcom (e.g. tip testing, axiom integration, extras, etc) as part of this repo, a common behavior that is found in several other bsp layers.

This would be a nice solution to being able to host more BSP related layers. I do warn about moving non-BSP layers like meta-qcom-distro into this, it makes coupling the DISTRO to the BSP a bit too easy, usually breaking standalone BSP usage.

PR #854, opened 2025-04-29
Add recipes-video: qcom-videocodec

malathi-gottam:
Change-Id: I6e7daae04a1bb4a49639b8777fed0d3c12973967

### DIFF
--- a/recipes-video/qcom-videocodec/qcom-videocodec_1.0.bb
+++ b/recipes-video/qcom-videocodec/qcom-videocodec_1.0.bb
@@ -1,0 +1,16 @@
+DESCRIPTION = "QCOM Iris Video drivers"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://vidc/src/resources.c;beginline=1;endline=2;md5=801f80980d171dd6425610833a22dbe6"
+SRCREV = "57f29f46b23b41c9211d542b58d2991c91fe59cf"
+
+SRC_URI = "${SRCPROJECT};branch=${SRCBRANCH};destsuffix=vendor/qcom/opensource/video-driver"
+
+SRCPROJECT = "git://git.codelinaro.org/clo/le/platform/vendor/opensource/video-driver.git;protocol=https"
+SRCBRANCH = "video.qclinux.0.0"
+S = "${WORKDIR}/vendor/qcom/opensource/video-driver"
+
+inherit module
+
+MAKE_TARGETS = "modules"
+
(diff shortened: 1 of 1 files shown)


recipes-video/qcom-videocodec/qcom-videocodec_1.0.bb
```diff
@@ -1,6 +1,6 @@
 DESCRIPTION = "QCOM Iris Video drivers"
 LICENSE = "GPL-2.0-only"
-LIC_FILES_CHKSUM = "file://vidc/src/resources.c;beginline=1;endline=2;md5=801f80980d171dd6425610833a22dbe6"
+LIC_FILES_CHKSUM = "file://vidc/src/resources.c;beginline=1;endline=2;md5=550794465ba0ec5312d6919e203a55f9"
 SRCREV = "57f29f46b23b41c9211d542b58d2991c91fe59cf"
 
 SRC_URI = "${SRCPROJECT};branch=${SRCBRANCH};destsuffix=vendor/qcom/opensource/video-driver"
```
vkraleti, line 6: I don't see a need to define SRCPROJECT and SRCBRANCH and use them in SRC_URI. The URL and branch can be directly mentioned in SRC_URI itself. Also, destsuffix is not required.

lumag:
> @malathi-gottam IMO this recipe should land in recipes-multimedia group. Don't introduce recipes-video section.

Maybe recipes-kernel? What do other BSP do with module recipes?

recipes-video/qcom-videocodec/qcom-videocodec_1.0.bb
lumag: I always assumed that those variables are to help with overriding. If it's not the case, please inline them.

recipes-video/qcom-videocodec/qcom-videocodec_1.0.bb
```diff
@@ -0,0 +1,16 @@
...
+
+SRCPROJECT = "git://git.codelinaro.org/clo/le/platform/vendor/opensource/video-driver.git;protocol=https"
+SRCBRANCH = "video.qclinux.0.0"
+S = "${WORKDIR}/vendor/qcom/opensource/video-driver"
+
+inherit module
+
+MAKE_TARGETS = "modules"
+
+MODULES_INSTALL_TARGET = "modules_install"
```
ricardosalveti: No need, already the default value.

ricardosalveti:
+1 for `recipes-kernel/foo-module`

PR #861, opened 2025-05-01
ci: Convert to yet another artifact upload API

doanac:
The current artifact upload logic requires the self-hosted runner and the file-server to share the same NFS volume as a means of uploading artifacts. This change decouples that so that we could use any runner and still upload artifacts.

We do this by leveraging the GITHUB_TOKEN included in every CI job (including PRs from forked repositories). This is an ephemeral token only valid during the lifetime of the workflow. By granting it read access to the "security-events" API, our file-serve can assert the token it gets is a valid qualcomm-linux token before allowing an upload to happen.

The upload logic gets a little hard to maintain with inline yaml and shell script, so I've created a Python helper to do the upload. It has a few handy features:

 * parallel uploads
 * upload progress
 * better logging/info messages

### DIFF
--- a/.github/workflows/publish_artifacts.py
+++ b/.github/workflows/publish_artifacts.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+# Copyright (c) 2025 Qualcomm Innovation Center, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
+
+from multiprocessing import Pool
+import os
+import sys
+from typing import List
+
+import requests
+
+gh_token = os.environ["GITHUB_TOKEN"]
+
+
+def upload_file(args):
+    try:
+        url, base, name = args
+
+        headers = {
+            "Authentication": f"Bearer {gh_token}",
+        }
+        r = requests.put(url, headers=headers, allow_redirects=False)
+        if not r.ok:
+            return name, f"Unable to get signed url HTTP_{r.status_code} - {r.text}"
+
+        url = r.headers["location"]
+        path = os.path.join(base, name)
+        r = requests.put(
+            url, data=open(path, "rb"), headers={"Content-type": "application/octet-stream"}
+        )
+        if not r.ok:
+            return name, f"Unable to upload content HTTP_{r.status_code} - {r.text}"
+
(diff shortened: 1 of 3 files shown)


doanac:
@ricardosalveti @ndechesne - I think this is ready to look at

ndechesne:
Looks ok to me. One question though. publish_artifacts.py is not specific to meta-qcom and will be needed in every repo which wants to publish files. should we host this file in a common location?

.github/workflows/publish_artifacts.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3
+
```
ricardosalveti, line 2: Please add:

```
# Copyright (c) 2025 Qualcomm Innovation Center, Inc. All rights reserved.
# SPDX-License-Identifier: MIT
```

ricardosalveti:
> Looks ok to me. One question though. publish_artifacts.py is not specific to meta-qcom and will be needed in every repo which wants to publish files. should we host this file in a common location?

Yeah, we already got demand to have the same script in at least 4 different repos.

doanac:
We can do this a couple ways:
* create a repo with common scripts, eg `qualcomm-linux/ci-helpers`
* create a GH action for publishing

I'm thinking a github action might work out best but I haven't done one in while - no strong opinion. Let me know what you want.

ricardosalveti:
+1 for a custom action.

doanac:
I agree a custom action is the correct approach. This is going to require us to get approval for a new repo to do this. 

Do we want to proceed with this for now or hold off until we get all the legal stuff in place?

ricardosalveti:
I'm ok to merge this now (had only one review/comment to add missing license) and later migrate to a custom action once the new repo is in place.

ndechesne:
sure. that works for me too. @doanac can you please fix the SPDX comment first, then we can merge

doanac:
force push and rebased on latest main.

ndechesne (approved):

ndechesne:
sorry.. clicked on 'disable auto merge' by error..

PR #867, opened 2025-05-06
Disable earlycon for all machines

lumag:
Use of earlycon makes it harder to debug issues (see [RB8 boot logs](https://github.com/qualcomm-linux/meta-qcom/actions/runs/14848561348/job/41689454218) for example). Disable earlycon in order to make it possible to get the full boot log.

### DIFF
--- a/conf/machine/include/qcom-qcm2290.inc
+++ b/conf/machine/include/qcom-qcm2290.inc
@@ -8,7 +8,7 @@ QCOM_VFAT_SECTOR_SIZE = "512"
 
 require conf/machine/include/arm/armv8a/tune-cortexa53.inc
 
-KERNEL_CMDLINE_EXTRA ?= "earlycon clk_ignore_unused pd_ignore_unused"
+KERNEL_CMDLINE_EXTRA ?= "clk_ignore_unused pd_ignore_unused"
 
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-qcom-boot-essential \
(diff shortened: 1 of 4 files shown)


vkraleti:
Sometimes, earlycon logs are needed for debugging. It might be worth to enable at least for debug builds. If you agree, I can submit a PR to enable them conditionally based on the DEBUG_BUILD flag.

lumag:
> Sometimes, earlycon logs are needed for debugging. It might be worth to enable at least for debug builds. If you agree, I can submit a PR to enable them conditionally based on the DEBUG_BUILD flag.

I think it's as easy to set `DEBUG_BUILD = "1"` as `KERNEL_CMDLINE_EXTRA:append = " earlycon"`, so I don't really see a point.

lumag:
I can probably expand / explain my answer. For me there is a difference between a debug build (where you get some useful extras, but generally you are debugging the image / software issue, the board boots fine) and the early kernel debugging (when nothing boots, you need some early debug, etc). Normal users are not expected to perform the second task. As such, there is no need to enable `earlycon` for anybody except kernel developers.
Moreover, as you can see from the linked log, enabling earlycon hides useful logs (and that's what prompted me to push this change).

vkraleti (approved):

ricardosalveti:
While it makes sense to drop it, it is the first time I hear it is causing garbage during boot, but looking it seems it is only fine in rb3gen2, there is quite a bit of garbage in rb1 and rb8.

ricardosalveti (approved):

PR #872, opened 2025-05-08
uki-esp-image: fix order of tasks

lumag:
Fix the order of tasks in uki-esp-image class so that the Linux UKI file has a chance to land in the generated VFAT image.


ricardosalveti (approved):
Thanks, was just starting to look at this issue and also noticed the actual UKI was not included in the esp image.

vkraleti (approved):

PR #722, opened 2025-01-28
esp-qcom-image: convert to OE-core uki.bbclass

koenkooi:
This new class uses the OE-core uki.bbclass infrastructure to assemble an
ESP image with systemd-boot + UKI. It uses method #2 for config-less booting.

Systemd-boot is installed through package management, the UKI is picked
up from deploy. 

Signed-off-by: Koen Kooi [koen.kooi@oss.qualcomm.com](mailto:koen.kooi@oss.qualcomm.com)
Fixes https://github.com/qualcomm-linux/meta-qcom/issues/707

PR #873, opened 2025-05-08
uki: add back support for console and KERNEL_CMDLINE_EXTRA in the kernel cmdline

ricardosalveti:
Both were removed as a side effect of https://github.com/qualcomm-linux/meta-qcom/pull/722

### DIFF
--- a/recipes-kernel/images/esp-qcom-image.bb
+++ b/recipes-kernel/images/esp-qcom-image.bb
@@ -12,7 +12,7 @@ inherit image uki uki-esp-image
 
 UKI_FILENAME = "${EFI_LINUX_IMG}"
 
-UKI_CMDLINE = "root=${QCOM_BOOTIMG_ROOTFS} rw rootwait"
+UKI_CMDLINE = "root=${QCOM_BOOTIMG_ROOTFS} rw rootwait console=${KERNEL_CONSOLE}"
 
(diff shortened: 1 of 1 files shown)


ricardosalveti:
Console might be ok to be removed, but would need to be validated with all targets (and different kernel recipes).

KERNEL_CMDLINE_EXTRA is actually used by rb1.

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -13,6 +13,7 @@ inherit image uki uki-esp-image
 UKI_FILENAME = "${EFI_LINUX_IMG}"
 
 UKI_CMDLINE = "root=${QCOM_BOOTIMG_ROOTFS} rw rootwait console=${KERNEL_CONSOLE}"
+UKI_CMDLINE += "${@'' + d.getVar('KERNEL_CMDLINE_EXTRA') if d.getVar('KERNEL_CMDLINE_EXTRA') else ''}"
```
lumag, line 16: Why do we need first `'' +` part ?

ricardosalveti: No need, updated with a simplified get or '' version, leftover from when I was playing with the console argument (which then required a string append).

vkraleti (approved):

lumag (approved):

PR #877, opened 2025-05-09
CI: disable fail-fast for CI job submission

lumag:
It is usually important to get the image tested on all boards rather than stopping on the first failure. Specifi fail-fast:false to the submit-job in order to let all image tests finish if one of them fails.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,6 +35,7 @@ jobs:
     needs: prepare-job-list
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


ricardosalveti (approved):

lumag:
Strangely enough, other jobs still got cancelled.

lumag:
https://github.com/qualcomm-linux/meta-qcom/actions/runs/14934277004

ricardosalveti:
Possible yeah, we can merge and double check again, won't cause any harm.

lumag:
Then it would be nice to merge it before #876 , we currently have an easy way to reproduce CI issues.

ndechesne (approved):

PR #878, opened 2025-05-12
HexagonRPC and libSSC support

lumag:
Add recipes for HexagomRPC and libSSC, intenting to test sensors on RB* platforms.

Note, this PR doesn't provide full integration, there is no DSP path detection, no extra sensors files, etc.


dynamic-layers/openembedded-layer/recipes-support/libssc/libssc_git.bb
```diff
@@ -0,0 +1,25 @@
+HOMEPAGE = "https://libssc.dylanvanassche.be/"
```
ricardosalveti, line 1: Why is this recipe under dynamic-layers and not in meta-qcom? This is a complete new recipe.

dynamic-layers/recipes-bsp/packagegroups/packagegroup-qcom.bbappend
```diff
@@ -0,0 +1,3 @@
+RRECOMMENDS:${PN}-miscellaneous += " \
```
ricardosalveti, line 1: Dynamic-layers but without specifying the layer it depends on.

dynamic-layers/openembedded-layer/recipes-support/libssc/libssc_git.bb
ricardosalveti: libqmi protobuf-c protobuf-c-native are all in openembedded-layer, mind adding that to the commit message?

ricardosalveti (approved):

ndechesne:
Looks good to me, would still like Srini to have a look too.

PR #882, opened 2025-05-13
Refresh qcs9075-iq-9075-evk DT patches

vkraleti:
A newer v6 version of the qcs9075-iq-9075-evk DT patches (https://lore.kernel.org/all/20250429054906.113317-1-quic_wasimn@quicinc.com/) has been submitted with updated DT naming. Replace the current v5 patches with  v6 to reflect the new DT name.


lumag (approved):

koenkooi:
Does this create a point in the revision history where the build/boot will fail because the kernel patches have been updated and the machine.conf hasn't? If so, it might make more sense to squash these two commits into a single one. It's a bit ugly, bit it would avoid creating a failure point during bisect.

lumag:
> Does this create a point in the revision history where the build/boot will fail because the kernel patches have been updated and the machine.conf hasn't? If so, it might make more sense to squash these two commits into a single one. It's a bit ugly, bit it would avoid creating a failure point during bisect.

Either squash, or remove obsolete names first, update patches, add new DTB names. However I thought that this doesn't warrant the issues. Squashing might be fine though.

ricardosalveti:
+1 to squash.

lumag (changes requested):
Okay, the suggested way is to squash both commits

vkraleti:
> Okay, the suggested way is to squash both commits

Done.

koenkooi (approved):

ricardosalveti (approved):

sbanerjee-quic (approved):

PR #889, opened 2025-05-14
Split ramdisk initramfs images

lumag:
Split `initramfs-firmware-image` and `initramfs-firmware-rb12-image` into corresponding per-platform images.


ricardosalveti (approved):

koenkooi:
Do we want to lock down these images with COMPATIBLE_MACHINE as well?

lumag:
> Do we want to lock down these images with COMPATIBLE_MACHINE as well?

No. I don't like using `COMPATIBLE_MACHINE` as I frequently build all packages with `qcom-armv8a` machine.

vkraleti (approved):

PR #890, opened 2025-05-14
Improve firmware packaging

lumag:
Simplify handling of board-specific firmware in preparation to handling tip versions of the firmware.


lumag:
This is a split from #852

koenkooi (approved):
The changes are straightforward and look good to me.

PR #893, opened 2025-05-16
qcs9075-iq-9075-evk: fix QCOM_DTB_DEFAULT

lumag:
Commit d55a42e05d42 ("linux-yocto-dev: refresh qcs9075-iq-9075-evk DT patches") has updated KERNEL_DEVICETREE, but failed to update QCOM_DTB_DEFAULT. Thus dtb.bin isn't being created, which makes tests on the RB8 board fail.

Point QCOM_DTB_DEFAULT to point to the same DTB as KERNEL_DEVICETREE does.

### DIFF
--- a/conf/machine/qcs9075-iq-9075-evk.conf
+++ b/conf/machine/qcs9075-iq-9075-evk.conf
@@ -6,7 +6,7 @@ require conf/machine/include/qcom-qcs9100.inc
 
 MACHINE_FEATURES += "efi pci"
 
-QCOM_DTB_DEFAULT ?= "qcs9075-rb8"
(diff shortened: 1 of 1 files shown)


ricardosalveti:
Thanks, a reminder that if we don't get test results published as comments, that means it failed to run the tests.

ricardosalveti (approved):

lumag:
> Thanks, a reminder that if we don't get test results published as comments, that means it failed to run the tests.

I think we should get test resuls as required for the PR to be merged... Can we do that? @mwasilew ?

mwasilew:
> I think we should get test resuls as required for the PR to be merged... Can we do that?

I don't know. I need to read about PR policies.

vkraleti (approved):

PR #897, opened 2025-05-19
initramfs-test-full: Add more utils

vivpuar:
Adding the utilities net-tools and rng-tools in 'initramfs-test-full-image.bb'

- net-tools is required by the [scripts](https://github.com/qualcomm-linux/qcom-linux-testkit/pull/35/) in PR#35 in qcom-linux-testkit
- rng-tools is required by the [script](https://github.com/qualcomm-linux/qcom-linux-testkit/blob/main/Runner/suites/Kernel/FunctionalArea/baseport/rngtest/run.sh) to run rngtest

### DIFF
--- a/recipes-test/images/initramfs-test-full-image.bb
+++ b/recipes-test/images/initramfs-test-full-image.bb
@@ -11,6 +11,8 @@ PACKAGE_INSTALL += " \
     ncurses \
     ncurses-terminfo \
     ncurses-terminfo-base \
+    net-tools \
+    rng-tools \
     stress-ng \
     util-linux \
     util-linux-chrt \


lumag:
Please add a recipe for the testsuite instead of pulling dependencies manually. Declare all dependencies for a testsuite (including those already being part of the initramfs). Then simply install the testsuite as a part of the initramfs

ndechesne:
> Please add a recipe for the testsuite instead of pulling dependencies manually. Declare all dependencies for a testsuite (including those already being part of the initramfs). Then simply install the testsuite as a part of the initramfs

Why? How is this PR different from what we already have in this image? I don't understand.

lumag:
> > Please add a recipe for the testsuite instead of pulling dependencies manually. Declare all dependencies for a testsuite (including those already being part of the initramfs). Then simply install the testsuite as a part of the initramfs
> 
> Why? How is this PR different from what we already have in this image? I don't understand.

From my PoV we have had a random collection of tools there. If we are using it to run a testsuite, I'd really like to know, which packages are required for the tests and which are just random tools. E.g. I don't want to duplicate the list if Neil starts using our testsuite in QCLT CI

lumag:
Another option might be define a packagegroup. Would that be better from your POV?

ndechesne:
it's an image, i am fine to keep things simple in there. net-tools and rng-tools are fairly standard utilities, so I am ok to have them listed there. a packagegroup is useful if we need to make several image variants so that we don't repeat the list, IMHO.

lumag (approved):

PR #903, opened 2025-05-20
mesa: update 25.0.5 -> 25.1.0

quaresmajose:
Replace with the new MR https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/34867
- 0001-freedreno-Add-initial-A702-support.patch

Which also includes the other patches:
- 0002-freedreno-A702-fixes-for-deqp-vk.patch
- 0003-freedreno-fix-compilation.patch


lumag:
If you are picking up a patch from mesa MR, then you don't need 2nd and 3rd patches, both of them were folded into the main patch.

lumag (approved):

ndechesne (approved):

PR #908, opened 2025-05-22
Add qcom-next kernel recipe

vkraleti:
Add linux-qcom-next 6.15 recipe to compile the qcom-next kernel tree, which includes additional patches intended for upstream submission on top of the mainline kernel.

The qcom-next kernel tree will coexist with the linux-qcom-staging tree for a few weeks, until it catches up with all required features. Post that, the linux-qcom-staging tree will be deprecated and the corresponding recipe will be removed.


recipes-kernel/linux/linux-qcom-next/0001-drivers-gpu-drm-msm-registers-improve-reproducibilit.patch
```diff
@@ -0,0 +1,54 @@
...
+based on where and when the kernel was built as the full path is
+captured.
+
+Since all of the files that this tool is run on is under the drivers
+directory, this modifies the application to strip all of the path before
+drivers.  Additionally it prints <stripped> instead of the date.
+
+Both changes solve the reproducibility issue.
+
+Upstream-Status: Inappropriate
```
lumag, line 17: This is incorrect. Please send the patch upstream.

lumag (changes requested):
Please enable it in the CI builds in addition to linux-yocto-dev.

recipes-kernel/linux/linux-qcom-next_6.15.bb
```diff
@@ -0,0 +1,37 @@
...
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
+
+inherit kernel
+
+COMPATIBLE_MACHINE = "(qcom)"
+
+SRCBRANCH ?= "qcom-next"
+SRC_URI = " \
+            git://github.com/qualcomm-linux/kernel.git;protocol=https;branch=${SRCBRANCH} \
```
ndechesne, line 13: qcom-next branch is expected to be rebased regularly, like very frequently. I don't think we should use qcom-next branch in the recipe. We need to find a way to use tags instead.

recipes-kernel/linux/linux-qcom-next/0001-drivers-gpu-drm-msm-registers-improve-reproducibilit.patch
vkraleti: I just picked this change from linux-yocto-dev. Probably should consult patch author "Ryan Eatmon <reatmon@ti.com>" before taking upstream.

recipes-kernel/linux/linux-qcom-next_6.15.bb
lumag: You can't use SRCREV if it won't be present on a branch after the next rebase.

lumag: Most likely the only way is to:
- use some fixed commit (like 6.15-rc7) as a default
- upgrade SRCREV to `${AUTOREV}` if it is selected as a `virtual/kernel` provider.
See how it's handled by linux-yocto-dev.

recipes-kernel/linux/linux-qcom-next/0001-drivers-gpu-drm-msm-registers-improve-reproducibilit.patch
lumag: > Posted: https://lore.kernel.org/dri-devel/20250523-binrep-v1-1-c3a446518847@oss.qualcomm.com/

... and you got a response. Just posting a patch isn't enough.

> I'll backport it to qcom-next tree and drop from here.

You can not backport a patch that isn't accepted. Please post it again, fixing the review.

recipes-kernel/linux/linux-qcom-next_6.15.bb
vkraleti: Updated SRCREV to point to "qcom-next-6.15-20250609" commit. Added devupstream support with SRCREV pointing to AUTOREV for users intended to build with tip of the branch.

recipes-kernel/linux/linux-qcom-next/0001-drivers-gpu-drm-msm-registers-improve-reproducibilit.patch
vkraleti: This change is now part of qcom-next-6.15-202050609 tag. Dropped the .patch.

conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -10,6 +10,12 @@ QCOM_DTB_DEFAULT ?= "qcs6490-rb3gen2"
 
 KERNEL_DEVICETREE ?= " \
                       qcom/qcs6490-rb3gen2.dtb \
+                      "
+
+# These DTBs are not upstreamed and currenty exist only as patches
+# against linux-yocto-dev. Conditionally build them to avoid
+# conflicts with other kernel providers like linux-qcom-next.
+KERNEL_DEVICETREE:append:pn-linux-yocto-dev = " \
```
ndechesne: it's merged in 6.16. I believe we are about to release 6.16-rc1 qcom-next, should we wait for it?

conf/machine/include/qcom-base.inc
```diff
@@ -1,7 +1,7 @@
 # Common configurations and variables for Qualcomm platforms.
 
 # Provider for linux kernel
-PREFERRED_PROVIDER_virtual/kernel ?= "linux-yocto-dev"
+PREFERRED_PROVIDER_virtual/kernel ?= "linux-qcom-next"
```
ndechesne, line 4: before we do that, I would like to see the list of patches that might be missing in qcom-next vs linux-yocto-dev. I am sure there are, and we need to add them into qcom-next.

conf/machine/qcs6490-rb3gen2-core-kit.conf
vkraleti: Sure we can wait. 6.16 tagging not done yet. Once done I'll update to it.

ricardosalveti: Yeah, better if we can wait for 6.16.

conf/machine/include/qcom-base.inc
ricardosalveti: Yeah, we can propose the remaining ones that are not yet in qcom-next, my guess is that the tc9563 series is not yet there.

recipes-kernel/linux/linux-qcom-next_6.15.bb
```diff
@@ -0,0 +1,33 @@
...
+
+SRC_URI = "git://github.com/qualcomm-linux/kernel.git;protocol=https;branch=qcom-next"
+
+# To build bleeding edge qcom-next kernel set preferred
+# provider of virtual/kernel to 'linux-qcom-next-tip'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "linux-qcom-next-tip"
+SRCREV:class-devupstream = "${AUTOREV}"
+
+# tag: qcom-next-6.15-20250609
```
shashim-quic: do we prefer to name qcom-next kernel recipe with a version (_6.15  here) ? qcom-next would keep evolving following upstream rc tags. 

Shouldn't we add recipe on similar lines on linux-yocto-dev and name it linux-qcom-next.bb ?

vkraleti: Thanks for the suggestion. Will consider during next push.

ndechesne: I, too, would prefer not have the version in the filename

vkraleti: @shashim-quic @ndechesne  please check latest push.

conf/machine/include/qcom-base.inc
vkraleti: Dropped switching to qcom-next by default for now. Will recheck all the patches and enable via another PR.

conf/machine/qcs6490-rb3gen2-core-kit.conf
vkraleti: Please check latest push.

recipes-kernel/linux/linux-qcom-next.bb
```diff
@@ -0,0 +1,33 @@
...
+
+# To build bleeding edge qcom-next kernel set preferred
+# provider of virtual/kernel to 'linux-qcom-next-tip'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "linux-qcom-next-tip"
+SRCREV:class-devupstream = "${AUTOREV}"
+
+# tag: qcom-next-6.16.rc1-20250616
+SRCREV = "fdb4295652600f45a895cdcd5703d981eea57b2d"
+PV = "6.16+git"
```
lumag: It's not 6.16 yet, so this is incorrect.

recipes-kernel/linux/linux-qcom-next.bb
```diff
@@ -0,0 +1,33 @@
...
+
+COMPATIBLE_MACHINE = "(qcom)"
+
+SRC_URI = "git://github.com/qualcomm-linux/kernel.git;protocol=https;branch=qcom-next"
+
+# To build bleeding edge qcom-next kernel set preferred
+# provider of virtual/kernel to 'linux-qcom-next-tip'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "linux-qcom-next-tip"
+SRCREV:class-devupstream = "${AUTOREV}"
```
lumag: Won't this break the `BB_NO_NETWORK` support?

recipes-kernel/linux/linux-qcom-next.bb
```diff
@@ -0,0 +1,33 @@
...
+S = "${WORKDIR}/git"
+
+KERNEL_CONFIG ?= "defconfig"
+
+do_configure:prepend() {
+    if [ ! -f "${S}/arch/${ARCH}/configs/${KERNEL_CONFIG}" ]; then
+        bbfatal "KERNEL_CONFIG '${KERNEL_CONFIG}' was specified, but not present in the source tree"
+    else
+        cp '${S}/arch/${ARCH}/configs/${KERNEL_CONFIG}' '${B}/.config'
+    fi
```
lumag: Please change `KERNEL_CONFIG_COMMAND` instead to run `oe_runmake_call -C ${S} O=${B} defconfig`.

conf/machine/qcs6490-rb3gen2-core-kit.conf
lumag: This DT is now a part of the `qcom-next` branch. Please drop this patch.

recipes-kernel/linux/linux-qcom-next.bb
vkraleti: How about "6.16-rc1+git" ?

recipes-kernel/linux/linux-qcom-next.bb
vkraleti: This will break, but qcom-next is a moving target. What's the best way to reliably get the tip commit? Ask it to have an ability to build 'tip of the branch'. https://github.com/qualcomm-linux/meta-qcom/pull/908#discussion_r2102924468 

Is it acceptable to emit a warning when a devupstream build is attempted with `BB_NO_NETWORK`? 
Or should we explicitly add the qcom-next kernel to `BB_ALLOWED_NETWORKS`?

recipes-kernel/linux/linux-qcom-next.bb
lumag: ack

recipes-kernel/linux/linux-qcom-next.bb
lumag: Default to a known-good tag here, update it to `${AUTOREV}` if devupstream kernel is selected as `virtual/kernel`.

vkraleti: It is mentioned in line 20, am I missing something?

```
# tag: qcom-next-6.16.rc1-20250616
SRCREV = "fdb4295652600f45a895cdcd5703d981eea57b2d"
```

lumag: Default to a tag _here_, so that it doesn't interact with `BB_NO_NETWORK` by default.

recipes-kernel/linux/linux-qcom-next_6.15.bb
ricardosalveti: Why not _git.bb here?

recipes-kernel/linux/linux-qcom-next.bb
ricardosalveti: I believe it makes sense to leverage linux-yocto.inc / kernel-yocto.bbclass at some point, that would allow specifying KBUILD_DEFCONFIG (besides better integrating with dynamic config fragments and what not).

For now changing KERNEL_CONFIG_COMMAND would be enough.

recipes-kernel/linux/linux-qcom-next_6.15.bb
vkraleti: done.

recipes-kernel/linux/linux-qcom-next.bb
vkraleti: I updated SRCREV to a tag, how ever with `BB_NO_NETWORK = "1" ` the error seems still existing as tags are also considered as  "floating" references. Need to find a better way to handle this.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -0,0 +1,28 @@
...
+PV = "${LINUX_VERSION}+git"
+
+# tag: qcom-next-6.16.rc1-20250616 fdb4295652600f45a895cdcd5703d981eea57b2d
+SRCREV ?= "qcom-next-6.16.rc1-20250616"
+
+# To build tip of qcom-next branch set preferred
+# virtual/kernel provider to 'linux-qcom-next-upstream'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "linux-qcom-next-upstream"
+SRCREV:class-devupstream ?= "${AUTOREV}"
```
lumag, line 29: This will still break `BB_NO_NETWORK`. Please make sure that the recipe works without network access unless the devupstream is actually selected to be built.

recipes-kernel/linux/linux-qcom-next.bb
lumag: Yeah... Please use SRCREV instead.

recipes-kernel/linux/linux-qcom-next_git.bb
ricardosalveti: `SRCREV:class-devupstream ?= '${@oe.utils.conditional("PREFERRED_PROVIDER_virtual/kernel", "linux-qcom-next-upstream", "${AUTOREV}", "fdb4295652600f45a895cdcd5703d981eea57b2d", d)}'` ?

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -0,0 +1,28 @@
...
+COMPATIBLE_MACHINE = "(qcom)"
+
+SRC_URI = "git://github.com/qualcomm-linux/kernel.git;branch=qcom-next;protocol=https"
+
+LINUX_VERSION ?= "6.16.rc1"
+
+PV = "${LINUX_VERSION}+git"
+
+# tag: qcom-next-6.16.rc1-20250616 fdb4295652600f45a895cdcd5703d981eea57b2d
+SRCREV ?= "qcom-next-6.16.rc1-20250616"
```
ricardosalveti: Yeah, better to go with the git revision instead.

recipes-kernel/linux/linux-qcom-next_git.bb
vkraleti: I already tried it and is not helping. 

`ERROR: meta-qcom/recipes-kernel/linux/linux-qcom-next_git.bb: AUTOREV/SRCPV set too late for the fetcher to work properly, please set the variables earlier in parsing. Erroring instead of later obtuse build failures.`

I see errors with linux-yocto-dev kernel as well. 

```
ERROR: ExpansionError during parsing oe-core/meta/recipes-kernel/linux/linux-yocto-dev.bb

bb.data_smart.ExpansionError: Failure expanding variable fetcher_hashes_dummyfunc[vardepvalue], 
expression was ${@bb.fetch.get_hashvalue(d)} which triggered exception NetworkAccess: Network access
disabled through BB_NO_NETWORK (or set indirectly due to use of BB_FETCH_PREMIRRORONLY) but access
requested with command git -c gc.autoDetach=false -c core.pager=cat -c safe.bareRepository=all -c
clone.defaultRemoteName=origin ls-remote https://git.yoctoproject.org/linux-yocto-dev.git  (for url
https://git.yoctoproject.org/linux-yocto-dev.git)
The variable dependency chain for the failure is: fetcher_hashes_dummyfunc[vardepvalue]
```

Seems OE builds got more stricter now and not allowing to use AUTOREV anymore in no network builds.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -0,0 +1,29 @@
...
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
+
+inherit kernel
+
+COMPATIBLE_MACHINE = "(qcom)"
+
+SRC_URI = "git://github.com/qualcomm-linux/kernel.git;branch=qcom-next;protocol=https"
+
+LINUX_VERSION ?= "6.16.rc1"
```
lumag: `-rc1`

lumag (changes requested):
You are still adding a recipe which is not being tested in the CI.

vkraleti:
> You are still adding a recipe which is not being tested in the CI.

Yes this not getting tested at this point of time but it will, once CI enhancements to build and store artifacts with multiple distros, kernels etc., in #935 and #922 gets concluded.  I attempted switching the default kernel to 'linux-qcom-next' but @ndechesne recommended not to as not all patches may have landed here yet. https://github.com/qualcomm-linux/meta-qcom/pull/908#discussion_r2139422260 

As others are waiting on this kernel for their dev activities, I'd request to accept this recipe even though CI is not in place.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -0,0 +1,28 @@
...
+
+# To build tip of qcom-next branch set preferred
+# virtual/kernel provider to 'linux-qcom-next-upstream'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "linux-qcom-next-upstream"
+SRCREV:class-devupstream ?= "${AUTOREV}"
+
+S = "${WORKDIR}/git"
+
+KERNEL_CONFIG_COMMAND ?= "oe_runmake_call -C ${S} O=${B} defconfig"
```
lumag: This should be `qcom_defconfig` for ARM machines.

ci/linux-qcom-next.yml
```diff
@@ -0,0 +1,6 @@
+header:
+  version: 14
+
+local_conf_header:
+  kernelprovider: |
+    PREFERRED_PROVIDER_virtual/kernel = "linux-qcom-next"
```
lumag: Please drop all unnecessary KAS fragments.

lumag:
So, neither of qcs6490-rb3gen2, qcs9075-iq-evk, qcs9075-ride-sx boots with qcom-next. @idlethread  could you please check, what happened?

ci/linux-qcom-next.yml
vkraleti: To support overlay verification, it's necessary to build kernel modules against both linux-qcom-next and linux-yocto-dev. Having KAS fragments for each kernel makes it convenient to switch between them. At a later point of time, these fragments are needed for building both kernels reliably in CI as well. If this PR isn't the right place, I'll move them to a separate PR. However, having them here simplifies kernel switching and testing workflows.

lumag: They are dead code from the point of this PR. I think that for testing modules it should be enough to build them against qcom-armv8a and against one of the new machine, so the KAS fragments are not necessary anyway.

vkraleti:
> So, neither of qcs6490-rb3gen2, qcs9075-iq-evk, qcs9075-ride-sx boots with qcom-next. @idlethread could you please check, what happened?

This is what I see in logs, debugging further, will update.
```
[    4.520871] Run /init as init process
Starting systemd-udevd version 257.6
root '/dev/disk/by-partlabel/rootfs' doesn't exist or does not contain a /dev.
```

lumag:
I'd vote for enabling both of them as built-in

vkraleti:
> I'd vote for enabling both of them as built-in

To do so either need to maintain qcom_defconfig or use configuration fragments in recipe. Which one do you suggest?

koenkooi:
> > I'd vote for enabling both of them as built-in
> 
> To do so either need to maintain qcom_defconfig or use configuration fragments in recipe. Which one do you suggest?

Fixing it upstream would be best, for everything else I’m a big fan of config fragments. They make it easy to understand which options were changed.

vkraleti:
> > > I'd vote for enabling both of them as built-in
> > 
> > To do so either need to maintain qcom_defconfig or use configuration fragments in recipe. Which one do you suggest?
> 
> Fixing it upstream would be best, for everything else I’m a big fan of config fragments. They make it easy to understand which options were changed.

Introducing custom qcom_defconfig need further discussion with kernel team. I'll take it up in parallel. For now added a configuration fragment in recipe to unblock this PR.

ndechesne:
> > > I'd vote for enabling both of them as built-in
> > 
> > 
> > To do so either need to maintain qcom_defconfig or use configuration fragments in recipe. Which one do you suggest?
> 
> Fixing it upstream would be best, for everything else I’m a big fan of config fragments. They make it easy to understand which options were changed.

Why do we need to do that? Is it a requirement to boot without an initramfs? It's definitely not the right thing to do. For sure I don't think qcom-next should change what upstream does. So if we really want our OE image to boot without an initramfs, then we should deal with config fragments here. but I would prefer not.

conf/machine/qrb2210-rb1-core-kit.conf
```diff
@@ -4,6 +4,8 @@
 
 require conf/machine/include/qcom-qcm2290.inc
 
+PREFERRED_PROVIDER_virtual/kernel ?= "linux-yocto-dev"
```
lumag: Why?? What is missing in qcom-next?

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -0,0 +1,56 @@
...
+
+    for fgmt in ${KERNEL_CONFIG_FRAGMENTS}; do
+        if [ -f ${S}/arch/${ARCH}/configs/${fgmt} ]; then
+            ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${S}/arch/${ARCH}/configs/${fgmt}
+        elif [ -f "${UNPACKDIR}/${fgmt}" ]; then
+            ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${UNPACKDIR}/${fgmt}
+        elif [ -f "${fgmt}" ]; then
+            ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${fgmt}
+        fi
+    done
```
lumag: Ugh. Too much trouble for a single fragment.

```
do_configure:append() {
    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${UNPACKDIR}/qcom.cfg
}
```

lumag (changes requested):

recipes-kernel/linux/linux-qcom-next_git.bb
vkraleti: It works for now, but I’m certain more fragments need to be added over time and some of which will likely come from the kernel tree itself.

lumag: I hope not. And I definitely don't expect any fragments to come from qcom-next.

lumag: In the end, the goal of qcom-next is to use config which is as close to upstream as possible.

recipes-kernel/linux/linux-qcom-next_git.bb
lumag: This still isn't addressed

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -0,0 +1,41 @@
...
+# virtual/kernel provider to 'linux-qcom-next-upstream'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "linux-qcom-next-upstream"
+SRCREV:class-devupstream ?= "${AUTOREV}"
+
+S = "${UNPACKDIR}/${BP}"
+
+do_configure:prepend() {
+    # Use a copy of the 'defconfig' from the actual repo to merge fragments
+    cp ${S}/arch/${ARCH}/configs/defconfig ${B}/.config
```
lumag: No, this will not work for armv7a

vkraleti: Updated recipe to use KBUILD_DEFCONFIG variable to conditionally define the defconfig.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -0,0 +1,44 @@
...
+
+do_configure:prepend() {
+    # Use a copy of the 'defconfig' from the actual repo to merge fragments
+    cp ${S}/arch/${ARCH}/configs/${KBUILD_DEFCONFIG} ${B}/.config
+
+    # Merge fragment for QCOM value add features
+    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${UNPACKDIR}/configs/qcom.cfg
+}
+
+KERNEL_CONFIG_COMMAND ?= "oe_runmake_call -C ${S} O=${B} defconfig"
```
lumag: This will override the .config that you've compiled few lines above.

vkraleti: This assignment is ineffective because of the weak assignment (?=) The default KERNEL_CONFIG_COMMAND of kernel.bblcass is what is being used. Removed it to avoid ambiguity.

lumag (approved):

sbanerjee-quic (approved):

PR #912, opened 2025-05-27
update meta-qcom url to generate kas lock file with commit sha

sbanerjee-quic:
Unless the url is added for meta-qcom the generated kas lock file is missing entry for meta-qcom commit.

Adding the 'url:' generates an additional commit sha for meta-qcom as follows in the lock.yml file.

`
        meta-qcom:
            commit: 11bc71051e2d486977a51cf7fc1a0143fec01109
`

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -11,6 +11,7 @@ defaults:
 
 repos:
   meta-qcom:
+    url: https://github.com/qualcomm-linux/meta-qcom
 
   oe-core:
     url: https://github.com/openembedded/openembedded-core


ndechesne:
hmm. it gets tricky... I tried locally, and I see to get the commit for meta-qcom on the remote URL, even if it does not match the commit I have checked out locally. which is error prone i think. 

maybe we need to think about what we really are trying to do here. 

the lock file is needed for 2 different purpose:
1. we want to make sure that all jobs in the same workflow use the same commits from all the remotes. Which I think works today, even without this patch. Since meta-qcom is checked out by github action, so we should be getting the same commit for meta-qcom, even if it's missing from the lock file.
2. we want to archive 'something' that will allow us to reproduce the same build in the future. and here clearly the lock file is missing meta-qcom data (but we have the commit from the workflow..).

ricardosalveti:
> hmm. it gets tricky... I tried locally, and I see to get the commit for meta-qcom on the remote URL, even if it does not match the commit I have checked out locally. which is error prone i think.
> 
> maybe we need to think about what we really are trying to do here.
> 
> the lock file is needed for 2 different purpose:
> 
> 1. we want to make sure that all jobs in the same workflow use the same commits from all the remotes. Which I think works today, even without this patch. Since meta-qcom is checked out by github action, so we should be getting the same commit for meta-qcom, even if it's missing from the lock file.
> 2. we want to archive 'something' that will allow us to reproduce the same build in the future. and here clearly the lock file is missing meta-qcom data (but we have the commit from the workflow..).

Yeah, this will create a bad side effect for developers using the same kas fragment when building locally, but I guess this also causes issues on PRs, since it will fetch the revision from the upstream repository instead.

Maybe we just extend the nightly definition to include an extra option that forces kas to use an additional config fragment that adds the url value? In the end, we probably just want the proper lock files for nightlys.

ndechesne:
> > hmm. it gets tricky... I tried locally, and I see to get the commit for meta-qcom on the remote URL, even if it does not match the commit I have checked out locally. which is error prone i think.
> > maybe we need to think about what we really are trying to do here.
> > the lock file is needed for 2 different purpose:
> > 
> > 1. we want to make sure that all jobs in the same workflow use the same commits from all the remotes. Which I think works today, even without this patch. Since meta-qcom is checked out by github action, so we should be getting the same commit for meta-qcom, even if it's missing from the lock file.
> > 2. we want to archive 'something' that will allow us to reproduce the same build in the future. and here clearly the lock file is missing meta-qcom data (but we have the commit from the workflow..).
> 
> Yeah, this will create a bad side effect for developers using the same kas fragment when building locally, but I guess this also causes issues on PRs, since it will fetch the revision from the upstream repository instead.
> 
> Maybe we just extend the nightly definition to include an extra option that forces kas to use an additional config fragment that adds the url value? In the end, we probably just want the proper lock files for nightlys.

I had another idea..  we can keep the lock file as it is, since it addresses quite well my first bullet. E.g. ensure that all run in the workflows use the same commit for all repos.

Then we can create a per-build-per-machine dump with 'kas dump --resolve-local XXX.yml' where XXX.yml is the exact set of yaml files used for the build. It will be a complete dump of all repo + commits + all extra config. That's actually more than what we have in the lock files.

ricardosalveti:
> I had another idea.. we can keep the lock file as it is, since it addresses quite well my first bullet. E.g. ensure that all run in the workflows use the same commit for all repos.
> 
> Then we can create a per-build-per-machine dump with 'kas dump --resolve-local XXX.yml' where XXX.yml is the exact set of yaml files used for the build. It will be a complete dump of all repo + commits + all extra config. That's actually more than what we have in the lock files.

Yeah, that is better indeed, want to send a pr?

sbanerjee-quic:
> 2\. we want to archive 'something' that will allow us to reproduce the same build in the future. and here clearly the lock file is missing meta-qcom data (but we have the commit from the workflow..).

My primary goal for this PR is use case #2, later reproducibility.

ricardosalveti:
@sbanerjee-quic closing based on #917, let us know if not enough for your use cases.

PR #914, opened 2025-05-28
recipe-bsp: modification to include qcom-iris in final image.

quic-satygupt:
adding kernel module qcom-iris in machine essential package group, so that it is present in device as part of final image.

PR #858, opened 2025-04-30
recipes-kernel: foo-module: add video iris driver

malathi-gottam:

PR #916, opened 2025-05-28
recipe-bsp: modification to include qcom-iris in final image.

quic-satygupt:
adding kernel module qcom-iris in machine essential package group,
so that it is present in device as part of final image.

Change-Id: I3597529a6ba81e883deb9c8c5e5807e72e733cda


koenkooi:
This depends on #858, right?

lumag:
Closing in favour of #914 , oriinal PR.

PR #919, opened 2025-05-29
Use RDEPENDS instead of RRECOMMENDS in packagegroup-qcom

vkraleti:
Switch to using RDEPENDS for including miscellaneous packages in packagegroup-qcom to ensure they are built by default.


lumag (changes requested):
No. The point of the group was to build those packages without including them in any of the images.

vkraleti:
> No. The point of the group was to build those packages without including them in any of the images.

Then only option that I could think of is to convert the RRECOMMENDS to RDEPENDS in packagegroup-qcom.bb. Is that fine?

sbanerjee-quic (approved):

PR #920, opened 2025-06-03
Workflows/dragonboard x20c

mwasilew:


lumag (approved):

koenkooi:
I have a related question about CI URLS: we want to start building testing multiple distributions per board, which means that we'll have 2 identically named, but actually different files each build: `core-image-foo-machine.qcomflash.tar.gz`, one for 'nodistro' and one for 'qcom-distro'. Where do you recommend adding the changes needed to support that?

mwasilew:
@ndechesne do we need to wait with this PR for @koenkooi's changes?

ndechesne:
> @ndechesne do we need to wait with this PR for @koenkooi's changes?

No. They can be merged. I think they will actually help @koenkooi. @koenkooi is going to add another folder in the path to publish both nodistro and qcomdistro binaries. So i think Koen will just need to ensure that input.url point to the 'base' url of either nodistro or qcomdistro, and everything else should still work ;)

PR #921, opened 2025-06-03
ci: publish buildchart.svg kas-build.yml on the deploy dir

quaresmajose:
Make buildchart and kas lock file machine specific
otherwise they collided with previously existing files
on the uploads dir.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -135,8 +135,8 @@ jobs:
           uploads_dir=./uploads
           mkdir $uploads_dir
           find $deploy_dir/ -maxdepth 1 -name "*.rootfs*.qcomflash" -exec rm -rf {} \;
+          cp buildchart.svg kas-build.yml $deploy_dir/
           mv $deploy_dir $uploads_dir/
-          cp buildchart.svg kas-build.yml $uploads_dir/
 
       - name: Upload private artifacts
         uses: qualcomm-linux/upload-private-artifact-action@v1


.github/workflows/build-yocto.yml
```diff
@@ -136,7 +136,8 @@ jobs:
           mkdir $uploads_dir
           find $deploy_dir/ -maxdepth 1 -name "*.rootfs*.qcomflash" -exec rm -rf {} \;
           mv $deploy_dir $uploads_dir/
-          cp buildchart.svg kas-build.yml $uploads_dir/
+          cp buildchart.svg $uploads_dir/buildchart-${{ matrix.machine }}.svg
```
ndechesne: hmm. looks like there is something I did not understand in how we publish. We indeed publish all artifact in a <machine> subfolder, but i missed that the <machine> is because we copy deploy/images/${{matrix.machine}}. 

So what I really wanted to do (and failed to!) was to put kas-build.yml and buildchart.svg in the <machine> folder. 
So I think the right fix is to do 
cp  buildchart.svg kas-build.yml $deploy_dir
before the 'mv' command instead.

quaresmajose: yeah the deploy_dir is already machine specific. Thanks

quaresmajose:
> Please follow Nico's suggestion

done

lumag (approved):

lumag:
Nit: 'This makes this published' -> 'Make buildchart and kas lock file'

quaresmajose:
> Nit: 'This makes this published' -> 'Make buildchart and kas lock file'

Fixed
Thanks

PR #922, opened 2025-06-03
Extend GitHub workflows to compile with multiple kernels

vkraleti:
Update the GitHub Actions workflow to support building with multiple kernel configuration in CI.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -96,8 +96,13 @@ jobs:
           - qrb2210-rb1-core-kit
           - qcom-armv8a
           - qcom-armv7a
+        distro:
+          - name: poky-altcfg
+            yamlfile: ""
+          - name: qcom
+            yamlfile: 'ci/distro.yml:'
     runs-on: [self-hosted, x86]
-    name: ${{ matrix.machine }}/poky-altcfg
+    name: ${{ matrix.machine }}/${{ matrix.distro.name }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -116,7 +121,7 @@ jobs:
           mkdir $KAS_WORK_DIR
           ${KAS_CONTAINER} dump --resolve-env --resolve-local --resolve-refs \
               ci/mirror.yml:ci/${{ matrix.machine }}.yml > kas-build.yml
-          ${KAS_CONTAINER} build ci/mirror.yml:ci/${{ matrix.machine }}.yml
+          ${KAS_CONTAINER} build ci/mirror.yml:${{ matrix.distro.yamlfile }}ci/${{ matrix.machine }}.yml
           ci/kas-container-shell-helper.sh ci/yocto-pybootchartgui.sh
           mv $KAS_WORK_DIR/build/buildchart.svg .
 
@@ -126,12 +131,12 @@ jobs:
 
       - uses: actions/upload-artifact@v4
         with:
-          name: buildchart-${{ matrix.machine }}
+          name: buildchart-${{ matrix.distro.name }}-${{ matrix.machine }}
           path: buildchart.svg
 
       - uses: actions/upload-artifact@v4
         with:
-          name: kas-build-${{ matrix.machine }}
+          name: kas-build-${{ matrix.distro.name }}-${{ matrix.machine }}
           path: kas-build.yml
 
       - name: Stage build artifacts for publishing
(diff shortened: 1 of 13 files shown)


.github/workflows/build-yocto.yml
```diff
@@ -89,8 +89,11 @@ jobs:
           - qrb2210-rb1-core-kit
           - qcom-armv8a
           - qcom-armv7a
+        kernel:
```
ndechesne: with this change, we never build/test the kernel that we set in the machine config file, it is always overriden with linux-yocto or linux-yocto-dev. that does not seem right to me.

.github/workflows/build-yocto.yml
```diff
@@ -108,8 +111,8 @@ jobs:
           export KAS_WORK_DIR=$PWD/../kas
           mkdir $KAS_WORK_DIR
           kas dump --resolve-env --resolve-local --resolve-refs \
-              ci/mirror.yml:ci/${{ matrix.machine }}.yml > kas-build.yml
-          kas build ci/mirror.yml:ci/${{ matrix.machine }}.yml
+              ci/mirror.yml:ci/${{ matrix.machine }}.yml:ci/${{ matrix.kernel }}.yml > kas-build.yml
```
ndechesne: if we create build variants, we certainly need to agree how we publish them. in the current situation, the next build will override the previous build.

.github/workflows/build-yocto.yml
vkraleti: Planning to add qcom-next kernel also to this list. Idea is to build all available machines with all available kernels in CI. That way the kernel set by machine config file is also covered. Is it an over kill?

.github/workflows/build-yocto.yml
vkraleti: #921 is supposed to address this. Isn't it?

ndechesne: no #921 is fixing a bug I introduced in #917.

.github/workflows/build-yocto.yml
ndechesne: I was not commenting on the build matrix we introduce, just saying that after this change, we never build in CI the default config we have in the tree. We can build 'additional' configs, but we should build the default config that we have set in MACHINE conf file.

lumag: I agree here

conf/machine/qrb2210-rb1-core-kit.conf
```diff
@@ -4,6 +4,8 @@
 
 require conf/machine/include/qcom-qcm2290.inc
 
+PREFERRED_PROVIDER_virtual/kernel ?= "linux-yocto-dev"
```
lumag, line 7: Which RB1 patches are not a part of the `linux-qcom-next`?

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -0,0 +1,28 @@
...
+# tag: qcom-next-6.16-rc2-20250623
+SRCREV ?= "3c20b557843734abc8fa0ea9ee6ed294eb52ece9"
+
+# To build tip of qcom-next branch set preferred
+# virtual/kernel provider to 'linux-qcom-next-upstream'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "linux-qcom-next-upstream"
+SRCREV:class-devupstream ?= "${AUTOREV}"
+
+S = "${UNPACKDIR}/${BP}"
```
lumag, line 26: Should not be necessary, please drop.

vkraleti:
I'll close PR and pull the commit to switch kernel preferred provider into #908. Lets discuss over there.

recipes-kernel/linux/linux-qcom-next_git.bb
vkraleti: It is needed to avoid below error. 

`linux-qcom-next-6.16-rc2+git-r0 do_populate_lic: QA Issue: linux-qcom-next: LIC_FILES_CHKSUM points to an invalid file: build/tmp/work-shared/qrb2210-rb1-core-kit/kernel-source/COPYING`

inheriting kernel bbclass sets `S = "${STAGING_KERNEL_DIR}"`. However this recipe is not unpacking sources under KERNEL_SRC_DIR, thus causing issue. For linux-yocto recipe, kernel-yocto bblcass takes care of unpacking under KERNEL_SRC_DIR.

lumag: You've kind of pinpointed the issue. The sources should go to `STAGING_KERNEL_DIR`, otherwise other recipes are going to be broken

PR #924, opened 2025-06-10
Fix ifc6410 support

lumag:
Update IFC6410 Kconfig fragments, enabling it to boot.


mwasilew:
I'm able to boot it on ifc6410 and it finds the `userdata` partition.

PR #925, opened 2025-06-10
pg-machine-essential: extend module list for better rb3gen2 support

ricardosalveti:
Extend list of generic qualcomm modules and include pci-pwrctrl-tc9593 for pcie / usb / ethernet support.


recipes-bsp/packagegroups/packagegroup-machine-essential.bb
```diff
@@ -56,14 +63,22 @@ RRECOMMENDS:${PN}-qcom-generic += " \
...
+    kernel-module-qcrypto \
     kernel-module-qrtr \
     kernel-module-qrtr-mhi \
     kernel-module-qrtr-smd \
     kernel-module-rmtfs-mem \
     kernel-module-rpmsg-ctrl \
     kernel-module-rtc-pm8xxx \
+    kernel-module-snd-q6apm \
+    kernel-module-snd-soc-lpass-va-macro \
+    kernel-module-snd-soc-lpass-wsa-macro \
```
lumag, line 79: Also -rx-macro and -tx-macro

recipes-bsp/packagegroups/packagegroup-machine-essential.bb
```diff
@@ -56,14 +63,22 @@ RRECOMMENDS:${PN}-qcom-generic += " \
...
+    kernel-module-snd-soc-sc8280xp \
+    kernel-module-snd-soc-wsa883x \
```
lumag: I think codecs are more or less platform-specific. E.g.
- wsa881x is sdm845, sm8150, sm8250
- wsa883x is sc8280xp,  sm8350, sm8450 (and qcm6490?)
- wsa884x is sm8550,  sm8650 and x1e
- wcd9335 is apq8096 / msm8996
- wcd934x is sdm845, sm8150
- wcd937x is qcm6490
- wcd938x is sc7280, sc8280xp, x1e8, sm8250-sm8550,
- wcd939x is sm8650

recipes-bsp/packagegroups/packagegroup-machine-essential.bb
```diff
@@ -101,6 +101,7 @@ RRECOMMENDS:${PN}-qcom-qcs6490-soc += " \
     kernel-module-dispcc-sc7280 \
     kernel-module-gpucc-sc7280 \
     kernel-module-lpassaudiocc-sc7280 \
+    kernel-module-pci-pwrctrl-tc9563 \
```
lumag: This is not specific to the QCS6490, so it should go to another group.

ricardosalveti: Yeah, was thinking if this should even be part of a group that we define here, since at the moment this is only used by rb3gen2 and rubikpi. But since we also have xhci-pci-renesas under board-generic, I guess we could add it there as well.

Otherwise this would need to be machine specific, and not sure if worth making it that granular for a kernel module.

recipes-bsp/packagegroups/packagegroup-machine-essential.bb
ricardosalveti: Added both.

recipes-bsp/packagegroups/packagegroup-machine-essential.bb
ricardosalveti: Moved wsa883x under qcm6490 since that is the only platform-specific pg we have here that includes support for it.

recipes-bsp/packagegroups/packagegroup-machine-essential.bb
ricardosalveti: Moved under board-generic, together with ax88179-178a.

ricardosalveti:
@mwasilew this should add some of the missing modules we had for rb3gen2, including pcie switch and usb network module (for ethernet).

If you manually add the renesas firmware everything should show up and work.

recipes-bsp/packagegroups/packagegroup-machine-essential.bb
```diff
@@ -35,15 +35,22 @@ RRECOMMENDS:${PN}-qcom-generic += " \
...
     kernel-module-phy-qcom-qmp-combo \
     kernel-module-phy-qcom-qmp-usbc \
     kernel-module-phy-qcom-qusb2 \
     kernel-module-phy-qcom-snps-femto-v2 \
+    kernel-module-q6apm-dai \
+    kernel-module-q6apm-lpass-dais \
+    kernel-module-q6prm-clocks \
+    kernel-module-qcom-battmgr \
     kernel-module-qcom-iris \
+    kernel-module-qcom-pd-mapper \
```
vkraleti, line 55: Before switching to the in-kernel pd-mapper, we need to ensure that it is functionally on par with the current userspace implementation. Previously, @lumag pointed out timing-related issues in the in-kernel solution that led to audio breakages. These need to be retested to avoid regressions.

mwasilew:
I used wrong build. With the last build from this PR I'm getting ethernet:

recipes-bsp/packagegroups/packagegroup-machine-essential.bb
lumag: The in-kernel pd-mapper has been provided for quite a while. If there are incompatibilities / regressions, please report them to the LAKML. Otherwise I'd consider it to be suitable for the production run.

PR #927, opened 2025-06-11
Github workflow: build both Poky and Qcom DISTRO configs

koenkooi:
Add another axis to the CI matrix: distro.name and have a companion variable to add optional extra yaml files. The DISTRO in meta-qcom-distro needs `distro.yml`, while the default Poky-that-isn't-poky-anymore does not need an extra yaml.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -89,8 +89,13 @@ jobs:
           - qrb2210-rb1-core-kit
           - qcom-armv8a
           - qcom-armv7a
+        distro:
+          - name: poky-systemd
+            yamlfile: ""
+          - name: qcom
+            yamlfile: 'ci/distro.yml:'
     runs-on: [self-hosted, x86]
-    name: ${{ matrix.machine }}/poky/systemd
+    name: ${{ matrix.machine }}/${{ matrix.distro.name }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -109,7 +114,7 @@ jobs:
           mkdir $KAS_WORK_DIR
           kas dump --resolve-env --resolve-local --resolve-refs \
               ci/mirror.yml:ci/${{ matrix.machine }}.yml > kas-build.yml
-          kas build ci/mirror.yml:ci/${{ matrix.machine }}.yml
+          kas build ci/mirror.yml:${{ matrix.distro.yamlfile }}ci/${{ matrix.machine }}.yml
           ci/yocto-pybootchartgui.sh && mv $KAS_WORK_DIR/build/buildchart.svg .
 
           if [ "${{ matrix.machine }}" = "qcom-armv8a" ]; then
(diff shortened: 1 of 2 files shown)


koenkooi:
> Something is not right. When you deploy the artifacts in 'stage build artifacts for publishing'
> 
> you have this:
> 
> ```
>       deploy_dir=../kas/build/tmp/deploy/images/${{matrix.machine}}
>       uploads_dir=./uploads
> ```
> 
> It's a bit tricky, but basically, we copy the <matrix.machine> build folder into the publishing area. Which is why in the publishing area we have machine subfolders.
> 
> Now that we are adding another axis, we need to publish to (for example) <matrix.distro.name>/<matrix.machine>, otherwise the 2nd distro build will overwrite the first one.

I've pushed that change as a separate commit.

.github/workflows/build-yocto.yml
```diff
@@ -136,7 +136,7 @@ jobs:
           # The upload-private-artifact-action runs from a container that
           # expects file to be relative to our PWD. deploy_dir is outside
           # that, so we move things around:
-          deploy_dir=../kas/build/tmp/deploy/images/${{matrix.machine}}
+          deploy_dir=../kas/build/tmp/deploy/images/${{ matrix.distro.name }}/${{matrix.machine}}
```
ndechesne: nope.. please check what this code is doing carefully.

deploy_dir is set to be the OE/kas output build folder, e.g. DEPLOY_DIR_IMAGES. we are 'moving' this folder in the uploads_dir folder a few lines below. what you want is to append matrix.distro.name to uploads_dir, so that the upload folder structure looks like:

```
-> poky-systemd
        -> machine1
        -> machine2
        -> machine3
-> qcom-distro
        -> machine1
        -> machine2
        -> machine3
```

koenkooi: I see, this likely needs some changes, as DEPLOY_DIR can be changed, e.g. meta-ti does that. I've changed the commit to update all uses of uploads_dir I could find.

ndechesne: I would be happy to see a patch where DEPLOY_DIR is retrieved from OE/KAS.

.github/workflows/build-yocto.yml
```diff
@@ -147,7 +147,7 @@ jobs:
         uses: qualcomm-linux/upload-private-artifact-action@v1
         id: upload_artifacts
         with:
-          path: ./uploads
+          path: ./uploads/${{ matrix.distro.name }}
```
ndechesne: hehe.. not here :) 
'path' here is what you want to upload. If you use ./uploads/{{matrix.distro.name}} then you will not upload the top level matrix.distro.name folder, but its content.. and we won't have the top level structure with distro/machine folders

koenkooi: Fixed

ndechesne:
the latest update looks good. Let's look at the resulting build and how it's published to confirm that we got everything right (and that LAVA can still run the tests!)

vkraleti:
Please take a look at #922 as well. I'm still working on this PR, but it will introduce 'kernel' axis. We might also add an "image" axis later on. The build storage approach we choose should be flexible enough to accommodate all these. Can something like below work? Will our build loading procedure can work with this structure?

<distro>/
  └── <machine>/
      └── <kernel>/
          └── <image>/

ricardosalveti:
https://github.com/qualcomm-linux/meta-qcom/pull/933 was just merged, please rebase.

koenkooi:
Due to this branch being in the main repo, I can't update it anymore *and* I can't edit the base for the PR, so move this to https://github.com/qualcomm-linux/meta-qcom/pull/935

PR #928, opened 2025-06-11
ci: correct download URLs for dragonboard 410c/820c

lumag:
Include machine name into the generated download URLs for db410c / db820c.

### DIFF
--- a/ci/lava/dragonboard-410c/boot.yaml
+++ b/ci/lava/dragonboard-410c/boot.yaml
@@ -7,11 +7,11 @@ actions:
       boot:
         headers:
           Authentication: Q_GITHUB_TOKEN
-        url: "{{BUILD_DOWNLOAD_URL}}/boot-apq8016-sbc-qcom-armv8a.img"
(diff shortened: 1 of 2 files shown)


ci/lava/dragonboard-410c/boot.yaml
```diff
@@ -7,11 +7,11 @@ actions:
       boot:
         headers:
           Authentication: Q_GITHUB_TOKEN
-        url: "{{BUILD_DOWNLOAD_URL}}/boot-apq8016-sbc-qcom-armv8a.img"
+        url: "{{BUILD_DOWNLOAD_URL}}/qcom-armv8a/boot-apq8016-sbc-qcom-armv8a.img"
```
ndechesne, line 10: would it be 'better' to handle the dragonboard410c->qcom-armv8 mapping in test.yml instead, and use DEVICE_TYPE here?

lumag: I don't want to overload DEVICE_TYPE: it is used in other places, where we actually mean the exact machine type (e.g. several lines below, in the test name: `"{{DEVICE_TYPE}}-smoke-test"` . I can probably add something like IMAGE_SUBDIR, which by default should be equal to DEVICE_TYPE, except for the dragonboards.

PR #932, opened 2025-06-16
linux-yocto-dev: import patches fixing venus probe crashes

ricardosalveti:
Backport and add fromlist patches required to fix venus probe crashes on rb3gen2, which happens quite often on boot and locks up the kernel completely on reboots.


koenkooi (approved):
The provenance of the patches looks good to me. Should we go over the existing ones and add FROMLIST/UPSTREAM/etc in their filenames as well, as a separate PR?

lumag:
The very obvious reason for not having extra changes in the filename / subject: currently you can simply `git format-patch` from the upstream kernel tree without any additional steps. Adding FROMLIST / FROMGIT / etc. requires extra changes and extra steps.

Second, it duplicates Upstream-Status. Duplication is always a bad idea, it's too easy to get it out of sync.

Those taggings make sense if we have patches coming from different sources and of different quality. We are using upstream-only patches (which should mean at least picking up the patch from the list). In such situation there is no need for extra tags.

Last, but not least, when the patch changes the status (e.g. the Submitted patch is changed to the Backport), currently you have to change one line in the patch. With the extra tags you also have to change the subject and to rename the file. This creates extra noise.

ricardosalveti:
> Please drop extra tags from the Subject / file name.

Removed.

ricardosalveti:
We can ignore the current test failures and merge once we have the required review approvals.

ndechesne (approved):

PR #935, opened 2025-06-18
Github workflow: build both Poky and Qcom DISTRO configs

koenkooi:
Add another axis to the CI matrix: distro.name and have a companion variable to add optional extra yaml files. The DISTRO in meta-qcom-distro needs `distro.yml`, while the default Poky-that-isn't-poky-anymore does not need an extra yaml.


koenkooi:
I've added a commit to unbreak meta-oe, it will get removed before merging.

ndechesne:
So, now we have the proper/expected failure, see https://github.com/qualcomm-linux/meta-qcom/actions/runs/15735817749. 

@mwasilew can you find out why the Test PR failure is not reported back to the PR, that would be nice. 

@koenkooi we need to decide what to do with the testing now that we have 2 distros. Are we duplicating the tests on both the poky and the qcom images?

koenkooi:
> So, now we have the proper/expected failure, see https://github.com/qualcomm-linux/meta-qcom/actions/runs/15735817749.
> 
> @mwasilew can you find out why the Test PR failure is not reported back to the PR, that would be nice.

We're seeing 2 issues here:

1. A failure that is caused by the machines being a folder deeper into the tree than before
2. A failure caused by rerunning only the failed tasks, most artefacts are in <id>-1, but the jobs looks for everything in <id>-2
 
> @koenkooi we need to decide what to do with the testing now that we have 2 distros. Are we duplicating the tests on both the poky and the qcom images?

Duplicate them or only run on them on the qcom-distro builds.

ricardosalveti:
> @koenkooi we need to decide what to do with the testing now that we have 2 distros. Are we duplicating the tests on both the poky and the qcom images?

I believe we will end up with 2 different sets of tests, with a bit of duplication, since we also want to test meta-qcom alone.

ndechesne:
> @rsalveti my idea is to move the test plan to a separate repository. This way we'll be able to cover poky, qcom-distro, debian and potentially any other builds coming in the future. Then we can use a tagged version of the test plan in meta-qcom workflow. I'll try to propose something shortly. There is already a parallel discussion about debian testing.

sounds like a good thing. 
In the mean time, @koenkooi can you please update this PR to ensure the LAVA test jobs use the poky-systemd builds (like it was before this PR). We can add more tests and test qcom distro once we have what @mwasilew said above. so that we can merge this PR. We would have build CI for poky and qcom distro, and run time tests for poky distro only. that's already an improvement.

ci/meta-oe-fix.patch
```diff
@@ -0,0 +1,160 @@
+From patchwork Tue Jun 17 06:15:52 2025
```
lumag: Nit: `b4 am 20250617061552.161683-1-Qi.Chen@windriver.com` produces a cleaner patch.

lumag (approved):
With the OE hack removed, LGTM

koenkooi:
> @koenkooi could you please drop the hack? I think we are ready to merge this.

Done!

ci/base.yml
```diff
@@ -31,7 +31,6 @@ local_conf_header:
   base: |
     CONF_VERSION = "2"
     INIT_MANAGER = "systemd"
-    PACKAGE_CLASSES = "package_ipk"
```
quaresmajose: looks like this is part of the https://github.com/qualcomm-linux/meta-qcom/pull/939

koenkooi: it is, let me fix that...

lumag:
@koenkooi this needs to be rebased in order to solve merge conflict

ricardosalveti:
> Did we not conclude that we need to update the LAVA template to pick up the downloads from the new location?

Yeah, we should fix that here as well.

vkraleti (approved):

quaresmajose (approved):

ndechesne (approved):

PR #936, opened 2025-06-18
workflows: switch test job to lava-action@v8

mwasilew:
lava-action@v8 provides the following features:
 - use lava_token for retrieving job results and logs. This allows to submit "personal" jobs where it's important to protect secrets (i.e. wifi passwords)
 - allow to save results with fixed file name. This change ensures that there is a clean set of results in case some test jobs have to be restarted.
 - add flag for Incomplete and Canceled jobs to not be considered "failures". This allows to produce full test report that is included in the PR comment even in case produced build doesn't boot on the hardware.


mwasilew:
@ndechesne @lumag I think this addresses most of the issues we've seen recently.

ricardosalveti (approved):

PR #937, opened 2025-06-19
workflow: stales: stop marking issues as stale automatically

ndechesne:
While it's a good idea to automatically mark PR as stale when inactive, we probably can live with inactive issues, or we can do a manual triage from time to time. An issue can still be marked as 'stalled' by adding the label manually, and will in turn be closed if no update is made until the 'close' idle delay is reached.


ricardosalveti:
I was thinking we could make the issues stale but not close them automatically, at least that would show that there is nobody working on the issue itself.

ricardosalveti (approved):

lumag (approved):

PR #939, opened 2025-06-20
ci/base.yml+workflows: switch to 'poky-altcfg'

koenkooi:
Testing meta-qcom* layers against upstream DISTROs is only useful if the CI doesn't fundamentally change the upstream policies and configuration. Since systemd is virtually indispensable for meta-qcom* machines, switch form `poky` to `poky-altcfg` and remove the initmanager changes from the KAS snippet.

This change will make CI result stop being misleading about how well an upstream DISTRO works with these layers.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -3,7 +3,7 @@
 header:
   version: 14
 
-distro: poky
+distro: poky-altcfg
 
 defaults:
   repos:
@@ -30,8 +30,6 @@ repos:
 local_conf_header:
   base: |
     CONF_VERSION = "2"
-    INIT_MANAGER = "systemd"
-    PACKAGE_CLASSES = "package_ipk"
     INHERIT += "buildstats buildstats-summary"
     INHERIT += "buildhistory"
     INHERIT += "rm_work"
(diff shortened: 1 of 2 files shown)


ci/base.yml
```diff
@@ -30,7 +30,6 @@ repos:
 local_conf_header:
   base: |
     CONF_VERSION = "2"
-    INIT_MANAGER = "systemd"
     PACKAGE_CLASSES = "package_ipk"
```
quaresmajose: You can also remove the PACKAGE_CLASSES since the package_ipk is the default package manager.
https://git.yoctoproject.org/poky/tree/meta-poky/conf/distro/poky-altcfg.conf#n6

ndechesne: +1

koenkooi: Done!

quaresmajose: the PACKAGE_CLASSES is still there so maybe it was uncommited on your machine

quaresmajose: I think it was unintentional moved to https://github.com/qualcomm-linux/meta-qcom/pull/935

quaresmajose (approved):

PR #940, opened 2025-06-20
Adds support for Whinlatter release series 

quaresmajose:
- layer.conf: Update to whinlatter release series
- Set S to be in UNPACKDIR in recipes that explicitly set S
- Remove S in recipes that fetch from git via setting BB_GIT_DEFAULT_DESTSUFFIX


koenkooi (dismissed):

dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-adbd-cmdline.bb
```diff
@@ -3,16 +3,15 @@ SECTION = "console/utils"
 LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
-S = "${WORKDIR}/sources"
-UNPACKDIR = "${S}"
+S = "${UNPACKDIR}"
```
ndechesne, line 6: are you sure this is needed? I have not looked deeply into the recent changes, but I got the feeling that we no longer need to set S at all, and that we can use the defaults value from bitbake.conf

recipes-devtools/qtestsign/qtestsign_git.bb
```diff
@@ -29,4 +27,4 @@ do_install() {
 FILES:${PN} += "${PYTHON_SITEPACKAGES_DIR}/qtestsign"
 RDEPENDS:${PN} = "python3 python3-cryptography"
 
-BBCLASSEXTEND = "native nativesdk"
\ No newline at end of file
+BBCLASSEXTEND = "native nativesdk"
```
ndechesne, line 30: unrelated diff.

dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-adbd-cmdline.bb
quaresmajose: For recipes with git sources we dont need to set S anymore but for the other they are still required. The previoes job fail was becasue of that:
```
S should be set relative to UNPACKDIR, e.g. replace WORKDIR with UNPACKDIR in "S = ${WORKDIR}/sources"
```

recipes-devtools/qc-image-unpacker/qc-image-unpacker_git.bb
```diff
@@ -13,7 +13,7 @@ SRC_URI = " \
 SRCREV = "28a783a9fc25dc87b7416b6d5b6f9ccd497d1c2e"
 
 PV = "0.2.0+"
-S = "${WORKDIR}/git/src"
+S = "${UNPACKDIR}/${BP}/src"
```
ndechesne, line 16: BP -> ${BB_GIT_DEFAULT_DESTSUFFIX}, no? Otherwise you are assuming that ${BB_GIT_DEFAULT_DESTSUFFIX} is always BP.

dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-adbd-cmdline.bb
ndechesne: but this is the default

S = "${UNPACKDIR}/${BP}"

why is it not enough?

recipes-devtools/qc-image-unpacker/qc-image-unpacker_git.bb
quaresmajose: If you take a look on the upstream commit https://git.openembedded.org/openembedded-core/commit/?id=f80c07019ddadaf9c5fb890faabfda7920ecd15e there are more references to this `${UNPACKDIR}/${BP}`, I followed the same pattern.

recipes-devtools/qtestsign/qtestsign_git.bb
quaresmajose: This caused probably because the file was previows edited on a windows machine or something else that don't add the end of line. So the change was caused by my editor that always add the end of file.

```patch
-BBCLASSEXTEND = "native nativesdk"
\ No newline at end of file
+BBCLASSEXTEND = "native nativesdk"
```

dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-adbd-cmdline.bb
quaresmajose: The second commit and your reference is wrong and should be:
```
Set S from UNPACKDIR in recipes that use only local files
```
https://git.openembedded.org/openembedded-core/commit/?id=f64b7e5fb3181734c8dde87b27e872a026261a74
please take a look ^ I will fix it when the current job finish.

recipes-devtools/qc-image-unpacker/qc-image-unpacker_git.bb
ndechesne: right. But I think they are recipes which don't use the git fetcher.

quaresmajose: fot this case maybe it wuld be better to set the fetcher subdir to `;subdir=${BB_GIT_DEFAULT_DESTSUFFIX}/src` and with that droping the S. I will test it.

recipes-devtools/qtestsign/qtestsign_git.bb
lumag: This should be fixed in a separate commit

recipes-devtools/qc-image-unpacker/qc-image-unpacker_git.bb
lumag: They are. See cracklib for example.

conf/layer.conf
```diff
@@ -11,7 +11,7 @@ BBFILE_PRIORITY_qcom = "6"
 
 LAYERDEPENDS_qcom = "core"
 LAYERRECOMMENDS_qcom = "openembedded-layer"
-LAYERSERIES_COMPAT_qcom = "styhead walnascar"
+LAYERSERIES_COMPAT_qcom = "styhead walnascar whinlatter"
```
lumag: Are we still compatible with styhead and walnascar?

quaresmajose: Yeah good point! With the above breaking changes we are only compatible with whinlatter.
Should I discard the others? We don't have those branches elsewhere.

lumag: Yes, please. If necessary, we can branch them later

recipes-devtools/qtestsign/qtestsign_git.bb
quaresmajose: done

conf/layer.conf
quaresmajose: done

lumag:
Dropping of old releases should come before backwards-incompatible changes

quaresmajose:
> Dropping of old releases should come before backwards-incompatible changes

Upstream in oe-core they come imediatly after the breaking changes.

koenkooi (approved):

PR #942, opened 2025-06-24
qcom-common: drop GL / EGL / GL ES providers

lumag:
Stop setting DEFAULT_PROVIDER in the qcom-common.inc. The default-providers.inc already sets them correctly and overriding those values in this file conflicts with the GLVND being enabled.


ricardosalveti (approved):

koenkooi (approved):
To add to this: this wasn't even a complete list! Looks good to me.

PR #944, opened 2025-06-26
ci: fix meta-qcom-distro builds

ricardosalveti:
Fixes for meta-qcom-distro builds, including extended lockfile, updated kas-build.yml, correct include order of the kas yml fragment and set qcom-multimedia-image as default distro image.

First two changes are general kas clean up fixes (removing deprecated options).

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -120,7 +120,7 @@ jobs:
           export KAS_WORK_DIR=$PWD/../kas
           mkdir $KAS_WORK_DIR
           ${KAS_CONTAINER} dump --resolve-env --resolve-local --resolve-refs \
-              ci/mirror.yml:ci/${{ matrix.machine }}.yml > kas-build.yml
+              ci/mirror.yml:ci/${{ matrix.machine }}.yml${{ matrix.distro.yamlfile }} > kas-build.yml
           ${KAS_CONTAINER} build ci/mirror.yml:ci/${{ matrix.machine }}.yml${{ matrix.distro.yamlfile }}
           ci/kas-container-shell-helper.sh ci/yocto-pybootchartgui.sh
           mv $KAS_WORK_DIR/build/buildchart.svg .
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -36,7 +36,7 @@ jobs:
 
       - name: Run kas lock
         run: |
-          ${KAS_CONTAINER} dump --update --lock --inplace ci/base.yml
+          ${KAS_CONTAINER} lock --update ci/base.yml
 
       - uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -36,7 +36,7 @@ jobs:
 
       - name: Run kas lock
         run: |
-          ${KAS_CONTAINER} lock --update ci/base.yml
+          ${KAS_CONTAINER} lock --update ci/base.yml:ci/distro.yml
 
       - uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -100,7 +100,7 @@ jobs:
           - name: poky-altcfg
             yamlfile: ""
           - name: qcom
-            yamlfile: 'ci/distro.yml:'
+            yamlfile: ':ci/distro.yml'
     runs-on: [self-hosted, x86]
     name: ${{ matrix.machine }}/${{ matrix.distro.name }}
     steps:
@@ -121,7 +121,7 @@ jobs:
           mkdir $KAS_WORK_DIR
           ${KAS_CONTAINER} dump --resolve-env --resolve-local --resolve-refs \
               ci/mirror.yml:ci/${{ matrix.machine }}.yml > kas-build.yml
-          ${KAS_CONTAINER} build ci/mirror.yml:${{ matrix.distro.yamlfile }}ci/${{ matrix.machine }}.yml
+          ${KAS_CONTAINER} build ci/mirror.yml:ci/${{ matrix.machine }}.yml${{ matrix.distro.yamlfile }}
           ci/kas-container-shell-helper.sh ci/yocto-pybootchartgui.sh
           mv $KAS_WORK_DIR/build/buildchart.svg .
 
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -20,7 +20,7 @@ repos:
   bitbake:
     url: https://github.com/openembedded/bitbake
     layers:
-      .: excluded
+      .: disabled
 
   meta-yocto:
     url: https://git.yoctoproject.org/meta-yocto
--- a/ci/distro.yml
+++ b/ci/distro.yml
@@ -19,3 +19,6 @@ repos:
       meta-oe:
       meta-python:
       meta-xfce:
+
+target:
+  - qcom-multimedia-image


mwasilew:
I booted the image from `qcom` directory and there is still no `adb`
https://quic-yocto-fileserver-1029608027416.us-central1.run.app/qualcomm-linux/meta-qcom/15910014242-1/qcom/qcs6490-rb3gen2-core-kit/core-image-base-qcs6490-rb3gen2-core-kit.rootfs.qcomflash.tar.gz
@koenkooi is this expected or we have a bug in artifact publishing?

ricardosalveti:
> I booted the image from `qcom` directory and there is still no `adb` https://quic-yocto-fileserver-1029608027416.us-central1.run.app/qualcomm-linux/meta-qcom/15910014242-1/qcom/qcs6490-rb3gen2-core-kit/core-image-base-qcs6490-rb3gen2-core-kit.rootfs.qcomflash.tar.gz @koenkooi is this expected or we have a bug in artifact publishing?

There is still an issue with the build setup, it seems it is not building the right distro / image, checking now.

.github/workflows/build-yocto.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
       - name: Run kas lock
         run: |
-          ${KAS_CONTAINER} dump --update --lock --inplace ci/base.yml
+          ${KAS_CONTAINER} lock --update ci/base.yml
```
ndechesne: seeing this line makes me realize we are locking only the repos mentioned in base.yml, and we've now added others in distro.yml. @quaresmajose , we need to fix that, right?

.github/workflows/build-yocto.yml
```diff
@@ -100,7 +100,7 @@ jobs:
           - name: poky-altcfg
             yamlfile: ""
           - name: qcom
-            yamlfile: 'ci/distro.yml:'
+            yamlfile: ':ci/distro.yml'
```
ndechesne, line 103: I probably missed that in the previous PR. but why do we have the colon here, and not later when we use the variable?

.github/workflows/build-yocto.yml
```diff
@@ -121,7 +121,7 @@ jobs:
           mkdir $KAS_WORK_DIR
           ${KAS_CONTAINER} dump --resolve-env --resolve-local --resolve-refs \
               ci/mirror.yml:ci/${{ matrix.machine }}.yml > kas-build.yml
-          ${KAS_CONTAINER} build ci/mirror.yml:${{ matrix.distro.yamlfile }}ci/${{ matrix.machine }}.yml
+          ${KAS_CONTAINER} build ci/mirror.yml:ci/${{ matrix.machine }}.yml${{ matrix.distro.yamlfile }}
```
ndechesne, line 124: d'oh... i am sorry we did not catch that in the review of #935... @koenkooi how was it tested in the first place? It turns out we never built with qcom-distro..

.github/workflows/build-yocto.yml
ndechesne: looks like @ricardosalveti saw that too, it's fixed later in the PR!

.github/workflows/build-yocto.yml
koenkooi: I rebased and rediffed it a lot, I suspect it broke along the way :/

.github/workflows/build-yocto.yml
```diff
@@ -36,7 +36,7 @@ jobs:
 
       - name: Run kas lock
         run: |
-          ${KAS_CONTAINER} lock --update ci/base.yml
+          ${KAS_CONTAINER} lock --update ci/base.yml:ci/distro.yml
```
koenkooi, line 39: Do we always want to lock everything, or should this use the ${{ matrix.distro.yamlfile }} variable?

quaresmajose: We need to lock every layer to build the same thing in all the jobs/workers builders

.github/workflows/build-yocto.yml
quaresmajose: I also prefer to use the `:` inplicity and not inside the variable

.github/workflows/build-yocto.yml
quaresmajose: yup we need to lock every layer to build the same thing in all the workers/jobs

.github/workflows/build-yocto.yml
quaresmajose: +1

quaresmajose (commented):
I had already seen that we had some deprecated warnings but I was waiting for the move to the IT managed self-hosted runners.

.github/workflows/build-yocto.yml
koenkooi: Even layers that aren't in BBLAYERS?

.github/workflows/build-yocto.yml
koenkooi: Since this one is optional and I don't know how kas handles two colons in a row, e.g. `foo.yml::bar.yaml`

lumag:
Build errors for `qcom-armv7a` should be fixed by https://github.com/qualcomm-linux/meta-qcom/pull/947

.github/workflows/build-yocto.yml
quaresmajose: kas can handles the `::` well. you can try with:
```
kas shell ci/mirror.yml::ci/base.yml
```
and check the mirror was seted correctly on the `conf/local.conf`

.github/workflows/build-yocto.yml
quaresmajose: In my opinion yes, regardless of whether it is used in BBLAYERS or not. The intention of this lock is to ensure that all layers will use the same _sha_ in all jobs.
Otherwise we run the risk of _jobA_ being able to use _sha1_ and _jobB_ that starts 1 minute later being able to use _sha2_ because in the meantime the layer was updated upstream.

.github/workflows/build-yocto.yml
ricardosalveti: Let me check, just assumed kas would not handle that.

.github/workflows/build-yocto.yml
ricardosalveti: Locking everything here won't cause any issue, even if the layer is not really used later on.

And we should lock everything in one single step otherwise we might end up with runs using different revisions of the same layer.

.github/workflows/build-yocto.yml
ricardosalveti:
```
$ kas build --skip repos_checkout ci/qcs6490-rb3gen2-core-kit.yml:
2025-06-27 12:24:12 - INFO     - kas 4.8.1 started
2025-06-27 12:24:12 - ERROR    - All concatenated config files must belong to the same repository or all must be outside of versioning control
```

The problem is not :: in a row, the problem is that for poky-altcfg the distro fragment is empty, and kas refuses that.

.github/workflows/build-yocto.yml
ricardosalveti: Kas can't handle `:` as the last argument.

.github/workflows/build-yocto.yml
quaresmajose: you can add a final `:` in the end of line and in this way it works
```
kas shell ci/mirror.yml::ci/base.yml::
```

.github/workflows/build-yocto.yml
quaresmajose: but can handle `::` so if we add `:[EMPTY_VAR]:` with always `:` in the end of line of the las argument, it will works.

.github/workflows/build-yocto.yml
ricardosalveti:
```
$ kas shell ci/qcs6490-rb3gen2-core-kit.yml::
2025-06-27 13:09:17 - INFO     - kas 4.8.1 started
2025-06-27 13:09:17 - ERROR    - All concatenated config files must belong to the same repository or all must be outside of versioning control
```

ndechesne (approved):

lumag (approved):

PR #946, opened 2025-06-27
Fix linux-yocto-dev build in qcom-armv7a

lumag:
Make linux-yocto-dev build for qcom-armv7a machines by pointing out a correct defconfig and fixing a missing include file.

### DIFF
--- a/recipes-kernel/linux/linux-yocto-dev/drivers/0008-PCI-pwrctrl-Add-power-control-driver-for-tc9563.patch
+++ b/recipes-kernel/linux/linux-yocto-dev/drivers/0008-PCI-pwrctrl-Add-power-control-driver-for-tc9563.patch
@@ -65,7 +65,7 @@ new file mode 100644
 index 00000000000000..547c764a6f405a
 --- /dev/null
 +++ b/drivers/pci/pwrctrl/pci-pwrctrl-tc9563.c
-@@ -0,0 +1,628 @@
+@@ -0,0 +1,629 @@
 +// SPDX-License-Identifier: GPL-2.0-only
 +/*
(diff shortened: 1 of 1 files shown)


recipes-kernel/linux/linux-yocto-dev/drivers/0008-PCI-pwrctrl-Add-power-control-driver-for-tc9563.patch
```diff
@@ -65,14 +65,15 @@ new file mode 100644
...
-@@ -0,0 +1,628 @@
+@@ -0,0 +1,629 @@
 +// SPDX-License-Identifier: GPL-2.0-only
 +/*
 + * Copyright (c) 2024 Qualcomm Innovation Center, Inc. All rights reserved.
 + */
 +
 +#include <linux/delay.h>
 +#include <linux/device.h>
++#include <linux/gpio/consumer.h>
```
ndechesne, line 76: why not a proper patch, with upstream-status?

lumag: I'm sorry, I don't understand the comment. There is an issue in the patch itself. I reported it as a response to the original patch, so hopefully it gets fixed in the next revision. It is more logical to fix the patch rather than adding more noise on top of it.

ndechesne: ah, sorry, I understand now.

ricardosalveti (approved):

PR #947, opened 2025-06-27
image-types-qcom: disable ESP for armv7 machines

lumag:
Qualcomm ARMv7 machines don't use EFI, ESP and UKI. Explicitly disable QCOM_ESP_IMAGE for those machines in order to let qcom-distro / qcom-multimedia-image work for qcom-armv7a.

### DIFF
--- a/classes-recipe/image_types_qcom.bbclass
+++ b/classes-recipe/image_types_qcom.bbclass
@@ -8,6 +8,9 @@ IMAGE_TYPES += "qcomflash"
 QCOM_ESP_IMAGE ?= "esp-qcom-image"
 QCOM_ESP_FILE ?= "${@'efi.bin' if d.getVar('QCOM_ESP_IMAGE') else ''}"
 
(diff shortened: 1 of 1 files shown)


classes-recipe/image_types_qcom.bbclass
```diff
@@ -8,6 +8,10 @@ IMAGE_TYPES += "qcomflash"
 QCOM_ESP_IMAGE ?= "esp-qcom-image"
 QCOM_ESP_FILE ?= "${@'efi.bin' if d.getVar('QCOM_ESP_IMAGE') else ''}"
 
+# ARMv7 machines don't use EFI / UKI
+QCOM_ESP_IMAGE:armv7a = ""
```
ndechesne: I would prefer qcom-armv7a, no?

lumag: ack

ndechesne (approved):

PR #948, opened 2025-06-27
qcom-armv[78]a: stop pulling in mesa-driver-msm

lumag:
Mesa drivers are not essential for the machine to boot. Also, it is a distro-level (or user-level) decision, which set of GL drivers to  pull into the resulting image. Stop listing mesa-driver-msm in MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS.


koenkooi:
What would be a better place to drag in this driver? Or does it get dragged in by something else already?

ricardosalveti (approved):

koenkooi (approved):

PR #952, opened 2025-07-01
workflows: save build_url for each machine separately

mwasilew:
Save build_url-<machine name> files as build artifacts. This change helps to re-run failed builds in PR. Till now the assumption was that all builds share the same upload URL. In case build jobs fail for infrastructure reasons and only failed builds are restarted, the URLs are not the same for all machines. This patch allows the test jobs to use appropriate URL for each machine separately.

It should be noted that machine names in kas files, build-yocto.yml and ci/lava directory must be kept in sync for this code to work properly.


mwasilew:
Please review carefully. The changes here will only be properly tested when merged to master branch. I tried to run a very similar set of workflows in my test repository and it seems to work.

.github/workflows/build-yocto.yml
```diff
@@ -156,25 +154,14 @@ jobs:
...
-          build_url: ${{ needs.compile.outputs.url }}
+      - name: Save URL
+        id: save-url
         run: |
-          echo "Downloads URL: ${build_url}"
-          echo "url=\"${build_url}\"" >> $GITHUB_OUTPUT
-          echo "${build_url}" > build_url
-          echo "## Download URL" >> $GITHUB_STEP_SUMMARY
-          echo "[${build_url}](${build_url})" >> $GITHUB_STEP_SUMMARY
+          echo "${{ steps.compile.outputs.url }}" > build_url-${{ matrix.machine }}
```
ndechesne: the matrix has 2 axis, for machine and distro. don't you need to take matrix.distro into account?

mwasilew: sorry, I overlooked that. Let me do some experiments and come back.

mwasilew:
I changed generating test jobs to `tuxlava`. Unfortunately I forgot to include db410c and db820c there so this PR is still incomplete.

mwasilew:
There is one more issue with jobs generated by tuxlava. They use archived version of test-definitions, for example:
https://storage.tuxboot.com/test-definitions/2025.06.tar.zst
Downloading this archive is not currently possible in Qualcomm. There is a ticket trying to address the problem.

mwasilew:
I changed a few things compared to the last version. The example workflow was executed from next branch https://github.com/qualcomm-linux/meta-qcom/actions/runs/18156877121

lumag (changes requested):
```
(node:1) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
```

lumag (approved):

mwasilew:
I separated the summary step from the build. This is how it looks now:
https://github.com/qualcomm-linux/meta-qcom/actions/runs/18202541790
I also added a build_url file for backward compatibility so the tests can complete. It can be removed after this PR is merged.

ricardosalveti:
@mwasilew please rebase it again.

ci/generatetests.py
```diff
@@ -0,0 +1,41 @@
+import logging
```
lumag, line 4: ✖ source-license-headers-exist:
  	✖ ci/generatetests.py: The first 60 lines do not contain the pattern(s): Copyright(\s)*(\(c\)|©)?, SPDX-License-

lumag (changes requested):

ci/generatetests.py
mwasilew: Yes, I noticed. Pushing a fix now

PR #954, opened 2025-07-03
Drop deprecated linux-qcom-staging kernel recipe

vkraleti:
The `linux-qcom-staging` kernel is now deprecated and no longer maintained.
Remove the recipe to reduce maintenance overhead.


lumag (commented):
Removing -> Remove.

LGTM otherwise

sbanerjee-quic (approved):

PR #957, opened 2025-07-04
linux-qcom-next: update LINUX_VERSION and SRCREV

vivpuar:
Moving to 6.16-rc4-20250704 tag for linux-qcom-next and update SRCREV as 6.16-rc3 is giving fetch error leading to build failure

### DIFF
--- a/recipes-kernel/linux/linux-qcom-next_git.bb
+++ b/recipes-kernel/linux/linux-qcom-next_git.bb
@@ -8,12 +8,12 @@ inherit kernel
 
 COMPATIBLE_MACHINE = "(qcom)"
 
-LINUX_VERSION ?= "6.16-rc3"
+LINUX_VERSION ?= "6.16-rc4"
 
 PV = "${LINUX_VERSION}+git"
 
-# tag: qcom-next-6.16-rc3-20250701
-SRCREV ?= "54f158d766ac86c95e97a17c9e51db880d2d30dd"
+# tag: qcom-next-6.16-rc4-20250704
+SRCREV ?= "52f18bdf8da147f2659862cfb77aeccddd40cb67"
 
 SRC_URI = "git://github.com/qualcomm-linux/kernel.git;branch=qcom-next;protocol=https"
 


recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -8,12 +8,12 @@ inherit kernel
 
 COMPATIBLE_MACHINE = "(qcom)"
 
-LINUX_VERSION ?= "6.16-rc3"
+LINUX_VERSION ?= "6.16-rc4"
```
koenkooi, line 11: This will cause issues when 6.16 gets released, `6.16-rcX` sorts higher than `6.16`. This will result in build failures due to the OE-core version checks ('version going backward') and if using package management, break upgrades.

vkraleti: I am afraid, we already introduced this issue. This change is not bringing it new.

lumag: @koenkooi what's the recommended version for RC releases? I see that some OE recipes use 'A.B.CrcN', without a dash before rc. Is that a better option?

koenkooi: @lumag the least bad way I know of is `current+new-rcX`, e.g. `6.15+6.16rc3`. Opkg (and very likely dpkg as well) has a utility called opkg-compare-versions (https://github.com/shr-project/opkg-utils/blob/master/opkg-compare-versions.c) that can help you with that. There is some nuance between opkg, dpkg+apt and rpm, but most version numbers in OE are simple enough to work the same will all 3 of those.

I'm trying to avoid dashes in version numbers, it's way too easy to put one in the 'wrong' place and dpkg+apt will behave in a way a non-Debian person doesn't expect :)

vkraleti: @koenkooi , @lumag I am fine with this proposal. 
As PV is already set as `6.16-rcX` in current recipe, going back to `6.15+6.16rcX` would break the build with [version-going-backwards] error. So can this be done while moving to 6.17-rc1? Till then want to continue upgrades just by bumping only rc number.

lumag: But if we follow your proposal the recipe will error out when 6.16 is released. So neither of paths is error-free.

koenkooi: I'm fine with fixing it for 6.17 and leaving 6.16 as is. Like @lumag says, something will break either way :/

lumag: Okay. This  got merged. Which version are we going to break?

koenkooi: @lumag 6.16.x will break, the 6.17rc recipe(s) will have proper versioning.

PR #958, opened 2025-07-04
linux-qcom-next: use 'nobranch=1' in SRC_URI to support retagged commits

vkraleti:
To avoid fetcher issues caused by the retargeting of the `qcom-next` branch on every tag upmerge, use `nobranch=1` in `SRC_URI`. This allows BitBake to search for the specified `SRCREV` across the entire repository, including commits that are no longer associated with the current branch tip but are still preserved by tag.

The `branch` parameter should be used only in cases like `devupstream`, where a stable and consistent branch reference is expected.

### DIFF
--- a/recipes-kernel/linux/linux-qcom-next_git.bb
+++ b/recipes-kernel/linux/linux-qcom-next_git.bb
@@ -15,7 +15,10 @@ PV = "${LINUX_VERSION}+git"
 # tag: qcom-next-6.16-rc3-20250701
 SRCREV ?= "54f158d766ac86c95e97a17c9e51db880d2d30dd"
 
-SRC_URI = "git://github.com/qualcomm-linux/kernel.git;branch=qcom-next;protocol=https"
+SRCBRANCH ?= "nobranch=1"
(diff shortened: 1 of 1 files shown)


lumag:
Please mention that those commits have a tag, otherwise it's not obvious that they won't be vacuum cleaned by GitHub.

lumag:
> > Please mention that those commits have a tag, otherwise it's not obvious that they won't be vacuum cleaned by GitHub.
> 
> How to do that? Is there a setting to preserve tagged commits in the branch?

No. the kernel repo preserves the tags, just mention that they exist in your commit message.

vkraleti:
> > > Please mention that those commits have a tag, otherwise it's not obvious that they won't be vacuum cleaned by GitHub.
> > 
> > 
> > How to do that? Is there a setting to preserve tagged commits in the branch?
> 
> No. the kernel repo preserves the tags, just mention that they exist in your commit message.

Updated commit message. Please check.

sbanerjee-quic (approved):

PR #941, opened 2025-06-24
WIP ci: Move to IT managed self-hosted runners

doanac:

PR #960, opened 2025-07-09
WIP ci: Move to IT managed AWS self-hosted runners

sampra2025:
These runners should be equivalent to the currently used in GCP.

IT provides two types of runners that are just like what is been used in in GCP:
During UAT, the label will be named with "*-dev"

 * `runs-on: [self-hosted, qcom-u2404, amd64-dev]`
 * `runs-on: [self-hosted, qcom-u2404, arm64-dev]`

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -8,14 +8,14 @@ on:
         value: ${{ jobs.create-output.outputs.url }}
 
 env:
-  CACHE_DIR: /srv/gh-runners/quic-yocto
-  KAS_REPO_REF_DIR: /srv/gh-runners/quic-yocto/kas-mirrors
-  KAS_CONTAINER: /srv/gh-runners/quic-yocto/kas-mirrors/kas-container
+  CACHE_DIR: /efs/qli/meta-qcom/
+  KAS_REPO_REF_DIR: /efs/qli/meta-qcom/kas-mirrors
+  KAS_CONTAINER: /efs/qli/meta-qcom/kas-mirrors/kas-container
 
 jobs:
   kas-setup:
     if: github.repository == 'qualcomm-linux/meta-qcom'
-    runs-on: [self-hosted, x86]
+    runs-on: [self-hosted, qcom-u2404, amd64-dev]
     steps:
       - name: Update kas-container
         run: |
@@ -46,7 +46,7 @@ jobs:
   yocto-check-layer:
     needs: kas-setup
     if: github.repository == 'qualcomm-linux/meta-qcom'
-    runs-on: [self-hosted, x86]
+    runs-on: [self-hosted, qcom-u2404, amd64-dev]
     steps:
       - uses: actions/checkout@v4
         with:
@@ -64,7 +64,7 @@ jobs:
   yocto-patchreview:
     needs: kas-setup
     if: github.repository == 'qualcomm-linux/meta-qcom'
-    runs-on: [self-hosted, x86]
+    runs-on: [self-hosted, qcom-u2404, amd64-dev]
     steps:
       - uses: actions/checkout@v4
         with:
@@ -102,7 +102,7 @@ jobs:
             yamlfile: ""
           - name: qcom
             yamlfile: ':ci/distro.yml'
-    runs-on: [self-hosted, x86]
+    runs-on: [self-hosted, qcom-u2404, amd64-dev]
     name: ${{ matrix.machine }}/${{ matrix.distro.name }}
     steps:
       - uses: actions/checkout@v4
@@ -162,7 +162,7 @@ jobs:
     needs: compile
     outputs:
       url: ${{ steps.print-output.outputs.url }}
-    runs-on: [self-hosted, x86]
+    runs-on: [self-hosted, qcom-u2404, amd64-dev]
     steps:
       - name: "Print output"
         id: print-output


lumag:
What's the difference from #941? We already had a similarly named PR.

koenkooi:
This PR now has 3 commits with the same headline, but different contents and 4 merge commits. Can this be rebased and cleaned up?

.github/workflows/build-yocto.yml
```diff
@@ -191,4 +191,4 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: build_url
-          path: build_url
+          path: build_url
```
ndechesne: can you please remove this hunk which is there by mistake?

ndechesne:
4 build jobs have been stuck and/or canceled. Do you know why @sampra2025 ? I would like to make progress on this PR, and merge it, but we need to verify that it's working fine, and it's stable.

sampra2025:
> > I tried to squash my commits but somehow it pulled other commits to my PR. I don't know how to clean-up now. Any help will be appreciated.
> 
> Please learn to use git: https://git-scm.com/book/en/v2

I have cleaned my branch after rebasing it to the tip of master.

lumag:
The patch is not consistent:

```
From: Satish Mhaske <smhaske@qti.qualcomm.com>
Signed-off-by: Satish Mhaske <smhaskey@qti.qualcomm.com>
```

ndechesne:
Good news, we have a first run in CI that succeeded using the new AWS runners.

However, the build is very slow. If we compare to the latest nightly build using the current GCP runners (https://github.com/qualcomm-linux/meta-qcom/actions/runs/16793280808/usage), we can see the following:

GCP runners:
- total compute time: 7h
- none of the individual/platform build takes more than 30 mins

AWS runners:
- total compute time: 50h
- Most of the builds are around 2h per platform

I will need to have a closer look at where we spend time, but @sampra2025  can you please check the instance configuration, something is clearly set up wrong. Are we sure we are using a fast, local disk for example? 

I briefly checked the log for qcs6490-rb3gen2-core-kit/qcom-distro for both the nightly build and this one, and they both seem to use the same amount of sstate content
Sstate summary: Wanted 4620 Local 4315 Mirrors 0 Missed 305 Current 0 (93% match, 0% complete)
vs
Sstate summary: Wanted 4620 Local 4604 Mirrors 0 Missed 16 Current 0 (99% match, 0% complete)

Both builds almost used the same commit of oe-core (a handful of changes difference)

I will run the jobs again now, to see if we get consistent slow builds

ndechesne:
The new build I re-ran was much better. It looks like the sstate cache was not properly filled in in the first build. In the most recent build, all builds took around 15 mins, and one job was stuck waiting for a runner for a couple of hours, but eventually took 15 min. 

In the previous instance, we went through a full rebuild (including gcc, llvm..). I don't know if 2h in this case is expected, i will research on that more. 

I will run the the build one time to test for stability.

.github/workflows/build-yocto.yml
```diff
@@ -85,7 +85,7 @@ jobs:
     outputs:
       url: ${{ steps.upload_artifacts.outputs.url }}
     strategy:
-      fail-fast: true
+      fail-fast: false
```
ndechesne: If we wanted to make this change, it should be its own commit. However I don't think we should make this change. We do want to fail-fast, there is no point in finish the gazillions of builds we are doing if any build fails, since we will need to update the PR anyway.

sampra2025: The only reason behind this proposal is to avoid compute cost on EC2 runners. As you know, each workflow spins up around 18–20 runners, and if any one target fails late in the process, we end up wasting a lot of compute time, sometime its like 60+ minutes per target/runner, which adds up to ~20 hours in total.

Instead if there is way to validate the the change in one or few of them, then we might avoid the wasting of compute time and cost. Just a thought.

ricardosalveti: Let's not change this for now and work on trying to optimize (e.g. building only qemuv8 / poky as the initial run, and trigger the rest once it finishes successfully) in parallel.

We do want fail-fast in any scenario anyway, we just need to find a better way to chain the runs to avoid multiple runs that are known to fail.

sampra2025: Thanks for clarifying on fail fast, I will revert my change to fail_fast: true

ndechesne: > The only reason behind this proposal is to avoid compute cost on EC2 runners. As you know, each workflow spins up around 18–20 runners, and if any one target fails late in the process, we end up wasting a lot of compute time, sometime its like 60+ minutes per target/runner, which adds up to ~20 hours in total.

Every time a package/recipe is built in the OE build system, its output is included in the persistent cache. So we are effectively never wasting compute time, even if we fail the job, everything we've done up to the time the job is canceled will be reused/leveraged when we restart the jobs. 

However as soon as one of the matrix builds fail, we for sure, need to cancel all other jobs to give the developer a chance to fix the problem. we will never merge a PR if any of the matrix build has failed.

ricardosalveti: @sampra2025 latest revision still has this change, can you update it again and remove this piece?

.github/workflows/build-yocto.yml
ricardosalveti: @sampra2025 please remove this piece so we can unblock this merge.

lumag: This chunk is still present. Please drop it.

.github/workflows/build-yocto.yml
```diff
@@ -191,5 +191,4 @@ jobs:
       - name: Upload build URL
         uses: actions/upload-artifact@v4
         with:
-          name: build_url
-          path: build_url
+          name: build_url
```
lumag: Extra chunk still present.

lumag (changes requested):

.github/workflows/build-yocto.yml
lumag: `Run actions/upload-artifact@v4
Error: Input required and not supplied: path`

.github/workflows/build-yocto.yml
ricardosalveti: Removed in the last rev.

ricardosalveti:
Current run is without any sstate cache, we'll see how it goes.

lumag (approved):
LGTM now, thanks!

ricardosalveti (approved):

ricardosalveti:
@ndechesne in theory we should be good to switch, do you want to give it a try with a few more builds and benchmark the results or should we merge and follow up as we see more build runs?

PR #962, opened 2025-07-10
linux-yocto-6.12: drop an applied patch

vkraleti:
Drop the patch 0004-arm64-dts-qcom-qcm2290-fix-some-of-QUP-interconnects.patch, which is part of linux-yocto v6.12.34


koenkooi:
The CI failure seems unrelated, retriggering might make it go away.

lumag (approved):

PR #963, opened 2025-07-14
GitHub workflow to compile qcom-armv8a with linux-yocto-dev

vkraleti:
Update the GitHub Actions workflow to support building qcom-armv8a machine with linux-yocto-dev kernel in CI.


.github/workflows/build-yocto.yml
```diff
@@ -120,9 +126,15 @@ jobs:
           export SSTATE_DIR=${CACHE_DIR}/sstate-cache
           export KAS_WORK_DIR=$PWD/../kas
           mkdir $KAS_WORK_DIR
-          ${KAS_CONTAINER} dump --resolve-env --resolve-local --resolve-refs \
-              ci/mirror.yml:ci/${{ matrix.machine }}.yml${{ matrix.distro.yamlfile }} > kas-build.yml
-          ${KAS_CONTAINER} build ci/mirror.yml:ci/${{ matrix.machine }}.yml${{ matrix.distro.yamlfile }}
+          if [ "${{ matrix.kernel }}" = "default" ]; then
```
koenkooi: Using if statements leads to a lot of duplication, could a way like ${{ matrix.distro.yamlfile }} be used, where the var being empty is a valid config?

vkraleti: I made a few modifications to avoid if-else usage. Can you check now?

koenkooi (approved):

lumag (approved):

PR #968, opened 2025-07-18
hexagon-dsp-binaries: upgrade 20250627 -> 20250708

vivpuar:
No upstream changes, just upgrade following OE-Core's bump of linux-firmware version


ci/base.yml
```diff
@@ -44,4 +44,4 @@ local_conf_header:
 machine: unset
 
 target:
-  - core-image-base
+  - hexagon-dsp-binaries
```
koenkooi: I don't think this change should be here

vivpuar: Yes correct, apology for this. Will update.

lumag (approved):

vkraleti (approved):

PR #969, opened 2025-07-18
enable graphicsdlkm module only for linux-yocto-dev

vkraleti:
The graphicsdlkm kernel module can only be compiled with the linux-yocto-dev kernel. To avoid build failures when using other kernel providers, enable this module conditionally based on the current kernel provider.


koenkooi:
Does it _only_ work linux-yocto-dev or does it work with other kernels as well? What's the long term plan for getting it to work with any kernel?

lumag:
I think disabling it for linux-qcom-next defeats the purpose of that kernel. We are expected to be running linux-qcom-next and (optionally) Qualcomm proprietary stack on supported devices.

PR #975, opened 2025-07-23
enable kernel debug features by default in CI

vkraleti:
CI images are primarily used for testing, and enabling ramdump collection, page owner tracking is helpful for debugging kernel oops issues. Update the kernel command line via `local.conf` to include the necessary options to support these features.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -33,6 +33,8 @@ local_conf_header:
     INHERIT += "buildstats buildstats-summary"
     INHERIT += "buildhistory"
     INHERIT += "rm_work"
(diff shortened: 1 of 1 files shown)


lumag (changes requested):
Your commit message mentions ramdump only, while you are enabling other debug-related options.

lumag:
I think ` reboot=panic_warm`  is a bad idea, it resets the board, loosing all the state and destroying the logs (if the log is simply directed to the terminal window).

sbanerjee-quic (approved):

PR #977, opened 2025-07-23
ci/qcom-distro.yml: enable meta-virtualization and its dependencies

koenkooi:
This enables recipes like `docker-moby` and `podman`.

Related to https://github.com/qualcomm-linux/meta-qcom-distro/issues/53

### DIFF
--- a/ci/qcom-distro.yml
+++ b/ci/qcom-distro.yml
@@ -13,12 +13,21 @@ repos:
   meta-openembedded:
     url: https://github.com/openembedded/meta-openembedded
     layers:
+      meta-filesystems:
       meta-gnome:
       meta-multimedia: 
       meta-networking:
       meta-oe:
       meta-python:
       meta-xfce:
+
+  meta-virtualization:
+    url: https://git.yoctoproject.org/git/meta-virtualization
+    branch: master
+
+local_conf_header:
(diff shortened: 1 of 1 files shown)


ci/qcom-distro.yml
```diff
@@ -13,12 +13,17 @@ repos:
...
       meta-gnome:
       meta-multimedia: 
       meta-networking:
       meta-oe:
       meta-python:
       meta-xfce:
 
+  meta-virtualization:
+    url: https://git.yoctoproject.org/git/meta-virtualization
+    branch: master
```
lumag, line 26: We also need to enable `virtualization` in `DISTRO_FEATURES`.

koenkooi: @lumag 'need' is the wrong word for that in our case, e.g. `docker` builds and runs fine without it. That DISTRO_FEATURE mainly deals with libvirt and enables things like KVM in the kernel configs. IMO it does too much to be worth enabling.
The downside is that every build now gets an annoying nag message about that DISTRO_FEATURE not being enabled....

lumag: Yeah. There is a `SKIP_META_VIRT_SANITY_CHECK` which we can set.

PR #983, opened 2025-07-29
ci: reduce the amount of files published for each build

ndechesne:
Each build exports close to 590 files, because it was initially decided to 'blindly' publish the deploy_dir folder which is bitbake output directory.

In a typical build, there are 25 symlinks which end up being duplicate files on the file server (1.3GB per platform). The symlinks are convenient/shortened filenames, and don't need to be published.

deploy_dir also contains 'partitions' folder which is an intermediate artifact used to generate the image GPT files, with 490 files.

To clean up the published files, and to avoid growing it by mistake in the future, this commit switches the logic. Instead of publishing the entire deploy_dir folder (with minimal cleanup), the CI job now only publishes real files (no symlinks) located in the root of deploy_dir. Any additional files needed (later) can be manually added.

For a local RB3G2 build, it's 26 files, for 900MB.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -160,11 +160,10 @@ jobs:
           # expects file to be relative to our PWD. deploy_dir is outside
           # that, so we move things around:
           deploy_dir=../kas/build/tmp/deploy/images/${{matrix.machine}}
-          uploads_dir=./uploads/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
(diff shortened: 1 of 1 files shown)


koenkooi (dismissed):
I like this a lot! Another reason to move non-useful files to subfolders.

lumag (dismissed):

ndechesne:
I removed this line:

`mv $deploy_dir $uploads_dir/`

which was also 'moving' the root folder which is the <machine>. so with the current match, all files for all machines end up in the same folder..

ndechesne:
The problem I found yesterday is fixed. We seem to have CI issues right now, but other than that, this patch looks good now.

PR #984, opened 2025-07-29
packagegroup-qcs615-adp-air: packagegroups for QCS615-ADP-Air

fangwu1:
Add firmware packagegroups for the QCS615-ADP-Air board.

### DIFF
--- a/recipes-bsp/packagegroups/packagegroup-qcs615-adp-air.bb
+++ b/recipes-bsp/packagegroups/packagegroup-qcs615-adp-air.bb
@@ -1,0 +1,14 @@
+SUMMARY = "Packages for the QCS615-ADP-Air platform"
+
+inherit packagegroup
+
+PACKAGES = " \
+    ${PN}-firmware \
+"
+
+RRECOMMENDS:${PN}-firmware = " \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'linux-firmware-qcom-adreno-a630 linux-firmware-qcom-qcs615-adreno', '', d)} \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'wifi', 'linux-firmware-ath11k-qca6698aq', '', d)} \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'bluetooth', 'linux-firmware-qca-qca6698', '', d)} \
+    linux-firmware-qcom-venus-5.4 \
+"
--- a/conf/machine/qcs615-adp-air.conf
+++ b/conf/machine/qcs615-adp-air.conf
@@ -12,5 +12,9 @@ KERNEL_DEVICETREE ?= " \
                       qcom/qcs615-ride.dtb \
                       "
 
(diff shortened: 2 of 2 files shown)


recipes-bsp/packagegroups/packagegroup-qcs615-adp-air.bb
```diff
@@ -0,0 +1,14 @@
...
+
+inherit packagegroup
+
+PACKAGES = " \
+    ${PN}-firmware \
+"
+
+RRECOMMENDS:${PN}-firmware = " \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'linux-firmware-qcom-adreno-a630 linux-firmware-qcom-qcs615-adreno', '', d)} \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'wifi', 'linux-firmware-ath11k-qca6698aq linux-firmware-ath10k-qca9377', '', d)} \
```
lumag: Why do we need two sets here?

fangwu1: Confirmed with WLAN team, this is the customer requirement to use different WLAN chip. It can be inserted on board.

lumag: Then the same customer might potentially insert Intel or some other M.2 card.
We should list here only the firmware that is required for the boards that we ship. If somebody swaps M.2 card, it's their responsibility to pull in necessary packages.

fangwu1: Thanks for your feedback! Updated to remove optional firmware entry. Please take another look.

ricardosalveti:
Please squash the changes, see go/OSSBestPractices.

fangwu1:
Hi, @ricardosalveti, already updated

recipes-bsp/packagegroups/packagegroup-qcs615-adp-air.bb
```diff
@@ -0,0 +1,14 @@
+SUMMARY = "Packages for the QCS615-ADP-Air platform"
+
+inherit packagegroup
+
+PACKAGES = " \
+    ${PN}-firmware \
+"
+
+RRECOMMENDS:${PN}-firmware = " \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'linux-firmware-qcom-adreno-a630 linux-firmware-qcom-qcs615-adreno', '', d)} \
```
lumag, line 10: Nit: this needs to be changed to a612 (to get the RGMU firmware) after the next linux-firmware release.
For now this is good (and correct).

ricardosalveti (approved):

recipes-bsp/packagegroups/packagegroup-qcs615-adp-air.bb
fangwu1: Sure, thanks for reminding

PR #987, opened 2025-07-31
Add LAVA template for qcs8300-ride-sx machine

mattface:
This is a clone of the template for qcs9100-ride-sx with the
device-type changed as the LAVA deployment method is the same.
I have tested with a manual job
https://lava.infra.foundries.io/scheduler/job/86608

### DIFF
--- a/ci/lava/qcs8300-ride-sx/boot.yaml
+++ b/ci/lava/qcs8300-ride-sx/boot.yaml
@@ -1,0 +1,63 @@
+actions:
+- deploy:
+    images:
+      image:
+        headers:
+          Authentication: Q_GITHUB_TOKEN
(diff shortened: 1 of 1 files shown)


lumag (changes requested):
Other than an obvious issue of the commit subject not following the pattern (hint: `git log ci/lava`), I have a bigger issue: this is a copy of the existing file with just one line change. Can we do something better than that?

lumag:
> LGTM, we expect to drop the duplication with tuxlava.

Ack. Then just the subject needs to be fixed.

ricardosalveti:
Mind rebasing this pr?

lumag (approved):

PR #992, opened 2025-08-07
Deploy contents.xml to enable automatic image flashing in Axiom

vkraleti:
The 'contents.xml' file maps images to their respective partitions in a format understood by Axiom. Deploy the same to enable Axiom automatic image flashing on supported boards.

### DIFF
--- a/classes-recipe/image_types_qcom.bbclass
+++ b/classes-recipe/image_types_qcom.bbclass
@@ -7,6 +7,8 @@ IMAGE_TYPES += "qcomflash"
 
 QCOM_BOOT_FIRMWARE ?= ""
 
+QCOM_CONTENTS_XML ?= ""
+
 QCOM_ESP_IMAGE ?= "esp-qcom-image"
 QCOM_ESP_FILE ?= "${@'efi.bin' if d.getVar('QCOM_ESP_IMAGE') else ''}"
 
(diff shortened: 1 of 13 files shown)


ricardosalveti:
There is a lot of content at https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/qualcomm_linux-spf-1-0/qualcomm-linux-spf-1-0_test_device_public/r1.0_00095.0/qcs9100-le-1-0/contents.xml that are related to the metabuild, and I know we were able to demonstrate with a minimal contents.xml file before, can we simplify and have only what we indeed need for axiom here?

ricardosalveti:
Depending on what we need in the xml file, we could install / deploy all xmls (from all targets) as part of a single recipe, and pick only one for the tarball, instead of having multiple recipes.

lumag:
> > There is a lot of content at https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/qualcomm_linux-spf-1-0/qualcomm-linux-spf-1-0_test_device_public/r1.0_00095.0/qcs9100-le-1-0/contents.xml that are related to the metabuild, and I know we were able to demonstrate with a minimal contents.xml file before, can we simplify and have only what we indeed need for axiom here?
> 
> Yeah this file is a bit heavy. How ever the plan is to start using the contents.xml file provided with the GA 1.5 release as-is. In parallel, work on dynamically generating contents.xml based on the actual images and partition layout. This enables the Test and IT teams to start Axiom validations and perform pipeline cleaning.

I don't like this idea. Test and IT teams can perform Axiom validation by using this PR as a base, but I don't think we should be landing it. These contents.xml files are bloated with files and locations we don't use and don't provide in our builds.

ricardosalveti:
> I've looked at the XML file, and the contents seem close to what qcom-ptool generates; perhaps this is where we should be generating contents.xml files for the various platforms.

https://github.com/qualcomm-linux/qcom-ptool/pull/29

vkraleti:
This contents.xml generation steps have significantly changed for the time this PR is created. Most of the changes in this PR are no longer relevant. To avoid confusion in reviews, I am closing this PR. Please follow https://github.com/qualcomm-linux/meta-qcom/pull/1066 for latest patches on this topic.

PR #1019, opened 2025-08-18
Add qcom-preflight-checks workflow

njjetha:
## Summary
qcom-preflight-checks calls a [reusable-workflow](https://github.com/qualcomm/qcom-reusable-workflows/blob/main/.github/workflows/qcom-preflight-checks-reusable-workflow.yml) that runs a series of preflight checks on your proposed contribution.

For more info check https://github.com/qualcomm/qcom-actions/blob/main/README.md


.github/workflows/qcom-preflight-checks.yml
```diff
@@ -0,0 +1,24 @@
...
+    branches: [ "master" ]
+  workflow_dispatch:
+
+permissions:
+ contents: read
+ security-events: write
+
+jobs:
+  qcom-preflight-checks:
+    uses: qualcomm/qcom-reusable-workflows/.github/workflows/qcom-preflight-checks-reusable-workflow.yml@v1.1.4
```
lumag: Can we use semantic tags and just require `@v1` here to get the latest one? E.g. when a new version gets released we definitely don't want to update all repos which use this workflow.

.github/workflows/qcom-preflight-checks.yml
```diff
@@ -0,0 +1,24 @@
...
+    with:
+        # ✅ Preflight Checkers
+        repolinter: true                   # default: true
+        semgrep: true                      # default: true
```
lumag: OE layers are a mixture of shell code, python snippets and patches. Does `semgrep` report anything sensible in such a case?

.github/workflows/qcom-preflight-checks.yml
```diff
@@ -0,0 +1,24 @@
+name: Qualcomm Preflight Checks
+on:
+  pull_request_target:
+    branches: [ "master" ]
+  push:
+    branches: [ "master" ]
```
lumag: Will this require updating for other branches? If so, why?

.github/workflows/qcom-preflight-checks.yml
lumag: Also, please restructure the workflow's repo in order to remove the need to specify the path inside the repo (like all other GH actions do).

.github/workflows/qcom-preflight-checks.yml
njjetha: @lumag [semgrep](https://semgrep.dev/docs/supported-languages) support python lang.
Yes if there is any vulnerability detected it will comment in the PR, semgrep job will be triggered for pull_request only.

.github/workflows/qcom-preflight-checks.yml
njjetha: Workflow will be triggered for `master `branch only if you want to trigger for all the branches `*` we can provide.

.github/workflows/qcom-preflight-checks.yml
lumag: Vulnerability in what? In the build recipe?
Also, what does it know about the OE Python libraries or other OE specifics?

.github/workflows/qcom-preflight-checks.yml
lumag: Yes, of course. We need similar handling for all stable branches.

.github/workflows/qcom-preflight-checks.yml
njjetha: @lumag 
One question if my minor and patch version are changing @v1 will always points to latest version Example
`v1, v1.1.1, v1.1.2,  v1.1.3,  v1.1.4`
In this case v1 will always points to v1.1.4 , 

I am not aware of this before. I need to test it.

I didn't understood your last comment
`Also, please restructure the workflow's repo in order to remove the need to specify the path inside the repo`

njjetha: @lumag Thanks, I read about it we would use @v1 it will points to latest minor and patch version.

lumag: > I didn't understood your last comment
> `Also, please restructure the workflow's repo in order to remove the need to specify the path inside the repo`

In all other actions we simple reference the repo and the tag. Here you require the full path inside the repo. Would it be possible to restructure preflight checks repo so that we don't have to specify that path?

njjetha: @lumag In reusable workflows, you must specify the full path to the workflow file within the repository. This is required by GitHub to locate the workflow. In contrast, composite actions can be referenced simply by the repository and tag, without needing to specify the internal file path

.github/workflows/qcom-preflight-checks.yml
njjetha: It’s a static analysis tool that scans source code to find patterns that may indicate security issues, bugs, or violations of coding standards.

.github/workflows/qcom-preflight-checks.yml
njjetha: What are stable branches in the repo please let me know I will add them.

lumag: Enable it for all branches so that we don't have to update config each time new branch is spawned.

.github/workflows/qcom-preflight-checks.yml
lumag: So, can it identify python code in OE recipes? Can it scan build recipes? What is the impact of having a "vulnerability" in OE recipe?

mynameistechno: Hi @lumag, @njjetha may not be the best person to answer questions about Semgrep. We worked with QPSI to enable this tool as an alternative to GitHub's codeQL. @igibek from QPSI can answer further questions regarding it. 

It supports Python for sure, and I think shell as well, but I don't think it natively supports bitbake. @igibek the OE recipe files can include snippets of python and shell. Do we think semgrep can parse these?

@lumag, depending on what @igibek finds out, we could disable the Semgrep check if it is interfering.

.github/workflows/qcom-preflight-checks.yml
njjetha: Resolved in [6a9d297](https://github.com/qualcomm-linux/meta-qcom/pull/1019/commits/6a9d2976ac3f6f40a0787304cb31850275f90392)

.github/workflows/qcom-preflight-checks.yml
njjetha: I had disabled semgrep check

.github/workflows/qcom-preflight-checks.yml
lumag: Ack. This still uses the full tag though. Can we shift to specifying it semantically; `@v1`?

.github/workflows/qcom-preflight-checks.yml
```diff
@@ -0,0 +1,26 @@
...
+    uses: qualcomm/qcom-reusable-workflows/.github/workflows/qcom-preflight-checks-reusable-workflow.yml@v1.1.4
+    with:
+        # ✅ Preflight Checkers
+        repolinter: true                   # default: true
+        semgrep: false                      # default: true
+        copyright-license-detector: true   # default: true
+        pr-check-emails: true              # default: true
+        dependency-review: true            # default: true
+    secrets:
+      SEMGREP_APP_TOKEN: ${{ secrets.SEMGREP_APP_TOKEN }}
```
lumag: with semgrep being gone, this should not be necessary.

.github/workflows/qcom-preflight-checks.yml
ricardosalveti: Yeah, please create a base tag (e.g. `@v1`) and use that instead here, then we won't have to keep this updated on minor workflow updates.

.github/workflows/qcom-preflight-checks.yml
njjetha: Resolved in 3689c78

PR #1020, opened 2025-08-18
ci: run a full sanity build first

ndechesne:
The matrix jobs results in a large amount of jobs being started at the same time. Running all jobs in parallel is not ideal to leverage the OE sstate cache. For any common tasks (native, armv8, .. ) when the object is missing from the sstate cache, it's likely going to be a miss for several jobs. As a result, each of them will build the missed object, and each of them will attempt to create the object in the sstate cache.

Instead, with this change, we first build for the generic qcom_armv8 machine, and if that builds succeeds we carry on with the other machines, with the hope that the sstate cache hit will be better.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -61,8 +61,42 @@ jobs:
         run: |
           ci/kas-container-shell-helper.sh ci/yocto-patchreview.sh
 
-  compile:
+  compile_warm_up:
     needs: kas-setup
+    if: github.repository == 'qualcomm-linux/meta-qcom'
+    runs-on: [self-hosted, qcom-u2404, amd64-ssd]
+    strategy:
+      fail-fast: true
+      matrix:
+        machine:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-yocto.yml
```diff
@@ -172,6 +163,83 @@ jobs:
         with:
           path: ./uploads
 
+  compile:
```
lumag, line 98: Do we have any chance to extract this to a template or external file in order to reduce c&p?

ndechesne: we could use a composite action stored in the local tree. i wanted to look into that and it did not look worth it, but if you really want I can !

lumag: Yes, please. I think that makes sense.

lumag (dismissed):
With the fixups squashed, LGTM

lumag:
This needs to be rebased now.

lumag (approved):

PR #1023, opened 2025-08-19
ci: upload build artifacts to S3 bucket

quic-viskuma:
Upload the build artifacts to the specified S3 bucket. This step is
essential to enable Axiom testing, as it relies on these artifacts
being available. Ensure the upload is completed before initiating tests

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -172,6 +172,13 @@ jobs:
         with:
           path: ./uploads
 
+      - name: Upload artifacts to S3 bucket
+        uses: qualcomm-linux/upload-private-artifact-action@aws-v1
+        with:
+          path: ./uploads
+          destination: ${{ github.repository_owner }}/${{ github.event.repository.name }}/${{ github.run_id }}-${{ github.run_attempt }}/
+          s3_bucket: qcom-prd-gh-artifacts
+
   create-output:
     needs: compile
     outputs:


lumag (changes requested):
Your commit message doesn't describe the reason for a change. Describe why, not what.

.github/workflows/build-yocto.yml
```diff
@@ -172,6 +172,13 @@ jobs:
         with:
           path: ./uploads
 
+      - name: Upload artifacts to S3 bucket
+        uses: qualcomm-linux/upload-private-artifact-action@aws-v1
+        with:
+          path: ./uploads
+          destination: ${{ github.repository_owner }}/${{ github.event.repository.name }}/${{ github.run_id }}-${{ github.run_attempt }}/
+          s3_bucket: qli-prd-foundri-gh-artifacts
```
ndechesne: why do we carry 'foundri' in the bucket name? that does not really bring anything, nor make any sense.

quic-viskuma:
> Your commit message doesn't describe the reason for a change. Describe why, not what.

Done

.github/workflows/build-yocto.yml
quic-viskuma: This naming was given by the IT team. If it makes sense, we can ask IT to rename it as qli-prd-distro-gh-artifacts presuming we'll use this for yocto and debian. Let me know if you have any other suggestion, I can check with IT.

ndechesne: qli-prd-gh-artifacts is probably enough. if we want (can?) host yocto and debian artifacts together. if we need to split them part, we can add debian and yocto in the name.

quic-viskuma: 'qcom-prd-gh-artifacts' will be better if we're considering common bucket for everyone and path name we're considering is -

`qcom-prd-gh-artifacts/<org>/<repo>/<run-id>-<attempt-no>/`

For example -

```
qcom-prd-gh-artifacts/qualcomm-linux/meta-qcom/26484859-1/
qcom-prd-gh-artifacts/qualcomm-linux/kernel/88384846-1/
qcom-prd-gh-artifacts/audioreach/audioreach-graphservices/12384886-1/
```

ndechesne: sounds good to me.

ricardosalveti: +1 for 'qcom-prd-gh-artifacts'.

quic-viskuma: updated S3 bucket - qcom-prd-gh-artifacts

ndechesne:
@lumag  I think it's ready to be merged. care to review the requested changes?

lumag (approved):

PR #1025, opened 2025-08-19
ci: clean up OE persistent storage from time to time

### LINKED ISSUE #1018: Management of AWS runners persistent storage
We've now merged our CI infrastructure to use AWS runners, managed by IT (#960). For OE builds we are using a persistent storage (mounted as /efs/qli/meta-qcom/) which is used for:
- OE sstate cache folder
- OE downloads folder
- KAS mirrors

This is done to support incremental builds, as opposed to cold build from scratch.

While the OE downloads and KAS mirrors folder sizes will be somehow reasonnable, the OE sstate cache is going to grow significantly in size, and need some careful cleanup. 

The simplest way to manage sstate is , when the file system supports atime, to delete files not accessed in the last X months, e.g. last 3 months:

`find XXX -name 'sstate*' -atime +3 -delete`

If the file system does not supports atime, we need to figure out a way to delete sstate folder completely from time to time. It can be either a manual intervention from IT, ideally we need to make sure there is no build running while we do that. 

Alternatively we could append the current month (or quarter) to the folder location, and deletes older cache, to ensure that we don't delete a folder while it's being used. 

Opening this issue to discuss it with everyone, in particular @sampra2025 , @doanac , @lumag , @ricardosalveti .

ndechesne:
sstate folder can grow significantly over time, and needs to be cleaned up. The simplest thing to do is to remove it regularly. To avoid any manual intervention, we append '-year-month' to the folder name, and we delete older folders. We keep current and last month to avoid deleting the 'last' month folder while it might still be in use by current running builds (around the first day of each month).

Closes #1018.


koenkooi (dismissed):

ricardosalveti:
```
    CACHE_DIR: /efs/qli/meta-qcom/
    KAS_CONTAINER: /efs/qli/meta-qcom/kas-mirrors/kas-container
    DL_DIR: /efs/qli/meta-qcom//downloads
    SSTATE_DIR: /efs/qli/meta-qcom//sstate-cache-2025-08
```

nit, there is a double slash as CACHE_DIR already contains a final slash to the path.

koenkooi:
Do we want to belt-and-suspenders and remove year-1 as well?

ndechesne:
> Do we want to belt-and-suspenders and remove year-1 as well?

I don't understand. in Jan 26, we will get the date '2 months ago' so delete 2025-11, then 2025-12 in Feb 26'.

ricardosalveti (approved):

koenkooi:
> > Do we want to belt-and-suspenders and remove year-1 as well?
> 
> I don't understand. in Jan 26, we will get the date '2 months ago' so delete 2025-11, then 2025-12 in Feb 26'.

If it works, yes. If it silently breaks for a month, we either need manual intervention or a more ruthless reaper.

ndechesne:
> > > Do we want to belt-and-suspenders and remove year-1 as well?
> > 
> > 
> > I don't understand. in Jan 26, we will get the date '2 months ago' so delete 2025-11, then 2025-12 in Feb 26'.
> 
> If it works, yes. If it silently breaks for a month, we either need manual intervention or a more ruthless reaper.

right. we could do 
rm -rf ${CACHE_DIR}/sstate-cache-$(date -d '1 year ago' '+%Y')-*
but I dont' think it's needed right now..

lumag (approved):

PR #1027, opened 2025-08-19
Readd KGSL kernel module

lumag:
Reapply the patch adding KGSL recipe and fix build errors with Linux 6.16.


ricardosalveti:
Is there any pressing needs to have this merged now? Otherwise I would rather wait get https://github.com/qualcomm-linux/kgsl/pull/9 merged first.

ndechesne:
> Is there any pressing needs to have this merged now? Otherwise I would rather wait get [qualcomm-linux/kgsl#9](https://github.com/qualcomm-linux/kgsl/pull/9) merged first.

+1. Especially because we can't really do anything without the UM blobs, right? Do we have them available in any form which would allow some testing?

lumag:
> > Is there any pressing needs to have this merged now? Otherwise I would rather wait get [qualcomm-linux/kgsl#9](https://github.com/qualcomm-linux/kgsl/pull/9) merged first.
> 
> +1. Especially because we can't really do anything without the UM blobs, right? Do we have them available in any form which would allow some testing?

Sure, let's get some feedback from the GPU team.

For the UMD I think @vkraleti or @sbanerjee-quic can comment better. There is a [recipe](https://github.com/qualcomm-linux/meta-qcom-hwe/tree/scarthgap/recipes-graphics/adreno) to bundle Adreno libs.

lumag:
I dropped all warning-related fixes, leaving only the actual build fixes. We need these two patches in order to cover all our build configurations (e.g. non-qcom 6.12).

recipes-graphics/kgsl-dlkm/kgsl-dlkm_git.bb
```diff
@@ -0,0 +1,28 @@
+inherit module
+
+DESCRIPTION = "Qualcomm KGSL driver for managing Adreno GPU"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://adreno.c;beginline=1;endline=1;md5=fcab174c20ea2e2bc0be64b493708266"
+
+SRCREV = "56d7a36e6b95d3e311fe22f93be98d9d5943036d"
```
koenkooi, line 8: This is missing an explicit PV for the version number.

lumag: done

Issue #1029, opened 2025-08-21
meta-qcom CI did not notice a yocto-check-layer failure


ndechesne:
meta-qcom CI runs yocto-check-layers, and the Yocto Project CI also runs it on its Autobuilder. The Yocto CI reported a failure in meta-qcom yocto-check-layer (see https://autobuilder.yoctoproject.org/valkyrie/#/builders/85/builds/606) which our own CI did not detect. Let's investigate why.

quaresmajose:
Cloning all repos separately it doesn't fail!
```bash
git clone https://github.com/openembedded/openembedded-core
cd openembedded-core
git checkout e861561b
git clone https://github.com/openembedded/bitbake
git -C bitbake checkout 7ad404fd
git clone https://git.yoctoproject.org/meta-yocto
git -C meta-yocto checkout 85967da3
git clone https://github.com/qualcomm-linux/meta-qcom.git
git -C meta-qcom checkout 5e444cab
source oe-init-build-env
bitbake-layers add-layer ../meta-yocto/meta-poky
bitbake-layers add-layer ../meta-qcom
yocto-check-layer-wrapper ../meta-qcom --no-auto-dependency
```

quaresmajose:
> > I have replicated localy the failure on the autobuilder using poky and meta-qcom
> > git clone https://git.yoctoproject.org/poky
> > cd poky
> > git checkout a2d9103b
> > git clone https://github.com/qualcomm-linux/meta-qcom.git
> > git -C meta-qcom checkout [5e444ca](https://github.com/qualcomm-linux/meta-qcom/commit/5e444cabe7539503bdac431180be0c463f3af03f)
> > source oe-init-build-env
> > bitbake-layers add-layer ../meta-qcom
> > yocto-check-layer-wrapper ../meta-qcom --no-auto-dependency
> 
> Doing the poky test but changing the DISTRO to be the same as the one of our test also fails as well
> 
> ```
> echo 'DISTRO = "nodistro"' >> conf/local.conf
> ```
> 
> It looks like a bug on the yocto-check-layer and not in meta-qcom.

Removing also the meta-yocto-bsp layer but the behaviour is the same.

ndechesne:
ok, i nailed it down.. 

When we run yocto-check-layer with oe-core + bitbake, we end up setting DISTRO=nodistro. When DISTRO=nodistro, then neither opengl nor vulkan is set, so mesa recipe is skipped. 

```
[...]

PR #1034, opened 2025-08-22
ci: runs the yocto-check-layer with nodistro and poky

quaresmajose:
By default our yocto-check-layer uses the oe-core nodisto.
We can specify the kas yml config to be used on the yocto-check-layer so we will add another job to run the yocto-check-layer using the Yocto poky distro.

The yocto-check-layer jobs don't take any adavantage of the bitbake sstate-cache and it will take some time when they runs sequential. So in parallel we reduce to half of what we would need if they were sequential.

Fixes https://github.com/qualcomm-linux/meta-qcom/issues/1029

Requires https://github.com/qualcomm-linux/meta-qcom/pull/1060

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -11,7 +11,6 @@ env:
   CACHE_DIR: /efs/qli/meta-qcom
   KAS_REPO_REF_DIR: /efs/qli/meta-qcom/kas-mirrors
   KAS_CONTAINER: /efs/qli/meta-qcom/kas-mirrors/kas-container
-  KAS_IMAGE_VERSION: latest
 
 jobs:
   kas-setup:
@@ -20,7 +19,8 @@ jobs:
     steps:
       - name: Update kas-container
         run: |
-          wget -qO ${KAS_CONTAINER} https://raw.githubusercontent.com/siemens/kas/refs/heads/master/kas-container
+          LATEST=$(git ls-remote --tags --refs --sort="v:refname" https://github.com/siemens/kas | tail -n1 | sed 's/.*\///')
+          wget -qO ${KAS_CONTAINER} https://raw.githubusercontent.com/siemens/kas/refs/tags/$LATEST/kas-container
           chmod +x ${KAS_CONTAINER}
 
       - name: Update kas mirrors


ci/yocto-check-layer.sh
```diff
@@ -27,6 +27,12 @@ BUILDDIR="$(mktemp -p $WORK_DIR -d -t build-yocto-check-layer-XXXX)"
 source $WORK_DIR/oe-core/oe-init-build-env $BUILDDIR
 git -c advice.detachedHead=false clone --quiet --shared $REPO_DIR meta-qcom
 
+# It is not possible to change the default configuration used by the yocto-check-layer-wrapper,
```
ndechesne: this is not exactly the right reason why we do that. the real question is what is the DISTRO that should be used when running yocto-check-layer, since it is obivous that it has an impact. Nowhere in the yocto documentation (see https://docs.yoctoproject.org/dev-manual/layers.html#making-sure-your-layer-is-compatible-with-yocto-project) this information is specified.

quaresmajose: Rigth, it is true that it is not written anywhere but to be compliant with Yocto project we need to test with same distro tested upstream. To be compatible with the OE-core we should also test with nodistro variant.
As you confirmed depending on the distro feature some recipes are not tested and this requirement obliges us to use the same distro feature of the distribution we want to be compliant with.

ndechesne: I am checking with RP to see if it can be clarified. If we need to test with poky, then we might want to clone the poky tree directly then. there are other differences, such as meta-yocto-bsp being included.. so it will be hard to keep track of that unless we use the same repo.

.github/workflows/build-yocto.yml
```diff
@@ -53,14 +53,42 @@ jobs:
...
+    runs-on: [self-hosted, qcom-u2404, amd64-ssd]
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/download-artifact@v4
+        with:
+          name: kas-lock
+          path: ci/
+
+      - name: Run yocto-check-layer poky
```
lumag: Can we run them from the old workflow rather than creating new jobs?

quaresmajose: We can but as this yocto-check-layer jobs don't take any adavantage of the bitbake sstate-cache it will take some time when they runs sequential. So in parallel we reduce to half of what we would need if they were sequential.

I can put the three jobs sequential again, I don't have a strong opinion on that.

lumag: Don't we spend time on winding the container / worker and checking out all the layers?

quaresmajose: Yeah we spend some time creating the new workers and fetching the container  but it's not ~10 minutes, which is the fastest case for the nodistro.

- **14m 19s** Run yocto-check-layer poky
- **11m 53s** Run yocto-check-layer nodistro

Without this parallelism, our lowest time would jump to be approximately 26 minutes.

ci/base.yml
```diff
@@ -9,6 +9,10 @@ defaults:
   repos:
     branch: master
 
+# export through BB_ENV_EXTRAWHITE
+env:
+    DISTRO:
```
ndechesne: not a fan of this change in base.yml.. 

can you not get around what you want to do with adding ci/poky.yml and set DISTRO=poky there. and use it when you want to default to poky?

0001-yocto-check-layer-show-the-distro-variant-used.patch
```diff
@@ -0,0 +1,27 @@
...
+
+diff --git a/scripts/yocto-check-layer b/scripts/yocto-check-layer
+index 67cc71950f..5751e139cc 100755
+--- a/scripts/yocto-check-layer
++++ b/scripts/yocto-check-layer
+@@ -35,6 +35,7 @@ logger = scriptutils.logger_create(PROGNAME, stream=sys.stdout)
+ def test_layer(td, layer, test_software_layer_signatures):
+     from checklayer.context import CheckLayerTestContext
+     logger.info("Starting to analyze: %s" % layer['name'])
++    logger.info('Distro variant: %s' % td['bbvars']["DISTRO"])
```
ndechesne: I like this patch. However I don't think it should be Distro variant, but just Distro;

ci/base.yml
quaresmajose: I am working on it and will create a new pull request to add ci/poky.yml support to build the poky distro.

0001-yocto-check-layer-show-the-distro-variant-used.patch
quaresmajose: I fixed it

lumag: It's not fixed yet.

.github/workflows/build-yocto.yml
lumag: Ack.

0001-yocto-check-layer-show-the-distro-variant-used.patch
quaresmajose: it is fixed now

ci/base.yml
quaresmajose: PR https://github.com/qualcomm-linux/meta-qcom/pull/1060 ci/poky: add kas config yml

ci/poky.yml
```diff
@@ -0,0 +1,19 @@
...
+repos:
+  oe-core:
+    layers:
+      meta: disabled
+
+  meta-yocto:
+    layers:
+      meta-poky: disabled
+
+  poky:
```
ndechesne: I don't think this is needed. We want to switch from nodistro to poky (to qcom) distro easily. I don't think we should clone poky. if a tree is missing we can add it directly instead.

ci/yocto-check-layer.sh
```diff
@@ -24,7 +24,12 @@ _is_dir "$WORK_DIR"
 # Creates a temporary build directory to run the yocto-check-layer
 # script to avoid a contaminated environment.
 BUILDDIR="$(mktemp -p $WORK_DIR -d -t build-yocto-check-layer-XXXX)"
-source $WORK_DIR/oe-core/oe-init-build-env $BUILDDIR
+if bitbake-layers show-layers | grep ^core | grep -q poky; then
```
ndechesne: this commit and the previous one are not needed if you don't clone poky

ci/kas-container-shell-helper.sh
```diff
@@ -14,7 +15,10 @@ fi
...
 SCRIPT=${SCRIPT#$TOPDIR/}
 
+# inject TOPDIR
+KAS_YML="$TOPDIR/${KAS_YML//:/:$TOPDIR\/}"
+
 # on ci the kas-container is not on the default path
 KAS_CONTAINER=${KAS_CONTAINER:-$(which kas-container)}
 
-exec $KAS_CONTAINER shell $TOPDIR/ci/base.yml --command "/repo/$SCRIPT /repo /work"
+exec $KAS_CONTAINER shell $KAS_YML --command "/repo/$SCRIPT /repo /work"
```
ndechesne, line 24: I don't understand this commit, and the commit message does not explain why you are making this change. 

I think we should run 
exec $KAS_CONTAINER shell $TOPDIR/ci/base.yml --command "/repo/$SCRIPT /repo /work"
or 
exec $KAS_CONTAINER shell $TOPDIR/ci/base.yml:$TOPDIR/ci/poky.yml --command "/repo/$SCRIPT /repo /work"
and set DISTRO=poky in poky.yml

ci/poky.yml
quaresmajose: The reson for this was because there are no easy way to change the config template. This makes it impossible to change the distribution from nodistro to yocto when we don't use poky repository.

quaresmajose: Either way the poky combo layer will be frozen as we can see on the status of this week. I will have to rethink this PR again without using the poky repo.

Yocto Project Status 16 September 2025 (WW37)
```
# We aim to merge bitbake-setup before building 5.3 M3, work is currently being done to migrate the autobuilder to use it.

# This change will mean “poky” as a combo-layer repository will no longer be needed or updated.
```
https://lists.yoctoproject.org/g/yocto-status/message/81

ci/yocto-check-layer.sh
quaresmajose: yes it is true

ci/kas-container-shell-helper.sh
quaresmajose: We can not set the disto for the yocto-check-layer. What I had before was a rather ugly hack to pass the DISTRO through an environment variable using the BB_ENV_EXTRAWHITE.
```
# export through BB_ENV_EXTRAWHITE
env:
    DISTRO:
```

quaresmajose: I will have to send something upstream to use in yocto-check-layer the distro that is defined in main configuration and not what comes from the template. This way we can set he distro on the kas yml file and this will be the one used in the yocto-check-layer.

ndechesne: If you do:

```
$ cat ci/poky.yml
# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json

header:
  version: 14

distro: poky

```

then: 

kas dump  ci/qcom-armv8a.yml
vs 
kas dump  ci/qcom-armv8a.yml:ci/poky.yml

will set DISTRO accordingly

quaresmajose: yes it set the distro corretly but our ci/yocto-check-layer.sh will creat a new clean bitbake environment where the distro is defined in the upstream configuration template.

This was the reason of the bug, the bitbake template cofiguration doesn't set the distro and uses the default one, which is `nodistro`, while the poky template cofiguration sets the distro to be `DISTRO ?= "poky"`.
We can see the template configuration diff of poky and oe-core+bitbake in https://github.com/qualcomm-linux/meta-qcom/issues/1029#issuecomment-3211432410

ndechesne: oh. that's interesting, and you're right. well it's a problem poky needs to fix for itself too, once they no longer use the mono repo!

ndechesne: have you tried setting TEMPLATECONF env variable in poky.yml?

quaresmajose: No, I can look on it.

quaresmajose: I think there are interesting changes coming in this area with the release of version YP 5.3 in the coming weeks, it is better to wait and see if this will be done. On the autobuilder this change of DISTRO will also be necessary.

PR #1038, opened 2025-08-26
Add linux-yocto-6.16 specific qcom.scc

vkraleti:
The recipe for linux-yocto 6.16 has landed in the Poky upstream, but the meta-qcom layer is not yet fully ready to consume it. Currently, the linux-yocto_%.bbappend in meta-qcom attempts to apply to all versions of linux-yocto, which causes build failures.

To mitigate this issue for now, a 6.16-specific qcom.scc file is added. Remaining patches can be added over time as needed.


lumag:
Well... No. Closing in favour of #1039

PR #953, opened 2025-07-03
libgbm: Add custom MSM backend support for Adreno GPU

quic-devanshi:
introduce a libgbm backend, which is required for buffer allocation and management for
Qualcomm Adreno GPU.

PR #1041, opened 2025-08-28
recipes-graphics: Add recipe for qcom-adreno

ssadhula:
Introduce a new recipe for qcom-adreno that fetches Qualcomm Adreno User
Mode libraries from Qartifactory and builds qcom-adreno package to enable
Adreno GPU support on top of KGSL. These libraries include a set of
user-mode drivers that support OpenGLES, Vulkan, and OpenCL APIs.

GPU component in the DRM driver is disabled once skip_gpu module parameter
is passed. However, support for this parameter in the DRM driver is
currently under development and not yet available.

### DIFF
--- a/recipes-bsp/packagegroups/packagegroup-qcom-graphics.bb
+++ b/recipes-bsp/packagegroups/packagegroup-qcom-graphics.bb
@@ -1,0 +1,12 @@
+SUMMARY = "QCOM Graphics package group"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+inherit packagegroup
+
+RDEPENDS:${PN} = " \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'glvnd', 'qcom-adreno-egl', '', d)} \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'vulkan', 'qcom-adreno-vulkan', '', d)} \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'opencl', 'qcom-adreno-cl', '', d)} \
+"
+
--- a/recipes-bsp/packagegroups/packagegroup-qcom-ci.bb
+++ b/recipes-bsp/packagegroups/packagegroup-qcom-ci.bb
@@ -21,4 +21,5 @@ RDEPENDS:${PN}:append:aarch64 = " \
     iris-video-dlkm \
     kgsl-dlkm \
     qcom-fastcv-binaries \
+    packagegroup-qcom-graphics \
 "


recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+LICENSE_PATH     +=  "${UNPACKDIR}"
+
+DESCRIPTION = "Qualcomm Adreno Graphics User Mode libraries"
+
+DEPENDS = "wayland glib-2.0 libdmabufheap libglvnd libdrm libxcb libx11 xcb-util-image virtual/libgbm msm"
+
+SRC_URI = "${ARTIFACTORY_URL}/${BPN}_${PV}_armv8-2a.tar.gz"
+SRC_URI[sha256sum] = "7d59e09799efaf7e80efae270341339042717d715ec78af0d6440925463159d6"
+
+S = "${UNPACKDIR}"
```
lumag: There should be no need for S assignment

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+
+LICENSE          = "NO.LOGIN.BINARY.LICENSE.QTI.pdf"
+LIC_FILES_CHKSUM = "file://${UNPACKDIR}/${LICENSE};md5=4ceffe94cb40cdce6d2f4fb93cc063d1"
+LICENSE_PATH     +=  "${UNPACKDIR}"
+
+DESCRIPTION = "Qualcomm Adreno Graphics User Mode libraries"
+
+DEPENDS = "wayland glib-2.0 libdmabufheap libglvnd libdrm libxcb libx11 xcb-util-image virtual/libgbm msm"
+
+SRC_URI = "${ARTIFACTORY_URL}/${BPN}_${PV}_armv8-2a.tar.gz"
```
lumag: Please inline the URL here.
Also, not all our targets are ARMv8.2-A. Some targets (APQ8016, QRB2210) are based on ARMv8-A only. Please provide support for both.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+
+ARTIFACTORY_URL = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/gfx-adreno.le.0.0/250828.1/prebuilt_yocto"
+
+LICENSE          = "NO.LOGIN.BINARY.LICENSE.QTI.pdf"
+LIC_FILES_CHKSUM = "file://${UNPACKDIR}/${LICENSE};md5=4ceffe94cb40cdce6d2f4fb93cc063d1"
+LICENSE_PATH     +=  "${UNPACKDIR}"
+
+DESCRIPTION = "Qualcomm Adreno Graphics User Mode libraries"
+
+DEPENDS = "wayland glib-2.0 libdmabufheap libglvnd libdrm libxcb libx11 xcb-util-image virtual/libgbm msm"
```
lumag: Are those DEPENDS or RDEPENDS?

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
+inherit pkgconfig
+
+ARTIFACTORY_URL = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/gfx-adreno.le.0.0/250828.1/prebuilt_yocto"
+
+LICENSE          = "NO.LOGIN.BINARY.LICENSE.QTI.pdf"
+LIC_FILES_CHKSUM = "file://${UNPACKDIR}/${LICENSE};md5=4ceffe94cb40cdce6d2f4fb93cc063d1"
+LICENSE_PATH     +=  "${UNPACKDIR}"
```
lumag: Why do you need LICENSE_PATH assignment?

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
+inherit pkgconfig
+
+ARTIFACTORY_URL = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/gfx-adreno.le.0.0/250828.1/prebuilt_yocto"
+
+LICENSE          = "NO.LOGIN.BINARY.LICENSE.QTI.pdf"
```
lumag: LICENSE = "CLOSED"

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+SRC_URI[sha256sum] = "7d59e09799efaf7e80efae270341339042717d715ec78af0d6440925463159d6"
+
+S = "${UNPACKDIR}"
+
+do_install () {
+    install -d ${D}${includedir}/adreno
+    cp -rf ${S}${includedir}/adreno/* ${D}${includedir}/adreno/
+
+    install -d ${D}/${libdir}
+    cp -r ${S}/${libdir}/* ${D}/${libdir}/
```
lumag: Please use fixed paths on the source side. If we change the libdir assignment, the path inside the archive will not change.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+
+    install -d ${D}/lib/firmware
+    cp -rf ${S}/lib/firmware/* ${D}/lib/firmware/
```
lumag: No, The firmware should be coming from linux-firmware or from BSP packages instead of the adreno archive.
And anyway, it should be `${nonarch_base_libdir}/firmware/`

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+
+    install -d ${D}/usr/share/glvnd/egl_vendor.d
+    cp ${S}/usr/share/glvnd/egl_vendor.d/EGL_adreno.json ${D}/usr/share/glvnd/egl_vendor.d/
```
lumag: `${datadir}`

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+
+    install -d ${D}/usr/share/vulkan/icd.d
+    cp ${S}/usr/share/vulkan/icd.d/adrenovk.json ${D}/usr/share/vulkan/icd.d/
```
lumag: `${datadir}`

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+
+    install -d ${D}/etc/OpenCL/vendors
+    cp ${S}/etc/OpenCL/vendors/adrenocl.icd ${D}/etc/OpenCL/vendors/
```
lumag: `${sysconfdir}`

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+
+    install -d ${D}/${sysconfdir}/profile.d
+    cp ${S}/${sysconfdir}/profile.d/qcom-adreno_env.sh ${D}/${sysconfdir}/profile.d/qcom-adreno_env.sh
+
+    install -d ${D}/${sysconfdir}/modprobe.d
+    cp ${S}/${sysconfdir}/modprobe.d/qcom-adreno.conf ${D}/${sysconfdir}/modprobe.d/qcom-adreno.conf
+}
+
+INSANE_SKIP:${PN} = "already-stripped arch usrmerge"
```
lumag: At least the comment is necessary. Neither `arch` nor `usrmerge` should be a part of this.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+               /lib/firmware/* \
+               /usr/share/glvnd/egl_vendor.d/EGL_adreno.json \
+               /usr/share/vulkan/icd.d/adrenovk.json \
+               /etc/OpenCL/vendors/adrenocl.icd \
+               ${sysconfdir}/profile.d/qcom-adreno_env.sh \
+               ${sysconfdir}/modprobe.d/qcom-adreno.conf "
+
+FILES:${PN}-dev = "${includedir}/adreno \
+                   ${libdir}/pkgconfig \
+                   ${libdir}/lib*.so "
```
lumag: Why do you need to override the defaults here? Please only add necessary iteams.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
+inherit pkgconfig
+
+ARTIFACTORY_URL = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/gfx-adreno.le.0.0/250828.1/prebuilt_yocto"
+
+LICENSE          = "NO.LOGIN.BINARY.LICENSE.QTI.pdf"
+LIC_FILES_CHKSUM = "file://${UNPACKDIR}/${LICENSE};md5=4ceffe94cb40cdce6d2f4fb93cc063d1"
+LICENSE_PATH     +=  "${UNPACKDIR}"
+
+DESCRIPTION = "Qualcomm Adreno Graphics User Mode libraries"
```
lumag: Missing RPROVIDES:${PN}.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+DEPENDS = "wayland glib-2.0 libdmabufheap libglvnd libdrm libxcb libx11 xcb-util-image virtual/libgbm msm"
+
+SRC_URI = "${ARTIFACTORY_URL}/${BPN}_${PV}_armv8-2a.tar.gz"
+SRC_URI[sha256sum] = "7d59e09799efaf7e80efae270341339042717d715ec78af0d6440925463159d6"
+
+S = "${UNPACKDIR}"
+
+do_install () {
+    install -d ${D}${includedir}/adreno
+    cp -rf ${S}${includedir}/adreno/* ${D}${includedir}/adreno/
```
lumag: Include files should be coming from `opencl-headers-dev`, `vulkan-headers-dev` and `libglvnd-dev`. Don't install generic headers here.

Don't install extra `/usr/include/adreno` copy, it's non-standard and should not be used.

Files to be installed:
- CL/cl_ext_qcom.h
- CL/cl_qcom_ml_ops.h
- EGL/eglextQCOM.h
- GLES/glextQCOM.h
- GLES2/gl2extQCOM.h
- GLES3/gl3esxQCOM.h
- GLES3/gl3extQCOM.h

Also please provide an explanation, while the mentioned files are not a part of the standard GLES / CL extensions registry?

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
lumag: Also please make sure that it's possible to build and install Adreno packages if DISRO has disabled X11 or Wayland features.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+    cp ${S}/${sysconfdir}/modprobe.d/qcom-adreno.conf ${D}/${sysconfdir}/modprobe.d/qcom-adreno.conf
+}
+
+INSANE_SKIP:${PN} = "already-stripped arch usrmerge"
+
+FILES:${PN} = "${libdir}/lib*.so.* \
+               /lib/firmware/* \
+               /usr/share/glvnd/egl_vendor.d/EGL_adreno.json \
+               /usr/share/vulkan/icd.d/adrenovk.json \
+               /etc/OpenCL/vendors/adrenocl.icd \
```
lumag: Split this package into CL. GL ES, GL ES2 and Vulkan packages

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
lumag:
```
ERROR: Required build target 'qcom-adreno' has no buildable providers.
Missing or unbuildable dependency chain was: ['qcom-adreno', 'msm']
```
This wasn't even tested.

ssadhula: msm is provided by https://github.com/qualcomm-linux/meta-qcom/pull/953

ssadhula: > Are those DEPENDS or RDEPENDS?

They are RDEPENDS

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: Okay.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: Okay.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: Okay.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: Okay.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: Okay.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: Okay.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: These are private QCOM specific extensions. Okay, we will include only these header files.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: GL ES cannot be a separate package as it depends on GL ES 2. We will create separate packages for CL, Vulkan, GL ES/GL ES2 and one common package which will be used by all other packages.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: > Also please make sure that it's possible to build and install Adreno packages if DISRO has disabled X11 or Wayland features.

Currently, libvulkan_adreno.so is a precompiled single binary with both X11 and Wayland supported. So, we have RDEPENDS on both X11 and Wayland.

lumag: I will check, how this is handled by other layers

lumag: Well. First, you have declared it in DEPENDS, so it will be pulled into the build environment for no reason. Second, It's not declared in RDEPENDS, so it will not be pulled into the image. Third, it's a backend, so it should not be depended upon at all. Pull it into the image in some other way, e.g. via packagegroups.

lumag: Last, but not least, this PR is not marked as a draft, which means that it's ready for review and it's intended to be tested and merged. However, one can't actually do that, building it breaks.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
lumag: I think, there are Qualcomm extensions in the GL registry. Please work with Kronos to get those extensions into the registry and into the corresponding extension headers. Without a public definition one should not be using those extensions.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
lumag: GL ES package can depend on ES2. But it still must be a separate package.

lumag:
For the reference, I made a graph of dependencies between libraries in the tarball. Dotted lines means that there is no DT_NEEDED link, but the name is mentioned in the library file (which might mean dynamic loading). If any of the branches is incorrect, feel free to correct me.
adreno
[adreno.dot.txt](https://github.com/user-attachments/files/22084838/adreno.dot.txt)

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,58 @@
...
+               ${sysconfdir}/profile.d/qcom-adreno_env.sh \
+               ${sysconfdir}/modprobe.d/qcom-adreno.conf "
+
+FILES:${PN}-dev = "${includedir}/adreno \
+                   ${libdir}/pkgconfig \
```
lumag: I took a glance. No, you should not be shipping pkgconfig files. Instead the software should use .pc files provided by `libglvnd-dev`.

lumag:
#953 is (hopefully) going to be merged soon. Could you please rebase your changes on top of it?

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,65 @@
+inherit pkgconfig
+
+LICENSE          = "CLOSED"
```
lumag: No extra whitespaces.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,65 @@
...
+DESCRIPTION = "Qualcomm Adreno Graphics User Mode libraries"
+
+SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/gfx-adreno.le.0.0/250908/prebuilt_yocto/${BPN}_${PV}_armv8-2a.tar.gz;subdir=${BPN}-${PV}"
+
+SRC_URI[sha256sum] = "02974d294597994a255f7aca6d6d61422fb44f11bd4cf6064e183c3888ea9def"
+
+PACKAGES += " ${PN}-fw ${PN}-common ${PN}-gles ${PN}-gles2 ${PN}-egl ${PN}-vulkan ${PN}-cl"
+
+RPROVIDES:${PN}-common   = "${PN}-common"
+RPROVIDES:${PN}-fw       = "${PN}-fw"
```
lumag: No. You must use firmware from linux-firmware-qcom-*.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,65 @@
...
+
+    install -d ${D}${sysconfdir}/profile.d
+    cp ${S}/etc/profile.d/qcom-adreno_env.sh ${D}/${sysconfdir}/profile.d/qcom-adreno_env.sh
+
+    install -d ${D}${sysconfdir}/modprobe.d
+    cp ${S}/etc/modprobe.d/qcom-adreno.conf ${D}/${sysconfdir}/modprobe.d/qcom-adreno.conf
+}
+
+FILES:${PN}-fw       = "${nonarch_base_libdir}/firmware/*"
+FILES:${PN}-common   = "${libdir}/libllvm-*.so.* ${libdir}/libgsl.so.1 ${libdir}/libadreno_utils.so.1 ${libdir}/libq3dtools_*.so.* ${sysconfdir}/modprobe.d/qcom-adreno.conf ${sysconfdir}/profile.d/qcom-adreno_env.sh"
```
lumag: One per line, please.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,65 @@
...
+RPROVIDES:${PN}-common   = "${PN}-common"
+RPROVIDES:${PN}-fw       = "${PN}-fw"
+RPROVIDES:${PN}-gles     = "${PN}-gles"
+RPROVIDES:${PN}-gles2    = "${PN}-gles2"
+RPROVIDES:${PN}-egl      = "${PN}-egl"
+RPROVIDES:${PN}-vulkan   = "${PN}-vulkan"
+RPROVIDES:${PN}-cl       = "${PN}-cl"
+
+RDEPENDS:${PN}-common += " zlib glib-2.0 libdmabufheap msm-gbm-backend ${PN}-fw kgsl-dlkm"
+RDEPENDS:${PN}-egl    += " glib-2.0 libgbm wayland libdmabufheap zlib libxcb xcb-util-image libx11-xcb ${PN}-common"
```
lumag: No, let OE find all RDEPENDS on its own. Desribe only non-shlibs dependencies here.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,65 @@
...
+    install -d ${D}${sysconfdir}/profile.d
+    cp ${S}/etc/profile.d/qcom-adreno_env.sh ${D}/${sysconfdir}/profile.d/qcom-adreno_env.sh
+
+    install -d ${D}${sysconfdir}/modprobe.d
+    cp ${S}/etc/modprobe.d/qcom-adreno.conf ${D}/${sysconfdir}/modprobe.d/qcom-adreno.conf
+}
+
+FILES:${PN}-fw       = "${nonarch_base_libdir}/firmware/*"
+FILES:${PN}-common   = "${libdir}/libllvm-*.so.* ${libdir}/libgsl.so.1 ${libdir}/libadreno_utils.so.1 ${libdir}/libq3dtools_*.so.* ${sysconfdir}/modprobe.d/qcom-adreno.conf ${sysconfdir}/profile.d/qcom-adreno_env.sh"
+FILES:${PN}-egl      = "${libdir}/libEGL_*.so.1 ${libdir}/libeglSubDriver*.so.* ${datadir}/glvnd/egl_vendor.d/EGL_adreno.json"
```
lumag: I'd really ask you to make it possible to build packages if there is no X11 or no Wayland enabled in DISTRO_FEATURES.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,65 @@
...
+
+LICENSE          = "CLOSED"
+
+DESCRIPTION = "Qualcomm Adreno Graphics User Mode libraries"
+
+SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/gfx-adreno.le.0.0/250908/prebuilt_yocto/${BPN}_${PV}_armv8-2a.tar.gz;subdir=${BPN}-${PV}"
+
+SRC_URI[sha256sum] = "02974d294597994a255f7aca6d6d61422fb44f11bd4cf6064e183c3888ea9def"
+
+PACKAGES += " ${PN}-fw ${PN}-common ${PN}-gles ${PN}-gles2 ${PN}-egl ${PN}-vulkan ${PN}-cl"
```
lumag: PACKAGE_BEFORE_PN

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,65 @@
...
+FILES:${PN}-vulkan   = "${libdir}/libvulkan_*.so.* ${datadir}/vulkan/icd.d/adrenovk.json"
+FILES:${PN}-cl       = "${libdir}/libOpenCL_*.so.* ${libdir}/libCB.so.1 ${sysconfdir}/OpenCL/vendors/adrenocl.icd"
+FILES:${PN}          = ""
+
+# Prebuilt libraries are already stripped
+INSANE_SKIP:${PN} = "already-stripped"
+
+# Firmware files architecture is not 64-bit
+INSANE_SKIP:${PN}-fw = "arch"
+
```
lumag: Please drop extra empty line.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,65 @@
+inherit pkgconfig
```
lumag: What for? You have prebuilt binaries, so we don't need pkg-config support.

lumag (changes requested):

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
pvgtejas-quic: The sqe and gmu firmware from linux-firmware will be available at /lib/firmware/qcom, but kgsl driver currently only checks /lib/firmware. The changes in kgsl to use /lib/firmware/qcom would be available soon. Can the firmware package be dropped once the kgsl support is available and Adreno is validated with linux-firmware-qcom-*?

lumag: I don't think so. Worst case you can create compat symlinks pointing to the existing files, but I definitely don't think we should be getting another copy of Adreno firmware.

lumag:
This wasn't tested.

```
ERROR: packagegroup-qcom-graphics-1.0-r0 do_package_write_ipk: An allarch packagegroup shouldn't depend on packages which are dynamically renamed (qcom-adreno-gles to libglesv1-cm-adreno1)
ERROR: packagegroup-qcom-graphics-1.0-r0 do_package_write_ipk: An allarch packagegroup shouldn't depend on packages which are dynamically renamed (qcom-adreno-vulkan to libvulkan-adreno1)
```

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: If we remove glib-2.0 from RDEPENDS, I am seeing below error, looks like OE is not able to find RDEPENDS on its own,

ERROR: qcom-adreno-1.838.1-r0 do_package_qa: QA Issue: /usr/lib/libvulkan_adreno.so.1 contained in package qcom-adreno-vulkan requires libglib-2.0.so.0()(64bit), but no providers found in RDEPENDS:qcom-adreno-vulkan? [file-rdeps]

lumag: I think (and I'd let @koenkooi correct me), that these packages should be listed in DEPENDS (as you did initially), but (!) please prepend that with a comment, saying that they are listed in order to identify RDEPENDS. Keep those separate from the actual DEPENDS, if there are any.

koenkooi: Correct, if they are in DEPENDS, the OE shlib code should add the rdepends automagically. If it isn't in DEPENDS, you will likely end up in a situation where the library+symbols are unknown and you get the message you showed.

Needing to add a library to RDEPENDS is a red flag and in most cases needs a different solution.

lumag: So, I guess, add DEPENDS line and add a comment that these are for automatically selecting RDEPENDS

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,71 @@
+LICENSE = "CLOSED"
+
+DESCRIPTION = "Qualcomm Adreno Graphics User Mode libraries"
+
+SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/gfx-adreno.le.0.0/250908/prebuilt_yocto/${BPN}_${PV}_armv8-2a.tar.gz;subdir=${BPN}-${PV}"
```
lumag, line 8: Why do you need subdir?

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,71 @@
...
+# These are listed here in order to identify RDEPENDS
+DEPENDS += " glib-2.0 libdmabufheap libglvnd libdrm virtual/libgbm msm-gbm-backend \
+             ${@bb.utils.contains('DISTRO_FEATURES', 'wayland', 'wayland', '', d)} \
+             ${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'libxcb libx11 xcb-util-image', '', d)}"
+
+PACKAGE_BEFORE_PN += " ${PN}-common ${PN}-gles ${PN}-gles2 ${PN}-egl ${PN}-vulkan ${PN}-cl"
+
+RPROVIDES:${PN}-common = "${PN}-common"
+RPROVIDES:${PN}-egl = "${PN}-egl"
+RPROVIDES:${PN}-cl = "${PN}-cl"
```
lumag: No need to

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,71 @@
...
+
+# Use dynamically renamed package names
+RPROVIDES:${PN}-gles = "libglesv1-cm-adreno1"
+RPROVIDES:${PN}-gles2 = "libglesv2-adreno2"
+RPROVIDES:${PN}-vulkan = "libvulkan-adreno1"
```
lumag: No, thank you. Don't try to overplay Yocto where it does its job correctly. It tells you that the packagegroup should be machine-specific when it really should be.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,69 @@
...
+
+RDEPENDS:${PN}-common += " kgsl-dlkm"
+RDEPENDS:${PN}-egl += " ${PN}-common"
+RDEPENDS:${PN}-gles2 += " ${PN}-common ${PN}-egl"
+RDEPENDS:${PN}-gles += " ${PN}-gles2"
+RDEPENDS:${PN}-vulkan += " vulkan-loader ${PN}-common"
```
lumag: No need to RDEPEND on vulkan-loader. It will be picked up by the applications which depend on libvulkan

recipes-bsp/packagegroups/packagegroup-qcom-graphics.bb
```diff
@@ -0,0 +1,14 @@
...
+
+inherit packagegroup
+
+PROVIDES = "${PACKAGES}"
+
+PACKAGES = "${PN}"
+
+RDEPENDS:${PN} = " \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'libglesv1-cm-adreno1', '', d)} \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'vulkan', 'libvulkan-adreno1', '', d)} \
```
lumag: Please use non-renamed package names here.

recipes-bsp/packagegroups/packagegroup-qcom-graphics.bb
```diff
@@ -0,0 +1,14 @@
+SUMMARY = "QCOM Graphics package group"
+
+inherit packagegroup
+
+PROVIDES = "${PACKAGES}"
```
lumag: Why?

recipes-bsp/packagegroups/packagegroup-qcom-graphics.bb
```diff
@@ -0,0 +1,14 @@
...
+
+PACKAGES = "${PN}"
```
lumag: Is this necessary?

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,69 @@
...
+SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/gfx-adreno.le.0.0/250908/prebuilt_yocto/${BPN}_${PV}_armv8-2a.tar.gz;subdir=${BPN}-${PV}"
+
+SRC_URI[sha256sum] = "02974d294597994a255f7aca6d6d61422fb44f11bd4cf6064e183c3888ea9def"
+
+# These are listed here in order to identify RDEPENDS
+DEPENDS += " glib-2.0 libdmabufheap libglvnd libdrm virtual/libgbm msm-gbm-backend \
+             ${@bb.utils.contains('DISTRO_FEATURES', 'wayland', 'wayland', '', d)} \
+             ${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'libxcb libx11 xcb-util-image', '', d)}"
+
+PACKAGE_BEFORE_PN += " ${PN}-common ${PN}-gles ${PN}-gles2 ${PN}-egl ${PN}-vulkan ${PN}-cl"
```
lumag:
```
RPROVIDES:${PN}-egl += "virtual-egl-icd"
```

I am submitting a patch adding two more virtuals, so you might want to add these two:

```
RPROVIDES:${PN}-cl += "virtual-opencl-icd"
RPROVIDES:${PN}-vulkan += "virtual-vulkan-icd"

File: recipes-graphics/adreno/qcom-adreno_1.838.1.bb [hunk 33]
```diff
@@ -0,0 +1,69 @@
...
+                      ${libdir}/libadreno_utils.so.1 \
+                      ${libdir}/libq3dtools_*.so.* \
+                      ${sysconfdir}/modprobe.d/qcom-adreno.conf \
+                      ${sysconfdir}/profile.d/qcom-adreno_env.sh"
+FILES:${PN}-egl = "${libdir}/libEGL_adreno.so.1 \
+                   ${@bb.utils.contains('DISTRO_FEATURES', 'wayland', '${libdir}/libeglSubDriverWayland.so.*', '', d)} \
+                   ${@bb.utils.contains('DISTRO_FEATURES', 'x11', '${libdir}/libeglSubDriverX11.so.*', '', d)} \
+                   ${datadir}/glvnd/egl_vendor.d/EGL_adreno.json"
+FILES:${PN}-gles2 = "${libdir}/libGLESv2*.so.*"
+FILES:${PN}-gles = "${libdir}/libGLESv1*.so.*"
```
lumag: How will these two packages be pulled in?
The application depends on `libEGL` and lib`GLESv2`, so it will get `libglvnd0` via RDEPENDS. The `libglvnd0` RDEPENDS on `virtual-egl-icd`, so it can pull EGL implementation (`${PN}-egl`). What about the GLES libs? GLVND doesn't pull in GLES v1 / v2 libs directly (it only loads EGL_vendor), so we can't add those dependencies.
Are they loaded by our libegl implementation?

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,69 @@
...
+             ${@bb.utils.contains('DISTRO_FEATURES', 'x11', 'libxcb libx11 xcb-util-image', '', d)}"
+
+PACKAGE_BEFORE_PN += " ${PN}-common ${PN}-gles ${PN}-gles2 ${PN}-egl ${PN}-vulkan ${PN}-cl"
+
+RPROVIDES:${PN}-common = "${PN}-common"
+RPROVIDES:${PN}-egl = "${PN}-egl"
+RPROVIDES:${PN}-cl = "${PN}-cl"
+
+# Use dynamically renamed package names
+RPROVIDES:${PN}-gles = "libglesv1-cm-adreno1"
```
lumag: gles1

recipes-bsp/packagegroups/packagegroup-qcom-graphics.bb
ssadhula: If I use non-renamed package, it is showing the same error - "An allarch packagegroup shouldn't depend on packages which are dynamically renamed"

lumag: And what does it mean? What are other ways to fix the error?

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: S is set to ${UNPACKDIR}/${BP}

But the tarball is extracted in ${UNPACKDIR}, so if I use ${S} in do_install step, it doesn't have the extracted files.

lumag: Please add a comment

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
lumag: Also `${PN}-gles1` instead of `${PN}-gles`

recipes-bsp/packagegroups/packagegroup-qcom-graphics.bb
lumag: as it seems to take a while.
Please follow the path that other had to take for their packagegroups and set `PACKAGE_ARCH = "${MACHINE_ARCH}" inside the packagegroup-qcom-graphics.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: Yes, we don't RDEPEND on vulkan-loader. But would it be useful to have vulkan-loader as part of multimedia image? Incase any vulkan application is pushed via adb

lumag: 1. Don't push random apps via ADB. Push the ipk and install it. 
2. Such 'nice to have' belongs to the qcom-multimedia-image rather than this recipe.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: Yes, you are right, our libEGL implementation loads libGLESv2. We will update RDEPENDS accordingly.

lumag:
```
ERROR: Nothing RPROVIDES 'qcom-adreno-cl' (but /work/build-yocto-check-layer-rx5k/meta-qcom/recipes-bsp/packagegroups/packagegroup-qcom-graphics.bb RDEPENDS on or otherwise requires it)
qcom-adreno RPROVIDES qcom-adreno-cl but was skipped: one of 'opengl vulkan' needs to be in DISTRO_FEATURES
NOTE: Runtime target 'qcom-adreno-cl' is unbuildable, removing...
Missing or unbuildable dependency chain was: ['qcom-adreno-cl']
ERROR: Required build target 'meta-world-pkgdata' has no buildable providers.
Missing or unbuildable dependency chain was: ['meta-world-pkgdata', 'packagegroup-qcom-graphics', 'qcom-adreno-cl']
```

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,75 @@
+LICENSE = "CLOSED"
+
+DESCRIPTION = "Qualcomm Adreno Graphics User Mode libraries"
```
lumag: This looks more like a SUMMARY. Please fix it and provide both SUMMARY and DESCRIPTION.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
```diff
@@ -0,0 +1,75 @@
...
+
+# S is set to ${UNPACKDIR}/${BP}, use subdir to unpack in S
```
lumag: It's rather 'no top-level dir in the archive, unpack to subdir to prevent UNPACKDIR pollution'.

lumag (changes requested):

ricardosalveti:
```
2025-09-16 13:24:05 - ERROR    - ERROR: Nothing PROVIDES 'libglvnd' (but /work/build/../../repo/recipes-graphics/adreno/qcom-adreno_1.838.1.bb DEPENDS on or otherwise requires it)
2025-09-16 13:24:05 - ERROR    - libglvnd was skipped: missing required distro feature 'glvnd' (not in DISTRO_FEATURES)
2025-09-16 13:24:05 - INFO     - NOTE: Runtime target 'qcom-adreno-egl' is unbuildable, removing...
```

ricardosalveti:
Recipe can only depend on libglvnd when glvnd is available in DISTRO_FEATURES, which is not the case for the pure meta-qcom build.

lumag:
> Recipe can only depend on libglvnd when glvnd is available in DISTRO_FEATURES, which is not the case for the pure meta-qcom build.

The problem is that the GL library depends on libglvnd. If glvnd is not provided, qcom-adreno can provide Vulkan and OpenCL, but not EGL / GL ES packages.

koenkooi (approved):
I prefer R* variables below do_install, since do_package happens after do_install, but this is good enough to go in.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: We will add other targets in future releases.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: I have removed all the headers.

recipes-graphics/adreno/qcom-adreno_1.838.1.bb
ssadhula: Resolving this conversation for now, as this PR is compiling fine. We will revisit this in future PRs.

Issue #1003, opened 2025-08-12
DSP - fastrpc - Testapp/utilities required in the meta-qcom initramfs and ramdisk


abbajaj806:
We need fastrpc binaries to be part of metq-qcom image fastrpc scenarios validations

https://github.com/quic/fastrpc

ekanshibu:
PR with fastrpc github switch is merged:
https://github.com/qualcomm-linux/meta-qcom/pull/843

PR #1043, opened 2025-08-29
initramfs-test-full-image: add fastrpc-tests package

ekanshibu:
Add `fastrpc-tests` package in initramfs-test-full-image for fastrpc
testing.

Closes: https://github.com/qualcomm-linux/meta-qcom/issues/1003

### DIFF
--- a/recipes-test/images/initramfs-test-full-image.bb
+++ b/recipes-test/images/initramfs-test-full-image.bb
@@ -6,6 +6,7 @@ PACKAGE_INSTALL += " \
     bootrr \
     coreutils \
     expect \
+    fastrpc-tests \
     hdparm \
     igt-gpu-tools-tests \
     kexec \


lumag:
Well... I don't think so. The package is not required for booting or functioning of the hardware. Please add test packages to the test images instead.

ekanshibu:
I see. Can I use the same PR to bring that change or should I create a new one?

lumag:
Up to you, I'm fine either way.

recipes-test/images/initramfs-test-full-image.bb
```diff
@@ -6,6 +6,7 @@ PACKAGE_INSTALL += " \
     bootrr \
     coreutils \
     expect \
+    fastrpc-tests \
```
vkraleti, line 9: If the FastRPC userspace library resides in the root filesystem, how does placing the test binaries in the ramdisk help? Once the rootfs is mounted, the ramdisk context is no longer available, which means the tests would not be accessible for execution. Has this approach been validated?

lumag: This initramfs is used to run tests, it doesn't mount / pivot to the on-disk rootfs.

ricardosalveti (approved):

PR #1047, opened 2025-09-01
action.yml: Upload OpenCV IPK for nightly private sanity testing

kraghava-qti:
The OpenCV IPK is built but not included in any image. This change uploads it for private nightly sanity testing of OpenCV packages built by packagegroup-qcom-ci. The IPK will be side-loaded to the target.

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -67,10 +67,14 @@ runs:
         # expects file to be relative to our PWD. deploy_dir is outside
         # that, so we move things around:
         deploy_dir=../kas/build/tmp/deploy/images/${{inputs.machine}}
+        ipk_dir=../kas/build/tmp/deploy/ipk
         uploads_dir=./uploads/${{ inputs.distro_name }}${{ inputs.kernel_dirname }}/${{ inputs.machine }}
+        uploads_ipk_dir=$uploads_dir/ipk
         mkdir -p $uploads_dir
+        mkdir -p $uploads_ipk_dir
         find $deploy_dir/ -maxdepth 1 -type f -exec cp {} $uploads_dir/ \;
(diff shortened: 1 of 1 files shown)


ricardosalveti:
What is special about opencv that we only want these ipks to be published? I think we want to better discuss if we should publish the ipks created during the build or if we should instead have the packages installed as part of an image.

lumag:
No. Please get it included into the testing image rather than publishing the ipk.

lumag:
> > What is special about opencv that we only want these ipks to be published? I think we want to better discuss if we should publish the ipks created during the build or if we should instead have the packages installed as part of an image.
> 
> There’s nothing special about OpenCV, except that it happens to be one of the first IPKs we want to preserve to demonstrate the capability of installing packages on pure upstream builds to showcase downstream features. This list can grow big pretty soon.
> 
> > No. Please get it included into the testing image rather than publishing the ipk.
> 
> Even if the pkg is part of the image, still we need to preserve IPKs for testing different configuration combinations planned.

We don't provide binary feeds. Please build the images with all IPKs installed as required.

sssanjee-quic:
Hi @lumag , Since we have OpenCV BB recipe which builds both test applications and libraries, can you please let us know if we can integrate the OpenCV into qcom-multimedia-image full distro image in https://github.com/qualcomm-linux/meta-qcom-distro/tree/main/recipes-products/images ?

sssanjee-quic:
can you please help to share more insights on testing image , i don't see testing image in https://github.com/qualcomm-linux/meta-qcom-distro/tree/main/recipes-products/images

vkraleti:
@sssanjee-quic You may consider adding to [packagegroup-qcom-test-pkgs.bb](https://github.com/qualcomm-linux/meta-qcom-distro/blob/main/recipes-products/packagegroups/packagegroup-qcom-test-pkgs.bb)  Images  interested in picking test pkgs will pick this group.

PR #1048, opened 2025-09-02
Add adbd support into ramdisk

vkraleti:
Move image-adbd.bbclass from the meta-qcom-distro layer to the meta-qcom layer, allowing initrd images to inherit it and support adbd usage in ramdisk environments. 
Add enable-adbd to IMAGE_FEATURES of initramfs-test-image, enabling adbd by default for improved usability and debugging in initramfs-based systems.

### DIFF
--- a/recipes-test/images/initramfs-test-image.bb
+++ b/recipes-test/images/initramfs-test-image.bb
@@ -2,6 +2,9 @@ require recipes-test/images/initramfs-tiny-image.bb
 
 DESCRIPTION = "Small ramdisk image for running tests (bootrr, etc)"
 
+IMAGE_FEATURES += "enable-adbd"
+inherit image-adbd
+
 PACKAGE_INSTALL += " \
     alsa-utils-alsaucm \
     alsa-utils-amixer \
@@ -37,8 +40,6 @@ PACKAGE_INSTALL += " \
 # on. This can be handled by .bbappends, but then image recipes including this
 # one would not get all these tools. So simulate dynamic bbappend here.
 PACKAGE_INSTALL_openembedded-layer += " \
-    android-tools-adbd \
-    android-tools-adbd-cmdline \
     cpufrequtils \
     cryptsetup \
     devmem2 \


recipes-test/images/initramfs-test-image.bb
```diff
@@ -2,6 +2,9 @@ require recipes-test/images/initramfs-tiny-image.bb
 
 DESCRIPTION = "Small ramdisk image for running tests (bootrr, etc)"
 
+IMAGE_FEATURES += "enable-adbd"
+inherit image-adbd
```
lumag, line 6: You can't do this by default. `android-tools` is not a part of OE-Core.

vkraleti: Thanks for pointing. Updated the bbclass to install `android-tools-adbd` package only when openembedded layer is available.

classes-recipe/image-adbd.bbclass
```diff
@@ -0,0 +1,22 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+# This class installs adbd into the target image when openembedded-layer is available.
+# The adbd daemon is disabled unless IMAGE_FEATURES contains the 'enable-adbd'
+# Also one can disable adbd by removing /etc/usb-debugging-enabled from rootfs manually.
+
+IMAGE_FEATURES[validitems] += "enable-adbd"
+
+PACKAGE_INSTALL_openembedded-layer += " \
```
lumag: This depends on the PACKAGE_INSTALL_foo infrastructure and thus it's not generic enough. If you want to have this in the class, please check BBLAYERS here too.

classes-recipe/image-adbd.bbclass
```diff
@@ -0,0 +1,27 @@
...
+"
+
+enable_adbd_at_boot () {
+	touch ${IMAGE_ROOTFS}/etc/usb-debugging-enabled
+}
+
+ROOTFS_POSTPROCESS_COMMAND += "${@bb.utils.contains('IMAGE_FEATURES', [ 'enable-adbd' ], 'enable_adbd_at_boot; ', '',d)}"
+
+python() {
+    if 'openembedded-layer' not in d.getVar('BBFILE_COLLECTIONS').split():
```
lumag: I don't think this is optimal. It requires `BBFILE_COLLECTION` checks both here and on image side. What about moving all the checks to this class and documenting that it enables adbd only if it's available.

vkraleti: Updated changes to throw a warning is openembedded-layer is not included in the mix. Please check.

PR #1053, opened 2025-09-03
Build both multimedia-image and multimedia-proprietary-image in CI

vkraleti:
Enhance CI to build both images provided by meta-qcom-distro.
- qcom-multimedia-image
- qcom-multimedia-proprietary-image

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -76,6 +76,8 @@ jobs:
             yamlfile: ""
           - name: qcom-distro
             yamlfile: ':ci/qcom-distro.yml'
+          - name: qcom-distro-prop-image
+            yamlfile: ':ci/qcom-distro-prop-image.yml'
         kernel:
           - type: default
             dirname: ""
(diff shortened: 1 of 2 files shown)


ricardosalveti:
`ERROR: kgsl-dlkm-0.0+git-r0 do_compile: oe_runmake failed
`

kgsl is compatible with aarch64 only, we need to reflect that in meta-qcom-distro.

ricardosalveti:
And there is a bug in kgsl as well, as it tried to compile it (COMPATIBLE_MACHINES instead of COMPATIBLE_MACHINE), let me send a patch for that.

ci/qcom-distro.yml
```diff
@@ -31,4 +31,5 @@ local_conf_header:
 
 target:
   - qcom-multimedia-image
+  - qcom-multimedia-proprietary-image
```
lumag: Only for arm64 machines.
Maybe the image should also be defined as compatible with aarch64 only.

vkraleti: Updated change to exclude qcom-armv7a + qcom-distro-prop-image combination.

ricardosalveti (approved):

PR #1060, opened 2025-09-05
ci/poky: add kas config yml

quaresmajose:
This kas configuration poky.yml file is intended to override the base.yml.

[kas|kas-container] {...} ci/base.yml:ci/poky.yml
[kas|kas-container] {...} ci/machine.yml:ci/poky.yml

With that we disbale the layers in the base.yml and we add the main poky layers.
This allows you to change the distribution from the nodistro to the poky.

### DIFF
--- a/ci/poky.yml
+++ b/ci/poky.yml
@@ -1,0 +1,19 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+
+repos:
+  oe-core:
+    layers:
+      meta: disabled
+
+  meta-yocto:
+    layers:
+      meta-poky: disabled
+
+  poky:
+    url: https://github.com/yoctoproject/poky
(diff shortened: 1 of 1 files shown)


quaresmajose:
The second patch is not need and are there mostly to fix a kas bug. The PR remains in draft until this issue is resolved.

quaresmajose:
One of the interstig points of this week Yocto Project Status 9 September 2025 (WW36)
https://lists.yoctoproject.org/g/yocto/message/65825

Key Status/Updates:
- This change will mean “poky” as a combo-layer repository will no longer be needed or updated. This means autobuilder changes before we build 5.3 M3.
_I understand from here that the autobuilder will start using the git repository meta-yocto and stop using poky_

It seems that the poky distribution will now live only in the meta-poky layer that we already track in our base.yml meta-yocto repository.

quaresmajose:
Either way the poky combo layer will be frozen then this doesn't make sense anymore.

PR #1062, opened 2025-09-06
audio: driver: add audioreach-dlkm kernel module recipe

nandamajay:
Add audioreach-dlkm_1.0.bb recipe for building the AudioReach dynamic loadable kernel module for Qualcomm platforms.

To include this module in your image, add the following to your local.conf:
    IMAGE_INSTALL:append = " audioreach-dlkm"
    
This pull request supersedes [previous PR #1061](https://github.com/qualcomm-linux/meta-qcom/pull/1061), which was closed because the recipe required a fix to the S variable after the source directory was renamed.
The current PR includes the corrected path and resolves the build error.

### DIFF
--- a/recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
+++ b/recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
@@ -1,0 +1,21 @@
+SUMMARY = "Qualcomm AudioReach Kernel Module"
+DESCRIPTION = "Kernel module for Qualcomm AudioReach"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://../LICENSE;md5=0a5a2ad232bafb6974f9a29d1ba0f488"
+
+inherit module
+
+SRC_URI = "git://github.com/AudioReach/audioreach-kernel.git;protocol=https;branch=master"
+SRCREV = "f414028b67971a830c956298338831dd8217969a"
+S = "${UNPACKDIR}/audioreach-dlkm-${PV}/audioreach-driver"
+
+EXTRA_OEMAKE += "DESTDIR=${D}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+# Restrict build to supported machines
+COMPATIBLE_MACHINE:aarch64 = "qcm6490|qcs6490|qcs8275|qcs8300|qcs9075|qcs9100"
+
+FILES:${PN} = "${nonarch_base_libdir}/modules/${KERNEL_VERSION}/updates/audioreach_driver.ko"
+RPROVIDES:${PN} += "kernel-module-audio-reach-driver kernel-module-audioreach-driver-${KERNEL_VERSION}"
+


recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,21 @@
+SUMMARY = "Qualcomm AudioReach Kernel Module"
+DESCRIPTION = "Kernel module for Qualcomm AudioReach"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://../LICENSE;md5=0a5a2ad232bafb6974f9a29d1ba0f488"
+
+inherit module
+
+SRC_URI = "git://github.com/AudioReach/audioreach-kernel.git;protocol=https;branch=master"
+SRCREV = "f414028b67971a830c956298338831dd8217969a"
+S = "${UNPACKDIR}/audioreach-dlkm-${PV}/audioreach-driver"
```
lumag: Please drop. Subdirs can go into SRC_URI.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,21 @@
...
+
+EXTRA_OEMAKE += "DESTDIR=${D}"
```
lumag: No need to.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,21 @@
...
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
```
lumag: Why do you need this one? What is the default for modules?

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,21 @@
...
+
+# Restrict build to supported machines
+COMPATIBLE_MACHINE:aarch64 = "qcm6490|qcs6490|qcs8275|qcs8300|qcs9075|qcs9100"
```
lumag: Definitely no. I hope it can be built for all 64 bit ARM machines, including qcom-armv8a. 
Don't be too creative.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,21 @@
...
+
+FILES:${PN} = "${nonarch_base_libdir}/modules/${KERNEL_VERSION}/updates/audioreach_driver.ko"
+RPROVIDES:${PN} += "kernel-module-audio-reach-driver kernel-module-audioreach-driver-${KERNEL_VERSION}"
```
lumag: Drop both. And also an extra empty line at the end of the file.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thank you for the feedback. I will update the commit message and PR description to remove the extra information and avoid suggesting changes to local.conf.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thanks for pointing this out. I will update the recipe to include subdirs in SRC_URI and remove the explicit S assignment

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Noted. I will remove the EXTRA_OEMAKE line from the recipe.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thank you for the clarification. I will update COMPATIBLE_MACHINE to support all 64-bit ARM machines as recommended.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: I will remove the unnecessary FILES and RPROVIDES lines and clean up the file formatting.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,21 @@
+SUMMARY = "Qualcomm AudioReach Kernel Module"
```
ricardosalveti: Is it going to be just for qualcomm?

nandamajay: Thanks Ricardo. In the recent commit, I’ve updated the field to remove the Qualcomm-specific reference.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,12 @@
+DESCRIPTION = "AudioReach driver"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=0a5a2ad232bafb6974f9a29d1ba0f488"
+
+SRC_URI = "git://github.com/AudioReach/audioreach-kernel.git;protocol=https;branch=master;destsuffix=audioreach-dlkm-${PV}"
```
lumag: No need for `destsuffix`, defaults should work good enough.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,12 @@
...
+
+SRCREV = "1caa146704f22dd90deaa0f8faf664905d975821"
```
lumag: Is it actually '1.0'? Or is `1.0` in the recipe file just a random default?

nandamajay: Yes, 1.0 is the actual package version defined in the recipe. It's not a placeholder — Yocto uses this version (${PV}) to manage source unpacking. For example, when cloning the source, Yocto expects the directory to follow the ${PN}-${PV} format, which in this case becomes audioreach-dlkm-1.0.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: I added destsuffix because during Yocto builds, the source was being cloned into a directory named only after the package (audioreach-dlkm). However, the build system was expecting the source to be in a directory named with both the package name and version (e.g., audioreach-dlkm-1.0).

Without destsuffix, this mismatch caused the build to fail. Adding it ensured the source directory matched what the build system expected, allowing the build to succeed.

If there's a recommended way to handle this layout without using destsuffix, I'm happy to update the recipe accordingly.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
lumag: I don't see any tag which would mark this commit as 1.0. What is the versioning strategy for the driver?

nandamajay: This driver is part of a stop-gap solution while we work in parallel on an upstream version of the Driver, in collaboration with Srinivas Kandagatla. Once the upstreaming effort stabilizes, we plan to adopt a proper versioning strategy.

lumag:
> > Why is this module being submitted to meta-qcom rather than being a part of meta-audioreach / meta-ar?
> 
> These drivers are specific to Qualcomm targets and while meta-ar (or meta-audioreach) includes parts of meta-qcom-distro, the drivers are platform-specific and should reside in meta-qcom

I'm not sure if I follow. Is meta-ar / meta-audioreach going to be platform-independent or platform-specifiic?

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
lumag: This doesn't make a sense to me. Skipping dstsuffix works (as much as it can work, see my next comment).

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
lumag: I read your response as 'in reality it's not 1.0'. Please update recipe name and set PV accordingly.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thanks for the clarification, @lumag 
You're right — after further analysis, I found the issue and confirmed that using subdir in SRC_URI resolves it cleanly. I'll update the recipe accordingly and remove the destsuffix usage. Thanks for the guidance!

nandamajay:
> I am very unhappy now. This recipe, as it's implemented now, doesn't produce any modules. The only thing that gets built and packaged is an empty `Module.symvers` file.
> 
> Please, validate your submissions before posting them.

You're absolutely right, @lumag  — this was a basic compilation mistake on my part, and I sincerely apologize for overlooking it.The issue stemmed from an incorrect Makefile setup, which caused the build system to skip module compilation and only generate an empty Module.symvers.

I've corrected the Makefile to properly invoke the kernel build system using KERNEL_SRC and M, and I’ll revalidate the recipe thoroughly before resubmitting.

Thanks again for your feedback and patience — I appreciate it.

lumag:
> Hi @lumag , Quick question: I noticed that unless I explicitly add the following line to the recipe: FILES:${PN} = "${nonarch_base_libdir}/modules/${KERNEL_VERSION}/updates/audioreach_driver.ko"
> 
> the .ko file doesn’t appear in the main IPK (audioreach-dlkm__.ipk). However, it does show up in the debug IPK (audioreach-dlkm-dbg__.ipk) under .debug/.
> 
> Is it recommended to keep this line to ensure proper packaging, or is there a cleaner way to handle this?

OE already handles everything for your. Please check the packages that are generated by the recipe.

> 
> Appreciate your guidance!

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,13 @@
+DESCRIPTION = "AudioReach driver"
```
lumag: Missing SUMMARY. DESCRIPTION is very terse.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,13 @@
...
+LIC_FILES_CHKSUM = "file://LICENSE;md5=0a5a2ad232bafb6974f9a29d1ba0f488"
+
+inherit module
+
+MAKE_TARGETS = "modules"
+
+SRC_URI = "git://github.com/AudioReach/audioreach-kernel.git;protocol=https;branch=master;subdir=audioreach-dlkm-1.0"
+SRCREV = "c494af74e420661591a3715f02e4f21f4944a0f2"
+
+COMPATIBLE_MACHINES = ""
```
lumag: ^$

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,13 @@
+DESCRIPTION = "AudioReach driver"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=0a5a2ad232bafb6974f9a29d1ba0f488"
+
+inherit module
+
+MAKE_TARGETS = "modules"
+
+SRC_URI = "git://github.com/AudioReach/audioreach-kernel.git;protocol=https;branch=master;subdir=audioreach-dlkm-1.0"
```
lumag: No subdirs, thank you.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,13 @@
+DESCRIPTION = "AudioReach driver"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=0a5a2ad232bafb6974f9a29d1ba0f488"
+
+inherit module
+
+MAKE_TARGETS = "modules"
```
lumag: Please add some logic to the order of the variables. You can't make the module before SRC_URI / SRCREV come into play.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,13 @@
...
+
+SRC_URI = "git://github.com/AudioReach/audioreach-kernel.git;protocol=https;branch=master;subdir=audioreach-dlkm-1.0"
+SRCREV = "c494af74e420661591a3715f02e4f21f4944a0f2"
```
lumag: Missing PV assignment

lumag (changes requested):
Squash into two commits.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thanks, @lumag ! I've added a SUMMARY and expanded the DESCRIPTION to better reflect the driver's functionality.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thanks for the review!
I've updated the build to restrict it to supported platforms using COMPATIBLE_MACHINE:
COMPATIBLE_MACHINE = "qcs6490.*|qcs8275.*|qcs8300.*|qcs9075.*|qcs9100.*"

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thanks, @lumag — updated the recipe to remove subdir from SRC_URI as suggested. The module sources are now pulled directly from the root of the repository:

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thanks, @lumag  — reordered the variables in the recipe to ensure logical flow. SRC_URI and SRCREV now appear before inherit and MAKE_TARGETS, so the source is defined before the build logic kicks in.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thanks, @lumag — added the missing PV assignment to the recipe

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
ricardosalveti: It is now saying  "QCOM AudioReach kernel module" again.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,15 @@
+SUMMARY = "QCOM AudioReach kernel module"
+DESCRIPTION = "Builds the AudioReach DLKM for advanced audio processing on Qualcomm SoCs."
+
+PV = "1.0"
```
ricardosalveti: I don't think PV should be 1.0 here, as raised before by @lumag.

There is no tag at https://github.com/AudioReach/audioreach-kernel/tags, and no indication that the version is indeed 1.0.

Makes more sense to move to `PV = "0.0+git"` and rename the recipe to `recipes-kernel/audioreach-module/audioreach-dlkm_git.bb` instead.

ricardosalveti:
```
WARNING: audioreach-dlkm-1.0-r0 do_install: Module.symvers not found in /home/rsalveti/build/qualcomm-linux/build/tmp/work/qcs9100_ride_sx-qcom-linux/audioreach-dlkm/1.0/sources/audioreach-dlkm-1.0/
WARNING: audioreach-dlkm-1.0-r0 do_install: Please consider setting MODULES_MODULE_SYMVERS_LOCATION to a
WARNING: audioreach-dlkm-1.0-r0 do_install: directory below B to get correct inter-module dependencies
```

Please set MODULES_MODULE_SYMVERS_LOCATION to the correct path.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,15 @@
...
+LIC_FILES_CHKSUM = "file://LICENSE;md5=0a5a2ad232bafb6974f9a29d1ba0f488"
+
+SRC_URI = "git://github.com/AudioReach/audioreach-kernel.git;protocol=https;branch=master"
+SRCREV = "c494af74e420661591a3715f02e4f21f4944a0f2"
+
+inherit module
+
+MAKE_TARGETS = "modules"
+
+COMPATIBLE_MACHINE = "qcs6490.*|qcs8275.*|qcs8300.*|qcs9075.*|qcs9100.*"
```
ricardosalveti: I don't like exposing the soc families here for a module, as we can build this module probably for any architecture. If really for aarch64 only, better to be specific to it instead of multiple machines families as this will have to be maintained overtime.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thanks @ricardosalveti , that makes sense. I'll update the recipe to use PV = "0.0+git" and rename it to audioreach-dlkm_git.bb under recipes-kernel/audioreach-module/ as suggested. Appreciate the clarification!
Just to add — this is a Qualcomm-specific stop-gap solution to enable internal integration. We're actively working on an upstreamable version and will align the recipe and versioning accordingly once that’s ready.

lumag: Upstremable version doesn't need versioning, it goes directly to upstream.
As we lightly discussed, upstreameable solution would likely require extending the upstream drivers in order to reload /program topology into the DSP (hint: emu10k1, emufx programming using hwdep device). The out-of-tree replacing modules can't become an upstreamable solution.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
lumag: We are really going in circles. This has been frowned upon, you have changed it, now this is back. Please stop doing that.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
lumag: Anyway, this is not "1.0". It is "0.0+git".

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
```diff
@@ -0,0 +1,15 @@
+SUMMARY = "QCOM AudioReach kernel module"
+DESCRIPTION = "Builds the AudioReach DLKM for advanced audio processing on Qualcomm SoCs."
```
lumag: No, this is not a description. Please take a look around then write a suitable one.

recipes-bsp/packagegroups/packagegroup-qcom-ci.bb
```diff
@@ -21,4 +21,5 @@ RDEPENDS:${PN}:append:aarch64 = " \
     iris-video-dlkm \
     kgsl-dlkm \
     qcom-fastcv-binaries \
+    audioreach-dlkm \
```
lumag: This list has been sorted. Please keep it so.

lumag:
> > > Why is this module being submitted to meta-qcom rather than being a part of meta-audioreach / meta-ar?
> > 
> > 
> > These drivers are specific to Qualcomm targets and while meta-ar (or meta-audioreach) includes parts of meta-qcom-distro, the drivers are platform-specific and should reside in meta-qcom
> 
> I'm not sure if I follow. Is meta-ar / meta-audioreach going to be platform-independent or platform-specifiic?

I don't think I got any response to this question.  I'd very much prefer to have these modules in the AudioReach layer, so that you have better control on them (and so that they don't break builds when upstream kernel API changes).

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Hi @ricardosalveti  @lumag  — thanks for the feedback.
Just to confirm before I proceed: would you prefer we keep COMPATIBLE_MACHINE = "aarch64" to avoid listing individual SoC families ?
I want to make sure we're aligned and avoid reverting this again. Please let me know if this is the direction you'd like to keep moving forward with.

lumag: No, we are not.This will not work. Earlier I've pointed out what you need to do.

nandamajay: Thanks @lumag — I’ve updated the recipe to use:
COMPATIBLE_MACHINE = "^$"

recipes-bsp/packagegroups/packagegroup-qcom-ci.bb
nandamajay: @lumag updated

recipes-kernel/audioreach-module/audioreach-dlkm_git.bb
```diff
@@ -0,0 +1,18 @@
+SUMMARY = "QCOM AudioReach kernel module"
+DESCRIPTION = "This module is designed for integration with the AudioReach framework, allowing flexible deployment of audio algorithms and efficient hardware utilization."
+
+PV = "0.0+git"
```
ricardosalveti: Please move PV after the license declaration (or after SRCREV, with an empty line separating it).

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thanks @lumag — I’ve revised the DESCRIPTION to better reflect the module’s purpose and aligned it with the style used in similar kernel module recipes. Please let me know if further adjustments are needed.

recipes-kernel/audioreach-module/audioreach-dlkm_1.0.bb
nandamajay: Thanks @ricardosalveti  and @lumag  — updated the recipe to PV = "0.0+git" and renamed it to audioreach-dlkm_git.bb under recipes-kernel/audioreach-module/.

recipes-kernel/audioreach-module/audioreach-dlkm_git.bb
nandamajay: Thanks @ricardosalveti  — I've updated the recipe to move PV after the license declaration. Let me know if anything else needs adjustment.

recipes-kernel/audioreach-module/audioreach-dlkm_git.bb
```diff
@@ -0,0 +1,18 @@
+SUMMARY = "QCOM AudioReach kernel module"
```
lumag: s/QCOM/Qualcomm/

recipes-kernel/audioreach-module/audioreach-dlkm_git.bb
```diff
@@ -0,0 +1,18 @@
...
+DESCRIPTION = "This module is designed for integration with the AudioReach framework, allowing flexible deployment of audio algorithms and efficient hardware utilization."
+
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=0a5a2ad232bafb6974f9a29d1ba0f488"
+PV = "0.0+git"
```
lumag: What has PV in common with LICENSE?

nandamajay: Thanks @lumag — updated the recipe to move PV after SRCREV with proper spacing for better logical grouping. Appreciate the guidance!

recipes-kernel/audioreach-module/audioreach-dlkm_git.bb
nandamajay: Thanks @lumag — updated SUMMARY to use "Qualcomm" instead of "QCOM".

recipes-kernel/audioreach-module/audioreach-dlkm_git.bb
```diff
@@ -0,0 +1,19 @@
...
+
+inherit module
+
+MAKE_TARGETS = "modules"
+
+MODULES_MODULE_SYMVERS_LOCATION = "audioreach-driver"
+
+COMPATIBLE_MACHINE = "^$"
+COMPATIBLE_MACHINE:aarch64 = "(.*)"
```
ndechesne, line 19: This is not restricted to the QCOM platforms mentioned in the commit message.

PR #1064, opened 2025-09-09
Fix COMPATIBLE_MACHINE checks

lumag:
Empty value of COMPATIBLE_MACHINE makes base.bbclass skip the compatibility check completely. Use "^$" (matches empty string) as a documented way to skip the recipe by default. While fixing this also fix the broken configurations for qcom-armv8a and qcom-armv7a machines which ended up having an empty token in MACHINEOVERRIDES, which matched the COMPATIBLE_MACHINE which was expected to filter out recipes.


ricardosalveti (dismissed):
Nice catch.

conf/machine/qcom-armv7a.conf
```diff
@@ -8,6 +8,7 @@
 #
 # Do not use this machine for SDXnn modems or for Snapdragon S1/S2/S3.
 
+SOC_FAMILY = "qcom-generic"
```
ndechesne: I am not sure I understand the actual problem. that said I don't quite like the solution.. 
For each SOC I think SOC_FAMILY should be "qcom:<soc>", that makes sense. I think it should be just "qcom" for the generic machine, and not "qcom:qcom-generic". what do you think?

conf/machine/include/qcom-common.inc
```diff
@@ -1,5 +1,5 @@
-SOC_FAMILY:prepend = "qcom:"
 require conf/machine/include/soc-family.inc
+MACHINEOVERRIDES =. "qcom:"
```
ndechesne, line 2: thanks! this is much better than previous, and quite a nice improvement!

PR #1066, opened 2025-09-10
Deploy contents.xml to enable image flashing in Axiom

vkraleti:
The 'contents.xml' file maps images to their respective partitions in a format understood by Axiom. Deploy the same to enable Axiom automatic image flashing on supported boards.


recipes-bsp/partition/qcom-partition-conf_git.bb
```diff
@@ -25,6 +25,8 @@ do_deploy() {
         install -m 0644 ${QCOM_PLATFORM_SUBDIR}/rawprogram*.xml -D ${DEPLOYDIR}/partitions/${QCOM_PLATFORM_SUBDIR}
         install -m 0644 ${QCOM_PLATFORM_SUBDIR}/zeros_*.bin -D ${DEPLOYDIR}/partitions/${QCOM_PLATFORM_SUBDIR}
         install -m 0644 ${QCOM_PLATFORM_SUBDIR}/wipe_rawprogram_PHY*.xml -D ${DEPLOYDIR}/partitions/${QCOM_PLATFORM_SUBDIR}
+        find "${QCOM_PLATFORM_SUBDIR}" -maxdepth 1 -name 'contents.xml' \
```
ndechesne: why did you use the find command here? QCOM_PLATFORM_SUBDIR is already the folder where contents.xml should exist.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -102,6 +102,10 @@ create_qcomflash_pkg() {
         install -m 0644 ${zeros} .
     done
 
+    [ -e "${DEPLOY_DIR_IMAGE}/${QCOM_PARTITION_FILES_SUBDIR}" -a \
+        ! -e "contents.xml" ] && \
```
ndechesne: why do you need this line? in which situation can contents.xml already exist?

classes-recipe/image_types_qcom.bbclass
```diff
@@ -102,6 +102,10 @@ create_qcomflash_pkg() {
         install -m 0644 ${zeros} .
     done
 
+    [ -e "${DEPLOY_DIR_IMAGE}/${QCOM_PARTITION_FILES_SUBDIR}" -a \
```
ndechesne: -d instead of -e for folder

classes-recipe/image_types_qcom.bbclass
vkraleti: Added it assuming that for incremental builds no need to run the task again.

recipes-bsp/partition/qcom-partition-conf_git.bb
vkraleti: Present implementation in qcom-ptool is template based and templates are not added for all platforms. So while compiling for some machines like apq8016, RB1, RB2 'contents.xml' is not available. Once a more generic solution is implemented, this can be dropped.

classes-recipe/image_types_qcom.bbclass
vkraleti: This was a mistake. The condition should check for the presence of a file rather than a folder, since not all machines support the generation of 'contents.xml'. Corrected it in latest changes.

recipes-bsp/partition/qcom-partition-conf_git.bb
ndechesne: then use if [ -e <> ] instead?

classes-recipe/image_types_qcom.bbclass
ricardosalveti: This is not done for the other files, so no need to do this just for one xml file.

ndechesne: there is no concept of incremental build at this stage. the task is executed if needed or not , and in which case it's fed from the sstate-cache. when it is executed it is executed in a clean folder.

recipes-bsp/partition/qcom-partition-conf_git.bb
```diff
@@ -25,6 +25,8 @@ do_deploy() {
         install -m 0644 ${QCOM_PLATFORM_SUBDIR}/rawprogram*.xml -D ${DEPLOYDIR}/partitions/${QCOM_PLATFORM_SUBDIR}
         install -m 0644 ${QCOM_PLATFORM_SUBDIR}/zeros_*.bin -D ${DEPLOYDIR}/partitions/${QCOM_PLATFORM_SUBDIR}
         install -m 0644 ${QCOM_PLATFORM_SUBDIR}/wipe_rawprogram_PHY*.xml -D ${DEPLOYDIR}/partitions/${QCOM_PLATFORM_SUBDIR}
+        ! [ -e ${QCOM_PLATFORM_SUBDIR}/contents.xml ] || \
```
ndechesne: sorry to be pedantic here.. but I find it much simpler to do

[ -e <file] ]; && do something

lumag: Note, this also sets `$?`, which might have unexpected consequences when running with `set -e`.

vkraleti: @ndechesne `[ -e <file>] && do something` is sending an unintended `exit 1` code thus causing errors in bitbake environment. Instead using a full if block is much more safer here.
```
if [ -e <file] ]; then 
    do something
fi
```

Issue #1015, opened 2025-08-14
kas-container is potentially downloaded in an unsafe way


ndechesne:
Since #846, kas-container is downloaded by a job and stored in the persistent storage (as KAS_CONTAINER). However this seems unsafe since each job will override kas-container in the persistent storage. We should download kas-container in a workflow specific folder instead.

PR #1077, opened 2025-09-15
[WIP] ci: store the kas-container the temporary directory

quaresmajose:
This will avoid possible race condiction.
i.e: between diferent versions in multiple PR

Fixes https://github.com/qualcomm-linux/meta-qcom/issues/1015


quaresmajose:
I rewrote commit messages and added the problem description

lumag (dismissed):

.github/workflows/build-yocto.yml
```diff
@@ -55,6 +62,18 @@ jobs:
...
+        uses: actions/download-artifact@v4
+        with:
+          name: kas-container
+          path: ${{ runner.temp }}
+
+      - name: Setting up kas-container
+        run: |
+          KAS_CONTAINER=$RUNNER_TEMP/kas-container
+          echo "KAS_CONTAINER=$KAS_CONTAINER" >> $GITHUB_ENV
+          chmod +x $KAS_CONTAINER
```
ndechesne, line 80: are you sure we need the chmod +x each time we download the file, since we did it when it was uploaded? 
also, I don't think we need to set KAS_CONTAINER and we could use $RUNNER_TEMP/kas-container everywhere directly. So we if can get rid of chmod+x and the variable we can completely remote the 'setting up kas container' job which would be nice..

ndechesne (commented):
please include Fixes: #1015 in the commit log.

.github/workflows/build-yocto.yml
quaresmajose: _File permissions are not maintained during artifact upload. All directories will have 755 and all files will have 644. For example, if you make a file executable using chmod and then upload that file, post-download the file is no longer guaranteed to be set as an executable._
_If you must preserve permissions, you can tar all of your files together before artifact upload. Post download, the tar file will maintain file permissions and case sensitivity._
https://github.com/actions/upload-artifact?tab=readme-ov-file#permission-loss

quaresmajose: Habving $RUNNER_TEMP/kas-container everywere is a bit bigger imo but I can drop the KAS_CONTAINER.

koenkooi:
> > please include Fixes: #1015 in the commit log.
> 
> The commit already mention that:
> 
> ```
> This change will fix [2] avoiding possible race condiction in multiple PR's.
> 
> [1] https://github.com/qualcomm-linux/meta-qcom/pull/846
> [2] https://github.com/qualcomm-linux/meta-qcom/issues/1015
> ```

Does that work with Github's auto-close logic that parses commit messages?

.github/workflows/build-yocto.yml
ndechesne: it's only used twice in the end, when we call the action.yml, so it's ok. 
can you please update this PR and also fix the commit message (see the fixes: xxx comment), so that we can merge it?

quaresmajose: Updated the commit message with the (fixes: xxx comment) and also removed the KAS_CONTAINER in an extra commit, can you just confirm if it is really to remove the KAS_CONTAINER envirement variable?

quaresmajose: The KAS_CONTAINER is need also is some of the ci script hepers so I added it again

quaresmajose:
> > > > please include Fixes: #1015 in the commit log.
> > > 
> > > 
> > > The commit already mention that:
> > > ```
> > > This change will fix [2] avoiding possible race condiction in multiple PR's.
> > > 
> > > [1] https://github.com/qualcomm-linux/meta-qcom/pull/846
> > > [2] https://github.com/qualcomm-linux/meta-qcom/issues/1015
> > > ```
> > 
> > 
> > Does that work with Github's auto-close logic that parses commit messages?
> 
> No, it requires (more or less) exact wording as documented by GH.

Lets see if the auto-close will works, added:
```
Fixes: #1015
```

.github/workflows/build-yocto.yml
```diff
@@ -10,15 +10,16 @@ on:
...
 
 jobs:
   kas-setup:
     if: github.repository == 'qualcomm-linux/meta-qcom'
     runs-on: [self-hosted, qcom-u2404, amd64-ssd]
     steps:
-      - name: Update kas-container
+      - name: Setting up kas-container
         run: |
+          KAS_CONTAINER=$RUNNER_TEMP/kas-container
```
ndechesne, line 26: can KAS_CONTAINER be defined in the env section at the top of the yml file? so that it is set only once.

.github/actions/compile/action.yml
```diff
@@ -27,6 +25,19 @@ runs:
         name: kas-lockfile
         path: ci/
 
+    - name: Download kas-container
```
ndechesne, line 31: is that needed here? we should download the artifact in build-yocto.yml instead, and leave that outside of the action file. it would be slightly nicer, no?

.github/workflows/build-yocto.yml
quaresmajose: The answer is no, that would actually be ideal and that's where I started. But it is not possible because variable expansion is not supported, so the following gives an error:
```
env:
  KAS_CONTAINER: ${{ runner.temp }}/kas-container

```

.github/actions/compile/action.yml
quaresmajose: If it is done in the action, it only needs to be done in a single location, which is what is currently implemented. If it is done in build-yocto.yml, it will have to be done every time a new worker is created. Then in build-yocto.yml it will have to be implemented in compile_warm_up and compile.

.github/workflows/build-yocto.yml
quaresmajose: [Invalid workflow file: .github/workflows/build-yocto.yml#L1](https://github.com/quaresmajose/meta-qcom/actions/runs/18224193177/workflow)
```
(Line: 13, Col: 18): Unrecognized named-value: 'runner'. Located at position 1 within expression: runner.temp
```

lumag: Pass the path or the temp path as an input?

quaresmajose: Using a variable is also good when we want to add some default command line args to the kas-container

mwasilew:
Last set of tests failed due to infrastructure issue

ndechesne (approved):

Issue #1079, opened 2025-09-16
bitbake ADBD parsing WARNING issue


quaresmajose:
Since https://github.com/qualcomm-linux/meta-qcom/pull/1048 we have now a warning on every build printed during bitbake parsing
```
2025-09-16 09:12:23 - INFO     - Parsing recipes...WARNING: /work/build/../../repo/recipes-test/images/initramfs-kerneltest-full-image.bb: 'image-adbd' is inherited but the meta-openembedded layer is not included in bblayers. ADBD may not work as expected.
2025-09-16 09:12:23 - INFO     - WARNING: /work/build/../../repo/recipes-test/images/initramfs-test-full-image.bb: 'image-adbd' is inherited but the meta-openembedded layer is not included in bblayers. ADBD may not work as expected.
2025-09-16 09:12:23 - INFO     - WARNING: /work/build/../../repo/recipes-test/images/initramfs-kerneltest-image.bb: 'image-adbd' is inherited but the meta-openembedded layer is not included in bblayers. ADBD may not work as expected.
2025-09-16 09:12:23 - INFO     - WARNING: /work/build/../../repo/recipes-test/images/initramfs-test-image.bb: 'image-adbd' is inherited but the meta-openembedded layer is not included in bblayers. ADBD may not work as expected.
2025-09-16 09:12:34 - INFO     - done.
```

It would be useful to remove this warning and only display it when necessary.

PR #1082, opened 2025-09-16
Replace the event handler-based check with a custom task in image-adbd.bbclass     

vkraleti:
The 'ConfigParsed' event is triggered when base configuration files
like bitbake.conf and global INHERIT statements are parsed. Since
these are common across all builds, the oelayer_check warning appears
in nearly every build, which is unwarranted.

To prevent this, replace the event handler-based check with a task that
verifies layer presence and shows a warning. As tasks are scheduled only
when the recipe inheriting image-adbd.bbclass is selected for build, this
change avoids unnecessary warnings.

Fixes https://github.com/qualcomm-linux/meta-qcom/issues/1079

### DIFF
--- a/classes-recipe/image-adbd.bbclass
+++ b/classes-recipe/image-adbd.bbclass
@@ -20,11 +20,11 @@ enable_adbd_at_boot () {
 
 ROOTFS_POSTPROCESS_COMMAND += "${@bb.utils.contains('IMAGE_FEATURES', [ 'enable-adbd' ], 'enable_adbd_at_boot; ', '',d)}"
 
-python oelayer_check() {
-    if 'openembedded-layer' not in e.data.getVar('BBFILE_COLLECTIONS').split():
+addtask oelayer_check before do_build
+do_oelayer_check[nostamp] = "1"
+python do_oelayer_check() {
+    if 'openembedded-layer' not in d.getVar('BBFILE_COLLECTIONS').split():
         bb.warn("'image-adbd' is inherited but the meta-openembedded layer"
                 " is not included in bblayers. ADBD may not work as expected.")
 }
 
-addhandler oelayer_check
-oelayer_check[eventmask] = "bb.event.RecipePreFinalise"


quaresmajose:
The warning are still there https://github.com/qualcomm-linux/meta-qcom/actions/runs/17773408844/job/50516456354?pr=1082#step:3:180

```
2025-09-16 17:27:50 - INFO     - Parsing recipes...WARNING: /work/build/../../repo/recipes-test/images/initramfs-test-full-image.bb: 'image-adbd' is inherited but the meta-openembedded layer is not included in bblayers. ADBD may not work as expected.
2025-09-16 17:27:50 - INFO     - WARNING: /work/build/../../repo/recipes-test/images/initramfs-kerneltest-image.bb: 'image-adbd' is inherited but the meta-openembedded layer is not included in bblayers. ADBD may not work as expected.
2025-09-16 17:27:50 - INFO     - WARNING: /work/build/../../repo/recipes-test/images/initramfs-test-image.bb: 'image-adbd' is inherited but the meta-openembedded layer is not included in bblayers. ADBD may not work as expected.
2025-09-16 17:27:50 - INFO     - WARNING: /work/build/../../repo/recipes-test/images/initramfs-kerneltest-full-image.bb: 'image-adbd' is inherited but the meta-openembedded layer is not included in bblayers. ADBD may not work as expected.
```

classes-recipe/image-adbd.bbclass
```diff
@@ -20,11 +20,11 @@ enable_adbd_at_boot () {
 
 ROOTFS_POSTPROCESS_COMMAND += "${@bb.utils.contains('IMAGE_FEATURES', [ 'enable-adbd' ], 'enable_adbd_at_boot; ', '',d)}"
 
-python oelayer_check() {
-    if 'openembedded-layer' not in e.data.getVar('BBFILE_COLLECTIONS').split():
+addtask oelayer_check before do_build
+do_oelayer_check[nostamp] = "1"
+python do_oelayer_check() {
+    if 'openembedded-layer' not in d.getVar('BBFILE_COLLECTIONS').split():
```
lumag, line 26: As a second thought, wouldn't it be enough to use bb.warn_once?

quaresmajose: It currently only appears once per component. Even using bb.warnonce the warning will appear once in the same way in the 3 components where it is used: initramfs-test-image, initramfs-kerneltest-image and initramfs-test-full-image.

lumag: It would be logical as we should be warning about all three recipes. I'm not sure what do you mean by the 'component' though.

vkraleti: At present, this warning appears for every image recipe that inherits the bbclass. Can update to "bb.warnonce" but not sure if there is a way to list all recipes that are inherit this class as part of that warning message.

lumag: I think, listing all recipes is a correct way to react: in the end, the user might not care about the warning about random image-adbd or initramfs-tiny-image, but it might notice the warning about initramfs-test-image which they are building

quaresmajose: By component I mean recipe/image. As mentioned in https://github.com/qualcomm-linux/meta-qcom/pull/1082#discussion_r2372745614 this warning appears for every image recipe that inherits the bbclass, if we change it to bb.warnonce the output will be exactly the same as with bb.warn. In my opinion using use bb.warnonce doesn't make sense in this scenario.

PR #1084, opened 2025-09-17
Enable GPU firmware even w/o OpenGL

lumag:
Enabling of GPU isn't limited to OpenGL-only cases. Install GPU firmware if the distro enabled 'vulkan' or 'opencl' (where applicable).

### DIFF
--- a/recipes-bsp/packagegroups/packagegroup-dragonboard820c.bb
+++ b/recipes-bsp/packagegroups/packagegroup-dragonboard820c.bb
@@ -8,7 +8,7 @@ PACKAGES = " \
 "
 
 RRECOMMENDS:${PN}-firmware = " \
-    ${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'linux-firmware-qcom-adreno-a530 linux-firmware-qcom-apq8096-adreno', '', d)} \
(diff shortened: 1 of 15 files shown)


ndechesne (approved):

recipes-bsp/packagegroups/packagegroup-dragonboard820c.bb
```diff
@@ -8,7 +8,7 @@ PACKAGES = " \
 "
 
 RRECOMMENDS:${PN}-firmware = " \
-    ${@bb.utils.contains('DISTRO_FEATURES', 'opengl', 'linux-firmware-qcom-adreno-a530 linux-firmware-qcom-apq8096-adreno', '', d)} \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'opencl opengl', 'linux-firmware-qcom-adreno-a530 linux-firmware-qcom-apq8096-adreno', '', d)} \
```
ricardosalveti: We should use bb.utils.contains_any then, otherwise it will only include when all are available in DISTRO_FEATURES.

lumag: Fixed

ricardosalveti (approved):

PR #1087, opened 2025-09-19
recipes-graphics: Update KGSL tip and drop merged patches from recipe

quic-ksanadhy:
Update the SRCREV to point to the latest commit in the KGSL source code repository. Additionally, remove patches that have become redundant, as their changes have already been merged in the source code and no longer required to be applied explicitly in kgsl-dlkm recipe.


quic-ksanadhy:
> Please also add a short description of the changes done as part of the revision bump.

Acknowledged, Updated commit text to add short description about the changes done as part of srcrev change.

quic-ksanadhy:
> Subject is incorrect: it should start from `qcom-adreno: update ...`. It changes qcom-adreno recipe, rather than everything in recipes-graphics. Also the subject is not the sentence, so it should not start from the capital letter.

Acknowledged.

lumag (approved):

PR #1090, opened 2025-09-19
linux-qcom-next config: make docker work, enable more network options

koenkooi:
Docker failed to launch due to missing iptables NAT compatibility:

```
Jan 01 01:34:17 qcs6490-rb3gen2-core-kit dockerd[1145]: failed to start daemon: Error initializing network controller: error obtaining controller instance: failed to register "bridge" driver: failed to create NAT chain DOCKER: iptables failed: iptables --wait -t nat -N DOCKER: modprobe: FATAL: Module ip_tables not found in directory /lib/modules/6.17.0-rc5-00207-g81052b1bbda7 
Jan 01 01:34:17 qcs6490-rb3gen2-core-kit dockerd[1145]: iptables v1.8.11 (legacy): can't initialize iptables table `nat': Table does not exist (do you need to insmod?) 
Jan 01 01:34:17 qcs6490-rb3gen2-core-kit dockerd[1145]: Perhaps iptables or your kernel needs to be upgraded. 
Jan 01 01:34:17 qcs6490-rb3gen2-core-kit dockerd[1145]:  (exit status 3) 
Jan 01 01:34:17 qcs6490-rb3gen2-core-kit systemd[1]: docker.service: Main process exited, code=exited, status=1/FAILURE
```

The `CONFIG_NETFILTER_XTABLES_LEGACY` symbol enables that compatibility layer, but while we're at it, enable all netfilter options, TCP congestion algorithms and queue disciplines.

Also enable `CONFIG_VETH`, docker needs that for networking as well.


ricardosalveti (approved):
I do wonder if we should eventually leverage the linux-yocto config fragments here, or the infrastructure at least, but this is fine for now.

lumag:
I start to dislike this config file. We need to sort out options which should go to qcom-next and options which should go to meta-qcom-distro. None of these options are really BSP-specific.

PR #1095, opened 2025-09-22
IQ-8275-EVK (RB4) enablement

vkraleti:
Added machine configuration, CI build support for IQ-8275-EVK.  Boot test changes will follow once the devices are setup in lab.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -107,6 +107,7 @@ jobs:
       fail-fast: true
       matrix:
         machine:
+          - iq-8275-evk
           - qcm6490-idp
           - qcs615-adp-air
           - qcs8300-ride-sx
--- a/ci/iq-8275-evk.yml
+++ b/ci/iq-8275-evk.yml
@@ -1,0 +1,8 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+  includes:
+  - ci/base.yml
+
+machine: iq-8275-evk


conf/machine/iq-8275-evk.conf
```diff
@@ -0,0 +1,26 @@
...
+MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
+    packagegroup-qcs8300-ride-firmware \
+    packagegroup-qcs8300-ride-hexagon-dsp-binaries \
+"
+
+QCOM_CDT_FILE = "cdt_qcs8275_iq_8275_evk"
+QCOM_BOOT_FILES_SUBDIR = "qcs8300"
+
+# FIXME provide device-specific table
+QCOM_PARTITION_FILES_SUBDIR = "partitions/qcs8300-ride-sx"
```
lumag, line 23: No device-specific parition table?

vkraleti: Not yet, I am adding it in qcom-ptool for both RB4 & RB8. Once ready 'll update machine conf to use it.

lumag: Ack

.github/workflows/build-yocto.yml
```diff
@@ -107,6 +107,7 @@ jobs:
         machine:
           - qcm6490-idp
           - qcs615-adp-air
+          - qcs8275-iq-8275-evk
```
ndechesne: I don't know why we ended up with qcs9075-iq-9075-evk earlier.. which seems wrong naming. 

I think we should not continue on this naming, and use a better name. The kernel has decided to go with lemans-evk and monaco-evk, which are different from the commercial names..  so perhaps we can use the commercial names, but at least not repeat the SoC name in there? 

e.g. iq-8275-evk?

ricardosalveti: Yeah, +1 for a shorter name, commercial related, as this is the evk for a particular soc, no need to replicate the soc name.

vkraleti: Guideline is to name machines using the format <SOC_ID>-<BOARD_NAME>. Followed the same for all existing machines as well as on 1.0.  For example `qcs9100`-`ride-sx`, `qcs8300`-`ride-sx`
For qcs9075-iq-9075-evk.conf:
- qcs9075 is the SOC_ID
- iq-9075-evk is the board name

Similarly, for qcs8275-iq-8275-evk.conf:
- qcs8275 is the SOC_ID
- iq-8275-evk is the board name

ndechesne: I understand the guideline when <BOARD_NAME> is generic, like ride-sx or idp, that makes sense. It is to ensure we can distinguish them. But when the board name already includes a reference to the onboard SoC, I would prefer to remove the duplication and keep things simpler.

vkraleti: Unfortunately I have no authority to take a final call here. @sbanerjee-quic can you confirm if this renaming is allowed?

lumag: More importantly, should we follow the upstream naming? In 6.18 that will become lemans-evk, monaco-evk, etc.

lumag: For the current quesrtion, I don't think we should be including the SoC name in the machine name, unless it's a part of the name on its own. E.g. it should be 'qcs9100-ride-sx', but 'rb1-core-kit'?

sbanerjeeqc: I am of opinion we go with upstream name and board name, monaco-iq-8275-evk, as kernel has got the DT as Monaco and lemans. IMO machine conf should be relatable to the kernel naming and commercial naming.

ndechesne: I very much prefer monaco-iq-8275-evk than the current name. I even like it :)
@lumag, ok?

lumag: I would prefer to use only commercial name as is (and add Monaco EVK in the description)

sbanerjeeqc: Based on discussion we will go with IQ-8275-EVK.
For Lemans as well we will use commercial name.

vkraleti: @ndechesne @lumag @ricardosalveti updated the machine conf name. Please check

conf/machine/iq-8275-evk.conf
```diff
@@ -0,0 +1,26 @@
...
+MACHINE_FEATURES += "efi pci"
+
+QCOM_DTB_DEFAULT ?= "monaco-evk"
+
+KERNEL_DEVICETREE ?= " \
+                      qcom/monaco-evk.dtb \
+                      "
+
+MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
+    packagegroup-qcs8300-ride-firmware \
```
lumag: I would rather not reuse ride-specific packages for a different board.

lumag (changes requested):

conf/machine/iq-8275-evk.conf
vkraleti: Created packagegroup-iq8275-evk-firmware.

lumag:
> Test failures due the rb3gen2 on the hyd lab.

If it's not reliable, it should be disabled for job submission.

recipes-bsp/hexagon-dsp-binaries/hexagon-dsp-binaries_20250917.bb
```diff
@@ -36,9 +36,9 @@ do_install () {
 
 PACKAGES =+ "\
     ${PN}-qcom-db820c-adsp \
-    ${PN}-qcom-qcs8300-ride-adsp \
-    ${PN}-qcom-qcs8300-ride-cdsp \
-    ${PN}-qcom-qcs8300-ride-gdsp \
+    ${PN}-qcom-qcs8300-adsp \
```
lumag: I'd prefer to keep these packages as ride-specific (because they include a path directly inside) and add symlinks for other platforms. This way if we get actually platform-specific files there (acdb, sensors data) we won't end up with a board trying to use files from another board..

conf/machine/iq-8275-evk.conf
lumag: Any progress here?

lumag (changes requested):

recipes-bsp/hexagon-dsp-binaries/hexagon-dsp-binaries_20250917.bb
vkraleti: Where do you recommend to add this symlink? In recipe or in dsp-binaries Makefiles?

conf/machine/iq-8275-evk.conf
vkraleti: Once https://github.com/qualcomm-linux/qcom-ptool/pull/42 merged into qcom-ptool, will update here.

recipes-bsp/hexagon-dsp-binaries/hexagon-dsp-binaries_20250917.bb
lumag: The config.txt file format supports Link: directive. Please use it

vkraleti: Thanks for the tip. Submitted https://github.com/linux-msm/hexagon-dsp-binaries/pull/20 and dropped hexagon-dsp-binaries addition for now. Will submit another PR once this merges.

lumag:
Test failures are not related to this PR. Merging it manually.

PR #1097, opened 2025-09-22
Align qcs615 boot firmware with qclinux 1.5 and add missing CDT

ricardosalveti:
Align boot firmware for qcs615 with the release used as part of qclinux 1.5 and include the missing CDT for the target.


ricardosalveti:
@vkraleti do you happen to know why there is a newer boot firmware rev just for qcs615? All the other ones are based on 00095.

lumag (approved):

PR #1101, opened 2025-09-25
Enable hardware watchdog support in CI images

vkraleti:
Enabling hardware watchdog in CI images is essential for collecting system dumps in case of hangs or crashes.
Set `WATCHDOG_RUNTIME_SEC` to 30 seconds via local.conf to ensure the `RuntimeWatchdogSec=` parameter 
in `/etc/systemd/system.conf` reflects this value, thereby activating the watchdog.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -42,6 +42,7 @@ local_conf_header:
     DISTRO_FEATURES:append = " efi pni-names"
     EXTRA_IMAGE_FEATURES = "allow-empty-password empty-root-password allow-root-login"
     IMAGE_ROOTFS_EXTRA_SPACE = "307200"
+    WATCHDOG_RUNTIME_SEC:pn-systemd = "30"
 
 machine: unset
 


koenkooi:
As a big fan of using watchdogs: yay! Should be consider globally enabling this as well for meta-qcom-distro?

ci/base.yml
```diff
@@ -42,6 +42,7 @@ local_conf_header:
     DISTRO_FEATURES:append = " efi pni-names"
     EXTRA_IMAGE_FEATURES = "allow-empty-password empty-root-password allow-root-login"
     IMAGE_ROOTFS_EXTRA_SPACE = "307200"
+    WATCHDOG_RUNTIME_SEC = "30"
```
quaresmajose: Since it is only used in systemd, maybe it will be better to set it local for that recipe with `WATCHDOG_RUNTIME_SEC:pn-systemd` and this also makes it more obvious that it is not a global variable.

ricardosalveti:
> As a big fan of using watchdogs: yay! Should be consider globally enabling this as well for meta-qcom-distro?

Yeah, @vkraleti please also add this at the distro layer itself.

ci/base.yml
vkraleti: Done

lumag (approved):

PR #1106, opened 2025-09-29
CI: reduce number of simultaneous builds jobs

lumag:
Our CI builds use a lot of parallel jobs, resulting in a high resource usage. Reduce the number of jobs being by merging two qcom-distro jobs into a single one and using special jobs for qcom-armv7a builds (since we don't support proprietary builds on qcom-armv7a).

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -75,8 +75,6 @@ jobs:
           - name: poky-altcfg
             yamlfile: ""
           - name: qcom-distro
-            yamlfile: ':ci/qcom-distro.yml'
-          - name: qcom-distro-prop-image
             yamlfile: ':ci/qcom-distro-prop-image.yml'
           # kvm is compatible with all the compile warm up machines
           - name: qcom-distro-kvm
@@ -123,15 +121,12 @@ jobs:
           - qcs615-ride
           - qcs8300-ride-sx
           - qcs9100-ride-sx
-          - qcom-armv7a
           - rb1-core-kit
           - sm8750-mtp
         distro:
           - name: poky-altcfg
             yamlfile: ""
           - name: qcom-distro
-            yamlfile: ':ci/qcom-distro.yml'
-          - name: qcom-distro-prop-image
             yamlfile: ':ci/qcom-distro-prop-image.yml'
           - name: qcom-distro-sota
             yamlfile: ':ci/qcom-distro-sota.yml'
@@ -147,7 +142,7 @@ jobs:
           - machine: qcom-armv8a
             distro:
                 name: qcom-distro
-                yamlfile: ':ci/qcom-distro.yml'
+                yamlfile: ':ci/qcom-distro-prop-image.yml'
             kernel:
                 type: additional
                 dirname: "+linux-yocto-dev"
@@ -161,11 +156,22 @@ jobs:
                 type: default
                 dirname: ""
                 yamlfile: ""
-        exclude:
-          # Incompatible builds
           - machine: qcom-armv7a
             distro:
-                name: qcom-distro-prop-image
+                name: poky-altcfg
+                yamlfile: ''
+            kernel:
+                type: default
+                dirname: ""
+                yamlfile: ""
+          - machine: qcom-armv7a
+            distro:
+                name: qcom-distro
+                yamlfile: ':ci/qcom-distro.yml'
+            kernel:
+                type: default
+                dirname: ""
+                yamlfile: ""
     name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
     steps:
       - uses: actions/checkout@v4
--- a/ci/qcom-distro-prop-image.yml
+++ b/ci/qcom-distro-prop-image.yml
@@ -5,4 +5,6 @@ header:
   includes:
    - ci/qcom-distro.yml
 
-target: qcom-multimedia-proprietary-image
+target:
+  - qcom-multimedia-image
+  - qcom-multimedia-proprietary-image


.github/workflows/build-yocto.yml
```diff
@@ -143,11 +139,19 @@ jobs:
                 type: additional
                 dirname: "+linux-yocto-lts"
                 yamlfile: ":ci/linux-yocto-lts.yml"
+          - machine: qcom-armv7a
+            distro:
+                name: qcom-distro-no-prop-image
+                yamlfile: ':ci/qcom-distro-no-prop-image.yml'
```
ndechesne: the idea to reduce the number of jobs is good. However I don't like having a yaml file to 'negate' something. in fact I don't think I like hardcoding the 'target' in the yaml file that much either, especially since it cannot be overriden easily (maybe something to improve in kas..).

the target can be provided on the command line, with kas build --target <>, how about we use that instead. e.g. for qcom-distro we don't set target in the yaml file (or only qcom-multimedia-image) and we deal with the prop image from the CLI?

lumag: I will take a look. What is the recommended way of testing CI scripts?

ndechesne: what do you mean 'testing the CI scripts' ?

mwasilew: There is no offline tool for testing github workflows. In theory pushing to "next" branch should be used. However I'm still struggling with the changes in test plans there. I'll try to have some version ready for review today or tomorrow so you can use the branch for your experiments.

lumag: I've updated CI config to drop targets but it's (almost) completely untested :-D

.github/actions/compile/action.yml
```diff
@@ -53,7 +55,7 @@ runs:
     - name: Kas build images
       shell: bash
       run: |
-        ${{inputs.kas}} build ci/mirror.yml:ci/${{ inputs.machine }}.yml${{ inputs.distro_yaml }}${{ inputs.kernel_yaml }}
+        ${{inputs.kas}} build ci/mirror.yml:ci/${{ inputs.machine }}.yml${{ inputs.distro_yaml }}${{ inputs.kernel_yaml }} -- ${{ inputs.targets }}
```
ndechesne: I believe you need kas build --target <TARGET> xxxx

.github/actions/compile/action.yml
```diff
@@ -14,6 +14,8 @@ inputs:
     required: true
   kas:
     required: true
+  targets:
```
ndechesne: target (without plural) seems better. The 'target' is a list of all images to build.

ndechesne (commented):
I like this change.. let's see what the other have to say. It does break users though, we need to be cautious, and we need to update the README as well.

.github/actions/compile/action.yml
lumag: done

.github/actions/compile/action.yml
lumag: `--target` doesn't seem to accept multiple targets. Passing them as 'raw bitbake args' seems to work.

vkraleti:
We could consider splitting the machines more evenly between the compile_warm_up & compile steps to reduce parallel load. Currently, the split is 2:7. Eventually, we can group 4–5 machines into a batch and run builds in multiple batches.

lumag:
> We could consider splitting the machines more evenly between the compile_warm_up & compile steps to reduce parallel load. Currently, the split is 2:7. Eventually, we can group 4–5 machines into a batch and run builds in multiple batches.

That would increase the amount of cache misses in the warm_up stage, so it's not logical

quaresmajose:
> > Removing the target from the yml is not good in my opinion because it will force the user to choose which image they want and this may not be easy for everyone. We have the envirment variable KAS_TARGET that overwrites the respective setting in the configuration file. Using this we wouldn't even need new arguments in kas and we can keep this only on CI
> 
> I'm not sure, why choosing an image is not easy. I think that specifying the image in the distro is counter-intuitive as it steers the user towards a particular image.

Because this forces the user to know the images from a predefined list and this makes it difficult for beginners to use. I think we should be the ones to make this choice and thus make the use of the layer simpler for new users. For someone experienced, changing the image is trivial, but I don't believe it is like that for everyone.

vkraleti:
> > We could consider splitting the machines more evenly between the compile_warm_up & compile steps to reduce parallel load. Currently, the split is 2:7. Eventually, we can group 4–5 machines into a batch and run builds in multiple batches.
> 
> That would increase the amount of cache misses in the warm_up stage, so it's not logical

In that case warm_up stage can remain as. However, I believe the compile step should be split into multiple batches. Since the number of machines is expected to grow over time, reducing the job count from 35 to 28 (as proposed in this PR) may only be a temporary relief. If the additional 7 jobs are already becoming a bottleneck, we might face the same issue again soon.

ndechesne:
> > We could consider splitting the machines more evenly between the compile_warm_up & compile steps to reduce parallel load. Currently, the split is 2:7. Eventually, we can group 4–5 machines into a batch and run builds in multiple batches.
> 
> That would increase the amount of cache misses in the warm_up stage, so it's not logical

If we have a problem with the CI runners and the number of jobs we create we should fix our infrastructure instead of limiting ourselves. I have yet to see a problem related to the number of jobs we create.

.github/workflows/build-yocto.yml
```diff
@@ -144,11 +140,20 @@ jobs:
                 type: additional
                 dirname: "+linux-yocto-lts"
                 yamlfile: ":ci/linux-yocto-lts.yml"
+          - machine: qcom-armv7a
```
ndechesne, line 159: I think this is very convoluted and complicated.. 

in the build matrix you have MACHINE=qcom-armv7a and DISTRO={poky,qcom-distro}. then you create a new config for qcom-armv7a here, then you exclude a config later. I think it's hard to read today, and will be impossible to understand in a few days from now. 

Let's keep things simple and 'explicit'. If qcom-armv7a is 'special' then we remove it from the main matrix, and we add the specific config we want to build.

ci/qcom-distro.yml
```diff
@@ -35,3 +35,4 @@ local_conf_header:
 
 target:
   - qcom-multimedia-image
+  - qcom-multimedia-proprietary-image
```
ndechesne: that means that a meta-qcom user trying to build qcom-armv7a + qcom-distro will fail, right? if so, that's NAK.

.github/workflows/build-yocto.yml
lumag: Fixed

ci/qcom-distro.yml
lumag: Fixed

vkraleti (commented):
@lumag please rebase to resolve conflicts. We need this optimization to expand CI coverage.

PR #1109, opened 2025-10-01
ci: clean up sstate cache with a dedicated monthly job

ndechesne:
We use a 'monthly' persistent sstate cache folder, to limit the disk usage. Up until now, we were cleaning up the 'old' folder (2 months back), in the the 'compile' step.

The CI failed on 10/1, with errors like this one:

rm: cannot remove
'/efs/qli/meta-qcom/sstate-cache-2025-08/96/8d/sstate:libxfixes:armv8-2a-qcom-linux:6.0.1:r0:armv8-2a:14:968dbabfeff8c1be8f64efbc2780ddb4e4e4d16759c27cef155838329d7de706_create_spdx.tar.zst': Stale file handle

This is because several compile jobs were started , for each machine, and they stepped onto each other. To avoid that, I am moving the cleanup of sstate cache folder into a proper monthly 'cron' job, which will be executed only once.

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -30,8 +30,6 @@ runs:
       shell: bash
       run: |
         # use a monthly sstate cache folder
-        # keep current and last month to allow smooth end of month transition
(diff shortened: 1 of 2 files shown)


ricardosalveti:
Test failures not related to this change.

PR #1115, opened 2025-10-08
opencv: enable FastCV only for ARMv8 machines

lumag:
The FastCV is a binary module, which is not provided for 32-bit ARM CPUs running Linux. Make sure that this PACKAGECONFIG option is enabled only for ARMv8 Qualcomm machines.


dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_%.bbappend
```diff
@@ -1 +1,4 @@
-PACKAGECONFIG:append:qcom = " tests fastcv"
+PACKAGECONFIG:append:qcom = " tests"
+
+# Only on ARMv8 Qualcomm machines
+PACKAGECONFIG:append:qcom:armel64 = " fastcv"
```
ricardosalveti: aarch64?

ricardosalveti (changes requested):

dynamic-layers/openembedded-layer/recipes-support/opencv/opencv_%.bbappend
lumag: Indeed

ricardosalveti (approved):

PR #1117, opened 2025-10-08
iq-9075-evk: rename from qcs9075-iq-9075-evk

lumag:
Follow the discussion on GitHub and rename IQ-9075 machine to use just the commercial name for the board, dropping the (duplicating) SoC name from the machine name.


ci/iq-9075-evk.yml
```diff
@@ -5,4 +5,4 @@ header:
   includes:
   - ci/base.yml
 
-machine: qcs9075-iq-9075-evk
+machine: iq-9075-evk
```
mwasilew, line 8: Is this RB8? I will need to add alias to LAVA devices to support this job

vkraleti: Yes, this is RB8.

ricardosalveti: @mwasilew would the alias keep the old name functional as well? Just thinking how to land this change with the testing done as part of it.

mwasilew: yes, alias just adds another name for the same board. I can remove the other name later if we don't need it any more.

mwasilew: alias added.

PR #1119, opened 2025-10-10
QCS615 firmware

shoudil:
Consume ADSP/CDSP firmware and dsp libraries from upstream for QCS615


recipes-bsp/hexagon-dsp-binaries/hexagon-dsp-binaries_20250917.bb
```diff
@@ -36,6 +36,8 @@ do_install () {
 
 PACKAGES =+ "\
     ${PN}-qcom-db820c-adsp \
+    ${PN}-qcom-qcs615-ride-adsp \
```
lumag, line 45: Split this commit into two.

shoudil: > Split this commit into two.

You mean one commit for firmware, and another for hexagon dsp binaries ?

lumag: One for hexagon-dsp-binaries, another for the packagegroup

recipes-bsp/packagegroups/packagegroup-qcs615-adp-air.bb
```diff
@@ -4,11 +4,18 @@ inherit packagegroup
...
     ${@bb.utils.contains_any('DISTRO_FEATURES', 'opencl opengl vulkan', 'linux-firmware-qcom-adreno-a630 linux-firmware-qcom-qcs615-adreno', '', d)} \
     ${@bb.utils.contains('DISTRO_FEATURES', 'wifi', 'linux-firmware-ath11k-qca6698aq', '', d)} \
     ${@bb.utils.contains('DISTRO_FEATURES', 'bluetooth', 'linux-firmware-qca-qca6698', '', d)} \
+    linux-firmware-qcom-qcs615-audio \
+    linux-firmware-qcom-qcs615-compute \
     linux-firmware-qcom-venus-5.4 \
 "
+RRECOMMENDS:${PN}-hexagon-dsp-binaries = " \
+    hexagon-dsp-binaries-qcom-qcs615-ride-adsp \
+    hexagon-dsp-binaries-qcom-qcs615-ride-cdsp \
```
lumag, line 20: What is the relationship between ADP Air and Ride?

ricardosalveti: ADP air is also an automotive reference platform (SA6155), probably why ride is also used here.

lumag: If it is a different device, so it should not be reusing packages 1:1. Instead it needs its own set of packages.
@shoudil see how it was done for IQ platforms

shoudil: @lumag @ricardosalveti 
Here I use "ride" to refer the upstream hexagon dsp binaries https://github.com/linux-msm/hexagon-dsp-binaries/tree/trunk/qcs615/Qualcomm/QCS615-RIDE. I have confusion as well why QCS615-RIDE used here. the board we use is "ADP AIR".

Both ADP AIR and ADP RIDE are auto reference platform, as I know start from Auto Gen4 (such as Lemans/Monaco) we call "RIDE", for old platform we call "AIR". For QCS615 (Gen3 chip), both used maybe due to some history reason. the kernel device tree was named "ride" as well.
https://github.com/qualcomm-linux/kernel/blob/qcom-next/arch/arm64/boot/dts/qcom/qcs615-ride.dts

shoudil: We will raise a PR to rename https://github.com/linux-msm/hexagon-dsp-binaries/tree/trunk/qcs615/Qualcomm/QCS615-RIDE and then I update this PR accordingly.

lumag: Would you mind also fixing the kernel? At least we should fix the name inside the file.

shoudil: > Would you mind also fixing the kernel? At least we should fix the name inside the file
@quic-tingweiz  any comment for this ?

quic-tingweiz: @lumag could you clarify what should be fixed in Kernel?

lumag: @quic-tingweiz if it's ADP AIR, then [model](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/boot/dts/qcom/qcs615-ride.dts?h=v6.18-rc1#n13) should be fixed.

PR #1122, opened 2025-10-10
CI: enable ccache

### LINKED ISSUE #1051: Enable ccache on CI builds
The ccache can improve significantly our builds speeds when we can't take advantage of the bitbake sstate cache, in sutch scenario we need to compile the code. The cache storage needs to be available from the previous builds so we need to store these ccache artifacts temporay on our permanent storage.

We need to inherit the ccache bbclass globaly in the base.yml or creating a specific new one for that propose:
```
INHERIT += "ccache.bbclass"
```

For the native class targets we need to install the ccache on the host container and: 
https://git.yoctoproject.org/poky/commit/?id=1722c4162c6c0830fa86461dd626d5385765b02c
```
ASSUME_PROVIDED += "ccache-native"
HOSTTOOLS += "ccache"
```

For the recipes that doesn't like the ccache we can disable it explicit with the `CCACHE_DISABLE` local variable
https://git.yoctoproject.org/poky/commit/?id=530ceee676cacc238def5b3cb109f9a1f2686892

lumag:
In order to speed up builds, enable ccache.

Fixes #1051

### DIFF
--- a/recipes-bsp/images/initramfs-firmware-rb5-image.bb
+++ b/recipes-bsp/images/initramfs-firmware-rb5-image.bb
@@ -1,9 +1,6 @@
 DESCRIPTION = "Tiny ramdisk image with Qualcomm Robotics RB5 firmware files"
 
-# Do not install anything by default
-PACKAGE_INSTALL = ""
-
-PACKAGE_INSTALL:qcom-armv8a = " \
+PACKAGE_INSTALL += " \
     packagegroup-rb5-firmware \
 "
 


.github/actions/compile/action.yml
```diff
@@ -32,6 +32,7 @@ runs:
         # use a monthly sstate cache folder
         echo "DL_DIR=${{inputs.cache_dir}}/downloads" >> $GITHUB_ENV
         echo "SSTATE_DIR=${{inputs.cache_dir}}/sstate-cache-$(date '+%Y-%m')" >> $GITHUB_ENV
+        echo "CCACHE_TOP_DIR=${{inputs.cache_dir}}/ccache" >> ${GITHUB_ENV
```
ndechesne: what's the max size ccache will use?

ricardosalveti: Missing } (or remove initial { for GITHUB_ENV.

ricardosalveti: And we probably have to also think how often we will be cleaning this up, should we handle it the same way as sstate-cache?

lumag: It is handled by ccache itself, unless we drop packages from the build

lumag: >Missing } (or remove initial { for GITHUB_ENV.

Fixed

lumag: > what's the max size ccache will use?

The default is 5 GB for each package (as each package uses its own dir)

ci/base.yml
```diff
@@ -32,6 +32,7 @@ local_conf_header:
     CONF_VERSION = "2"
     INHERIT += "buildstats buildstats-summary"
     INHERIT += "buildhistory"
+    INHERIT += "ccache"
```
ricardosalveti, line 35: This will enable ccache for all users using our kas files, but it should be fine, the default ccache folder is under tmpdir.

ndechesne: is ccache always installed by yocto/oe instructions?

lumag: Yes, ccache-native is being built by OE.

quaresmajose: When we dont have the ccache instaled on the host/contianer the ccache-native will be build and used. I will propose installing it on oficcial upstream kas conteiner.

quaresmajose: And so we can also use it in all native recipes.

.github/actions/compile/action.yml
```diff
@@ -32,6 +32,7 @@ runs:
         # use a monthly sstate cache folder
         echo "DL_DIR=${{inputs.cache_dir}}/downloads" >> $GITHUB_ENV
         echo "SSTATE_DIR=${{inputs.cache_dir}}/sstate-cache-$(date '+%Y-%m')" >> $GITHUB_ENV
+        echo "CCACHE_TOP_DIR=${{inputs.cache_dir}}/ccache" >> $GITHUB_ENV
```
quaresmajose, line 35: This neeed to be exported to be used inside the container, maybe using the BB_ENV_PASSTHROUGH_ADDITIONS

quaresmajose: Perhaps the best thing would be to declare CCACHE_TOP_DIR as a global variable and pass it to kas-container using the BB_ENV_PASSTHROUGH_ADDITIONS

lumag: @quaresmajose I had a more weird idea. Stop using `GITHUB_ENV` for `DL_DIR` and `SSTATE_DIR` (and `CCACHE_TOP_DIR`)  and pass those vars through an additional kas fragment generated by this script. WDYT?

quaresmajose: I think this will not work because the paths are not the same inside and ouside the container and we still need some mechanism to bind mount it on the container.
The DL_DIR and SSTATE_DIR are special ones and the kas-container helper script bind mount it inside the container but was not the case for the CCACHE_TOP_DIR.
I need to play a bit with the CCACHE_TOP_DIR to see what is needed to make it available inside the container, afaik this is not supported in the kas-container helper.

PR #1118, opened 2025-10-09
selinux: Enable SELinux support and initial policy integration.

sasikuma-qti:

PR #1125, opened 2025-10-14
Add kernel config fragment support for linux-qcom-next recipe

rsiddoji:
Certain features like selinux, require specific configuration options
that are not present in the default kernel  configuration. Having 
kernel config fragment support from recipe helps to explicitly define
such configurations without modifying default configuration.

### DIFF
--- a/recipes-kernel/linux/linux-qcom-next_git.bb
+++ b/recipes-kernel/linux/linux-qcom-next_git.bb
@@ -4,7 +4,7 @@ DESCRIPTION = "Linux ${PV} kernel for QCOM devices"
 LICENSE = "GPL-2.0-only"
 LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
 
-inherit kernel
+inherit kernel cml1
 
 COMPATIBLE_MACHINE = "(qcom)"
 
@@ -41,5 +41,5 @@ do_configure:prepend() {
     cp ${S}/arch/${ARCH}/configs/${KBUILD_DEFCONFIG} ${B}/.config
 
     # Merge fragment for QCOM value add features
-    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${UNPACKDIR}/configs/qcom.cfg
+    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${@" ".join(find_cfgs(d))}
 }


recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,10 +36,25 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
+KERNEL_CONFIG_FRAGMENTS:append = " ${UNPACKDIR}/configs/qcom.cfg"
```
lumag: Why do you need an `:append` here? Just set it directly.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,10 +36,25 @@ S = "${UNPACKDIR}/${BP}"
...
+
 do_configure:prepend() {
     # Use a copy of the 'defconfig' from the actual repo to merge fragments
     cp ${S}/arch/${ARCH}/configs/${KBUILD_DEFCONFIG} ${B}/.config
 
-    # Merge fragment for QCOM value add features
-    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${UNPACKDIR}/configs/qcom.cfg
+    if [ ! -z "${KERNEL_CONFIG_FRAGMENTS}" ]
```
lumag: We _know_ that it won't be empty.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,10 +36,22 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
+KERNEL_CONFIG_FRAGMENTS = " ${UNPACKDIR}/configs/qcom.cfg"
```
lumag: No need for the whitespace

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,10 +36,22 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
+KERNEL_CONFIG_FRAGMENTS = "${UNPACKDIR}/configs/qcom.cfg"
```
ndechesne: it's not clear how we intend to use this new variable. especially because it's initialized without =?

sasikuma-qti: It will be used as below. For example, if we need to add selinux kernel config fragment :
**KERNEL_CONFIG_FRAGMENTS:append = " ${@bb.utils.contains('DISTRO_FEATURES', 'selinux', '${UNPACKDIR}/selinux.cfg', '', d)}"**

I checked this with "**bitbake -e linux-qcom-next**" to check the value of "KERNEL_CONFIG_FRAGMENTS".

It was **KERNEL_CONFIG_FRAGMENTS="/local/mnt/workspace/sasikuma/kodiak/meta-qcom-rfc2/build/tmp/work/qcs9100_ride_sx-poky-linux/linux-qcom-next/6.17+git/sources/configs/qcom.cfg /local/mnt/workspace/sasikuma/kodiak/meta-qcom-rfc2/build/tmp/work/qcs9100_ride_sx-poky-linux/linux-qcom-next/6.17+git/sources/selinux.cfg"** as expected

ndechesne: sure. but where would you add the 'append' line? Why? what's the use case?

it's typically better to initialize with =? if the variable is expected to be modified.

sasikuma-qti: Sure, we will consider you suggestion. 

We will be creating a dynamic layer for selinux in meta-qcom and then we will add selinux config fragment using **linux-qcom-next.bbappend** recipe.

Pull request for selinux enablement https://github.com/qualcomm-linux/meta-qcom/pull/1118 is under development. Waiting for this fragment support change to be merged to add our selinux kernel fragment.

Selinux fragment in upstream meta-selinux: https://git.yoctoproject.org/meta-selinux/tree/recipes-kernel/linux/files/selinux.cfg

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,10 +36,22 @@ S = "${UNPACKDIR}/${BP}"
...
+        if [ ! -e "$f" ]
+        then
+            echo "Could not find kernel config fragment $f"
+            exit 1
+        fi
+    done
+
     # Merge fragment for QCOM value add features
-    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${UNPACKDIR}/configs/qcom.cfg
+    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${KERNEL_CONFIG_FRAGMENTS}
```
ricardosalveti: Can't we automatically search and add all config fragments by default (.cfg files), as done by u-boot (https://git.openembedded.org/openembedded-core/tree/meta/recipes-bsp/u-boot/u-boot-configure.inc#n51)?

That way we won't have to maintain KERNEL_CONFIG_FRAGMENTS.

quaresmajose: This approach looks good and we can still control which (.cfg files) will be used with the `SRC_URI`, we can remove some from the list that we don't want with `:remove` and add with `:append`

ricardosalveti: The annoying part, comparing with other recipes which are also using config fragments (e.g. u-boot), is that the user would have to extend SRC_URI and also update the KERNEL_CONFIG_FRAGMENTS variable, which can be error prone (there might also be cases of conditionals, like only using a fragment when a certain layer is available).

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -4,10 +4,12 @@ DESCRIPTION = "Linux ${PV} kernel for QCOM devices"
 LICENSE = "GPL-2.0-only"
 LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
 
-inherit kernel
+inherit kernel cml1
 
 COMPATIBLE_MACHINE = "(qcom)"
 
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
```
quaresmajose: why is this necessary?

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -40,6 +42,7 @@ do_configure:prepend() {
     # Use a copy of the 'defconfig' from the actual repo to merge fragments
     cp ${S}/arch/${ARCH}/configs/${KBUILD_DEFCONFIG} ${B}/.config
 
+
     # Merge fragment for QCOM value add features
-    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${UNPACKDIR}/configs/qcom.cfg
+    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${@" ".join(find_cfgs(d))}
```
quaresmajose, line 44: Maybe will be better to save all the fragments on a variable local bash variable first and call the merge_config.sh only when there is something to be configured.

recipes-kernel/linux/linux-qcom-next_git.bb
siddojigari: We have qcom cfg in local path (https://github.com/qualcomm-linux/meta-qcom/tree/master/recipes-kernel/linux/linux-qcom-next)  will not get picked without this.

recipes-kernel/linux/linux-qcom-next_git.bb
siddojigari: so you want this to be  assigned to a variable and get this sanity checked before passed to merge_config ?

recipes-kernel/linux/linux-qcom-next_git.bb
quaresmajose: This is not necessary by default bitbake search the ${THISDIR}/${PN} when we have a recipe named ${PN}.bb on ${THISDIR}. You can delete it.

recipes-kernel/linux/linux-qcom-next_git.bb
quaresmajose: Yeah is what I am sugesting but if the configs/qcom.cfg will always be present, maybe it won't be necessary. let's leave it as you have.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -40,6 +42,7 @@ do_configure:prepend() {
     # Use a copy of the 'defconfig' from the actual repo to merge fragments
     cp ${S}/arch/${ARCH}/configs/${KBUILD_DEFCONFIG} ${B}/.config
 
+
```
quaresmajose: This extra line can also be removed.

lumag (approved):

siddojigari:
> 
Thanks for the  comments , I had updated the commit text as requested.

vkraleti:
db410c failures are not related to this PR. Merging manually.

PR #1126, opened 2025-10-14
ci: Add sota.yml for sota feature support

ysolanki11:
Add ci/sota.yml to enable sota integration in CI by including the meta-updater repository(branch: master) for uptane-based support and appending sota to
DISTRO_FEATURES in local_conf_header for proper build configuration.

### DIFF
--- a/ci/sota.yml
+++ b/ci/sota.yml
@@ -1,0 +1,13 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+
+repos:
+  meta-updater:
+    url: https://github.com/uptane/meta-updater
+    branch: master
+
+local_conf_header:
+  extra: |
+    DISTRO_FEATURES:append = " sota"


ricardosalveti:
> @ysolanki11 qcom-distro+linux-yocto-dev is failing

It is not related to this PR, we're investigating the issue.

sbanerjee-quic (approved):

lumag (changes requested):
- Please fix your Git config and write your name in an expected way
- Please describe how this fragment is supposed to be used / tested.

ricardosalveti:
We should at least add a CI configuration that would use this fragment.

ci/sota.yml
```diff
@@ -0,0 +1,13 @@
...
+  version: 14
+
+repos:
+  meta-updater:
+    url: https://github.com/uptane/meta-updater
+    branch: master
+
+local_conf_header:
+  extra: |
+    DISTRO_FEATURES:append = " sota"
```
ricardosalveti, line 13: Thinking again on this, we should instead have a sota-compatible qcom-distro in meta-qcom-distro instead of expanding DISTRO_FEATURES here, since we do want to support users and setups that are not specific to kas.

This kas fragment should really be to build qcom-distro with sota.

PR #1135, opened 2025-10-16
recipes-support: Add userspace-resource-manager recipe

jpagadal:
recipes-support: Add userspace-resource-manager recipe.

userspace-resource-manager is a lightweight daemon that monitors and dynamically regulates CPU, memory, and I/O usage of user-space processes. It leverages kernel interfaces like procfs, sysfs and cgroups to enforce runtime policies, ensuring system stability and performance in embedded and resource-constrained environments.

### DIFF
--- a/recipes-support/userspace-resource-manager/userspace-resource-manager_0.1.bb
+++ b/recipes-support/userspace-resource-manager/userspace-resource-manager_0.1.bb
@@ -1,0 +1,37 @@
+HOMEPAGE = "https://github.com/qualcomm/userspace-resource-manager"
+SUMMARY = "Userspace daemon for dynamic System resource management"
+DESCRIPTION = "Userspace Resource Manager(URM) is a lightweight userspace \
+daemon that monitors system resources and enforces policies using \
+Linux kernel interfaces such as cgroups and sysfs."
+
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+SRC_URI = "git://github.com/qualcomm/userspace-resource-manager.git;protocol=https;branch=main;tag=v${PV}"
+SRCREV = "2e4383baad53acbeb93ab63b409b8cb47f996e6d"
+
+inherit cmake pkgconfig systemd
+
+DEPENDS += "libyaml"
+
+PACKAGECONFIG ??= "\
+    ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'state-detector systemd', '', d)} \
+    tests \
+"
+
+PACKAGECONFIG[classifier] = "-DBUILD_CLASSIFIER=ON,-DBUILD_CLASSIFIER=OFF"
+PACKAGECONFIG[state-detector] = "-DBUILD_STATE_DETECTOR=ON,-DBUILD_STATE_DETECTOR=OFF"
+PACKAGECONFIG[systemd] = ",,systemd"
+PACKAGECONFIG[tests] = "-DBUILD_TESTS=ON,-DBUILD_TESTS=OFF"
+
+SYSTEMD_SERVICE:${PN} = "urm.service"
+FILES:${PN} += "${sysconfdir}/urm/*"
+
+PACKAGE_BEFORE_PN += "${PN}-tests"
+FILES:${PN}-tests += " \
+    ${sysconfdir}/urm/tests/* \
+    ${bindir}/RestuneComponentTests \
+    ${bindir}/RestuneIntegrationTests \
+    ${libdir}/libRestuneTestUtils.so* \
+    ${libdir}/libRestunePlugin.so* \
+"


recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
+SUMMARY = "resource-tuner"
+DESCRIPTION = "A lightweight userspace daemon for dynamic resource monitoring \
+               and policy enforcement using kernel interfaces like cgroups and sysfs."
+
+SRC_URI = "git://github.com/qualcomm/resource-tuner.git;protocol=https;branch=main"
```
lumag: Is it Qualcomm-specific or is it generic enough to work on other platforms? If it is generic, it should go to meta-oe so that other platforms can also benefit from it.

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+SRCREV = "26ff78ec039431538c5fa0c6b6fd30d9883fe486"
+S = "${UNPACKDIR}/${PN}-${PV}/Build"
```
lumag: "S" should point to the source location rather than the build location

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+
+DEPENDS += "libyaml"
+DEPENDS += "systemd"
+
+inherit cmake
+inherit pkgconfig
+inherit systemd
+
+EXTRA_OECMAKE += "-DCMAKE_C_FLAGS='-fdebug-prefix-map=${WORKDIR}=/usr/src/debug/resource-tuner'"
+EXTRA_OECMAKE += "-DCMAKE_CXX_FLAGS='-fdebug-prefix-map=${WORKDIR}=/usr/src/debug/resource-tuner'"
```
lumag: Why? This should be handled by OE

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+
+EXTRA_OECMAKE += "-DBUILD_SIGNALS=ON -DBUILD_CLI=ON -DBUILD_TESTS=ON -DBUILD_STATE_DETECTOR=ON"
```
lumag: PACKAGECONFIG?

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+
+EXTRA_OECMAKE += "-DCMAKE_VERBOSE_MAKEFILE=ON"
+EXTRA_OEMAKE += "VERBOSE=1"
```
lumag: Why?

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+
+EXTRA_OEMAKE += "CFLAGS='-g -O0 --coverage -fprofile-arcs -ftest-coverage'"
+EXTRA_OEMAKE += "LDFLAGS='-lgcov'"
```
lumag: This makes a debugging, coverage-enabled build. Why are you sending this for a production build?

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+
+EXTRA_OECMAKE += "-DCMAKE_INSTALL_SYSCONFDIR=/etc"
```
lumag: Should be handled by OE

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+
+do_install:append() {
+    install -m 0644 ${S}/../Services/resource-tuner.service -D ${D}${systemd_system_unitdir}/resource-tuner.service
```
lumag: Fix your build scripts in order to let 'make install' work.

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+    install -m 0644 ${S}/../Core/Configs/*.yaml ${D}/etc/resource-tuner/common/
+    install -m 0644 ${S}/../Signals/Configs/*.yaml ${D}/etc/resource-tuner/common/
+
+    install -d ${D}/etc/resource-tuner/custom
+    install -d ${D}/etc/resource-tuner/tests/Configs/ResourceSysFsNodes
+    install -m 0644 ${S}/../Tests/Configs/*.yaml ${D}/etc/resource-tuner/custom/
+    install -m 0644 ${S}/../Tests/Configs/ResourceSysFsNodes/*.txt ${D}/etc/resource-tuner/tests/Configs/ResourceSysFsNodes/
+}
+
+INSANE_SKIP:${PN} += "buildpaths dev-so dev"
```
lumag: That's insane. No.

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+FILES:${PN}-dev = ""
+FILES:${PN} += "${libdir}"
+FILES:${PN} += "/usr/include/**"
```
lumag: Definitely no

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+FILES:${PN} += "${systemd_system_unitdir}"
```
lumag: Should be handled by OE>

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+FILES:${PN} += "/etc/resource-tuner/**"
+SOLIBS = ".so"
+FILES_SOLIBSDEV = ""
```
lumag: Why?

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+SYSTEMD_AUTO_ENABLE:${PN} = "enable"
```
lumag: Does it differ from the default?

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
+SUMMARY = "resource-tuner"
+DESCRIPTION = "A lightweight userspace daemon for dynamic resource monitoring \
+               and policy enforcement using kernel interfaces like cgroups and sysfs."
```
lumag: Drop the whitespaces in the beginning of the line.

lumag (commented):
On top of that:
- Commit message is missing.
- Commit subject doesn't follow the established practice
- The PV is invalid, the relase is 0.1, not 1.0.0
- SRCREV doesn't point to the release commit.

koenkooi:
Please avoid using `+=` when a regular multiline variable works just fine. The `+=` operator adds a lot of parsing overhead than multiline variables don't.

recipes-support/urm/resource-tuner_1.0.0.bb
jpagadal: Currently we are targeting Qualcomm-specific SoCs.

lumag:
No stray merge commits, thank you.

recipes-support/urm/resource-tuner_1.0.0.bb
koenkooi: Also, `SUMMARY` should be a proper, albeit terse sentence, so at the very least you need to start it with a capital. Doing `SUMMARY=$PN` is completely unhelpful.

jpagadal: ACK, will update in next patch.

recipes-support/urm/resource-tuner_1.0.0.bb
jpagadal: ACK, will update in next patch.

recipes-support/urm/resource-tuner_1.0.0.bb
jpagadal: ACK, will update in next patch.

recipes-support/urm/resource-tuner_1.0.0.bb
jpagadal: ACK, will update in next patch.

recipes-support/urm/resource-tuner_1.0.0.bb
jpagadal: ACK

recipes-support/urm/resource-tuner_1.0.0.bb
jpagadal: Fixed.

recipes-support/urm/resource-tuner_1.0.0.bb
jpagadal: Done.

recipes-support/urm/resource-tuner_1.0.0.bb
jpagadal: ACK, will update in next patch.

recipes-support/urm/resource-tuner_1.0.0.bb
jpagadal: ACK.

recipes-support/urm/resource-tuner_1.0.0.bb
jpagadal: ACK.

lumag: Please stop spamming with acks. Just implement those.

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
+SUMMARY = "resource-tuner"
```
lumag: What is it? Please see other recipes.

recipes-support/urm/resource-tuner_1.0.0.bb
```diff
@@ -0,0 +1,54 @@
...
+DESCRIPTION = "A lightweight userspace daemon for dynamic resource monitoring \
+               and policy enforcement using kernel interfaces like cgroups and sysfs."
+
+SRC_URI = "git://github.com/qualcomm/resource-tuner.git;protocol=https;branch=main"
+SRCREV = "26ff78ec039431538c5fa0c6b6fd30d9883fe486"
```
lumag: This SRCREV doesn't belong to any of the tags. At the same time you've named the recipe `foo_1.0.0`, whcih means that you've packaged the `1.0.0` release of the tuner.

recipes-bsp/packagegroups/packagegroup-qcom.bb
```diff
@@ -15,6 +15,7 @@ RDEPENDS:${PN}-boot-essential = " \
     qrtr \
     rmtfs \
     tqftpserv \
+    resource-tuner \
```
lumag: It's definitely not boot essential. Not to mention that it should be a separate commit.

recipes-support/urm/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,40 @@
...
+SUMMARY = "Userspace daemon for dynamic resource management via cgroups"
+DESCRIPTION = "Resource Tuner is a lightweight userspace daemon that monitors system \
+resources and enforces policies using Linux kernel interfaces such as cgroups and sysfs."
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+SRC_URI = "git://github.com/qualcomm/resource-tuner.git;protocol=https;branch=main"
+SRCREV = "562e2e19472b6c5c51886a22869bc4c4973b4a6f"
+S = "${UNPACKDIR}/${PN}-${PV}"
```
lumag: Why do you need this?

recipes-support/urm/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,40 @@
...
+
+DEPENDS += "libyaml systemd"
+
+inherit cmake pkgconfig systemd
+
+EXTRA_OECMAKE += "${S}/Build"
```
lumag: Why?

recipes-support/urm/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,40 @@
...
+PACKAGECONFIG ??= "signals cli state-detector tests"
+PACKAGECONFIG[signals]        = "-DBUILD_SIGNALS=ON,-DBUILD_SIGNALS=OFF"
+PACKAGECONFIG[cli]            = "-DBUILD_CLI=ON,-DBUILD_CLI=OFF"
+PACKAGECONFIG[state-detector] = "-DBUILD_STATE_DETECTOR=ON,-DBUILD_STATE_DETECTOR=OFF"
+PACKAGECONFIG[tests]          = "-DBUILD_TESTS=ON,-DBUILD_TESTS=OFF"
+
+SYSTEMD_SERVICE:${PN} = "resource-tuner.service"
+
+do_install:append() {
+    install -d ${D}/etc/resource-tuner/common
```
lumag: Here and later: use `${sysconfdir}` instead of /etc.

recipes-support/urm/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,40 @@
...
+    install -m 0644 ${S}/Core/Configs/*.yaml ${D}/etc/resource-tuner/common/
+    install -m 0644 ${S}/Signals/Configs/*.yaml ${D}/etc/resource-tuner/common/
+
+    install -d ${D}/etc/resource-tuner/custom
+    install -d ${D}/etc/resource-tuner/tests/Configs/ResourceSysFsNodes
+    install -m 0644 ${S}/Tests/Configs/*.yaml ${D}/etc/resource-tuner/custom/
+    install -m 0644 ${S}/Tests/Configs/ResourceSysFsNodes/*.txt ${D}/etc/resource-tuner/tests/Configs/ResourceSysFsNodes/
```
lumag: Move all installation to `make install`

recipes-support/urm/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,40 @@
...
+}
+
+FILES:${PN} += "${libdir}"
```
lumag: Why?

recipes-support/urm/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,40 @@
...
+FILES:${PN} += "/etc/resource-tuner/**"
+SOLIBS = ".so"
```
lumag: Why?

recipes-bsp/packagegroups/packagegroup-qcom.bb
jpagadal: Agreed. Could you please suggest a better location for the resource-tuner package? 
I’m considering placing it in the overlay, but open to your recommendation.

recipes-support/resource-tuner/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,24 @@
...
+inherit cmake pkgconfig systemd
+
+PACKAGECONFIG ??= "signals cli state-detector tests"
+PACKAGECONFIG[signals]        = "-DBUILD_SIGNALS=ON,-DBUILD_SIGNALS=OFF"
+PACKAGECONFIG[cli]            = "-DBUILD_CLI=ON,-DBUILD_CLI=OFF"
+PACKAGECONFIG[state-detector] = "-DBUILD_STATE_DETECTOR=ON,-DBUILD_STATE_DETECTOR=OFF"
+PACKAGECONFIG[tests]          = "-DBUILD_TESTS=ON,-DBUILD_TESTS=OFF"
+
+SYSTEMD_SERVICE:${PN} = "resource-tuner.service"
+FILES:${PN} += "${sysconfdir}/resource-tuner/**"
```
koenkooi: Double asterisk

recipes-support/resource-tuner/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,24 @@
...
+SUMMARY = "Userspace daemon for dynamic System resource management"
+DESCRIPTION = "Resource Tuner is a lightweight userspace daemon that monitors system \
+resources and enforces policies using Linux kernel interfaces such as cgroups and sysfs."
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+SRC_URI = "git://github.com/qualcomm/resource-tuner.git;protocol=https;branch=main"
+SRCREV = "02645fc2e7d649e979f9b073b14ed35ca8af5ceb"
+OECMAKE_SOURCEPATH = "${S}/Build"
```
koenkooi: Either move `inherit cmake` up, or `OECMAKE_SOURCEPATH = `down. It is confusing to use variables before having inherited the class. The parser doesn't care, but humans do :)
We try to keep variables and methods roughly in the order they are used, especially variables from classes.

recipes-support/resource-tuner/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,24 @@
...
+
+DEPENDS += "libyaml systemd"
+
+inherit cmake pkgconfig systemd
+
+PACKAGECONFIG ??= "signals cli state-detector tests"
```
koenkooi: Nit: we try to keep lists sorted, swapping _signals_ and _cli_ would help with that. I've marked this as a nit since PACKAGECONFIG entries often benefit from being sorted by effect.

recipes-bsp/packagegroups/packagegroup-qcom.bb
koenkooi: For future reference: we try to keep such lists sorted alphabetically.

lumag: Add it in meta-qcom-distro to the images where it is required.

recipes-support/resource-tuner/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,25 @@
...
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+SRC_URI = "git://github.com/qualcomm/resource-tuner.git;protocol=https;branch=main"
+SRCREV = "02645fc2e7d649e979f9b073b14ed35ca8af5ceb"
+
+inherit cmake pkgconfig systemd
+
+OECMAKE_SOURCEPATH = "${S}/Build"
```
lumag: Why do you need this?

jpagadal: I have CMakeLists.txt file under Build directory not at the root of source - https://github.com/qualcomm/resource-tuner/tree/main/Build

recipes-support/resource-tuner/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,25 @@
...
+
+DEPENDS += "libyaml systemd"
+
+PACKAGECONFIG ??= "cli signals state-detector tests"
+PACKAGECONFIG[cli]            = "-DBUILD_CLI=ON,-DBUILD_CLI=OFF"
```
ricardosalveti: Please remove these extra whitespaces.

recipes-support/resource-tuner/resource-tuner_0.2.bb
lumag: Why is it not in the source root? Have you seen this in other, typical CMake projects?

jpagadal: Moved CMakeList file to source root. Thanks.

recipes-support/resource-tuner/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,23 @@
...
+DEPENDS += "libyaml systemd"
+
+PACKAGECONFIG ??= "cli signals state-detector tests"
+PACKAGECONFIG[cli] = "-DBUILD_CLI=ON,-DBUILD_CLI=OFF"
+PACKAGECONFIG[signals] = "-DBUILD_SIGNALS=ON,-DBUILD_SIGNALS=OFF"
+PACKAGECONFIG[state-detector] = "-DBUILD_STATE_DETECTOR=ON,-DBUILD_STATE_DETECTOR=OFF"
+PACKAGECONFIG[tests] = "-DBUILD_TESTS=ON,-DBUILD_TESTS=OFF"
+
+SYSTEMD_SERVICE:${PN} = "resource-tuner.service"
+FILES:${PN} += "${sysconfdir}/resource-tuner/*"
```
ricardosalveti: Do we need the files from /etc/resource-tuner/tests/Configs/ResourceSysFsNodes?

jpagadal: These are test config files; our plan is to use these in Lava or Axiom CI pipeline.

lumag: If they are not necessary for the normal work course, they should be moved to a separate package.

ricardosalveti: Yeah, please move them to a separated package.

jpagadal: Done. Addressed by disabling the tests switch from PACKAGECONFIG.

lumag: We asked to _move_ files to a separate package, not to disable their installation.

ricardosalveti: If you are planning to use these files then it should be enabled by default, and as asked, just have them in a separated package.

recipes-support/resource-tuner/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,23 @@
...
+
+DEPENDS += "libyaml systemd"
+
+PACKAGECONFIG ??= "cli signals state-detector"
+PACKAGECONFIG[cli] = "-DBUILD_CLI=ON,-DBUILD_CLI=OFF"
+PACKAGECONFIG[signals] = "-DBUILD_SIGNALS=ON,-DBUILD_SIGNALS=OFF"
+PACKAGECONFIG[state-detector] = "-DBUILD_STATE_DETECTOR=ON,-DBUILD_STATE_DETECTOR=OFF"
+PACKAGECONFIG[tests] = "-DBUILD_TESTS=ON,-DBUILD_TESTS=OFF"
+
+SYSTEMD_SERVICE:${PN} = "resource-tuner.service"
```
anujm1: If `SYSTEMD_SERVICE` is not set, I think the default behavior is for this to be set to `<package-name>.service` so this statement might not be needed.

jpagadal: It is not picking "resource-tuner.service" after adding tests package.
Without tests package, it works fine.

recipes-support/resource-tuner/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,29 @@
...
+
+DEPENDS += "libyaml systemd"
+
+PACKAGECONFIG ??= "cli signals state-detector tests"
+PACKAGECONFIG[cli] = "-DBUILD_CLI=ON,-DBUILD_CLI=OFF"
+PACKAGECONFIG[signals] = "-DBUILD_SIGNALS=ON,-DBUILD_SIGNALS=OFF"
+PACKAGECONFIG[state-detector] = "-DBUILD_STATE_DETECTOR=ON,-DBUILD_STATE_DETECTOR=OFF"
+PACKAGECONFIG[tests] = "-DBUILD_TESTS=ON,-DBUILD_TESTS=OFF"
+
+FILES:${PN} += "${sysconfdir}/resource-tuner/common/*"
```
lumag: This still should be a separate package, if these are only for tests.

recipes-support/resource-tuner/resource-tuner_0.2.bb
lumag: It's fine, keep the assignment please.

recipes-support/resource-tuner/resource-tuner_0.2.bb
jpagadal: These are intended for the main package because they are part of the core functionality required by the daemon. 
Custom and test configs are only needed for running test-cases, while in prod builds configs present in common directory are needed.

lumag: This line disagrees with you:

```
FILES:${PN}-tests += "${sysconfdir}/resource-tuner/custom/* \
```

recipes-support/resource-tuner/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,30 @@
+HOMEPAGE = "https://github.com/qualcomm/resource-tuner"
+SUMMARY = "Userspace daemon for dynamic System resource management"
+DESCRIPTION = "Resource Tuner is a lightweight userspace daemon that monitors system \
+resources and enforces policies using Linux kernel interfaces such as cgroups and sysfs."
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+SRC_URI = "git://github.com/qualcomm/resource-tuner.git;protocol=https;branch=main"
```
koenkooi: Can you add `;tag=v${PV}` to the SRC_URI?

koenkooi (changes requested):

recipes-support/resource-tuner/resource-tuner_0.2.bb
```diff
@@ -0,0 +1,30 @@
...
+DESCRIPTION = "Resource Tuner is a lightweight userspace daemon that monitors system \
+resources and enforces policies using Linux kernel interfaces such as cgroups and sysfs."
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+SRC_URI = "git://github.com/qualcomm/resource-tuner.git;protocol=https;branch=main;tag=v${PV}"
+SRCREV = "681a19ba3f6731f3acf1e1022fcd94298eecff82"
+
+inherit cmake pkgconfig systemd
```
ricardosalveti: Do you really have a strict dependency on systemd?

Can we make this optional and enable based on PACKAGECONFIG? You could check DISTRO_FEATURES and only enable systemd when it is used by the distro.

ricardosalveti:
```
ERROR: Nothing PROVIDES 'systemd' (but /work/build-yocto-check-layer-kAXB/meta-qcom/recipes-support/resource-tuner/resource-tuner_0.2.bb DEPENDS on or otherwise requires it)
systemd was skipped: using DISTRO 'nodistro', which is missing required DISTRO_FEATURES: 'usrmerge systemd'
```

This is because nodistro is used in yocto-check-layer and you are explicitly depending on systemd.

lumag:
> This is because nodistro is used in yocto-check-layer and you are explicitly depending on systemd.

I think I wrote this in another recipe review, but anyway: recipes MAY NOT depend on `systemd`.

recipes-support/resource-tuner/resource-tuner_0.2.bb
lumag: I'd say, it should be dropped. I think it's another case of systemd dependency just to get the system unit location.

anujm1: > I'd say, it should be dropped. I think it's another case of systemd dependency just to get the system unit location.

It seems to be using it for more than that: https://github.com/qualcomm/resource-tuner/blob/600e1ca306de1ddfa8b4a1c0a4676e65e71fa311/StateDetector.cpp#L5 

And, compilation of this is controlled by `state-detector` PACKAGECONFIG in this recipe so the explicit systemd dependency can probably go there.

jpagadal: We are using systemd to get the Display states (ON/OFF) in this module, we will control systemd dependency using DISTRO_FEATURES.

jpagadal (commented):
Renamed resource-tuner as userspace-resource-manager.

jpagadal:
@sbanerjee-quic, could you please clarify the reason for moving urm.service to the source repo? I’ve seen similar patterns in open-source projects like fastrpc, where service files are kept in recipes, so I’d like to understand the rationale.

recipes-support/userspace-resource-manager/files/urm.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
```
lumag: Why is it not a part of the git tree?

recipes-support/userspace-resource-manager/userspace-resource-manager_0.3.bb
```diff
@@ -0,0 +1,47 @@
...
+    git://github.com/qualcomm/userspace-resource-manager.git;protocol=https;branch=main;tag=v${PV} \
+    file://urm.service \
+"
+SRCREV = "d6d559bf824ef57888a58bf6a3311dfa488b537e"
+
+inherit cmake pkgconfig systemd
+
+DEPENDS += "\
+    libyaml \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'systemd', '', d)} \
```
lumag: Use `PACKAGECONFIG` for it.

lumag (changes requested):

recipes-support/userspace-resource-manager/files/urm.service
```diff
@@ -0,0 +1,10 @@
...
+Description=Userspace Resource Manager(URM) Service
+
+[Service]
+Restart=on-failure
+ExecStart=/usr/bin/urm
```
koenkooi: This hardcodes ${bindir}

recipes-support/userspace-resource-manager/files/urm.service
jpagadal: Moved to source git tree.

recipes-support/userspace-resource-manager/userspace-resource-manager_0.1.bb
```diff
@@ -0,0 +1,38 @@
...
+    ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'systemd', '', d)} \
+    tests \
+"
+
+PACKAGECONFIG[classifier] = "-DBUILD_CLASSIFIER=ON,-DBUILD_CLASSIFIER=OFF"
+PACKAGECONFIG[state-detector] = "-DBUILD_STATE_DETECTOR=ON,-DBUILD_STATE_DETECTOR=OFF"
+PACKAGECONFIG[systemd] = ",,systemd"
+PACKAGECONFIG[tests] = "-DBUILD_TESTS=ON,-DBUILD_TESTS=OFF"
+
+SYSTEMD_SERVICE:${PN} = "urm.service"
```
koenkooi, line 27: The ensure the unit works, you'll need something like this:

do_install_prepend() {
    sed -i s:/usr/bin/:${bindir}:g urm.service
}

Without this DISTROs that set bindir to e.g. /bin will fail.

jpagadal: Thanks for the input. I switched to do_install:append() because the service file isn’t installed yet during prepend, which caused sed to fail.

lumag: Why can't it be generated properly so that you don't have to patch it?

ricardosalveti: Yes, I would prefer this to be done automatically as part of the upstream project build process (have urm.service.in and generate the final service file based on the bindir value used).

jpagadal: Incorporated the changes at CMake level, service.in https://github.com/qualcomm/userspace-resource-manager/blob/main/urm.service.in

lumag (dismissed):

recipes-support/userspace-resource-manager/userspace-resource-manager_0.1.bb
```diff
@@ -0,0 +1,38 @@
...
+SRC_URI = "git://github.com/qualcomm/userspace-resource-manager.git;protocol=https;branch=main;tag=v${PV}"
+SRCREV = "2e4383baad53acbeb93ab63b409b8cb47f996e6d"
+
+inherit cmake pkgconfig systemd
+
+DEPENDS += "libyaml"
+
+PACKAGECONFIG ??= "\
+    ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'state-detector', '', d)} \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'systemd', '', d)} \
```
quaresmajose: This two lines can be join:
`${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'systemd state-detector', '', d)}`

jpagadal: Done.

lumag:
Please rebase on top of tip to solve build issues.

PR #1139, opened 2025-10-16
CI: unbreak building of images

lumag:
Apply several fixes in order to let CI build our images

### DIFF
--- a/recipes-kernel/iris-video-module/iris-video-dlkm/0001-video-driver-follow-v4l2_fh_add-_del-API-changes.patch
+++ b/recipes-kernel/iris-video-module/iris-video-dlkm/0001-video-driver-follow-v4l2_fh_add-_del-API-changes.patch
@@ -1,0 +1,189 @@
+From 8946d5d577d36faf45317057889036a236d9265f Mon Sep 17 00:00:00 2001
+From: Dmitry Baryshkov <dmitry.baryshkov@oss.qualcomm.com>
+Date: Mon, 20 Oct 2025 13:17:52 +0300
+Subject: [PATCH] video: driver: follow v4l2_fh_add() / _del() API changes
+
+Since Linux 6.18 functions v4l2_fh_add() and v4l2_fh_del() take care of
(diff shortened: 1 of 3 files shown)


koenkooi (approved):

koenkooi (approved):

koenkooi:
There's a typo in the commit header: 'mirros'

.github/actions/compile/action.yml
```diff
@@ -39,24 +39,24 @@ runs:
       run: |
         mkdir $KAS_WORK_DIR
         ${{inputs.kas}} dump --resolve-env --resolve-local --resolve-refs \
-          ci/mirror.yml:ci/${{ inputs.machine }}.yml${{ inputs.distro_yaml }}${{ inputs.kernel_yaml }} > kas-build.yml
```
quaresmajose, line 42: This is not the real issue otherwise yocto autobuilder master branch will fail in the same way as well because they also use the same sstate-cache.

To fix this type of issues we should invalidate the current sstate-cache of the refered package with HASHEQUIV_HASH_VERSION. In this case is the busybox. so something like this:

```
cat <<_EOF_ >> recipes-core/busybox/busybox_1.37.0.bb
# remove at next version upgrade or when output changes
HASHEQUIV_HASH_VERSION .= ".1"'
_EOF_
```

take this as ref: https://git.yoctoproject.org/poky/commit/?id=a320a45220ee4cbf527e13ba89382c9d13942d55

lumag: There is a difference. You have quoted the case where the recipe has changed. Here the global metadata was changed in an incompatible way with our builder picking up sstate produced against master-next. It's not a problem of a single package, SPDX data for all packages wasn't matching (with busybox being the first one to trigger the error).

quaresmajose: Yestarday I did some tests, using the mirror, and I couldn't reproduce the SPDX problem so it seems like a sporadic. Removing it completely is by no means the correct solution. Since it was a blocker for us I understand the solution but it must be temporary.

PR #1142, opened 2025-10-17
ci: base: workaround to temporary disable the libsolv in opkg 

quaresmajose:
The libsolv crashes the opkg in some scenarios not yet fully known.
The proper fix is ​​to disable the libsolv in opkg while looking for
the solution to the problem.

It was observed during debug testing that are still on going that
the issue does not exist when the libsolv is disabled in opkg.

For now it fixes the segfault [1]:
```
kernel: opkg[3077180]: segfault at 7fd957c88b28 ip 00007fd957e162f2 sp 00007fffb6068490 error 4 in libsolv.so.1[142f2,7fd957e0e000+7c000] likely on CPU 4 (core 8, socket 0)
```
[1] https://bugzilla.yoctoproject.org/show_bug.cgi?id=16010

### DIFF
--- a/recipes-bsp/images/initramfs-firmware-sm8650-hdk-image.bb
+++ b/recipes-bsp/images/initramfs-firmware-sm8650-hdk-image.bb
@@ -2,7 +2,6 @@ DESCRIPTION = "Tiny ramdisk image with SM8650 HDK devices firmware files"
 
 PACKAGE_INSTALL += " \
     packagegroup-sm8650-hdk-firmware \
-    packagegroup-sm8550-hdk-firmware \
 "
 
 BAD_RECOMMENDATIONS = " \


ci/base.yml
```diff
@@ -43,6 +43,9 @@ local_conf_header:
     EXTRA_IMAGE_FEATURES = "allow-empty-password empty-root-password allow-root-login"
     IMAGE_ROOTFS_EXTRA_SPACE = "307200"
     WATCHDOG_RUNTIME_SEC:pn-systemd = "30"
+  fixme: |
+    # the libsolv crashes the opkg in some scenarios not yet fully known
+    PACKAGECONFIG:remove:pn-opkg-native = "libsolv"
```
ricardosalveti, line 48: What is the implication of disabling libsolv here? It will use the internal solver, but not sure what are the side effects.

Also, this is changing the kas config, and the issue is outside the kas context, so it should be done at the opkg-native level directly.

quaresmajose: The internal resolver is deprecated and with this change we will get the warning:
```
The opkg internal solver will be deprecated in future opkg releases. Consider enabling 'libsolv' in PACKAGECONFIG.
```
The full impact I don't know but without this change this segmentfault can happen again at any time.

> Also, this is changing the kas config, and the issue is outside the kas context, so it should be done at the opkg-native level directly.

Do you mean do that with a bbappend to change the opkg-native?

ricardosalveti: Yes, better with a bbappend.

koenkooi: And that bbappend should not be in a BSP layer, but in a DISTRO layer.

quaresmajose: The reason I place it fisrt in the kas config is beacuse the workaround is mostly to fix possible issues in our CI builds until we continue searching the real solution to the problem. This will involve debuging the libsolv and it may take some time.
In the end, I think this shouldn't be in any of our layers because it doesn't make sense to condition our users to use opkg in this way.

quaresmajose:
I am getiing some `error: opkg_solver_install: Cannot install package` and so I put the PR in draft mode

quaresmajose:
We're really going to have to fix the libsolve and hope it doesn't happen again while we work on the fix.

PR #1144, opened 2025-10-21
qcs615-adp-air: update the default storage type

fangwu1:
Set the default storage variant to emmc for qcs615-adp-air

### DIFF
--- a/conf/machine/qcs615-adp-air.conf
+++ b/conf/machine/qcs615-adp-air.conf
@@ -18,4 +18,4 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
 
 QCOM_CDT_FILE = "cdt_adp_air_sa6155p"
 QCOM_BOOT_FILES_SUBDIR = "qcs615"
-QCOM_PARTITION_FILES_SUBDIR ?= "partitions/qcs615-adp-air/ufs"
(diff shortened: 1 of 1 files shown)


fangwu1:
Hi, @lumag , the commit message updated

lumag:
> Hi, @lumag , the commit message updated

What is 'QCLinux Nightly'? Also why was it set to ufs? Should there be a Fixes tag?

lumag:
Is there a UFS-based hardware? How was it tested before?

fangwu1:
On Talos, both UFS and eMMC configurations are supported, but eMMC has prioritized support. This is also the case in the downstream QCLinux, as shown in the following code section:
xref: /LE.QCLINUX.1.0/layers/meta-qti-bsp/conf/machine/qcs615-adp-air.conf
```
# Support emmc build with priority
MACHINEOVERRIDES .= ":emmc-storage"
```

how to use it
for EMMC: S3-4 OFF
image

for UFS: S3-4 ON

vkraleti:
@ricardosalveti yes, IQ6 Beta EVK has eMMC storage. IMO, we need to continue existing ADP-AIR with UFS and introduce a new machine conf for IQ6 with eMMC storage type.

ricardosalveti:
Yeah, then I think we're ok not adding a new machine configuration for emmc here, validating adp with ufs and the evk with emmc is enough already.

PR #1145, opened 2025-10-21
Revert "CI: stop using mirrors list"

quaresmajose:
This reverts commit 79d45251b5120999d26f68261be56b84377b2ca9.

Removing the mirror completely is by no means the correct solution.
The SPDX issue is identified and fixed [1], it was also necessary to
include the SPDX_LICENSES content on the task hash calculation.

[1] https://git.openembedded.org/openembedded-core/commit/?id=10669f6f615058293671fb16454601580b7b34e9

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -39,24 +39,24 @@ runs:
       run: |
         mkdir $KAS_WORK_DIR
         ${{inputs.kas}} dump --resolve-env --resolve-local --resolve-refs \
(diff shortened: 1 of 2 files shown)


quaresmajose:
I did a last change of the patch without testing before send it upstream and I made a mess.
Need to restest it again with the version I have validated before send it upstream to verify that it really works.

quaresmajose:
Done

koenkooi (approved):

ricardosalveti (approved):

PR #1148, opened 2025-10-21
CI/Fix broken login and missing boards

mwasilew:
Fix CI LAVA jobs:
 - add login details to qcom-distro jobs
 - bring back dragonboard jobs which were missing after last refactoring

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,15 +49,17 @@ jobs:
               if [ "${{ matrix.machine }}" = "qcom-armv8a" ]; then
                   python3 ci/generatetests.py --templates ci/lava --os ${{ matrix.distro.name }} --device dragonboard-410c --build-url "${BUILD_URL}"
                   python3 ci/generatetests.py --templates ci/lava --os ${{ matrix.distro.name }} --device dragonboard-820c --build-url "${BUILD_URL}"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -49,15 +49,17 @@ jobs:
...
                   python3 ci/generatetests.py --templates ci/lava --os ${{ matrix.distro.name }} --device ${{ matrix.machine }} --build-url "${BUILD_URL}"
+                  echo "MACHINE=${{ matrix.machine }}" >> $GITHUB_ENV
               fi
           fi
       - name: 'Upload test jobs'
         uses: actions/upload-artifact@v4
         with:
           name: testjobs-${{ matrix.machine }}-${{ matrix.distro.name }}
-          path: ${{ matrix.machine }}-${{ matrix.distro.name }}-*.yaml
+          path: ${{ env.MACHINE}}*-${{ matrix.distro.name }}-*.yaml
```
ricardosalveti, line 62: Do we need the extra '*' here?

mwasilew: it's to add support `dragonboard-410c` and `dragonboard-820c`. `MACHINE` variable is set to `dragonboard` as we're producing job templates for both in the same github action.

PR #1156, opened 2025-10-24
Update linux-qcom-next to tag qcom-next-6.18-rc1-20251022

vkraleti:
Move to the latest linux-qcom-next release tag qcom-next-6.18-rc1-20251022, which corresponds to kernel version 6.18-rc1.


lumag:
Broken on db820c.

PR #1157, opened 2025-10-24
README: Use shields.io to fix broken badges

ndechesne:
GitHub badges require authentication for projects that one is a member of. Members of the project who should really see these badges will have to login with GitHub before seeing these. Use shields.io as a badge proxying and caching service.

Largely based on https://github.com/qualcomm-linux/qcom-deb-images/pull/189

Suggested-by: Loïc Minier <loic.minier@oss.qualcomm.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # meta-qcom
 
-[![Build on push](https://github.com/qualcomm-linux/meta-qcom/actions/workflows/push.yml/badge.svg)](https://github.com/qualcomm-linux/meta-qcom/actions/workflows/push.yml)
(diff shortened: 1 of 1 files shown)


lool:
> What is broken with the current approach? I opened the repo in the incognito window and I can see all the badges.

This is when one is logged into GitHub but not into the extra SSO step to access Qualcomm repositories (Qualcomm SSO) which has to reauth every day.

There is a screenshot in https://github.com/qualcomm-linux/qcom-deb-images/pull/189 (review there appreciated too)

ricardosalveti (approved):

PR #1162, opened 2025-10-28
qcom-fastcv-binaries: FastCV DSP Support Enhancements

quic-ptak:
This PR introduces improvements to DSP-based FastCV functionality and streamlines deployment and validation processes.

### DIFF
--- a/recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
+++ b/recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
@@ -12,7 +12,7 @@ SRC_URI = "${ARTIFACTORY_URL}/${BPN}_${PV}_${PBT_ARCH}.tar.gz"
 SRC_URI[sha256sum] = "72accb95a1b29cd704db0b4a52e2e47b9905f80275599bc6b898a520a2663295"
 S = "${UNPACKDIR}"
 
-RDEPENDS:${PN} += "glib-2.0"
+DEPENDS += "glib-2.0 fastrpc"
 
 # This package is currently only used and tested on ARMv8 (aarch64) machines.
 # Therefore, builds for other architectures are not necessary and are explicitly excluded.
@@ -23,13 +23,40 @@ do_install() {
     install -d ${D}${libdir}/pkgconfig
     install -d ${D}${datadir}/doc/${PN}
     install -d ${D}${includedir}/fastcv
+    install -d ${D}${datadir}/qcom/qcs8300/Qualcomm/QCS8300-RIDE/dsp/cdsp
+    install -d ${D}${datadir}/qcom/sa8775p/Qualcomm/SA8775P-RIDE/dsp/cdsp
+    install -d ${D}${datadir}/qcom/qcm6490/Thundercomm/RB3gen2/dsp/cdsp
+
 
     install -m 0755 ${S}/usr/lib/libfastcvopt.so.1.8.0 ${D}${libdir}
-    cp -d ${S}/usr/lib/libfastcvopt.so* ${D}${libdir}
+    install -m 0755 ${S}/usr/lib/libfastcvdsp_stub.so.1.8.0 ${D}${libdir}
+    cp -d ${S}/usr/lib/libfastcvopt.so.1 ${D}${libdir}
+    cp -d ${S}/usr/lib/libfastcvopt.so ${D}${libdir}
+    cp -d ${S}/usr/lib/libfastcvdsp_stub.so.1 ${D}${libdir}
+    cp -d ${S}/usr/lib/libfastcvdsp_stub.so ${D}${libdir}
 
     install -m 0644 ${S}/usr/lib/pkgconfig/qcom-fastcv-binaries.pc ${D}${libdir}/pkgconfig/
     install -m 0644 ${S}/usr/share/doc/${PN}/NOTICE ${D}${datadir}/doc/${PN}
     install -m 0644 ${S}/usr/share/doc/${PN}/NOLOGINBINARYLICENSEQTI.pdf ${D}${datadir}/doc/${PN}
-    install -m 0644 ${S}/usr/include/fastcv/*.h ${D}${includedir}/fastcv/
+    install -m 0644 ${S}/usr/include/fastcv/fastcv.h ${D}${includedir}/fastcv/
+    install -m 0644 ${S}/usr/include/fastcv/fastcvExt.h ${D}${includedir}/fastcv/
+
+    install -m 0644 ${S}/usr/lib/dsp/cdsp/cv/v75/MONACO/*.so ${D}${datadir}/qcom/qcs8300/Qualcomm/QCS8300-RIDE/dsp/cdsp
+    install -m 0644 ${S}/usr/lib/dsp/cdsp/cv/v73/LEMANS/*.so ${D}${datadir}/qcom/sa8775p/Qualcomm/SA8775P-RIDE/dsp/cdsp
+    install -m 0644 ${S}/usr/lib/dsp/cdsp/cv/v68/KODIAK/*.so ${D}${datadir}/qcom/qcm6490/Thundercomm/RB3gen2/dsp/cdsp
 
 }
+
+PACKAGES += "\
+    ${PN}-qcs8300-ride-dsp \
+    ${PN}-sa8775p-ride-dsp \
+    ${PN}-thundercomm-rb3gen2-cdsp \
+"
+
+INSANE_SKIP:${PN}-qcs8300-ride-dsp = "arch libdir"
+INSANE_SKIP:${PN}-sa8775p-ride-dsp = "arch libdir"
+INSANE_SKIP:${PN}-thundercomm-rb3gen2-cdsp = "arch libdir"
+
+FILES:${PN}-qcs8300-ride-dsp += "${datadir}/qcom/qcs8300/Qualcomm/QCS8300-RIDE/dsp"
+FILES:${PN}-sa8775p-ride-dsp += "${datadir}/qcom/sa8775p/Qualcomm/SA8775P-RIDE/dsp"
+FILES:${PN}-thundercomm-rb3gen2-cdsp += "${datadir}/qcom/qcm6490/Thundercomm/RB3gen2/dsp"


recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
```diff
@@ -23,14 +23,51 @@ do_install() {
...
 }
+
+INSANE_SKIP:${PN} += " arch already-stripped"
+FILES:${PN} += "\
+    ${libdir}/libfastcvopt.so* \
+    ${libdir}/libfastcvdsp_stub.so* \
+    ${libdir}/dsp/cdsp/cv/v68/KODIAK/ \
+    ${libdir}/dsp/cdsp/cv/v69/KAILUA/ \
+    ${libdir}/dsp/cdsp/cv/v73/LEMANS/ \
+    ${libdir}/dsp/cdsp/cv/v75/MONACO/ \
```
lumag: Split per platform.

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
```diff
@@ -23,14 +23,51 @@ do_install() {
...
+    install -m 0755 ${S}/usr/lib/dsp/cdsp/cv/v68/KODIAK/libfastcvdsp_skel.so ${D}${libdir}/dsp/cdsp/cv/v68/KODIAK/
+    install -m 0755 ${S}/usr/lib/dsp/cdsp/cv/v69/KAILUA/libdspCV_skel.so ${D}${libdir}/dsp/cdsp/cv/v69/KAILUA/
+    install -m 0755 ${S}/usr/lib/dsp/cdsp/cv/v69/KAILUA/libfastcvadsp.so ${D}${libdir}/dsp/cdsp/cv/v69/KAILUA/
+    install -m 0755 ${S}/usr/lib/dsp/cdsp/cv/v69/KAILUA/libfastcvdsp_skel.so ${D}${libdir}/dsp/cdsp/cv/v69/KAILUA/
+    install -m 0755 ${S}/usr/lib/dsp/cdsp/cv/v73/LEMANS/libdspCV_skel.so ${D}${libdir}/dsp/cdsp/cv/v73/LEMANS/
+    install -m 0755 ${S}/usr/lib/dsp/cdsp/cv/v73/LEMANS/libfastcvadsp.so ${D}${libdir}/dsp/cdsp/cv/v73/LEMANS/
+    install -m 0755 ${S}/usr/lib/dsp/cdsp/cv/v73/LEMANS/libfastcvdsp_skel.so ${D}${libdir}/dsp/cdsp/cv/v73/LEMANS/
+    install -m 0755 ${S}/usr/lib/dsp/cdsp/cv/v75/MONACO/libdspCV_skel.so ${D}${libdir}/dsp/cdsp/cv/v75/MONACO/
+    install -m 0755 ${S}/usr/lib/dsp/cdsp/cv/v75/MONACO/libfastcvadsp.so ${D}${libdir}/dsp/cdsp/cv/v75/MONACO/
+    install -m 0755 ${S}/usr/lib/dsp/cdsp/cv/v75/MONACO/libfastcvdsp_skel.so ${D}${libdir}/dsp/cdsp/cv/v75/MONACO/
```
lumag: @basak-qcom what is your opinion here?
My 2c is that Hexagon binaries do not belong to `${libdir}` anyway.

quic-ptak: Hi, 
We are installing all hexagon libs in /usr/lib/dsp/cdsp/cv/<version>/<target> folders, hence used ${libdir}/dsp/cdsp/cv/. Hope that clarifies it

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
quic-ptak: Do you mean we need to segregate the install rules according to target ?

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
lumag: That's what I argue here. Hexagon binaries do not belong to `${libdir}`. This dir is reserved for the _host_ libraries, the ones to be loaded on the SoC. These binaries are binary data, so they belong to `${datadir}` aka `/usr/share/`.
How are these libraries handled? Does FastCV set `DSP_LIBRARY_PATH` manually?

lumag: Hit 'Comment' too quickly. Also, will those libraries work on fused devices? Or will fused devices require specially signed libraries?

My generic preference would be to coinstall these libraries to the same locations as those used by `hexagon-dsp-binaries`. Note meta-qcom and Debian will use slightly different paths there.

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
lumag: No, I mean to separate these platform-specific libraries into separate packages. Don't install data which is not going to be used by the platform.

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
basak-qcom: I agree with Dmitry on `${datadir}`. _Where_ within datadir I don't think we've concluded yet.

quic-ptak: Hi Dmitri,
 
FastCV sets DSP_LIBRARY_PATH dynamically based on target soc_ids. Currently, DSP skeleton libraries are released as part of the HLOS build, not within CDSP builds.
We follow the structure:  /usr/lib/dsp/cdsp/cv/<Hexagon_version>/<TARGET>/
 
Given this setup, could you suggest an appropriate and consistent location to host the DSP skeleton libraries, ensuring compatibility with FastCV’s path resolution logic and alignment with Yocto packaging best practices?

can we install the target based dsp skel libs in the below paths ?
/usr/share/qcom/cv/<Hexagon_version>/<TARGET>/

Also, there is a PR from fastrpc: [https://github.com/qualcomm/fastrpc/pull/234](url) which allows rpc to look for paths in /usr/share/qcom/../ paths. If we want to do it on similar lines, what is your suggestion on that ?

lumag: We are still in process of discussing. Could you please provide a feedback from your side:
- Is there a well-defined triplet for various hexagon versions?
- Do you depend on exact CDSP features (e.g. vector extensions, etc)?

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
```diff
@@ -4,15 +4,15 @@ LICENSE = "CLOSED"
 LIC_FILES_CHKSUM = "file://${UNPACKDIR}/usr/share/doc/${PN}/NOLOGINBINARYLICENSEQTI.pdf;md5=4ceffe94cb40cdce6d2f4fb93cc063d1 \
                     file://${UNPACKDIR}/usr/share/doc/${PN}/NOTICE;md5=4b722aa0574e24873e07b94e40b92e4d "
 
-PBT_BUILD_DATE = "250826"
+PBT_BUILD_DATE = "251015"
```
ricardosalveti, line 7: Why is the build date being updated but the version is still the same 1.8.0? We should at least reflect the revision in PV (or recipe name).

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
```diff
@@ -23,14 +23,51 @@ do_install() {
...
+    install -d ${D}${libdir}/dsp/cdsp
+    install -d ${D}${libdir}/dsp/cdsp/cv
+    install -d ${D}${libdir}/dsp/cdsp/cv/v68
+    install -d ${D}${libdir}/dsp/cdsp/cv/v68/KODIAK
+    install -d ${D}${libdir}/dsp/cdsp/cv/v69
+    install -d ${D}${libdir}/dsp/cdsp/cv/v69/KAILUA
+    install -d ${D}${libdir}/dsp/cdsp/cv/v73
+    install -d ${D}${libdir}/dsp/cdsp/cv/v73/LEMANS
+    install -d ${D}${libdir}/dsp/cdsp/cv/v75
+    install -d ${D}${libdir}/dsp/cdsp/cv/v75/MONACO
```
ricardosalveti: Can't the tarball contain a makefile with a install rule that just do the right thing (e.g. based on DESTDIR)?

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
ricardosalveti: Also, shouldn't these changes also be aligned with the 1.8.2 update? You modified a substantial part of the recipe based on 1.8.0 and later renamed it to 1.8.2.

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
quic-ptak: Hi dmitri, 

Can you please explain "triplet for various hexagon versions"? We have no idea on this. Are you referring to target triplet in hlos like aarch64-linux-gnu, aarch64-linux-android etc.?

lumag: Yes

lumag:
We had a brief discussion with the FastRPC team regarding these (and other Qualcomm-provided FastRPC libraries). They seem to be specific to the platform and, up to some point, to the device.

As such, I recommend installing these libraries into the `/usr/share/qcom/SoC/Vendor/Device/dsp/Ndsp` structure and letting FastRPC daemons to pick up them from the default paths. The library can override the `DSP_LIBRARY_PATH` on its own, but that won't be used inside meta-qcom.

sssanjee-quic:
Hi @lumag  ,  Have a small query , Since fastCV DSP skel libs are target dependent but they are board agnostic , are we allowed to install the libraries in /usr/share/qcom/SoC/**Vendor** folder i.e., target-wise as below ?
(for example : 
for qcs8300 target : dst path is /usr/share/qcom/**qcs8300**/Qualcomm/**<qcs8300 related fastcv dsp skel libs>**
for sa8775p target : dst path is /usr/share/qcom/**sa8775p**/Qualcomm/**<sa8775p related fastcv dsp skel libs>**
)

At runtime , we can append the above path to the DSP_LIBRARY_PATH from the fastcv library.

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
quic-ptak: done

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
```diff
@@ -4,12 +4,12 @@ LICENSE = "CLOSED"
 LIC_FILES_CHKSUM = "file://${UNPACKDIR}/usr/share/doc/${PN}/NOLOGINBINARYLICENSEQTI.pdf;md5=4ceffe94cb40cdce6d2f4fb93cc063d1 \
```
lumag, line 4: In the subject: `s/Bumped/bump`: imperative language, start with small letter, drop the full-stop: it's not a sentence.

Please provide a high-level changelog, "internal improvements" tells us nothing

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
```diff
@@ -25,12 +25,11 @@ do_install() {
     install -d ${D}${includedir}/fastcv
 
     install -m 0755 ${S}/usr/lib/libfastcvopt.so.1.8.0 ${D}${libdir}
-    cp -d ${S}/usr/lib/libfastcvopt.so.1 ${D}${libdir}
-    cp -d ${S}/usr/lib/libfastcvopt.so ${D}${libdir}
+    cp -d ${S}/usr/lib/libfastcvopt.so* ${D}${libdir}
```
lumag: Why?

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
```diff
@@ -1,3 +1,4 @@
+inherit pkgconfig
```
lumag: Out of place

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
```diff
@@ -12,24 +13,71 @@ SRC_URI = "${ARTIFACTORY_URL}/${BPN}_${PV}_${PBT_ARCH}.tar.gz"
...
+RDEPENDS:${PN} += "glib-2.0 fastrpc"
 
 # This package is currently only used and tested on ARMv8 (aarch64) machines.
 # Therefore, builds for other architectures are not necessary and are explicitly excluded.
 COMPATIBLE_MACHINE = "^$"
 COMPATIBLE_MACHINE:aarch64 = "(.*)"
 
+python () {
+    soc_id = d.getVar("SOC_FAMILY")
+    machine = d.getVar("MACHINE")
```
lumag: And what happens on `qcom-armv8a` machine?

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
```diff
@@ -12,24 +13,71 @@ SRC_URI = "${ARTIFACTORY_URL}/${BPN}_${PV}_${PBT_ARCH}.tar.gz"
...
 
     install -m 0644 ${S}/usr/lib/pkgconfig/qcom-fastcv-binaries.pc ${D}${libdir}/pkgconfig/
     install -m 0644 ${S}/usr/share/doc/${PN}/NOTICE ${D}${datadir}/doc/${PN}
     install -m 0644 ${S}/usr/share/doc/${PN}/NOLOGINBINARYLICENSEQTI.pdf ${D}${datadir}/doc/${PN}
     install -m 0644 ${S}/usr/include/fastcv/*.h ${D}${includedir}/fastcv/
 
+    install -d ${D}${datadir}/qcom/${SOC_ID}/Qualcomm/${BOARD}/dsp/cdsp
+    install -m 0644 ${S}/usr/lib/dsp/cdsp/cv/${HEXAGON_VERSION}/${TARGET}/libfastcvadsp.so ${D}${datadir}/qcom/${SOC_ID}/Qualcomm/${BOARD}/dsp/cdsp/
+    install -m 0644 ${S}/usr/lib/dsp/cdsp/cv/${HEXAGON_VERSION}/${TARGET}/libdspCV_skel.so ${D}${datadir}/qcom/${SOC_ID}/Qualcomm/${BOARD}/dsp/cdsp/
+    install -m 0644 ${S}/usr/lib/dsp/cdsp/cv/${HEXAGON_VERSION}/${TARGET}/libfastcvdsp_skel.so ${D}${datadir}/qcom/${SOC_ID}/Qualcomm/${BOARD}/dsp/cdsp/
```
lumag: Install here for all targets. Then split it to the per-target packages.

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
```diff
@@ -12,24 +13,71 @@ SRC_URI = "${ARTIFACTORY_URL}/${BPN}_${PV}_${PBT_ARCH}.tar.gz"
...
 }
+
+PACKAGES += "${PN}-dsp"
```
lumag: Will FastCV work without the DSP code?

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
Srinivas-Kandagatla: @quic-ptak @lumag AFAIU, these libraries should live in  <LIBRARY_PATH>/dsp/cdsp/<Hexagon_version>/  there should be no cv folder. And Am expecting other common libraries like qnn to be part of this path too, There is no point in having SoC or board specific path for these libraries, unless there is strong reason and not covered as part of Hexagon-Version.

lumag: @Srinivas-Kandagatla 
It's defintely should not be under `${libdir}`.
The question of having SoC is basically the question if any of the libraries is going to use something from the SoC-specific memory map.

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
quic-ptak: removed these soc nad machine based install rules

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
quic-ptak: removed it

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
quic-ptak: restored it

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
quic-ptak: done

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
quic-ptak: fastcv can work on cpu , but to run on dsp we need DSP skel libs installed on the device.

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
quic-ptak: done

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
Srinivas-Kandagatla: I did not exactly system libdir here but some dsp libdir whereever we decide to put them. I will be surprised to see if these generic libraries use any SoC specific memory map.
These libraries should be  PIC.

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
```diff
@@ -23,13 +23,49 @@ do_install() {
...
+"
+INSANE_SKIP:${PN}-iq9075-evk-dsp = "arch libdir"
+INSANE_SKIP:${PN}-iq8275-evk-dsp = "arch libdir"
+INSANE_SKIP:${PN}-qcs8300-ride-dsp = "arch libdir"
+INSANE_SKIP:${PN}-sa8775p-ride-dsp = "arch libdir"
+INSANE_SKIP:${PN}-thundercomm-rb3gen2-cdsp = "arch libdir"
+
+FILES:${PN}-iq9075-evk-dsp += "${datadir}/qcom/sa8775p/Qualcomm/IQ9075-EVK/dsp"
+FILES:${PN}-iq8275-evk-dsp += "${datadir}/qcom/qcs8300/Qualcomm/IQ8275-EVK/dsp"
+FILES:${PN}-qcs8300-ride-dsp += "${datadir}/qcom/qcs8300/Qualcomm/QCS8300-RIDE/dsp"
```
lumag, line 60: This will most likely break or misbehave once the user installs these two packages and `hexagon-dsp-binaries-qcom-iq8275-cdsp`.  Please install those only for qcs8300-ride and sa8775p-ride. The `hexagon-dsp-binaries-qcom-....` package will provide a compatibility symlink

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
```diff
@@ -12,7 +12,7 @@ SRC_URI = "${ARTIFACTORY_URL}/${BPN}_${PV}_${PBT_ARCH}.tar.gz"
 SRC_URI[sha256sum] = "72accb95a1b29cd704db0b4a52e2e47b9905f80275599bc6b898a520a2663295"
 S = "${UNPACKDIR}"
 
-RDEPENDS:${PN} += "glib-2.0"
+RDEPENDS:${PN} += "glib-2.0 fastrpc"
```
lumag: This should only be dependent by the dsp packages.

quic-ptak: fastrpc is a dependency of the HLOS library libfastcvdsp_stub.so, which enables DSP offloading of FastCV functions from the CPU. Therefore, it is included as a dependency in the main package. Target-specific FastCV packages are intended only for installing Hexagon-based DSP skeleton libraries.

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
quic-ptak: done

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
lumag: Libraries, yes. Does Abby of the libraries access hardware directly?

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
quic-ptak: edited the install rules,

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.2.bb
lumag: Is the daemon a dependency or is it a library? In the latter case, please add fastrpc to DEPENDS, letting OE pick up the correct library resolution rule on its own. In the former case we need to find a way to declare dependency on either of DSP daemons (fastrpc or hexagonrpcd).

quic-ptak: It is a library so added as dependency.

quic-ptak: In nutshell, fastcv lib: libfastcvdsp_stub.so has dependency on libcdsprpc.so which comes from fastrpc pkg.

lumag: Then, as I wrote, add fastrpc to `DEPENDS` (actually, I think both `fastrpc` and `glib-2.0` should be `DEPENDS`). Then let OE figure out library dependencies on its own.

lumag (approved):

recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.0.bb
lumag: Any of the libraries, of course.

lumag (approved):
`

ricardosalveti (approved):

PR #1165, opened 2025-10-29
enable CONFIG_SM_TCSRCC_8750 via qcom.cfg

vkraleti:
The TCSR clock controller is required during boot to provide the ref clocks to the UFS controller.
Setting CONFIG_SM_TCSRCC_8750 to y ensures the UFS driver successfully probe and initialize the device.


recipes-kernel/linux/linux-yocto-dev/configs/qcom.cfg
```diff
@@ -1,6 +1,7 @@
 # for UFS enumeration
 CONFIG_PHY_QCOM_QMP=y
 CONFIG_SCSI_UFS_QCOM=y
+CONFIG_SM_TCSRCC_8750=y
```
sbanerjeeqc, line 4: IMO a comment stating which all targets need this CONFIG will help. 
As of now this is required for sm8750 and sm8850.

vkraleti: I updated commit message to be a bit more explicit. Please check.

lumag (approved):

PR #1166, opened 2025-10-29
qcbor: library to expose Qualcomm Security Package

jiaxli-QC:
Add a recipe to build qcbor, library to work with security functionality on Qualcomm platforms.

The libqcbor, which are only available through meta-oe, is depended by Qcom-teec and MinkIPC. Thus add libqcbor recipe to the corresponding dynamic-layers subdir.

The qcbor library is used for decoding and encoding data structures.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-extended/qcbor/qcbor_1.5.3.bb
+++ b/dynamic-layers/openembedded-layer/recipes-extended/qcbor/qcbor_1.5.3.bb
@@ -1,0 +1,28 @@
+DESCRIPTION = " \
+    QCBOR is a powerful, commercial-quality CBOR encoder/decoder that \
+    implements these RFCs: RFC8949, RFC7049, RFC8742, RFC8943 \
+"
+
+HOMEPAGE = "https://github.com/laurencelundblade/QCBOR"
+
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=9abe2371333f4ab0e62402a486f308a5"
+
+SRC_URI = "git://github.com/laurencelundblade/QCBOR;protocol=https;branch=master;tag=v${PV}"
+
+SRCREV = "4ace4620d549f22c1163c5b00d3ae0c0dae1d207"
+
+
+inherit pkgconfig
+
+CFLAGS += " \
+    -DUSEFULBUF_DISABLE_ALL_FLOAT \
+"
+
+do_install(){
+    install -d ${D}${libdir}
+    install -m 755 ${S}/libqcbor.a ${D}${libdir}/
+    install -d ${D}${includedir}/qcbor
+    install -m 644 ${S}/inc/*.h ${D}${includedir}
+    install -m 644 ${S}/inc/qcbor/*.h ${D}${includedir}/qcbor
+}


recipes-security/qcbor/qcbor_git.bb
```diff
@@ -0,0 +1,29 @@
+SUMMARY = "Qualcomm QCBOR library"
+DESCRIPTION = " \
+QCBOR is a powerful, commercial-quality CBOR encoder-decoder. \
+The qcbor library is used for decoding and encoding data structures. \
+"
+HOMEPAGE = "https://github.com/laurencelundblade/QCBOR.git"
+SECTION = "libs"
+
+LICENSE = "BSD-3-Clause"
```
vksharma-oss: qcbor is using the BSD-3-clause license: 
ref: https://github.com/laurencelundblade/QCBOR?tab=License-1-ov-file

recipes-security/qcbor/qcbor_git.bb
```diff
@@ -0,0 +1,29 @@
...
+
+inherit cmake
+
+SRC_URI = "\
+    git://github.com/laurencelundblade/QCBOR.git;branch=master;protocol=https \
+"
+SRCREV = "7d9f0b787150d739ab50805008bc7142bc9b7822"
+PV = "0.0+git"
+
+EXTRA_OECMAKE = " -DBUILD_QCBOR_TEST=APP"
```
vksharma-oss: what is use of -DBUILD_QCBOR_TEST ? enabling this flag means you are generating a unit test client too along with library ?

recipes-security/qcbor/qcbor_git.bb
```diff
@@ -0,0 +1,29 @@
...
+
+do_install:append() {
+    install -d ${D}${bindir}
```
vksharma-oss: where the library is being installed ?

recipes-security/qcbor/qcbor_git.bb
```diff
@@ -0,0 +1,29 @@
...
+    install -m 0755 ${WORKDIR}/build/test/qcbortest ${D}${bindir}/
+}
+
+PACKAGES += "${PN}-test"
```
vksharma-oss: in which package the qcbor.so going ?

recipes-security/qcbor/qcbor_git.bb
lumag: Isn't it what is stated here?

recipes-security/qcbor/qcbor_git.bb
```diff
@@ -0,0 +1,29 @@
...
+    git://github.com/laurencelundblade/QCBOR.git;branch=master;protocol=https \
+"
+SRCREV = "7d9f0b787150d739ab50805008bc7142bc9b7822"
+PV = "0.0+git"
+
+EXTRA_OECMAKE = " -DBUILD_QCBOR_TEST=APP"
+
+do_install:append() {
+    install -d ${D}${bindir}
+    install -m 0755 ${WORKDIR}/build/test/qcbortest ${D}${bindir}/
```
lumag: Please patch the CMakeList to use `make install` instead. Send the patch upstream

recipes-security/qcbor/qcbor_git.bb
lumag: What goes to this package?

recipes-security/qcbor/qcbor_git.bb
vkraleti: If intention is to build test application along with the libs, use a PACKAGECONFIG check and use the same in installing the test bin as well as to add ${PN}-test to PACAKGES. Otherwise this recipe will end up with an empty test package and do_install failures when user wants to drop BUILD_QCBOR_TEST from OECMAKE.

recipes-security/qcbor/qcbor_git.bb
jiaxli-QC: BSD 3-Clause License

Redistribution and use in source and binary forms, with or without
modification, are permitted provided that the following conditions are met:

1. Redistributions of source code must retain the above copyright notice, this
   list of conditions and the following disclaimer.

2. Redistributions in binary form must reproduce the above copyright notice,
   this list of conditions and the following disclaimer in the documentation
   and/or other materials provided with the distribution.

3. Neither the name of the copyright holder nor the names of its
   contributors may be used to endorse or promote products derived from
   this software without specific prior written permission.

THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.

recipes-security/qcbor/qcbor_git.bb
jiaxli-QC: The library will be installed by `make install`

recipes-security/qcbor/qcbor_git.bb
jiaxli-QC: Hi @lumag , reviewers, what is the recommended way to include qcbor in the image? The build command `kas build meta-qcom/ci/qcs6490-rb3gen2-core-kit.yml` compiles the `core-image-base` target, which is defined in the OE-core layer.

recipes-security/qcbor/qcbor_git.bb
```diff
@@ -0,0 +1,29 @@
...
+}
+
+PACKAGES += "${PN}-test"
+
+FILES:${PN}-bin += " ${bindir}/qcbortest"
```
lumag: This package is not defined.

recipes-security/qcbor/qcbor_git.bb
lumag: The question of using PACKAGECONFIG is orthogonal. No, the user will not end up with an empty package, OE handles it internally. None of the recipes I know check PACKAGECONFIG when populating PACKAGES.

recipes-security/qcbor/qcbor_git.bb
```diff
@@ -0,0 +1,29 @@
...
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=9abe2371333f4ab0e62402a486f308a5"
+
+inherit cmake
+
+SRC_URI = "\
+    git://github.com/laurencelundblade/QCBOR.git;branch=master;protocol=https \
+"
+SRCREV = "7d9f0b787150d739ab50805008bc7142bc9b7822"
+PV = "0.0+git"
```
lumag: This is plain wrong. It is not 0.0. There was a 1.8.3 release. Please use the release instead of packaging random git commit (and lying about the version).

recipes-security/qcbor/qcbor_git.bb
jiaxli-QC: > If intention is to build test application along with the libs, use a PACKAGECONFIG check and use the same in installing the test bin as well as to add ${PN}-test to PACAKGES. Otherwise this recipe will end up with an empty test package and do_install failures when user wants to drop BUILD_QCBOR_TEST from OECMAKE.

Updated. PACKAGECONFIG is a boolean variable, so because we have three options (OFF / LIB / APP), we created a custom variable called QCBOR_TEST_MODE.

recipes-security/qcbor/qcbor_git.bb
```diff
@@ -0,0 +1,35 @@
...
+inherit cmake ptest
+
+SRC_URI = "\
+    git://github.com/laurencelundblade/QCBOR.git;branch=master;protocol=https \
+"
+SRCREV = "7d9f0b787150d739ab50805008bc7142bc9b7822"
+PV = "0.0+git"
+
+# QCBOR_TEST_MODE：OFF / LIB / APP
+QCBOR_TEST_MODE ?= "APP"
```
lumag: What does this mean? Why is it selectable?

recipes-security/qcbor/qcbor_git.bb
```diff
@@ -0,0 +1,35 @@
...
+
+EXTRA_OECMAKE += "-DBUILD_QCBOR_TEST=${QCBOR_TEST_MODE}"
+
+do_install:append() {
+    if [ "${QCBOR_TEST_MODE}" = "APP" ]; then
```
lumag: Ignoring review will bring you nowhere. Fold this into `make install`.

recipes-security/qcbor/qcbor_git.bb
lumag: Do you mean the library or the test utils? 
For the library you don't need to "include" it. Let it be pulled in by the dependencies. For the test utils, please include it into the test image. You had a nice idea of a security-testing packagegroup. I don't understand, why you dropped it .

recipes-security/qcbor/qcbor_git.bb
```diff
@@ -0,0 +1,35 @@
+SUMMARY = "Qualcomm QCBOR library"
```
ricardosalveti: Why Qualcomm? It is just the upstream project.

recipes-security/qcbor/qcbor_git.bb
jiaxli-QC: I indecated the test utils. How can we include the qcbortest into the .img file and flash it onto the device?

In this pull request, qcbor is not required as a dependency, so the packagegroup-secure package has been removed to enable faster code merging, we will add this package in the Minkipc pull request.

jiaxli-QC: Updated.

`FILES:${PN}-ptest += " ${bindir}/qcbortest"`

recipes-security/qcbor/qcbor_git.bb
jiaxli-QC: Updated.
`FILES:${PN}-ptest += " ${bindir}/qcbortest"`

recipes-security/qcbor/qcbor_git.bb
jiaxli-QC: Alright, Have updated the version as 1.5.3.
ref: https://github.com/laurencelundblade/QCBOR/releases
`PV = "1.5.3+git"`

ricardosalveti:
> Being a BSP layer, meta-qcom is expected to be built standalone using oe-core without dependency on any other layers like meta-oe. If QTEE has dependencies on recipes from meta-oe, may be it can be added to [dynamic-layers](https://github.com/qualcomm-linux/meta-qcom/tree/master/dynamic-layers/openembedded-layer) section to conditionally compile.

Right, let's use it via dynamic-layers for now until we can validate the workflow and the builds, later on we can either propose for oe-core (harder in this scenario as we should have at least a common dependency there), or bring the recipe as part of meta-qcom but syncing it with the recipe available under meta-oe.

What I really want to avoid is multiple versions and variants of the same recipe in multiple layers, because meta-oe is also available with meta-qcom-distro.

lumag:
I think Ricardo meant to drop QCBOR from this layer and use qcbor dynamically from meta-oe.

ricardosalveti:
> Hi @lumag , if I use qcbor dynamically from meta-oe, could you share the compilation command? The README only mentions the `kas build meta-qcom/ci/qcs6490-rb3gen2-core-kit.yml` command. Many thanks.

Meta-oe is added when building with meta-qcom-distro, so you will have to use `kas build ci/qcs6490-rb3gen2-core-kit.yml:ci/qcom-distro.yml`.

jiaxli-QC:
> > Hi @lumag , if I use qcbor dynamically from meta-oe, could you share the compilation command? The README only mentions the `kas build meta-qcom/ci/qcs6490-rb3gen2-core-kit.yml` command. Many thanks.
> 
> Meta-oe is added when building with meta-qcom-distro, so you will have to use `kas build ci/qcs6490-rb3gen2-core-kit.yml:ci/qcom-distro.yml`.

How can we include minkipc and qcomtee in the image for validation purposes?

jiaxli-QC:
> 
> This is inline with expectations. Security features are typically tied to distributions and are not default offerings in BSP layers. For example, features like SELinux and kernel module signing are enabled through distribution-level configurations but not by default in BSP.
> 
> Similarly for this case, you need to :
> 
> 1. Remove the qcbor recipe.
> 2. Add the qcomtee and minkipc recipes under dynamic-layers in meta-qcom.
>    To use these, include the meta-qcom-distro layer while building. This will ensure that dependencies from meta-oe are correctly pulled in.

Hi @vkraleti @lumag , why we should move the minkipc and qcbor in dynamic-layer? This two repo is owned by qcom, can we create the `recipes-security` in the same directory as `recipes-bsp` directory. 
Because in the future we will need to manage a whole set of security-related components. Furthermore, in Mete-oe, `recipes-security` exists independently of `recipes-bsp` and `recipes-core`.

vkraleti:
> > Meta-oe is added when building with meta-qcom-distro, so you will have to use `kas build ci/qcs6490-rb3gen2-core-kit.yml:ci/qcom-distro.yml`.
> 
> Have we confirmed the build target? I'm wondering why the device isn't booting up after flashing the build with no modifications.

Yes, multimedia-image built with 'qcom-distro' is booting fine on rb3gen2-core-kit. This combination is in CI as well.

PR #1170, opened 2025-10-30
conf/machine: rename qcs615-adp-air to qcs615-ride

lumag:
According to what I understood from the explanations, the ADP AIR is an older name for the Automotive Development Platform devices, all of which were later renamed to RIDE. Rename the qcs615-adp-air to qcs615-ride in order to unify this machine and other ex-automotive devices.


lumag:
Note: I'm not sure if it should be just ride or ride-sx.

mwasilew:
All other RIDE platforms are named `ride-sx`. I don't know what `sx` stands for. The physical board looks different than other `ride-sx` boards.

koenkooi (approved):

ricardosalveti (approved):

PR #1172, opened 2025-10-30
Rb3gen2: Add support for Open Boot firmware (TF-A, OP-TEE and U-Boot) build

b49020:
Currently the open boot firmware stack has only been enabled on RB3Gen2 platform which can be built using following:

```
$ kas build ci/rb3gen2-core-kit-open-fw.yml:ci/meta-arm.yml
```

And currently the boot stack includes TF-A, OP-TEE and U-Boot. In future the plan is to enable upstream edk2 as well. Along with that the next target for open boot firmware is Lemans based IoT EVK platform.

Right now the build generates 2 firmware payloads as bl2.elf (unsigned) and fip.elf (test signed using qtestsign). It is required to sign bl2.elf with QTI signature using sectools but in future the plan is to drop QTI signature requirement with an updated release of XBL/XBL-SEC.

Once signing is done, one need to update following binaries in qcomflash tarball and then proceed with QDL flashing:
- tz.mbn -> bl2.mbn
- uefi.elf -> fip.elf

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -235,6 +235,14 @@ jobs:
                 type: u-boot-qcom-6.18
                 dirname: "_linux-qcom-6.18_u-boot-qcom"
                 yamlfile: ":ci/linux-qcom-6.18.yml:ci/u-boot-qcom.yml"
+          - machine: rb3gen2-core-kit-open-fw
+            distro:
+                name: qcom-distro-kvm
+                yamlfile: ':ci/qcom-distro-kvm.yml'
+            kernel:
+                type: default
+                dirname: ""
+                yamlfile: ""
     name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
     steps:
       - uses: actions/checkout@v4
--- a/ci/rb3gen2-core-kit-open-fw.yml
+++ b/ci/rb3gen2-core-kit-open-fw.yml
@@ -1,0 +1,9 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+  includes:
+  - ci/base.yml
+  - ci/meta-arm.yml
+
+machine: rb3gen2-core-kit-open-fw
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -10,16 +10,18 @@ BBFILE_PATTERN_qcom := "^${LAYERDIR}/"
 BBFILE_PRIORITY_qcom = "6"
 
 LAYERDEPENDS_qcom = "core"
-LAYERRECOMMENDS_qcom = "openembedded-layer"
+LAYERRECOMMENDS_qcom = "openembedded-layer meta-arm"
 LAYERSERIES_COMPAT_qcom = "whinlatter"
 
 LICENSE_PATH += "${LAYERDIR}/licenses"
 
 BBFILES_DYNAMIC += " \
-    openembedded-layer:${LAYERDIR}/dynamic-layers/openembedded-layer/*/*/*.bb \
-    openembedded-layer:${LAYERDIR}/dynamic-layers/openembedded-layer/*/*/*.bbappend \
+    meta-arm:${LAYERDIR}/dynamic-layers/meta-arm/*/*/*.bb \
+    meta-arm:${LAYERDIR}/dynamic-layers/meta-arm/*/*/*.bbappend \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bb \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bbappend \
+    openembedded-layer:${LAYERDIR}/dynamic-layers/openembedded-layer/*/*/*.bb \
+    openembedded-layer:${LAYERDIR}/dynamic-layers/openembedded-layer/*/*/*.bbappend \
     qcom-distro:${LAYERDIR}/dynamic-layers/qcom-distro/*/*/*.bb \
     qcom-distro:${LAYERDIR}/dynamic-layers/qcom-distro/*/*/*.bbappend \
     selinux:${LAYERDIR}/dynamic-layers/selinux/*/*/*.bb \
--- a/conf/machine/rb3gen2-core-kit-open-fw.conf
+++ b/conf/machine/rb3gen2-core-kit-open-fw.conf
@@ -1,0 +1,13 @@
+#@TYPE: Machine
+#@NAME: Qualcomm RB3Gen2 Development Kit (Core Kit with open boot firmware)
+#@DESCRIPTION: Machine configuration for Qualcomm RB3Gen2 development Kit, with QCS6490
+
+require conf/machine/rb3gen2-core-kit.conf
+
+EXTRA_IMAGEDEPENDS += "trusted-firmware-a-qcom"
+
+MACHINE_FEATURES += "optee"
+MACHINE_EXTRA_RRECOMMENDS += "${@bb.utils.contains('BBFILE_COLLECTIONS', 'meta-arm', 'packagegroup-optee', '', d)}"
+
+UBOOT_CONFIG = "qcs6490-rb3gen2"
+UBOOT_CONFIG[qcs6490-rb3gen2] = "qcm6490_defconfig"
--- a/dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom_git.bb
+++ b/dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom_git.bb
@@ -1,0 +1,8 @@
+require recipes-bsp/trusted-firmware-a/trusted-firmware-a.inc
+
+PV = "2.14.0-qcom+git"
+
+SRC_URI = "git://github.com/qualcomm-linux/trusted-firmware-a.git;protocol=https;name=tfa;branch=qcom-next"
+SRCREV_tfa = "78d883ea408d58786287102bc704a5bfce2cbd90"
+
+require trusted-firmware-a-qcom.inc
--- a/dynamic-layers/meta-arm/recipes-security/optee/optee-os-qcom_git.bb
+++ b/dynamic-layers/meta-arm/recipes-security/optee/optee-os-qcom_git.bb
@@ -1,0 +1,8 @@
+require recipes-security/optee/optee-os.inc
+
+PV = "4.9.0-qcom+git"
+
+SRC_URI = "git://github.com/qualcomm-linux/optee_os.git;protocol=https;name=optee;branch=qcom-next"
+SRCREV_optee = "c2b0684fcd89929976a8726e6e3af922b48dd2c7"
+
+require optee-qcom.inc
--- a/dynamic-layers/meta-arm/recipes-security/packagegroups/packagegroup-optee.bb
+++ b/dynamic-layers/meta-arm/recipes-security/packagegroups/packagegroup-optee.bb
@@ -1,0 +1,7 @@
+SUMMARY = "Packages for the OP-TEE support"
+
+inherit packagegroup
+
+RRECOMMENDS:${PN} = " \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'optee-os-tadevkit optee-client optee-test', '', d)} \
+"
--- a/ci/meta-arm.yml
+++ b/ci/meta-arm.yml
@@ -1,0 +1,9 @@
+header:
+  version: 14
+
+repos:
+  meta-arm:
+    url: https://git.yoctoproject.org/meta-arm
+    layers:
+      meta-arm:
+      meta-arm-toolchain:
--- a/dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
+++ b/dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
@@ -1,0 +1,33 @@
+DEPENDS += "qtestsign-native optee-os-qcom"
+
+SRC_URI += "git://github.com/coreboot/qc_blobs.git;protocol=https;name=qc_blobs;subdir=qc_blobs;branch=main"
+SRCREV_qc_blobs = "16207f367ddbf280a0c2c8a3d3ee454a59710a25"
+
+LICENSE += "& LICENSE.qcom"
+LIC_FILES_CHKSUM:qcm6490 += "file://${UNPACKDIR}/qc_blobs/sc7280/qtiseclib/LICENSE;md5=fa83f30385e617b56ef0934f13645621"
+
+COMPATIBLE_MACHINE:qcm6490 = "qcm6490"
+TFA_PLATFORM:qcm6490 = "rb3gen2"
+
+TFA_UBOOT ?= "1"
+TFA_BUILD_TARGET = "bl2 fip"
+TFA_SPD = "opteed"
+
+EXTRA_OEMAKE:append = " \
+    BL32=${RECIPE_SYSROOT}/${nonarch_base_libdir}/firmware/tee-raw.bin \
+   "
+
+EXTRA_OEMAKE:append:qcm6490 = " \
+    QTISECLIB_PATH=${UNPACKDIR}/qc_blobs/sc7280/qtiseclib/libqtisec.a \
+   "
+
+do_install:append:qcm6490() {
+    export CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1
+
+    ${OBJCOPY} -I binary -B aarch64 -O elf64-littleaarch64 ${D}${FIRMWARE_DIR}/fip.bin ${D}${FIRMWARE_DIR}/fip.o
+    ${LD} ${D}${FIRMWARE_DIR}/fip.o -o ${D}${FIRMWARE_DIR}/fip_unsigned.elf -EL -T ${S}/tools/qti/fip-elf.lds --defsym=ELFENTRY=0x9fc00000 -Ttext=0x9fc00000
+    rm -f ${D}${FIRMWARE_DIR}/fip.o
+
+    qtestsign -v6 aboot -o ${D}${FIRMWARE_DIR}/fip.elf ${D}${FIRMWARE_DIR}/fip_unsigned.elf
+    rm -f ${D}${FIRMWARE_DIR}/fip_unsigned.elf
+}
--- a/dynamic-layers/meta-arm/recipes-security/optee/optee-os-tadevkit_%.bbappend
+++ b/dynamic-layers/meta-arm/recipes-security/optee/optee-os-tadevkit_%.bbappend
@@ -1,0 +1,4 @@
+MACHINE_OPTEE_REQUIRE ?= ""
+MACHINE_OPTEE_REQUIRE:qcom = "optee-qcom.inc"
+
+require ${MACHINE_OPTEE_REQUIRE}
--- a/dynamic-layers/meta-arm/recipes-security/optee/optee-qcom.inc
+++ b/dynamic-layers/meta-arm/recipes-security/optee/optee-qcom.inc
@@ -1,0 +1,2 @@
+COMPATIBLE_MACHINE:qcm6490 = "qcm6490"
+OPTEEMACHINE:qcm6490 = "qcom-kodiak"
--- a/dynamic-layers/meta-arm/recipes-security/optee/optee-test_%.bbappend
+++ b/dynamic-layers/meta-arm/recipes-security/optee/optee-test_%.bbappend
@@ -1,0 +1,6 @@
+# Machine specific configurations
+
+MACHINE_OPTEE_REQUIRE ?= ""
+MACHINE_OPTEE_REQUIRE:qcom = "optee-qcom.inc"
+
+require ${MACHINE_OPTEE_REQUIRE}


koenkooi:
> Any ideas how can I fix the yocto run checks here? I have already added `meta-arm` as dependency for `meta-qcom` and added `meta-arm` repo to be part of repos in the KAS `base.yml` file. However, the build still says:
> 
> ```
> ERROR: Layer meta-qcom depends on meta-arm and isn't found.
> ```

I'm not sure if ordering matters in KAS, but could you try changing the repo section to this?
```yaml
  meta-arm:
    url: https://git.yoctoproject.org/git/meta-arm
    layers:
      meta-arm-toolchain:
      meta-arm:
```

That's the only thing that stood out to me as a difference from a working config with meta-arm included.

ci/open-boot-firmware.yml
```diff
@@ -0,0 +1,12 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+
+local_conf_header:
+  firmware: |
+    EXTRA_IMAGEDEPENDS += "trusted-firmware-a"
```
ndechesne: these configs seem to qualify to be a MACHINE FEATURE, so that they can be set per machine instead of globally. and then we can define an 'open' RB3g2?

b49020: That sounds like a good idea, I will work on machine features instead.

recipes-security/optee/optee-test-qcm6490.inc
```diff
@@ -0,0 +1,3 @@
+# qcm6490 specific configuration
+
+COMPATIBLE_MACHINE = "qcm6490"
```
ricardosalveti: Do we need a compatible machine for optee-test? It should be generic.

ci/open-boot-firmware.yml
```diff
@@ -0,0 +1,12 @@
...
+    PREFERRED_PROVIDER_virtual/bootloader = "u-boot"
+    TFA_UBOOT:pn-trusted-firmware-a = "1"
+  optee: |
+    IMAGE_INSTALL:append = " optee-test optee-client optee-os-ta"
```
ricardosalveti: We can probably define a packagegroup for these packages and include when optee is available in machine features.

recipes-bsp/trusted-firmware-a/trusted-firmware-a_%.bbappend
quaresmajose, line 1: This will break compatible with all other layers, because you are changing all trusted-firmware-a recipes in other layers if they exist. It will be better to make our changes only when the `qcom` override is in place, something like this:
```
TRUSTED_FIRMWARE_A_QCOM = ""
TRUSTED_FIRMWARE_A_QCOM:qcom = "trusted-firmware-a-qcom.inc"

require ${TRUSTED_FIRMWARE_A_QCOM}
```

and move everything from this bbappend to trusted-firmware-a-qcom.inc

recipes-security/optee/optee-os-tadevkit_%.bbappend
quaresmajose, line 1: Same here

recipes-security/optee/optee-os_%.bbappend
quaresmajose, line 1: Same here

recipes-security/optee/optee-test-qcm6490.inc
b49020: `meta-arm` does enforce the `COMPATIBLE_MACHINE` here since all the TAs being built in `optee-test` have a dependency on `optee-os`. It will fail otherwise to build.

ci/open-boot-firmware.yml
b49020: Sure, I can do that.

recipes-bsp/trusted-firmware-a/trusted-firmware-a_%.bbappend
b49020: Sounds reasonable, I will update accordingly.

recipes-security/optee/optee-test-qcm6490.inc
ricardosalveti: Should have a dependency on optee-os-tadevkit, which should also be generic, but I see that everything ended up being machine specific in meta-arm, so we can keep this way indeed.

b49020:
PR updated to incorporate review comments.

ci/yocto-check-layer.sh
```diff
@@ -34,7 +34,7 @@ CMD="$CMD meta-qcom"
 # Disable auto layer discovery
 CMD="$CMD --no-auto"
 # Layers to process for dependencies
-CMD="$CMD --dependency $WORK_DIR/oe-core/meta"
+CMD="$CMD --dependency $WORK_DIR/oe-core/meta $WORK_DIR/meta-arm/meta-arm"
```
quaresmajose: You also need the .../meta-arm/meta-arm-toolchain and I think is better to keep one layer per line.

b49020:
Although I have fixed the CI issues which were related to this PR however the current CI failure:

```
Parsing recipes...ERROR: /work/oe-core/meta/recipes-kernel/linux/linux-yocto_6.17.bb: Unable to get checksum for linux-yocto SRC_URI entry qcom.scc: file could not be found
```

doesn't seem to be related to this PR, any ideas?

b49020:
The CI is all green now, a gentle ping for any further comments here.

conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -13,9 +13,14 @@ KERNEL_DEVICETREE ?= " \
...
                       "
 
+EXTRA_IMAGEDEPENDS += "${@bb.utils.contains('MACHINE_FEATURES', 'tfa', 'trusted-firmware-a', '', d)}"
+PREFERRED_PROVIDER_virtual/bootloader = "${@bb.utils.contains('MACHINE_FEATURES','u-boot','u-boot','',d)}"
+TFA_UBOOT:pn-trusted-firmware-a = "${@bb.utils.contains('MACHINE_FEATURES','u-boot','1','0',d)}"
+
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-rb3gen2-firmware \
     packagegroup-rb3gen2-hexagon-dsp-binaries \
+    packagegroup-optee \
```
ndechesne: we probably don't want when optee is not enabled/available. no?

conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -13,9 +13,14 @@ KERNEL_DEVICETREE ?= " \
                       qcom/qcs6490-rb3gen2-vision-mezzanine.dtb \
                       "
 
+EXTRA_IMAGEDEPENDS += "${@bb.utils.contains('MACHINE_FEATURES', 'tfa', 'trusted-firmware-a', '', d)}"
```
ndechesne: is tfa a new MACHINE_FEATURE?

ci/open-boot-firmware.yml
```diff
@@ -0,0 +1,8 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+
+local_conf_header:
+  firmware: |
+    MACHINE_FEATURES:qcm6490:append = " tfa optee u-boot"
```
ndechesne: I don't know if it's a functional difference or not.. but we usually put the :append or :prepend first in the list of overrides.

ndechesne: I don't like changing the MACHINE config in the KAS config file. I would rather have a difference machine, what do you think?

.github/actions/compile/action.yml
```diff
@@ -10,6 +10,8 @@ inputs:
     required: true
   kernel_dirname:
     required: true
+  firmware_dirname:
```
ndechesne: having a new machine would simplify the CI config here.

ci/open-boot-firmware.yml
b49020: I don't think having a different machine makes sense here. Since here we are building a different boot firmware (open vs closed source) for the same machine like we do for supporting different kernel versions for the same machine.

> but we usually put the :append or :prepend first in the list of overrides.

Sure, I will fix that.

conf/machine/qcs6490-rb3gen2-core-kit.conf
b49020: Yeah, it is a new machine feature since you asked for it in previous review.

.github/actions/compile/action.yml
b49020: Yeah, but we already have different kernel versions built. I think similarly we can support open vs closed firmware builds. However, adding new machine for just open boot firmware will just make all the supported CI artifacts 2x the current number. So, I think current approach in this PR is more scalable.

conf/machine/qcs6490-rb3gen2-core-kit.conf
b49020: I added checks for "optee" as a machine feature within the "packagegroup-optee" since yocto-check-layers complained about it. Adding a machine feature check here felt redundant but I can add it if you want.

ndechesne: i see. I think it should be the other way around. packagegroup-optee should include them always, and we should include the packagegroup here only when optee is enabled.

b49020: I did that initially, but yocto-check-layers complained about "packagegroup-optee" as an independent entity since a machine compatible is required to build `optee-test` and others.

.github/actions/compile/action.yml
ndechesne: I don't understand why it would double the artifacts. 

I don't really like how we add 'variants' for kernel and now firmware in the CI config. Perhaps we can combine the kernel and firmware into a 'variant' attribute, so that we don't keep adding more and more in there?

conf/machine/qcs6490-rb3gen2-core-kit.conf
ndechesne: we should document them, somehow, somewhere..

conf/machine/qcs6490-rb3gen2-core-kit.conf
ndechesne: we need to figure out if this is the right thing to do..

conf/layer.conf
```diff
@@ -9,7 +9,7 @@ BBFILE_COLLECTIONS += "qcom"
 BBFILE_PATTERN_qcom := "^${LAYERDIR}/"
 BBFILE_PRIORITY_qcom = "6"
 
-LAYERDEPENDS_qcom = "core"
+LAYERDEPENDS_qcom = "core meta-arm"
```
lumag: Well... If OP-TEE is not a required feature, this should be handled via dynamic layers rather than layer dependency

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
```diff
@@ -0,0 +1,9 @@
+SRC_URI += "https://github.com/coreboot/qc_blobs/raw/refs/heads/master/sc7280/qtiseclib/libqtisec.a;name=qtiseclib"
```
lumag: What is the licence for the blob? Does it change the licence of the generated binary?

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -0,0 +1,22 @@
+SRCBRANCH = "qcom-next"
```
lumag: inline

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -0,0 +1,22 @@
...
+SRC_URI = "git://github.com/qualcomm-linux/trusted-firmware-a.git;protocol=https;name=tfa;branch=${SRCBRANCH}"
+SRCREV_tfa = "164d443c8e92c18fd2cca03661dd1a2b9ac37848"
+LIC_FILES_CHKSUM = "file://docs/license.rst;md5=6ed7bace7b0bc63021c6eba7b524039e"
+
+DEPENDS += "qtestsign-native"
+
+MACHINE_TFA_QCOM_REQUIRE ?= ""
+MACHINE_TFA_QCOM_REQUIRE:qcm6490 = "trusted-firmware-a-qcm6490.inc"
```
lumag: Will it work on other qcm6490-based machines (e.g. IDP, Tachyon or RubkPi)?

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
```diff
@@ -1,9 +1,13 @@
...
 
+DEPENDS += "optee-os"
+
 COMPATIBLE_MACHINE = "qcm6490"
 TFA_PLATFORM = "rb3gen2"
 TFA_BUILD_TARGET = "bl2 fip"
+TFA_SPD = "opteed"
 EXTRA_OEMAKE:append = " \
     QTISECLIB_PATH=${UNPACKDIR}/libqtisec.a \
+    BL32=${RECIPE_SYSROOT}/${nonarch_base_libdir}/firmware/tee-raw.bin \
```
lumag: If you change the order of the commits, you won't have to change the file that you have just added.

recipes-security/optee/optee-os-qcm6490.inc
```diff
@@ -0,0 +1,6 @@
+# RB3Gen2 specific OP-TEE support
+
+require optee-os-qcom.inc
+
+COMPATIBLE_MACHINE = "qcm6490"
```
lumag: Same comment: I'd prefer if we can build binaries for all supported machines

recipes-security/optee/optee-os-qcom.inc
```diff
@@ -0,0 +1,3 @@
+SRCBRANCH = "qcom-next"
```
lumag: inline

recipes-security/packagegroups/packagegroup-optee.bb
```diff
@@ -0,0 +1,9 @@
+SUMMARY = "Packages for the OP-TEE support"
+
+inherit packagegroup
+
+RRECOMMENDS:${PN} = " \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'optee-client', '', d)} \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'optee-test', '', d)} \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'optee-os-ta', '', d)} \
```
lumag: Why do we need three lines here?

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -0,0 +1,22 @@
+SRCBRANCH = "qcom-next"
+SRC_URI = "git://github.com/qualcomm-linux/trusted-firmware-a.git;protocol=https;name=tfa;branch=${SRCBRANCH}"
+SRCREV_tfa = "164d443c8e92c18fd2cca03661dd1a2b9ac37848"
```
lumag: This breaks the idea of the orignial TF-A recipes. Inside meta-arm there are 4 different versions of TF-A (2.10.17, 2.12.3, 2.13.0, git aka tip). Machine-specific includes don't change the URI and the SRCREV. In other words: if the user sets `PREFERRED_VERSION_trusted-firmware-a`, it is expected to get _that_ TF-A version, not some random machine-specific version

recipes-security/optee/optee-os-qcom.inc
```diff
@@ -0,0 +1,3 @@
...
+SRC_URI = "git://github.com/qualcomm-linux/optee_os.git;protocol=https;name=optee;branch=${SRCBRANCH}"
+SRCREV_optee = "a427f12bc60e74ccdd2ae771b50ed237397a0782"
```
lumag: The same question. We rewrite SRC_URI and SRCREV here. It breaks user assumptions.

recipes-bsp/u-boot/files/0001-dts-qcs6490-rb3gen2-u-boot-Add-OP-TEE-node.patch
```diff
@@ -0,0 +1,36 @@
...
++++ b/arch/arm/dts/qcs6490-rb3gen2-u-boot.dtsi
+@@ -15,6 +15,13 @@
+ 		      <0 0xC3400000 0 0x3CC00000>,
+ 		      <1 0x00000000 1 0x00000000>;
+ 	};
++
++	firmware {
++		optee {
++			compatible = "linaro,optee-tz";
++			method = "smc";
```
lumag: This node should only be enabled if we actually provide OP-TEE for the platform

conf/machine/qcs6490-rb3gen2-core-kit.conf
lumag: Please keep the list sorted

.github/actions/compile/action.yml
b49020: > I don't understand why it would double the artifacts.

I mean if we add a new machine here then all the different distro and kernel versions will have to be built twice.

> Perhaps we can combine the kernel and firmware into a 'variant' attribute, so that we don't keep adding more and more in there?

I am mostly interested to run open boot firmware builds using qcom-next kernel since for fully feature complete we will be contributing patches in kernel for OP-TEE support to enable all the peripherals. So, let me know how you would like to see the variants here.

conf/machine/qcs6490-rb3gen2-core-kit.conf
```diff
@@ -13,9 +13,14 @@ KERNEL_DEVICETREE ?= " \
...
+PREFERRED_PROVIDER_virtual/bootloader = "${@bb.utils.contains('MACHINE_FEATURES','u-boot','u-boot','',d)}"
+TFA_UBOOT:pn-trusted-firmware-a = "${@bb.utils.contains('MACHINE_FEATURES','u-boot','1','0',d)}"
```
lumag: It feels strange at a first glance that machine configuration (which should also define MACHINE_FEATURES) also depends on them. Please add a comment here, how is it expected to be enabled.

ci/base.yml
```diff
@@ -27,6 +27,12 @@ repos:
     layers:
       meta-poky:
 
+  meta-arm:
+    url: https://git.yoctoproject.org/meta-arm
+    layers:
+      meta-arm:
+      meta-arm-toolchain:
```
lumag: So, building has beek broken for all previous commits. That's not nice

ci/open-boot-firmware.yml
lumag: > I don't know if it's a functional difference or not.. but we usually put the :append or :prepend first in the list of overrides.

It is: `foo:soc:append` appends to `foo:soc`, which then override full `foo` value. `foo:append:soc`: makes `foo:append` work only on `soc`, which then appends to the existing `foo` variable without redefining it completely.

I also don't like the idea of changing MACHINE_FEATURES here. I'd prefer if we have enabled MACHINE_FEATURES for all machines that support TF-A and then used `COMBINED_FEATURES` in order to determine whether to use a normal boot flow or a TF-A based one.

.github/actions/compile/action.yml
```diff
@@ -10,6 +10,8 @@ inputs:
     required: true
```
lumag, line 10: > At this point due to QTI signature requirement, the LAVA boot test for RB3Gen2
can't be enabled. 

What does it mean? Can users build and deploy TF-A firmware? If not, this is just a dead code.

lumag (changes requested):

.github/workflows/build-yocto.yml
```diff
@@ -144,6 +145,14 @@ jobs:
...
                 dirname: "+linux-yocto-lts"
                 yamlfile: ":ci/linux-yocto-lts.yml"
+          - machine: qcs6490-rb3gen2-core-kit
+            distro:
+                name: qcom-distro
+                yamlfile: ':ci/qcom-distro.yml'
+            firmware:
+                type: additional
+                dirname: "+open-boot-firmware"
+                yamlfile: ":ci/open-boot-firmware.yml"
```
ndechesne: the yamlfile is unused. does not sound right.

.github/workflows/build-yocto.yml
```diff
@@ -162,6 +171,7 @@ jobs:
           distro_name: ${{matrix.distro.name}}
           kernel_yaml: ${{matrix.kernel.yamlfile}}
           kernel_dirname: ${{matrix.kernel.dirname}}
+          firmware_dirname: ${{matrix.firmware.dirname}}
```
ndechesne: what I meant in my other comment is that instead of adding more and more config here (e.g. machine, distro, kernel, firmware, ...) I would prefer to have machine, distro and 'variants'

e.g.

```
- machine: qcs6490-rb3gen2-core-kit
            distro:
                name: qcom-distro
                yamlfile: ':ci/qcom-distro.yml'
            firmware:
                type: additional
                dirname: "+open-boot-firmware"
                yamlfile: ":ci/open-boot-firmware.yml"

```
could be

```
- machine: qcs6490-rb3gen2-core-kit
            distro:
                name: qcom-distro
                yamlfile: ':ci/qcom-distro.yml'
            variants:
                type: additional
                dirname: "+open-boot-firmware"
                yamlfile: ":ci/open-boot-firmware.yml"
```

so, combining a custom kernel and custom firmware:

```
- machine: qcs6490-rb3gen2-core-kit
            distro:
                name: qcom-distro
                yamlfile: ':ci/qcom-distro.yml'
            variants:
                type: additional
                dirname: "+linux-yocto-lts+open-boot-firmware"
                yamlfile: ":ci/linux-yocto-lts.yml:ci/open-boot-firmware.yml"

```

ci/open-boot-firmware.yml
ndechesne: I like the COMBINED_FEATURES. 
We will then need to decide if we have an 'open boot' distro ;)

conf/machine/qcs6490-rb3gen2-core-kit.conf
b49020: Okay, I can document them here `conf/machine/include/qcom-common.inc`.

conf/machine/qcs6490-rb3gen2-core-kit.conf
b49020: Okay, let me add machine check here as well for consistency.

conf/layer.conf
b49020: No please, boot firmware is an essential component of any Yocto BSP layer. The fact that meta-qcom uses proprietary boot firmware blobs is unfortunate. Here, meta-arm is providing the common recipes for TF-A, OP-TEE as well as edk2 for Arm based platforms is an essential dependency as we are starting to support open boot stack on Qcom platforms.

Have a look at meta-ti for your reference here [1].

[1] https://git.yoctoproject.org/meta-ti/tree/meta-ti-bsp/conf/layer.conf

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
b49020: The license for this blob can be found here: https://github.com/coreboot/qc_blobs/blob/master/sc7280/qtiseclib/LICENSE.

> Does it change the licence of the generated binary?

AFAIK, it doesn't change the overall TF-A project license but rather it is required for the binary components to be compliant with the TF-A license terms, see here: https://trustedfirmware-a.readthedocs.io/en/latest/process/contributing.html#binary-components

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
b49020: Yeah, the TF-A/OP-TEE port is specific to the SoC. However, we need to make sure that the OEM hardware is running a QLI based boot firmware release where we can replace boot chain components with open-source ones.

BTW, I have been waiting to get my hands-on with one of these boards as well to run open boot stack.

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
b49020: This commit adds OP-TEE support to the TF-A recipe added by previous commit. So, I am not sure what do you meant by reordering commits here.

recipes-security/packagegroups/packagegroup-optee.bb
b49020: Okay, I will convert them in a single line.

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
b49020: That's a fair point, I will add proper versioned recipes instead.

recipes-bsp/u-boot/files/0001-dts-qcs6490-rb3gen2-u-boot-Add-OP-TEE-node.patch
b49020: Okay, I can add the OP-TEE machine feature check for this patch.

ci/base.yml
b49020: That's a fair point, I will move this diff to first patch instead.

.github/actions/compile/action.yml
b49020: Currently the developers within Qualcomm are able to generate QTI signed TF-A BL2 and deploy the TF-A firmware. So, it's definitely not a dead code.

For developers outside Qualcomm, we are expecting in the next QLI proprietary firmware blobs release (especially XBL_SEC), the QTI signature requirement for TF-A BL2 to be dropped. Then we should just be able to use `qtestsign` for that too.

.github/workflows/build-yocto.yml
b49020: Ah, I will fix it.

.github/workflows/build-yocto.yml
b49020: Sounds reasonable, I will use this format.

ci/open-boot-firmware.yml
b49020: It feels like we are moving in a direction to build open boot firmware and package closed boot firmware during a common build like for RB3Gen2 in this case. The build artifacts will contain both but what is present in the device flashing tarball is based on `open-boot` feature present in `COMBINED_FEATURES`. Does that sound reasonable?

.github/actions/compile/action.yml
lumag: > Currently the developers within Qualcomm are able to generate QTI signed TF-A BL2 and deploy the TF-A firmware. So, it's definitely not a dead code.

It is dead until one can use it to generate a fully working solution. If it requires extra signing step, it is of very limited use.

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
lumag: - Add OP-TEE as a recipe
- Add TF-A, which uses OP-TEE among other dependencies

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
lumag: Still the licence should be included in SBOM, etc., so it should be properly annotated.
Also we have to comply to the licensing terms. E.g.:

> (iii) redistribution of the Redistributable Binary Code must include the .txt file setting forth the terms and condition of this Agreement;

.github/actions/compile/action.yml
b49020: > If it requires extra signing step, it is of very limited use.

Sure, it's a temporary extra step until we enable the OEM only signed TZ image. But still, it's a good enough first step for people working within Qualcomm to collaborate while making the open boot firmware stack a reality on Qcom silicon.

conf/layer.conf
lumag: I'd still ask it to be handled via a dynamic layer inside meta-qcom. We try to depend only on OE-Core, however we use other layers to provide extended functionality.

b49020: Sorry but this isn't extended functionality but an essential function of the BSP layer to allow building boot firmware from source.

lumag: At this point, the basic functionality is to use the blobs. I welcome the usage of OS components and OP-TEE, but I don't want to enforce that dependency on everybody.

b49020: > At this point, the basic functionality is to use the blobs.

Is that what `meta-qcom` as an open source BSP layer would enforce its users to use by default even if the open boot stack becomes a viable option in future?

> but I don't want to enforce that dependency on everybody.

But you are ready to enforce a practice not at all followed by other open source BSP layers? What extra burden do you think meta-arm brings which isn't there for say meta-ti layer?

lumag: > > At this point, the basic functionality is to use the blobs.
> 
> Is that what `meta-qcom` as an open source BSP layer would enforce its users to use by default even if the open boot stack becomes a viable option in future?
>
> > but I don't want to enforce that dependency on everybody.
> 
> But you are ready to enforce a practice not at all followed by other open source BSP layers? What extra burden do you think meta-arm brings which isn't there for say meta-ti layer?

At this point OP-TEE & TF-A - based stack:
- is not enabled by default on any of the platforms
- doesn't work out of the box outside of QTI
- is provided for a single SoC

Having all of that in mind, as I wrote, I'd welcome the introduction of TF-A & OP-TEE based boot stack (and documenting that in the README), but I don't want to make everybody pull in that dependency. When it's decided that the open boot stack is a default option for any of the targets, we can flip it.

b49020: > is not enabled by default on any of the platforms

What do you mean by default here? One that is enforced by Qualcomm for every OEM? Or rather we give OEMs the choice to select among open vs closed boot firmware stacks. There are already various IoT and Auto OEMs asking for the open boot firmware stack.

> doesn't work out of the box outside of QTI

As I said this is being worked upon to enable it outside QTI. Also, the next step for me is to add support for upstream edk2 whose recipes are also currently provided by meta-arm which will work out of the box as of now.

> is provided for a single SoC

As I said in the PR itself, the plan is to support Kodiak and Lemans initially, but it will be scalable to other SoCs too with less effort.

> but I don't want to make everybody pull in that dependency.

Sooner or later, we need to add that dependency as otherwise how can we enable OEMs to build open boot stack in the same build environment.

> When it's decided that the open boot stack is a default option for any of the targets, we can flip it.

As I said above, we should really be offering a choice for OEMs to build open boot stacks. And I don't understand when you would feel it's the right time to flip. You need to really understand; the open boot firmware stack is an essential feature of any open source Yocto BSP layer which people expect to be available by default.

lumag: > > is not enabled by default on any of the platforms
> 
> What do you mean by default here? One that is enforced by Qualcomm for every OEM? Or rather we give OEMs the choice to select among open vs closed boot firmware stacks. There are already various IoT and Auto OEMs asking for the open boot firmware stack.

OEM devices are not supported by this layer. Inside this layer we provide generic code and also support for Qualcomm-manufactured devices. I know that there are use-cases which require OP-TEE and TF-A. But I really don't understand what you are arguing here about.

> 
> > doesn't work out of the box outside of QTI
> 
> As I said this is being worked upon to enable it outside QTI.

That doesn't change the current status of these changes.

> Also, the next step for me is to add support for upstream edk2 whose recipes are also currently provided by meta-arm which will work out of the box as of now.

That's perfect and really interesting.

> 
> > is provided for a single SoC
> 
> As I said in the PR itself, the plan is to support Kodiak and Lemans initially, but it will be scalable to other SoCs too with less effort.

Great.

> 
> > but I don't want to make everybody pull in that dependency.
> 
> Sooner or later, we need to add that dependency as otherwise how can we enable OEMs to build open boot stack in the same build environment.
> 
> > When it's decided that the open boot stack is a default option for any of the targets, we can flip it.
> 
> As I said above, we should really be offering a choice for OEMs to build open boot stacks. And I don't understand when you would feel it's the right time to flip. You need to really understand; the open boot firmware stack is an essential feature of any open source Yocto BSP layer which people expect to be available by default.

As I wrote, I don't understand, what we are arguing here. Please move your changes under `dynamic-layers/meta-arm-bsp/` and add `meta-arm-bsp` to `LAYERRECOMMENDS`. I'm not asking you to drop the changes. I'm asking you to allow people to use meta-qcom without pulling in extra layers if they don't need them.

b49020: > As I wrote, I don't understand, what we are arguing here.

I think here the main point of argument is to keep meta-arm layer in the dependency list of meta-qcom since we need to reuse the BSP recipes for TF-A, OP-TEE and edk2. Let others also chime in about what they think here. I think I have made my arguments here.

> Please move your changes under dynamic-layers/meta-arm-bsp/ and add meta-arm-bsp to LAYERRECOMMENDS.

The TF-A, OP-TEE and edk2 changes belong to the Qcom BSP layer and not to the Arm BSP layer. I will reiterate the example for meta-ti here.

lumag: > > Please move your changes under dynamic-layers/meta-arm-bsp/ and add meta-arm-bsp to LAYERRECOMMENDS.
> 
> The TF-A, OP-TEE and edk2 changes belong to the Qcom BSP layer and not to the Arm BSP layer. I will reiterate the example for meta-ti here.

`meta-qcom/dynamic-layers/meta-arm-bsp/`. This is a standard way of making recipes / appends depend on another layer without making it a layer dependency. See https://github.com/openembedded/meta-openembedded/tree/master/meta-oe/dynamic-layers

nrclark:
Just wanted to leave a comment here - I'm very glad to see this develop. GM and many of Qualcomm's other customers would strongly prefer a `TF-A`+`OP-TEE`+`U-boot` stack over the current proprietary mess. You would make my day if you could get the automotive group to adopt it for Lemans and Nord.

One extra item to note: if you don't have one already, you should consider setting up a uboot config that can be loaded with Sahara. That would allow customers to get away from Firehose and flatbuilds in favor of more cross-platform solutions inside of uboot.

quaresmajose:
Yes, this work tree can be broken because all the others have been updated in the meantime. The meta-yocto 
is the one which provides the linux-yocto.

Please rebase and pick the latest tip of all layers with:`kas build --update ci/qcs6490-rb3gen2-core-kit.yml:ci/open-boot-firmware.yml`.

The other way arround is locking the revision of all the layers when the PR runs last time and extract the [kas-lock.zip](https://github.com/qualcomm-linux/meta-qcom/actions/runs/19125459442/artifacts/4483017157) to `ci/base.lock.yml`. Then run `kas build ci/qcs6490-rb3gen2-core-kit.yml:ci/open-boot-firmware.yml`. Don't forget to remove the `ci/base.lock.yml` when it is no longer necessary.

ldts:
> Yes, this work tree can be broken because all the others have been updated in the meantime. The meta-yocto is the one which provides the linux-yocto.
> 
> Please rebase and pick the latest tip of all layers with:`kas build --update ci/qcs6490-rb3gen2-core-kit.yml:ci/open-boot-firmware.yml`.
> 
> The other way arround is locking the revision of all the layers when the PR runs last time and extract the [kas-lock.zip](https://github.com/qualcomm-linux/meta-qcom/actions/runs/19125459442/artifacts/4483017157) to `ci/base.lock.yml`. Then run `kas build ci/qcs6490-rb3gen2-core-kit.yml:ci/open-boot-firmware.yml`. Don't forget to remove the `ci/base.lock.yml` when it is no longer necessary.

thanks @quaresmajose ! yeah I forgot it is a moving target. I am chasing an issue with a secure watchdog that is being affected by the kernel boot sequence since 6.18-rc4 (still not sure what the heck is going on but the 6 second barking period programmed by TF-A is resulting in >22 during linux boot leading to a watchdog bite and reset). anyway thanks for the options!

quaresmajose:
There is a proposal to move OP-TEE from meta-arm to oe-core on the mailing list [[RFC] Moving OP-TEE support from meta-arm to oe-core
](https://lists.yoctoproject.org/g/yocto/message/66178)

conf/layer.conf
b49020: Added meta-arm as an optional dependency.

conf/machine/qcs6490-rb3gen2-core-kit.conf
b49020: Add a separate machine with open boot stack

recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
b49020: Added support for separate QTISECLIB license

.github/actions/compile/action.yml
b49020: Add a separate machine with open boot stack

conf/layer.conf
```diff
@@ -22,6 +22,8 @@ BBFILES_DYNAMIC += " \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bbappend \
     selinux:${LAYERDIR}/dynamic-layers/selinux/*/*/*.bb \
     selinux:${LAYERDIR}/dynamic-layers/selinux/*/*/*.bbappend \
+    meta-arm:${LAYERDIR}/dynamic-layers/meta-arm/*/*/*.bb \
```
ndechesne: unless there is a reason (if so, let's document it), please keep list sorted.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
```diff
@@ -0,0 +1,31 @@
+SRC_URI += " \
+    https://github.com/coreboot/qc_blobs/raw/refs/heads/master/sc7280/qtiseclib/libqtisec.a;name=qtiseclib \
+    https://github.com/coreboot/qc_blobs/raw/refs/heads/master/sc7280/qtiseclib/LICENSE;name=qtiseclib_lic \
```
lumag: You are basically using `${AUTOREV}` here. If binaries change in the master branch, we will face build failures. Can we please use the fixed commit IDs instead?

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
```diff
@@ -0,0 +1,31 @@
...
+   "
+SRC_URI[qtiseclib.sha256sum] = "6860dda0701c8709530608cc0e5a61b76484ae16cb673ba9a23510cf4b3d57bf"
+SRC_URI[qtiseclib_lic.sha256sum] = "ab177444ebe998e10534dc4bd80dbebac56301978dbaf78339b22bd3aa170aa7"
```
lumag: Could you please also `LICENSE += "LICENSE.qcom-2"`

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom_git.bb
```diff
@@ -0,0 +1,16 @@
+require recipes-bsp/trusted-firmware-a/trusted-firmware-a.inc
+
+DEPENDS += "qtestsign-native"
```
lumag: This should be a part of the qcm6490 include.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom_git.bb
```diff
@@ -0,0 +1,16 @@
...
+
+PV = "2.14.0-qcom+git"
+
+SRC_URI = "git://github.com/qualcomm-linux/trusted-firmware-a.git;protocol=https;name=tfa;branch=qcom-next"
+SRCREV_tfa = "78d883ea408d58786287102bc704a5bfce2cbd90"
+
+LICENSE = "BSD-2-Clause & BSD-3-Clause & MIT & Apache-2.0 & CLOSED"
```
lumag: Please drop CLOSED.

ci/rb3gen2-core-kit-open-fw.yml
```diff
@@ -0,0 +1,19 @@
...
+    url: https://git.yoctoproject.org/meta-arm
+    layers:
+      meta-arm:
+      meta-arm-toolchain:
+
+machine: rb3gen2-core-kit-open-fw
+
+local_conf_header:
+  optee: |
+    MACHINE_EXTRA_RRECOMMENDS += "packagegroup-optee"
```
lumag: This should be a part of the machine config

lumag (changes requested):
Overall looks great.

dynamic-layers/meta-arm/recipes-security/optee/optee-os-qcom_git.bb
```diff
@@ -0,0 +1,11 @@
+require recipes-security/optee/optee-os.inc
+
+PV = "4.9.0-qcom+git"
+
+SRC_URI = "git://github.com/qualcomm-linux/optee_os.git;protocol=https;name=optee;branch=qcom-next"
+SRCREV_optee = "c2b0684fcd89929976a8726e6e3af922b48dd2c7"
+
+MACHINE_OPTEE_REQUIRE ?= ""
+MACHINE_OPTEE_REQUIRE:qcm6490 = "optee-qcm6490.inc"
```
ricardosalveti: COMPATIBLE_MACHINE will only be set in case of qcm6490, while it should always be set to that, as we know it currently only builds on kodiak.

dynamic-layers/meta-arm/recipes-security/optee/optee-qcm6490.inc
```diff
@@ -0,0 +1,4 @@
+# RB3Gen2 specific OP-TEE support
+
+COMPATIBLE_MACHINE = "qcm6490"
+OPTEEMACHINE = "qcom-kodiak"
```
ricardosalveti: I would probably just make this include generic to qcom and define these variables for qcm6490 only.

dynamic-layers/meta-arm/recipes-security/packagegroups/packagegroup-optee.bb
```diff
@@ -0,0 +1,7 @@
+SUMMARY = "Packages for the OP-TEE support"
+
+inherit packagegroup
+
+RRECOMMENDS:${PN} = " \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'optee-os-tadevkit optee-client optee-test', '', d)} \
+"
```
ricardosalveti, line 7: Please break this packagegroup into a separated commit.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom_git.bb
ricardosalveti: What are the licenses for the closed components included as part of the build?

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom_git.bb
```diff
@@ -0,0 +1,16 @@
...
+LIC_FILES_CHKSUM = "file://docs/license.rst;md5=6ed7bace7b0bc63021c6eba7b524039e"
+
+MACHINE_TFA_QCOM_REQUIRE ?= ""
+MACHINE_TFA_QCOM_REQUIRE:qcm6490 = "trusted-firmware-a-qcm6490.inc"
```
ricardosalveti: Same comment to op-tee, this will only set COMPATIBLE_MACHINE for kodiak when kodiak is used, while it should be set for everyone.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
ricardosalveti: This license is probably qcom o qcom-2, need to compare and see.

conf/machine/rb3gen2-core-kit-open-fw.conf
```diff
@@ -0,0 +1,12 @@
...
+#@DESCRIPTION: Machine configuration for Qualcomm RB3Gen2 development Kit, with QCS6490
+
+require conf/machine/rb3gen2-core-kit.conf
+
+EXTRA_IMAGEDEPENDS += "trusted-firmware-a-qcom"
+
+MACHINE_FEATURES += "optee"
+
+UBOOT_CONFIG = "qcs6490-rb3gen2"
+UBOOT_CONFIG[qcs6490-rb3gen2] = "qcm6490_defconfig"
```
ricardosalveti, line 13: Why do we have UBOOT here?

ci/rb3gen2-core-kit-open-fw.yml
```diff
@@ -0,0 +1,19 @@
...
+  version: 14
+  includes:
+  - ci/base.yml
+
+repos:
+  meta-arm:
+    url: https://git.yoctoproject.org/meta-arm
+    layers:
+      meta-arm:
+      meta-arm-toolchain:
```
ricardosalveti: As we plan to add other targets later on, I would prefer this to be a kas fragment on its own.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
b49020: Sure

ci/rb3gen2-core-kit-open-fw.yml
b49020: Yocto check layers doesn't allow me to put this as part of machine config as all the OP-TEE components can't be provided by `meta-qcom` layer since `meta-arm` support comes via dynamic layers.

dynamic-layers/meta-arm/recipes-security/optee/optee-os-qcom_git.bb
b49020: The default meta-arm recipes allow support for Qemu targets and I would like to support them with our TF-A and OP-TEE Qcom staging trees as well. So this is not just Kodiak specific recipe.

dynamic-layers/meta-arm/recipes-security/optee/optee-qcm6490.inc
b49020: Sounds reasonable, will do that.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom_git.bb
b49020: Please see my response above.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
b49020: I don't see an exact match with qcom or qcom-2, there are minor wording differences. Let me rather follow Dmitry's suggestion with: `LICENSE += "LICENSE.qcom-2"` and also keep this additional license file explicitly.

conf/machine/rb3gen2-core-kit-open-fw.conf
b49020: TF-A recipe requires BL33 to be specified and upstream U-Boot recipe from oe-core works on RB3Gen2 but it doesn't enable OP-TEE features as of now due to missing patch-set carried by the u-boot-qcom tree. That's why I say there isn't direct merge dependency with https://github.com/qualcomm-linux/meta-qcom/pull/1473.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom_git.bb
b49020: The only closed component used is QTISECLIB from here: https://github.com/coreboot/qc_blobs/tree/main/sc7280/qtiseclib. The future plan is to migrate all the drivers upstream and then it shouldn't be required as well.

dynamic-layers/meta-arm/recipes-security/optee/optee-os-qcom_git.bb
b49020: Exploring further I see TF-A recipes allows Qemu machine support via `trusted-firmware-a_%.bbappend` which won't apply to qcom recipe variant. So I would just by default include the *.inc file.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
lumag: It can't be qcom-2. I think it should be qcom license.

koenkooi:
> @koenkooi I needed an initramfs change to break a circular dependency reported by CI, here: [3bdcf07](https://github.com/qualcomm-linux/meta-qcom/commit/3bdcf07275cd1e5cf2e7a7b5a9a2908109f1e4c2). Not sure why this isn't a problem with initramfs recipes provided by OE core. Or is meta-qcom initramfs recipe special in some sense here?

Our recipe is functionally the same as `openembedded-core/meta/recipes-core/images/core-image-initramfs-boot.bb`, I suspect the difference is in that we use both u-boot and TF-A which change the dependencies. I need to have a better look at those to judge if `IMAGE_DEPENDS = ""` is the proper solution.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcm6490.inc
b49020: Okay switched to qcom license.

b49020:
@koenkooi 

> I suspect the difference is in that we use both u-boot and TF-A which change the dependencies. I need to have a better look at those to judge if IMAGE_DEPENDS = "" is the proper solution.

Okay I have switched to your suggestion instead, let me know if that's fine.

lumag (dismissed):

quaresmajose (dismissed):

ci/rb3gen2-core-kit-open-fw.yml
ricardosalveti: Yeah, we need this in the machine conf, not every user will use kas.

Can you try like:

```
diff --git a/conf/machine/rb3gen2-core-kit-open-fw.conf b/conf/machine/rb3gen2-core-kit-open-fw.conf
index 3b0da0e..b5e0b49 100644
--- a/conf/machine/rb3gen2-core-kit-open-fw.conf
+++ b/conf/machine/rb3gen2-core-kit-open-fw.conf
@@ -10,3 +10,5 @@ MACHINE_FEATURES += "optee"

 UBOOT_CONFIG = "qcs6490-rb3gen2"
 UBOOT_CONFIG[qcs6490-rb3gen2] = "qcm6490_defconfig"
+
+MACHINE_EXTRA_RRECOMMENDS += "${@bb.utils.contains('BBFILE_COLLECTIONS', 'meta-arm', 'packagegroup-optee', '', d)}"
```

.github/workflows/build-yocto.yml
```diff
@@ -219,6 +219,14 @@ jobs:
                 type: default
                 dirname: ""
                 yamlfile: ""
+          - machine: rb3gen2-core-kit-open-fw
+            distro:
+                name: qcom-distro-kvm
+                yamlfile: ':ci/qcom-distro-kvm.yml'
```
ricardosalveti, line 241: Mind extending the commit message explaining why qcom-distro-kvm is used here?

ricardosalveti:
> Okay I have switched to your suggestion instead, let me know if that's fine.

Which build failed because of this? I want to reproduce locally.

b49020:
> > Okay I have switched to your suggestion instead, let me know if that's fine.
> 
> Which build failed because of this? I want to reproduce locally.

Just try following, it will report the circular dependency:

```
$ kas build ../meta-qcom/ci/rb3gen2-core-kit-open-fw.yml:../meta-qcom/ci/qcom-distro-kvm.yml
```

.github/workflows/build-yocto.yml
b49020: I thought it was obvious with open boot stack, Linux kernel booting in EL2 mode with KVM enabled is the default and only supported mode. Anyhow, will extend commit message to put this info.

ci/rb3gen2-core-kit-open-fw.yml
b49020: > Yeah, we need this in the machine conf, not every user will use kas.

Sure, users need to add the meta-arm layer somehow. Kas was suggested as the preferred approach for users already.

> Can you try like:

Anyhow, this looks like a nice trick to keep the kas file to a minimum. Will see if yocto check layers is happy with it.

ricardosalveti:
> That I plan to do as a follow up PR. Right now I need to figure out how to enable OEM signing for BL2 TZ image. We have been using sectools internally but there isn't a public release of profiles yet. With qtestsign, I need to see if that can support OEM only signing for TZ image.

Sure, sounds good.

.github/workflows/build-yocto.yml
ricardosalveti: It is obvious to us only :-)

Thanks for updating it.

lumag (approved):

CodeChenL:
Hello everyone, I am an operating systems engineer from Radxa. We are currently looking to begin using Open Boot firmware on our Qualcomm products such as [SBCs](https://docs.radxa.com/dragon/q6a), so our users and customers can customize their own boot firmware and, together with more people, make them better.
But when I tried to use this boot chain, it entered an infinite reboot when booting to BL31. Logs:
https://gist.github.com/CodeChenL/6b48eeaf33fc5f0a5738cde699c09941#file-gistfile1-txt
After my debugging I found it might be stuck at the final cpucp-related call: https://gist.github.com/CodeChenL/6b48eeaf33fc5f0a5738cde699c09941?permalink_comment_id=6101921#gistcomment-6101921
I would be very grateful if anyone in this thread could give me some help

b49020:
Thanks @CodeChenL for your report, can you rather create an issue under Qcom TF-A staging tree (https://github.com/qualcomm-linux/trusted-firmware-a/issues) here such that the status can be tracked?

CodeChenL:
> Thanks @CodeChenL for your report, can you rather create an issue under Qcom TF-A staging tree (https://github.com/qualcomm-linux/trusted-firmware-a/issues) here such that the status can be tracked?

Thank you for your reply. After our testing and discussion, we have decided to shift our development focus to upstream EDK2. I have already ported an MMC driver and successfully booted the system using it. Thank you all for your work; it's truly great and has set a good precedent for us

b49020:
@CodeChenL I assume you would have used following command to sign TF-A BL2:

```
sectools secure-image bl2.elf \
    --outfile bl2.mbn \
    --image-id TZ-TEE \
    --security-profile kodiak_tz_security_profile.xml \
    --sign --signing-mode TEST
```
actually I am able to reproduce the issue you are seeing here on RB3Gen2. Looks like OEM only signing feature in XBL_SEC is leading to this issue. The QTI signed BL2 works fine though. We will keep digging into this issue, I have filed one here: https://github.com/qualcomm-linux/trusted-firmware-a/issues/19 so you can keep track of status there.

> Thank you for your reply. After our testing and discussion, we have decided to shift our development focus to upstream EDK2. I have already ported an MMC driver and successfully booted the system using it. Thank you all for your work; it's truly great and has set a good precedent for us

PRs are always welcome to upstream edk2 project extending Qcom platforms support. I hope the open boot stack effort is valuable to your customers/developers.

CodeChenL:
> @CodeChenL I assume you would have used following command to sign TF-A BL2:
> 
> ```
> sectools secure-image bl2.elf \
>     --outfile bl2.mbn \
>     --image-id TZ-TEE \
>     --security-profile kodiak_tz_security_profile.xml \
>     --sign --signing-mode TEST
> ```
> 
> actually I am able to reproduce the issue you are seeing here on RB3Gen2. Looks like OEM only signing feature in XBL_SEC is leading to this issue. The QTI signed BL2 works fine though. We will keep digging into this issue, I have filed one here: [qualcomm-linux/trusted-firmware-a#19](https://github.com/qualcomm-linux/trusted-firmware-a/issues/19) so you can keep track of status there.
> 
> > Thank you for your reply. After our testing and discussion, we have decided to shift our development focus to upstream EDK2. I have already ported an MMC driver and successfully booted the system using it. Thank you all for your work; it's truly great and has set a good precedent for us
> 
> PRs are always welcome to upstream edk2 project extending Qcom platforms support. I hope the open boot stack effort is valuable to your customers/developers.

When i complete the basic functionality and pass our internal team's code quality review, I will submit them to the upstream EDK2

Daniel224455:
Hello,
Is there any plan to have MDP/Display working on U-Boot (Kodiak platform)?

b49020:
> Hello, Is there any plan to have MDP/Display working on U-Boot (Kodiak platform)?

I am not aware of anyone actively working on this feature for Kodiak. But feel free to raise a feature request here: https://github.com/qualcomm-linux/u-boot/issues such that anybody working upstream can share the implementation.

PR #1179, opened 2025-10-31
Camx downstream

gkhose-qipl:
camx: add camera userspace libs
Add recipe for the camera userspace libs.

### DIFF
--- a/recipes-multimedia/camx/camx-kt_1.0.bb
+++ b/recipes-multimedia/camx/camx-kt_1.0.bb
@@ -1,0 +1,32 @@
+SUMMARY = "Qualcomm camera User space libraries"
+
+DESCRIPTION = "Collection of prebuilt libraries to support camera downstream functionality."
+
+LICENSE = "CLOSED"
+
+# no top-level dir in the archive, unpack to subdir to prevent UNPACKDIR pollution
+SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/camx.qclinux.0.0/251030/prebuilt_yocto/${BPN}_${PV}_armv8-2a.tar.gz;subdir=${BPN}-${PV}"
+
+SRC_URI[sha256sum] = "6deebd1220b22c0381a0d32b09e2413ca6e0ab8445c6c5ef43b5cec5c25fe62c"
+
+DEPENDS += "camxlib-kt fastrpc libxml2 protobuf-native protobuf abseil-cpp"
+
+
+do_install() {
+    install -d ${D}${libdir}
+    cp -r ${S}/usr/lib/* ${D}${libdir}
+
+    install -d ${D}${includedir}
+    cp -r ${S}/usr/include/* ${D}${includedir}
+
+}
+
+FILES:${PN} = "\
+    /usr/lib/* "
+FILES:${PN}-dev = "/usr/include/*"
+
+
+INSANE_SKIP:${PN} = "dev-so"
+INSANE_SKIP:${PN} += "file-rdeps"
+
+
--- a/recipes-multimedia/camx/camx_1.0.bb
+++ b/recipes-multimedia/camx/camx_1.0.bb
@@ -1,0 +1,39 @@
+SUMMARY = "Qualcomm camera User space libraries"
+
+DESCRIPTION = "Collection of prebuilt libraries to support camera downstream functionality."
+
+LICENSE = "CLOSED"
+
+# no top-level dir in the archive, unpack to subdir to prevent UNPACKDIR pollution
+SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/camx.qclinux.0.0/251030/prebuilt_yocto/${BPN}_${PV}_armv8-2a.tar.gz;subdir=${BPN}-${PV}"
+
+SRC_URI[sha256sum] = "0cc7b668b512de857910c17605f4c10d645ec02fc53f96a21a63980ff33e868c" 
+
+DEPENDS += "camxlib fastrpc protobuf-native protobuf "
+
+# Install lib file at right location
+do_install() {
+    install -d ${D}${libdir}
+    cp -r ${S}/usr/lib/* ${D}${libdir}
+
+    install -d ${D}${includedir}
+    cp -r ${S}/usr/include/* ${D}${includedir}
+
+
+}
+
+
+#Skips check for archtecture
+#INSANE_SKIP:${PN} = "arch"
+# The modules require .so to be dynamicaly loaded
+#INSANE_SKIP:${PN} += "dev-so"
+# Prebuilt libraries are already stripped
+#INSANE_SKIP:${PN} += "already-stripped"
+
+FILES:${PN} = "\
+    /usr/lib/*"
+FILES:${PN}-dev = "/usr/include/*"
+
+#Skips check for .so symlinks
+INSANE_SKIP:${PN} = "dev-so"
+
--- a/recipes-multimedia/camx/camxcommon_1.0.bb
+++ b/recipes-multimedia/camx/camxcommon_1.0.bb
@@ -1,0 +1,26 @@
+SUMMARY = "Qualcomm camera User space libraries"
+
+DESCRIPTION = "Collection of prebuilt libraries to support camera downstream functionality."
+
+LICENSE = "CLOSED"
+
+# no top-level dir in the archive, unpack to subdir to prevent UNPACKDIR pollution
+SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/camx.qclinux.0.0/251030/prebuilt_yocto/${BPN}_${PV}_armv8-2a.tar.gz;subdir=${BPN}-${PV}"
+
+SRC_URI[sha256sum] = "adc2507ed33e55500c24db90944296550c4c344d5596aa0a9064e709ade8df70"
+
+DEPENDS += "glib-2.0"
+
+do_install() {
+    install -d ${D}${libdir}
+    cp -r ${S}/usr/lib/* ${D}${libdir}
+
+    install -d ${D}${includedir}
+    cp -r ${S}/usr/include/* ${D}${includedir}
+}
+
+FILES:${PN}-dev = "/usr/include/*"
+FILES:${PN} = "/usr/lib/*"
+
+#Skips check for .so symlinks
+INSANE_SKIP:${PN} = "dev-so"
--- a/recipes-multimedia/camx/camxlib-kt_1.0.bb
+++ b/recipes-multimedia/camx/camxlib-kt_1.0.bb
@@ -1,0 +1,43 @@
+SUMMARY = "Qualcomm camera User space libraries"
+
+DESCRIPTION = "Collection of prebuilt libraries to support camera downstream functionality."
+
+LICENSE = "CLOSED"
+
+# no top-level dir in the archive, unpack to subdir to prevent UNPACKDIR pollution
+SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/camx.qclinux.0.0/251030/prebuilt_yocto/${BPN}_${PV}_armv8-2a.tar.gz;subdir=${BPN}-${PV}"
+
+SRC_URI[sha256sum] = "a4d6ac09e44a2f1cafe37031c1dd0ab7788b987258bfcd42d368ce5b6c8fdba1"
+
+DEPENDS += "fastrpc protobuf-native protobuf protobuf-native libxml2  virtual/egl virtual/libgles2"
+
+# Install lib file at right location
+do_install() {
+    install -d ${D}${libdir}
+    cp -r ${S}/usr/lib/* ${D}${libdir}
+
+    install -d ${D}/lib/firmware
+    cp -r ${S}/lib/firmware/* ${D}/lib/firmware
+}
+
+relocate_firmware_files () {
+    install -d ${D}${nonarch_base_libdir}/firmware/
+    mv ${D}/lib/* ${D}${nonarch_base_libdir}/
+    rm -rf ${D}/lib/
+}
+do_install[postfuncs] += "relocate_firmware_files"
+
+FILES:${PN} = "\
+    /usr/lib/* \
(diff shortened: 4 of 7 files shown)


recipes-multimedia/camx/camx-kt_1.0.bb
```diff
@@ -0,0 +1,32 @@
...
+    cp -r ${S}/usr/lib/* ${D}${libdir}
+
+    install -d ${D}${includedir}
+    cp -r ${S}/usr/include/* ${D}${includedir}
+
+}
+
+FILES:${PN} = "\
+    /usr/lib/* "
+FILES:${PN}-dev = "/usr/include/*"
```
lumag, line 26: No, thank you. PLease keep the package split as is.

recipes-multimedia/camx/camx-kt_1.0.bb
```diff
@@ -0,0 +1,32 @@
...
+
+
+INSANE_SKIP:${PN} = "dev-so"
+INSANE_SKIP:${PN} += "file-rdeps"
```
lumag, line 30: Why?

recipes-multimedia/camx/camx_1.0.bb
```diff
@@ -0,0 +1,39 @@
...
+
+}
+
+
+#Skips check for archtecture
+#INSANE_SKIP:${PN} = "arch"
+# The modules require .so to be dynamicaly loaded
+#INSANE_SKIP:${PN} += "dev-so"
+# Prebuilt libraries are already stripped
+#INSANE_SKIP:${PN} += "already-stripped"
```
lumag, line 31: Remove the commented parts, please.

recipes-multimedia/camx/camx-kt_1.0.bb
```diff
@@ -0,0 +1,32 @@
...
+
+
```
lumag, line 32: Extra empty lines

recipes-multimedia/camx/camx_1.0.bb
```diff
@@ -0,0 +1,39 @@
...
+    install -d ${D}${libdir}
+    cp -r ${S}/usr/lib/* ${D}${libdir}
+
+    install -d ${D}${includedir}
+    cp -r ${S}/usr/include/* ${D}${includedir}
+
+
+}
+
+
```
lumag, line 25: Extra empty line

recipes-multimedia/camx/camx_1.0.bb
```diff
@@ -0,0 +1,39 @@
...
+#Skips check for archtecture
+#INSANE_SKIP:${PN} = "arch"
+# The modules require .so to be dynamicaly loaded
+#INSANE_SKIP:${PN} += "dev-so"
+# Prebuilt libraries are already stripped
+#INSANE_SKIP:${PN} += "already-stripped"
+
+FILES:${PN} = "\
+    /usr/lib/*"
+FILES:${PN}-dev = "/usr/include/*"
```
lumag, line 35: No need to.

recipes-multimedia/camx/camx_1.0.bb
```diff
@@ -0,0 +1,39 @@
...
+
+#Skips check for .so symlinks
+INSANE_SKIP:${PN} = "dev-so"
```
lumag, line 38: Why? Don't put no-SONAME files into the lib package.

recipes-multimedia/camx/camxcommon_1.0.bb
```diff
@@ -0,0 +1,26 @@
...
+
+    install -d ${D}${includedir}
+    cp -r ${S}/usr/include/* ${D}${includedir}
+}
+
+FILES:${PN}-dev = "/usr/include/*"
+FILES:${PN} = "/usr/lib/*"
+
+#Skips check for .so symlinks
+INSANE_SKIP:${PN} = "dev-so"
```
lumag, line 26: There is a certain pattern. Either provide a reason or drop all of theese.

recipes-multimedia/camx/camxlib-kt_1.0.bb
```diff
@@ -0,0 +1,43 @@
...
+
+DEPENDS += "fastrpc protobuf-native protobuf protobuf-native libxml2  virtual/egl virtual/libgles2"
+
+# Install lib file at right location
+do_install() {
+    install -d ${D}${libdir}
+    cp -r ${S}/usr/lib/* ${D}${libdir}
+
+    install -d ${D}/lib/firmware
+    cp -r ${S}/lib/firmware/* ${D}/lib/firmware
```
lumag, line 20: No. All the firmware should be coming from linux-firmware. We can accept it temporarily here, but please send a pull request to linux-firmware.

recipes-multimedia/camx/camxlib-kt_1.0.bb
```diff
@@ -0,0 +1,43 @@
...
+}
+
+relocate_firmware_files () {
+    install -d ${D}${nonarch_base_libdir}/firmware/
+    mv ${D}/lib/* ${D}${nonarch_base_libdir}/
+    rm -rf ${D}/lib/
+}
+do_install[postfuncs] += "relocate_firmware_files"
```
lumag, line 28: Can't you just put that do `do_install`?

recipes-multimedia/camx/camxlib_1.0.bb
```diff
@@ -0,0 +1,41 @@
...
+    rm -rf ${D}/lib/
+}
+do_install[postfuncs] += "relocate_firmware_files"
+
+FILES:${PN} = "\
+    /usr/lib/* \
+    ${nonarch_base_libdir}/firmware/*"
+
+#Skips check for archtecture
+INSANE_SKIP:${PN} = "arch"
```
lumag, line 37: Move the firmware to the separate package.

lumag:
Also, is it actually 1.0? Can we assume that the next upload will be 1.1? Or 1.0.1? Or something similar? We really expect versions to be semantic, not just random `1.0` all over the place.

recipes-multimedia/camx/camxcommon_1.0.bb
gkhose-qipl: @lumag ,
Below is a reference error that appears in each recipe since all libraries are versioned, and all .so files along with their symlinks are included in the regular package. As a result, we are adding the above line to avoid making changes.
: QA Issue: non -dev/-dbg/nativesdk- package camxlib contains symlink .so '/usr/lib/camera/components/com.qti.node.swregistration.so' [dev-so]

recipes-multimedia/camx/camxlib-kt_1.0.bb
gkhose-qipl: Sure, I'll move it. In the internal recipe, we don't need do_install() since we're using the internal bbclass, so only relocate_firmware_files was included. However, in meta-qcom, we are using do_install, so I'll move it to that function.

recipes-multimedia/camx/camxcommon_1.0.bb
lumag: Neither commit message nor recipe mentions anything. Please provide a comment in both places. Provide more exact path patterns in the FILES sections together with the comment there too.

ricardosalveti:
Drop the merge commits.

recipes-multimedia/camx/camx-kt_1.0.bb
```diff
@@ -0,0 +1,32 @@
+SUMMARY = "Qualcomm camera User space libraries"
+
```
ricardosalveti, line 2: Please remove this empty line.

recipes-multimedia/camx/camx-kt_1.0.bb
```diff
@@ -0,0 +1,32 @@
...
+DESCRIPTION = "Collection of prebuilt libraries to support camera downstream functionality."
+
```
ricardosalveti, line 4: Same.

recipes-multimedia/camx/camx-kt_1.0.bb
```diff
@@ -0,0 +1,32 @@
...
+LICENSE = "CLOSED"
+
+# no top-level dir in the archive, unpack to subdir to prevent UNPACKDIR pollution
+SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/camx.qclinux.0.0/251030/prebuilt_yocto/${BPN}_${PV}_armv8-2a.tar.gz;subdir=${BPN}-${PV}"
+
+SRC_URI[sha256sum] = "6deebd1220b22c0381a0d32b09e2413ca6e0ab8445c6c5ef43b5cec5c25fe62c"
+
+DEPENDS += "camxlib-kt fastrpc libxml2 protobuf-native protobuf abseil-cpp"
+
```
ricardosalveti, line 13: Drop extra empty line.

recipes-multimedia/camx/camx-kt_1.0.bb
```diff
@@ -0,0 +1,32 @@
...
+
+do_install() {
+    install -d ${D}${libdir}
+    cp -r ${S}/usr/lib/* ${D}${libdir}
+
+    install -d ${D}${includedir}
+    cp -r ${S}/usr/include/* ${D}${includedir}
+
```
ricardosalveti, line 21: Again.

recipes-multimedia/camx/camx_1.0.bb
```diff
@@ -0,0 +1,39 @@
+SUMMARY = "Qualcomm camera User space libraries"
```
ricardosalveti, line 1: Same summary as camx-kt?

recipes-multimedia/camx/camx_1.0.bb
```diff
@@ -0,0 +1,39 @@
...
+
+DESCRIPTION = "Collection of prebuilt libraries to support camera downstream functionality."
```
ricardosalveti, line 3: Same.

recipes-multimedia/camx/camxcommon_1.0.bb
```diff
@@ -0,0 +1,26 @@
+SUMMARY = "Qualcomm camera User space libraries"
+
+DESCRIPTION = "Collection of prebuilt libraries to support camera downstream functionality."
+
```
ricardosalveti, line 4: Same comments.

recipes-multimedia/camx/camxlib-kt_1.0.bb
```diff
@@ -0,0 +1,43 @@
+SUMMARY = "Qualcomm camera User space libraries"
+
+DESCRIPTION = "Collection of prebuilt libraries to support camera downstream functionality."
+
```
ricardosalveti, line 4: Same.

recipes-multimedia/camx/camxlib-kt_1.0.bb
ricardosalveti: Firmware in a lib package?

At least create a separated package just for the firmware, until it gets proposed and merged upstream. Also, can you explain why this was not yet submitted to linux-firmware?

ricardosalveti:
A bunch of duplication here, please be more specific about why we have so many recipes, their description, purposes, etc.

Also, can't we just include a makefile of some sort in the tarball itself? Then we could avoid a bunch of custom do_install functions (also helps packaging the same thing for other distros).

lumag:
Okay. I took a (slightly) deeper glance.
We discussed several months ago at Hyderabad, that all different CamX versions should be co-installable. Instead, `camxcommon` conflicts with `camxlib-kt`. 

The recipe is marked as having the version 1.0. However a quick check points out that CamX artifacts have been uploaded to QArtifactory several times, all having 1.0 in the filename, but with the different binary contents. The history of meta-qcom-hwe also shows that the binaries were blindly uploaded to the new location without increasing the version. This will break user's download cache (and I'm surprised that nobody complained up to now).

CamX packages include firmware files (they should be clearled and uploaded to linux-firmware).

CamX packages include DSP modules installed to `/usr/lib/rfsa`, where nothing will read them.

All the libraries have the `.so.0.1.0` suffix. Granted that you want to load them using the `.so` file rather than `.so.0` all those versions are just dummy.

The cherry on the top is a binary copy of the OpenCV library, however the recipe doesn't mention that at all.If I read it correctly, such a redistribution breaks OpenCV licence.

Last, but not least, the camx module doesn't build. After adding a missing dt-bindings header, the build still fails becaue if CONFIG_DMABUF_HEAPS is not enabled (and it is usually not enabled), the drivers silently tries to use ION (which is not available in upstream).

Having all of those issues in mind, these PRs don't look like code preview nor like an RFC. I'm going to close all CamX-related PRs. Please reopen them when:
- the code actually builds
- ... for all the kernels in meta-qcom
- the code is actually integrated for coexistence
- ... yes, that's a requirement.
- the tarballs are properly versioned
- tarball contents is also properly versioned and uses SOVERSIONs in a sensible way
- you have clear assessment of all the software which is included and of all the license terms

recipes-multimedia/camx/camxlib-kt_1.0.bb
ndechesne: @lumag this is an interesting situation... the firmware is needed by a driver which is not going upstream (the ISP driver). So it does not really make sense to submit to linux-firmware, since there is no 'user' upstream of the firmware. if/when there is a upstream driver for ISP, it would of course be different.

lool: It would be nice to manage all these extra firmware in a homogenous way, we could package them in one go as Yocto recipe or deb package; qcom-linux-firmware-extras or something

lumag: I'd prefer having a separate, camx-firmware recipe, providing one package per target platform.
Note, firmware should go into a correct location, `/lib/firmare/qcom/SoC/`

lumag: To bring it from the email:

We have discussed this internally. Let's lift up the linux-firmware
requirement for now, as camss doesn't need this firmware. The rest of
the issues with distribution and containers still apply so we'd kindly
ask to:
- a separate tarball (or tarballs) on QArtifactory with ICP firmware files
- firmware files need to be properly licensed, allowing redistribution,
e.g. LICENSE.qcom-2
- separate recipes handling the firmware, independently of the libraries

PR #1180, opened 2025-10-31
linux-qcom-next/linux-yocto-dev: enable more use cases

koenkooi:
Enable the following kernel features:

- ZRAM, when configured by userspace, can provide a compressed swapspace backed by regular RAM. This generally increases the amount of available memory, avoiding the need for disk-backed swap. When it isn't configured by userspace, this does nothing.
- In-kernel wifi infrastructure, this makes the `iwd` backend for connman/networkmanager/systemd work
- Binfmt-misc, this makes podman, FEX and other software that wants transparent interpreter remapping work


lumag:
> > Should it be also enabled for linux-yocto / linux-yocto-dev? Also, please rebase to fix build issue.
> 
> I've added it to linux-yocto-dev and rebased. Do we want this in linux-yocto non-dev as well? The current .scc is empty, so it seems to be out of sync already :/

The scc is there to pull out the rest of the files.

koenkooi:
> How likely are we to get these config changes accepted upstream?

I am unsure, Linus himself has been pushing back against various additions to the defconfigs saying distros should enable and test the features they want. I don't know where he draws the line and how much attention he pays to this.

lool:
> > How likely are we to get these config changes accepted upstream?
> 
> This sums up the situation quite nicely, https://lwn.net/Articles/1026337/, especially this excerpt:
> 
> > Torvalds, in turn, [strongly told Molnar to stop this work](https://lwn.net/ml/all/CAHk-=winzTt3SCzv8BWGMm0fzrXS0gb59gK0h4dAe0L6hj3X_w@mail.gmail.com), saying that the default configuration should be for "normal people". Options that are useful to cloud providers (such as the virtualization subsystems) should not be enabled, he said. The fact that all distributors enable a specific option is also not, in his mind, an argument for enabling that option in the default configuration.

I had Yocto as an upstream in mind here :)

lumag (approved):

recipes-kernel/linux/linux-qcom-next/configs/bsp-additions.cfg
```diff
@@ -328,3 +328,29 @@ CONFIG_ZRAM_WRITEBACK=y
 CONFIG_ZRAM_TRACK_ENTRY_ACTIME=y
 CONFIG_ZRAM_MEMORY_TRACKING=y
 CONFIG_ZRAM_MULTI_COMP=y
+
+# Wifi
+CONFIG_ASYMMETRIC_KEY_TYPE=y
```
lumag, line 337: I think most of these should be built as modules already. Could you please check / confirm?

koenkooi: I'll check later today, this section of the config is a copy/paste of what iwd told me to enable. I can turn all tristate options in `=m`, but I'd like to keep this list of options to make it easier to lift it into different kernels, that won't have anything enabled yet.

lumag: I'd rather not keep the full list here. It is ridiculously long and verbose. Most of the options are enabled by default. If you list all of them, for me it means that all of them are set incorrectly in the qcom-next config or in the defconfig.

recipes-kernel/linux/linux-qcom-next/configs/bsp-additions.cfg
```diff
@@ -307,3 +307,52 @@ CONFIG_NFT_TPROXY=m
 CONFIG_NFT_TUNNEL=m
 CONFIG_PACKET_DIAG=y
 CONFIG_VETH=m
+
+# ZRAM
+CONFIG_ZRAM=y
```
lumag: Likewise. Is `CONFIG_ZRAM=y` not enough?

recipes-kernel/linux/linux-qcom-next/configs/bsp-additions.cfg
koenkooi: The options that are enabled in linux-qcom-next in its current form are these:
```
CONFIG_ASYMMETRIC_KEY_TYPE=y
CONFIG_ASYMMETRIC_PUBLIC_KEY_SUBTYPE=y
CONFIG_CFG80211_CRDA_SUPPORT=y
CONFIG_CRYPTO_AES=y
CONFIG_CRYPTO_CBC=m
CONFIG_CRYPTO_CMAC=m
CONFIG_CRYPTO_DES=m
CONFIG_CRYPTO_ECB=y
CONFIG_CRYPTO_HMAC=y
CONFIG_CRYPTO_MD5=m
CONFIG_CRYPTO_RSA=y
CONFIG_CRYPTO_SHA1=m
CONFIG_CRYPTO_SHA256=m
CONFIG_CRYPTO_SHA512=y
CONFIG_PKCS7_MESSAGE_PARSER=y
CONFIG_X509_CERTIFICATE_PARSER=y
```

Which means that this specific kernel only needs the following:
```
CONFIG_CRYPTO_ARC4=m
CONFIG_CRYPTO_MD4=m
CONFIG_CRYPTO_USER_API_AEAD=m
CONFIG_CRYPTO_USER_API_HASH=m
CONFIG_CRYPTO_USER_API_SKCIPHER=m
CONFIG_KEY_DH_OPERATIONS=y
CONFIG_PKCS8_PRIVATE_KEY_PARSER=y
```

I'd still like to keep the complete list, since that makes it easier and less error-prone to copy this between kernels (linux-qcom-next, linux-yocto-dev, linux-yocto and vendor kernels).

I also don't trust upstream linux to not regress defconfigs, especially for things we currently don't test.

lumag: If you want a comprehensive list, please also enable ARM-specific crypto acceleration options.
At the same time, enabling obsolete & broken algorithms (DES, MD4, MD5) sounds very strange to me.

koenkooi: EAP-PEAP with MsCHAPv2 still uses MD4, EAP-MD5 uses, well MD5 and 3DES is sadly still allowed as an alternative to AES when using 802.1x :(

I'll have a look at the ARM-specific crypto options.

recipes-kernel/linux/linux-qcom-next/configs/bsp-additions.cfg
```diff
@@ -311,3 +311,24 @@ CONFIG_NFT_TPROXY=m
...
+CONFIG_ZRAM_BACKEND_LZ4=y
+CONFIG_ZRAM_BACKEND_LZ4HC=y
+CONFIG_ZRAM_BACKEND_ZSTD=y
+CONFIG_ZRAM_BACKEND_DEFLATE=y
+CONFIG_ZRAM_BACKEND_842=y
+CONFIG_ZRAM_BACKEND_LZO=y
+# CONFIG_ZRAM_DEF_COMP_LZORLE is not set
+CONFIG_ZRAM_DEF_COMP_LZO=y
```
ricardosalveti: Can't we move all zram options to module instead?

koenkooi: Sure!

PR #1181, opened 2025-10-31
Include downstream camera dtb into image

gkhose-qipl:
8275-iq-8275-evk.conf: include downstream camera dtb into image
qcs8300-ride-sx.conf: include downstream camera dtb into image
9075-iq-9075-evk.conf: include downstream camera dtb into image
qcs9100-ride-sx.conf: include downstream qcs9100 camera dtb into image

### DIFF
--- a/conf/machine/iq-8275-evk.conf
+++ b/conf/machine/iq-8275-evk.conf
@@ -13,6 +13,10 @@ KERNEL_DEVICETREE ?= " \
                       qcom/monaco-evk-camera-imx577.dtb \
                       "
 
+KERNEL_DEVICETREE:append:pn-linux-qcom-next = " \
+                      qcom/monaco-evk-camx.dtb \
+                      "
+
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-iq-8275-evk-firmware \
     packagegroup-iq-8275-evk-hexagon-dsp-binaries \


conf/machine/qcs8300-ride-sx.conf
```diff
@@ -9,6 +9,7 @@ MACHINE_FEATURES += "efi pci"
 QCOM_DTB_DEFAULT ?= "qcs8300-ride"
 
 KERNEL_DEVICETREE ?= " \
+                      qcom/qcs8300-ride-camx.dtb \
```
ricardosalveti: Add bellow qcs8300-ride.dtb, don't make it the first option.

conf/machine/qcs9100-ride-sx.conf
```diff
@@ -9,10 +9,14 @@ MACHINE_FEATURES += "efi pci"
 QCOM_DTB_DEFAULT ?= "qcs9100-ride-r3"
 
 KERNEL_DEVICETREE ?= " \
-                      qcom/qcs9100-ride.dtb \
+                      qcom/qcs9100-ride-camx.dtb \
+                      qcom/qcs9100-ride-r3-camx.dtb \
```
ricardosalveti: Same here.

gkhose-qipl:
camera compilation may fail and these PRs are being raised solely for code review purposes.
will move PR to open state once compilation is working fine.

gkhose-qipl:
> Which recipe will be responsible for providing these dtbs?
> 
> Here they are added by default, so it should work basically with any kernel the user might end up using.

recipe name:- linux-qcom-next
Yes, it should work, since the core kernel recipe will generate all of these camx DTBs.

lumag (commented):
```
Co-developed-by: Chandan Jha <cjha@qti.qualcomm.com>
Signed-off-by: Chandan Jha <cjha@qti.qualcomm.com>
Signed-off-by: Ganesh Khose <gkhose@qti.qualcomm.com>
```

Is this for the one-line patch? Two developers?

gkhose-qipl:
> > recipe name:- linux-qcom-next
> > Yes, it should work, since the core kernel recipe will generate all of these camx DTBs.
> 
> Right, but if an user ends up selecting a different kernel recipe (e.g. linux-yocto, linux-yocto-dev) then the build will probably fail due missing dtb/dtbo.
> 
> If such files are indeed only available when linux-qcom-next is used, then it needs to be bound to it.

done, we have added check to bound with linux-qcom-next

PR #1182, opened 2025-11-03
sm8750-mtp: add initial changes for sm8750-mtp Board

vivpuar:
Adding below changes for sm8750-mtp Board:

- KAS configuration file
- machine configuration
- set KERNEL_DEVICETREE with default dtb to pack DTB inside UKI

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -115,6 +115,7 @@ jobs:
           - qcs9100-ride-sx
           - qrb2210-rb1-core-kit
           - qcom-armv7a
+          - sm8750-mtp
         distro:
           - name: poky-altcfg
             yamlfile: ""
--- a/ci/sm8750-mtp.yml
+++ b/ci/sm8750-mtp.yml
@@ -1,0 +1,8 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+  includes:
+  - ci/base.yml
+
+machine: sm8750-mtp


conf/machine/include/qcom-sm8750.inc
```diff
@@ -0,0 +1,12 @@
+# Configurations and variables for SM8750 SoC family.
+
+SOC_FAMILY = "sm8750"
+require conf/machine/include/qcom-base.inc
+require conf/machine/include/qcom-common.inc
+
+DEFAULTTUNE = "armv8-2a-crypto"
+require conf/machine/include/arm/arch-armv8-2a.inc
```
lumag: Why is it just ARMv8.2-A?

recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -17,6 +17,7 @@ UKI_CMDLINE += "${@d.getVar('KERNEL_CMDLINE_EXTRA') or ''}"
 
 # Remove 'upstream' dtb, rely on EFI provided one
 KERNEL_DEVICETREE = ""
+KERNEL_DEVICETREE:sm8750-mtp = "${QCOM_DTB_DEFAULT}.dtb"
```
lumag, line 20: How does it work on other machines?

lumag (changes requested):
Also, missing commit messages.

conf/machine/include/qcom-sm8750.inc
vivpuar: SM8750 uses Oryon CPU architecture which is based on Nuvia cores, so added support for armv8-2a. Please let me know if its not correct, will update accordingly.

recipes-kernel/images/esp-qcom-image.bb
vivpuar: > How does it work on other machines?

On other machines UEFI has a support to look for dtb partition, while this support is not there on SM8750, so we need to pack dtb inside UKI.

conf/machine/include/qcom-sm8750.inc
lumag: It is not.

vivpuar: Is it armv9-2a?

lumag: No, as far as I know.

recipes-kernel/images/esp-qcom-image.bb
lumag: => commit message. It needs to explain, why do you need the change.

vivpuar: Updated the commit message

conf/machine/include/qcom-sm8750.inc
vivpuar: Referred internal docs and checked with the POC regarding the arm version support on Pakala, updated the version.

vivpuar:
> Please use imperative language. Embed instead of 'we embed', 'configure' rather than 'will be configured', etc.

Updated commit msg

lumag (approved):

sbanerjee-quic (approved):

lumag:
> Partition definition and boot firmware is still missing here, should we really merge as is?

I would say, the firmware at all is a big question. I don't think we have anything for SM8750 at this point.

PR #1183, opened 2025-11-04
linux-qcom-next configs:Set CONFIG_CLK_KAANAPALI_TCSRCC as enabled

yijiyang:


recipes-kernel/linux/linux-qcom-next/configs/qcom.cfg
```diff
@@ -307,3 +307,4 @@ CONFIG_NFT_TPROXY=m
 CONFIG_NFT_TUNNEL=m
 CONFIG_PACKET_DIAG=y
 CONFIG_VETH=m
+CONFIG_CLK_KAANAPALI_TCSRCC=y
```
vkraleti: This section isn't appropriate for this config. Assuming it is needed for UFS on Kaanapali add it under "# for UFS enumeration"

yijiyang: Updated according to your suggestion.

koenkooi:
"No description provided." Can we get at least some form of description in the commit message saying why this change is made and which platforms are affected?

ricardosalveti:
> This configuration is specific to Qualcomm and will not be upstreamed.

What do you mean Qualcomm specific? Even when available upstream it could be specific to Qualcomm, which is perfectly fine.

Please explain why this is not going to be enabled / available upstream as part of the commit message.

yijiyang:
> > Commit message upated
> 
> It still doesn't explain, why the change is performed here and not in the qcom-next itself.

Will explain.

> 
> Also, please take a look at git log, your message doesn't follow established practice. E.g. the 'configs' in the subject is totally out of place.

I followed this commit as a reference. If this approach is not considered standard, could you please point me to the established best practice?
https://github.com/qualcomm-linux/meta-qcom/commit/2717bd9566546b15cafa0d3e5f1a16fdc79544c9

aiquny (commented):
linux-qcom-next:configs:Set CONFIG_CLK_KAANAPALI_TCSRCC as enabled

Kaanapali is not sharing same config with sm8750 of CONFIG_CLK_8750_TCSRCC, and instead it has its own CONFIG_CLK_KAANAPALI_TCSRCC. So apart from current CONFIG_CLK_8750_TCSRCC enabled, CONFIG_CLK_KAANAPALI_TCSRCC is enabled for kaanapali.

Signed-off-by: Yijie Yang <yijie.yang@oss.qualcomm.com>

lumag:
@aiquny I didn't understand your comment regarding internal GitHub. 
Anyway: 
- please drop 'configs' from the subject
- in the commit message please describe the issue
- please briefly mention, why it is being fixed here and not in qcom-next.

yijiyang:
> @aiquny I didn't understand your comment regarding internal GitHub. Anyway:
> 
> * please drop 'configs' from the subject
> * in the commit message please describe the issue
> * please briefly mention, why it is being fixed here and not in qcom-next.

Updated, please check.

lumag (commented):
> CONFIG_CLK_KAANAPALI_TCSRCC is a required dependency for
CONFIG_SCSI_UFS_QCOM

No, it's not. Otherwise it would have been handles by Kconfig.

aiquny:
> > > > Kaanapali does not share the same configuration as SM8750 for
> > > > CONFIG_CLK_8750_TCSRCC. Instead, it uses its own configuration:
> > > > CONFIG_CLK_KAANAPALI_TCSRCC.
> > > 
> > > 
> > > Is this a problem you are trying to solve?
> > 
> > 
> > It is a reason why current change is needed. The previous issue was resolved by the commit [[2717bd9](https://github.com/qualcomm-linux/meta-qcom/commit/2717bd9566546b15cafa0d3e5f1a16fdc79544c9)]
> 
> @aiquny how is this different from the patches on the kernel mailing lists? I'd have understood your comment if we were not discussing this with your team regarding the kernel commits for the last several months. The commit message should start with the description of a problem or an issue that is to be solved.

Could you help to have specific how the problem would be here? The whole history is:
1. the commit here [https://github.com/qualcomm-linux/meta-qcom/commit/2717bd9566546b15cafa0d3e5f1a16fdc79544c9] have the issue description already to support kaanapali root filesystem needed. Since the solution given is to have "CONFIG_CLK_8750_TCSRCC" accepted. And it is explicitly says support kaanapali in this commit as well.

2. The current patch is to "Fixes: 2717bd9566546b15cafa0d3e5f1a16fdc79544c9", since "CONFIG_CLK_8750_TCSRCC" for kaanapali is not true. It is changed to CONFIG_CLK_KAANAPALI_TCSRCC.

lumag:
> > > > > Kaanapali does not share the same configuration as SM8750 for
> > > > > CONFIG_CLK_8750_TCSRCC. Instead, it uses its own configuration:
> > > > > CONFIG_CLK_KAANAPALI_TCSRCC.
> > > > 
> > > > 
> > > > Is this a problem you are trying to solve?
> > > 
> > > 
> > > It is a reason why current change is needed. The previous issue was resolved by the commit [[2717bd9](https://github.com/qualcomm-linux/meta-qcom/commit/2717bd9566546b15cafa0d3e5f1a16fdc79544c9)]
> > 
> > 
> > @aiquny how is this different from the patches on the kernel mailing lists? I'd have understood your comment if we were not discussing this with your team regarding the kernel commits for the last several months. The commit message should start with the description of a problem or an issue that is to be solved.
> 
> Could you help to have specific how the problem would be here? The whole history is:
> 
> 1. the commit here [https://github.com/qualcomm-linux/meta-qcom/commit/2717bd9566546b15cafa0d3e5f1a16fdc79544c9] have the issue description already to support kaanapali root filesystem needed. Since the solution given is to have "CONFIG_CLK_8750_TCSRCC" accepted. And it is explicitly says support kaanapali in this commit as well.
> 2. The current patch is to "Fixes: [2717bd9](https://github.com/qualcomm-linux/meta-qcom/commit/2717bd9566546b15cafa0d3e5f1a16fdc79544c9)", since "CONFIG_CLK_8750_TCSRCC" for kaanapali is not true. It is changed to CONFIG_CLK_KAANAPALI_TCSRCC.

Something in the style of: "On Kaanapali the UFS uses clocks/resets/something else provided by the TCSR clock controller. Before qcom-next 5.9.10 corresponding driver was shared with the TCSR clock controller driver for foo. In qcom-next 6.10.11 support for TCSR CC on Kaanapli was split to a separate driver becase of ....., which made UFS to fail to probe on name-the-device. Enable CONFIG_BAR_BAZ in order to let name-the-device boot from UFS."

And then, as you can see from this description, the Fixes tag is incorrect, because the mentioned commit wasn't broken. It was some other commit in another recipe which broke boot support.

yijiyang:
> 1. https://github.com/qualcomm-linux/meta-qcom/commit/2717bd9566546b15cafa0d3e5f1a16fdc79544c9

Updated

aiquny (approved):

koenkooi (approved):

PR #1184, opened 2025-11-04
linux-yocto: add support to unversioned FILESEXTRAPATHS

quaresmajose:
The FILESEXTRAPATHS content may be agnostic to the version, so add the corresponding path.
The search continues prioritizing the versioned if it exists.

- Move the qcom.scc and the bsp/* fragments to the unversioned path.


recipes-kernel/linux/linux-yocto_%.bbappend
```diff
@@ -3,7 +3,8 @@ COMPATIBLE_MACHINE:qcom = "qcom-armv8a|qcom-armv7a"
 
 BASEVER = "${@ d.getVar('LINUX_VERSION').rpartition('.')[0]}"
 
-FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}-${BASEVER}:"
+# prioritizing the versioned version path if it exists and fallback to the unversioned path
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}-${BASEVER}:${THISDIR}/${PN}"
```
koenkooi: You missed the trailing `:`, so `${THISDIR}/${PN}` will get concatenated with the first entry of the original `${FILESEXTRAPATHS}`

quaresmajose: Thanks, I fixed the missed `:` it in the meantime. There are new linux-yocto 6.16 and 1.17 failures:
```
ERROR: Could not locate BSP definition for qcom-armv8a/standard and no defconfig was provided 
```
something is still broken in linux-yocto I have to pin the version on 6.12

lumag:
I will suggest duplicating the files. We depend on the linux-yocto pulling all configs when we specify just one.

recipes-kernel/linux/linux-yocto_%.bbappend
quaresmajose: the BSP definition was fixed

quaresmajose:
> Continues in #1186

this one is merged

quaresmajose:
I disgree with you. yes, I am "adding support" because before this patch the path on FILESEXTRAPATHS is versioned and now it is also unversioned. All the metadata is the same and it only moves from three versioned paths (6.12, 6.16 and 6.17) to a new one unversioned path. The data on this three difrents paths are duplicated and are the same.
We still have the option to add diferent patches and configs but I agree that having it distributed across different paths can make it more difficult to interpret. On the other hand, having the content duplicated three times doesn't seem very elegant to me either.

quaresmajose:
We will push all this kernel fragments upstream to the yocto-kernel-cache https://github.com/qualcomm-linux/meta-qcom/issues/1198

PR #1186, opened 2025-11-04
linux-yocto: add support for 6.17

quaresmajose:
Adding a 6.17 kernel bbappend and fragments


lumag (changes requested):
Could you please also pickup a workaround patch from 6.16?

PR #1188, opened 2025-11-05
Add resize-rootfs_1.0.bb recipe to install rootfs resize script

vkraleti:
To be able to flash across various boards, the root filesystem is generated 
with a minimal size. This `resize-rootfs.sh` script, automatically resizes the 
rootfs on first boot to utilize the full available disk space.

### DIFF
--- a/recipes-bsp/partition/files/resize-rootfs.sh
+++ b/recipes-bsp/partition/files/resize-rootfs.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-clear
+
+[ $(whoami) = "root" ] || { echo "ERROR: Must be a root user to resize rootfs." && exit 1; }
+
+RESIZE2FS=$(which resize2fs) || { echo "ERROR: 'resize2fs' utility not found." && exit 1; }
+
+ROOT_DEVICE="$(readlink -f "$(findmnt / -o source -n -v)")"
+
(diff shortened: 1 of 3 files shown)


recipes-bsp/partition/files/resize-rootfs.sh
```diff
@@ -0,0 +1,11 @@
...
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-clear
+
+[ $(whoami) = "root" ] || { echo "ERROR: Must be a root user to resize rootfs." && exit 1; }
+
+RESIZE2FS=$(which resize2fs) || { echo "ERROR: 'resize2fs' utility not found." && exit 1; }
+
+ROOT_DEVICE="$(readlink -f "$(findmnt / -o source -n -v)")"
+
+${RESIZE2FS} "${ROOT_DEVICE}"
```
koenkooi, line 11: This is missing a check to see if it's actually ext2/3/4 before running resize2fs.

vkraleti:
> Is this something for the BSP or should it be be in a DISTRO layer?

Added recipe in BSP layer so that it is available for all DISTROs. However the actual inclusion into image is via meta-qcom-distro layer. [qualcomm-linux/meta-qcom-distro/pull/84]

ricardosalveti:
> Is this something for the BSP or should it be be in a DISTRO layer?

Yeah, I also believe this logic should be available and applied by the distro layer instead, there is nothing BSP specific here (and distros might already have a similar logic in place, as we have in meta-lmp).

PR #1191, opened 2025-11-05
Add KVM support in image packaging

### LINKED ISSUE #1154: Allow users to select the KVM-enabled XBL config
Add a switch that would let users select `xbl_config_kvm.elf` (instead of requiring them to go through archives, manually selecting the correct binary, etc.).

This can not be made a default, as it will break some of the video use-cases, but we should make it easy to switch to KVM-enabled builds.

kraghava-qti:
## 🧩 Summary

This pull request introduces support for KVM-aware packaging and build
configuration in the Qualcomm image generation workflow.

---

### 1. **KVM-aware xbl_config.elf packaging**
- **File modified**: `image_types_qcom.bbclass`
- Adds logic to detect if `kvm` is present in `DISTRO_FEATURES`.
- Sets a `KVM` variable accordingly.
- During `qcomflash` packaging, if KVM is enabled:
  - Verifies the presence of `xbl_config_kvm.elf`.
  - Fails the build with an error if the file is missing.
- Ensures that KVM-enabled builds use the correct `xbl_config.elf` variant.

### 2. Refs qualcomm-linux/meta-qcom-distro#86
---

## 🔍 Motivation

These changes are necessary to:
- Enable KVM support in builds that require virtualization.
- Ensure the correct bootloader configuration (`xbl_config_kvm.elf`) is used
  when KVM is enabled.
- Provide a clean and reusable kas configuration for KVM-enabled CI builds.

---

## 🧪 Testing

- Verified that builds with `qcom-distro-kvm.conf` correctly enable virtualization features.
- Confirmed that the packaging step fails gracefully if
  `xbl_config_kvm.elf` is missing.
- Ensured backward compatibility for non-KVM builds.

---

## Fixes #1154

### DIFF
--- a/classes-recipe/image_types_qcom.bbclass
+++ b/classes-recipe/image_types_qcom.bbclass
@@ -104,7 +104,7 @@ create_qcomflash_pkg() {
 
     # boot firmware
     for bfw in `find ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR} -maxdepth 1 -type f \
-            \( -name '*.elf' ! -name 'abl2esp*.elf' \) -o \
+            \( -name '*.elf' ! -name 'abl2esp*.elf' ! -name 'xbl_config*.elf' \) -o \
             -name '*.mbn*' -o \
             -name '*.fv' -o \
             -name 'sec.dat'` ; do
@@ -114,6 +114,16 @@ create_qcomflash_pkg() {
     # abl2esp
     if [ -e "${DEPLOY_DIR_IMAGE}/abl2esp-${ABL_SIGNATURE_VERSION}.elf" ]; then
         install -m 0644 "${DEPLOY_DIR_IMAGE}/abl2esp-${ABL_SIGNATURE_VERSION}.elf" .
+    fi
+
+    # xbl_config
+    xbl_config="xbl_config.elf"
+    if ${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}; then
+        xbl_config="xbl_config_kvm.elf"
+    fi
+
+    if [ -f "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${xbl_config}" ]; then
+        install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${xbl_config}" xbl_config.elf
     fi
 
     # sail nor firmware


classes-recipe/image_types_qcom.bbclass
```diff
@@ -36,6 +36,8 @@ do_image_qcomflash[depends] += "${@ ['', '${QCOM_PARTITION_CONF}:do_deploy'][d.g
 				${@'${QCOM_ESP_IMAGE}:do_image_complete' if d.getVar('QCOM_ESP_IMAGE') != '' else  ''}"
 IMAGE_TYPEDEP:qcomflash += "${IMAGE_QCOMFLASH_FS_TYPE}"
 
+KVM = "${@'1' if 'kvm' in d.getVar('DISTRO_FEATURES', True).split() else '0'}"
```
lumag: Any reason for not using `bb.utils.contains`?

classes-recipe/image_types_qcom.bbclass
```diff
@@ -110,6 +112,14 @@ create_qcomflash_pkg() {
         install -m 0644 ${bfw} .
     done
 
+    # if KVM is enabled, swap the default xbl_config.elf with xbl_config_kvm.elf
```
lumag: No, it doesn't ensure anything. The build will still use the standard XBL config.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -125,3 +135,4 @@ create_qcomflash_pkg() {
 }
 
 create_qcomflash_pkg[vardepsexclude] += "DATETIME"
+create_qcomflash_pkg[vardeps] += "KVM"
```
lumag: Why?

ci/qcom-distro-kvm.yml
```diff
@@ -0,0 +1,8 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+
+local_conf_header:
+  kvm: |
+    DISTRO_FEATURES:append = " virtualization kvm"
```
lumag: I would prefer to have a separate distro instead of patching it via KAS config.

lumag (changes requested):
Please rewrite your commit messages. Start by describing the problem or the issue that be exists. Then describe how to fix it.

classes-recipe/image_types_qcom.bbclass
kraghava-qti: Will update. No reason, just forgot.

classes-recipe/image_types_qcom.bbclass
kraghava-qti: Oops! Thank you. Will add the cp command.

classes-recipe/image_types_qcom.bbclass
kraghava-qti: My original intent was to keep this task dependent only on the kvm aspect DISTRO_FEATURES. But after some thinking this dependency no longer makes sense as changing any aspect of DISTRO_FEATURES may require a complete image rebuild for the flags to take effect. Will remove this.

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
```diff
@@ -110,6 +112,16 @@ create_qcomflash_pkg() {
         install -m 0644 ${bfw} .
     done
 
+    if [ "${KVM_ENABLED}" = "1" ]; then
+        if [ ! -f xbl_config_kvm.elf ]; then
+            bberror "xbl_config_kvm.elf NOT found! KVM is NOT supported on current machine ${MACHINE}"
+            exit 1
+        fi
```
quaresmajose: you can use the `bbfatal`  to avoid the exit call

classes-recipe/image_types_qcom.bbclass
```diff
@@ -36,6 +36,8 @@ do_image_qcomflash[depends] += "${@ ['', '${QCOM_PARTITION_CONF}:do_deploy'][d.g
 				${@'${QCOM_ESP_IMAGE}:do_image_complete' if d.getVar('QCOM_ESP_IMAGE') != '' else  ''}"
 IMAGE_TYPEDEP:qcomflash += "${IMAGE_QCOMFLASH_FS_TYPE}"
 
+KVM_ENABLED = "${@bb.utils.contains('DISTRO_FEATURES', 'kvm', '1', '0', d)}"
```
quaresmajose: This variable is only use one time so better to use the bb.utils.contains directly on the if clause below and drop the variable

classes-recipe/image_types_qcom.bbclass
```diff
@@ -110,6 +112,16 @@ create_qcomflash_pkg() {
...
     done
 
+    if [ "${KVM_ENABLED}" = "1" ]; then
+        if [ ! -f xbl_config_kvm.elf ]; then
+            bberror "xbl_config_kvm.elf NOT found! KVM is NOT supported on current machine ${MACHINE}"
+            exit 1
+        fi
+
+        # if KVM is enabled, overwrite the default xbl_config.elf with xbl_config_kvm.elf
+        cp xbl_config_kvm.elf xbl_config.elf
```
lumag: The rest of the file uses `install`. Why are you using `cp`?

classes-recipe/image_types_qcom.bbclass
lumag: Please make sure that only one XBL config gets installed.

classes-recipe/image_types_qcom.bbclass
kraghava-qti: `install` preserves the file attributes and permissions supplied on the cmd line. `cp` does not. I did not see the need to preserve the file meta data attributes. Either I use `cp -p` or I use `install` to preserve the src file attributes to the destination and prevent their change, that may happen by using a raw `cp`.

classes-recipe/image_types_qcom.bbclass
kraghava-qti: I guess no overwrite :)

classes-recipe/image_types_qcom.bbclass
lumag: Okay, it's more of 'why you are using cp, while the rest of the class uses install'. Please make your code consistent with the rest of the file.

classes-recipe/image_types_qcom.bbclass
lumag: I don't follow your comment.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -102,14 +104,28 @@ create_qcomflash_pkg() {
...
-            -name '*.mbn*' -o \
-            -name '*.fv' -o \
-            -name 'sec.dat'` ; do
+    for bfw in $(find ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR} \
+        -maxdepth 1 -type f \
+        \( -name '*.elf' \
+        -o -name '*.mbn*' \
+        -o -name '*.fv' \
+        -o -name 'sec.dat' \) \
+        ! -name 'xbl_config*.elf'); do
```
lumag: Move this to the first line.  Only *.elf files match the `xbl_config*.elf` pattern.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -102,14 +104,28 @@ create_qcomflash_pkg() {
...
         install -m 0644 ${bfw} .
     done
 
+    xbl_config_file="${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config.elf"
+    kvm_enabled="${@bb.utils.contains('DISTRO_FEATURES', 'kvm', '1', '0', d)}"
+    if [ "${kvm_enabled}" = "1" ]; then
```
lumag: `if [ "${@bb.utils.contains('DISTRO_FEATURES"....`

classes-recipe/image_types_qcom.bbclass
```diff
@@ -102,14 +104,28 @@ create_qcomflash_pkg() {
...
+        xbl_config_file="${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config_kvm.elf"
+    fi
+
+    if [ -f "${xbl_config_file}" ]; then
```
lumag: We don't perform checks in other places. Let `install` complain on its own.

classes-recipe/image_types_qcom.bbclass
kraghava-qti: no you have uefi.elf, shrm.elf, imagefv.elf that are NON xbl_config*.elf that need to be copied.

lumag: And?

kraghava-qti: and many more... Never mind. Please ignore. Will move to the top.

kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: changed the logic

classes-recipe/image_types_qcom.bbclass
kraghava-qti: changed it to install. done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
```diff
@@ -102,14 +102,26 @@ create_qcomflash_pkg() {
...
-    for bfw in `find ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR} -maxdepth 1 -type f \
-            -name '*.elf' -o \
-            -name '*.mbn*' -o \
-            -name '*.fv' -o \
-            -name 'sec.dat'` ; do
+    for bfw in $(find ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR} \
+            -maxdepth 1 -type f \
+            ! -name 'xbl_config*.elf' \
+            \( \
+                -name '*.elf' \
```
lumag: `find .. -type f \( -name '*.elf' ! -name 'xbl_config*.elf' \) -o -name '*.mbn' ....`

kraghava-qti: you wanted to group elf related matching together. Will change.

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
```diff
@@ -102,14 +102,23 @@ create_qcomflash_pkg() {
     fi
```
quaresmajose, line 104: I belive you can do the same with only one move before the for loop and you can discard everything else
```
if [ "${@bb.utils.contains('DISTRO_FEATURES', 'kvm', '1', '0', d)}" = "1" ]; then
    mv ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config_kvm.elf \
        ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config.elf 
fi
```

kraghava-qti: The concern I have about this is that mv when done might not allow the src file to be present for further image rebuilds, unless src is again regenerated as part of the build. Say I clean the image recipe and rebuild the image recipe again, since mv has moved the src it would attempt to move a non existing src on the 2nd build. Please correct if wrong.

lumag: Yep. Don't touch ${DEPLOY_DIR_IMAGE}, those files are managed by sstate and not by the recipes.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -102,14 +102,23 @@ create_qcomflash_pkg() {
...
 
     # boot firmware
-    for bfw in `find ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR} -maxdepth 1 -type f \
-            -name '*.elf' -o \
-            -name '*.mbn*' -o \
-            -name '*.fv' -o \
-            -name 'sec.dat'` ; do
+    for bfw in $(find ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR} \
```
lumag: Please don't introduce unnecessary changes.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -102,14 +102,23 @@ create_qcomflash_pkg() {
...
+            \( -name '*.elf' ! -name 'xbl_config*.elf' \) \
+            -o -name '*.mbn*' \
+            -o -name '*.fv' \
+            -o -name 'sec.dat'); do
+
         install -m 0644 ${bfw} .
     done
 
+    xbl_config_file="${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config.elf"
+    if [ "${@bb.utils.contains('DISTRO_FEATURES', 'kvm', '1', '0', d)}" = "1" ]; then
```
lumag: Just if KVM install this else install that fi.

kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done. removed unneeded changes

ci/qcom-distro-kvm.yml
kraghava-qti: done. created a seperate kas yml for kvm based qcom distro that inherits from qcom distro.

classes-recipe/image_types_qcom.bbclass
quaresmajose: Rigth! forgot this piont that this elf files were not created here in this image step and the move will also break the sstate.

ci/qcom-distro-kvm.yml
lumag: The comment still applies.

vkraleti: @kraghava-qti ask is to create a new qcom-distro-kvm.conf in meta-qcom-distro.

kraghava-qti: ack. in progress

kraghava-qti: @lumag @vkraleti 

https://github.com/qualcomm-linux/meta-qcom-distro/pull/86

classes-recipe/image_types_qcom.bbclass
ricardosalveti: You can use `if ${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}; then` instead.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -103,13 +103,19 @@ create_qcomflash_pkg() {
...
-            -name '*.elf' -o \
+            \( -name '*.elf' ! -name 'xbl_config*.elf' \) -o \
             -name '*.mbn*' -o \
             -name '*.fv' -o \
             -name 'sec.dat'` ; do
         install -m 0644 ${bfw} .
     done
 
+    if [ "${@bb.utils.contains('DISTRO_FEATURES', 'kvm', '1', '0', d)}" = "1" ]; then
+        install -m 0644 ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config_kvm.elf .
```
ricardosalveti: This won't make it the default one that is actually flashed, since partitions.conf is still looking for xbl_config.elf. If you want it to be the default you will have to also name it xbl_config.elf.

kraghava-qti: thank you. forgot the stashed changes in my local repo. done.

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

lumag (dismissed):

koenkooi (dismissed):
FWIW, `"${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}" = "true"` can be simplified to `${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}` (note the lack of quotation marks)
That makes run the shell builtins 'true' or 'false' instead of doing a string comparison.

The current form is good enough to go in as-is, though.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -103,13 +103,19 @@ create_qcomflash_pkg() {
...
-            -name '*.elf' -o \
+            \( -name '*.elf' ! -name 'xbl_config*.elf' \) -o \
             -name '*.mbn*' -o \
             -name '*.fv' -o \
             -name 'sec.dat'` ; do
         install -m 0644 ${bfw} .
     done
 
+    if [ "${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}" = "true" ]; then
+        install -m 0644 ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config_kvm.elf xbl_config.elf
```
kraghava-qti: This will fail on generic machine builds like qcom-armv8a as it does not have the concept of boot binaries. Need to conditionally install it based on presence of the file. Before it was not an issue as the find command would not find the elf files and install command would not trigger on generic arch machine. This is causing the dry run on this PR to fail.

CC: @lumag @vkraleti

classes-recipe/image_types_qcom.bbclass
```diff
@@ -103,13 +103,19 @@ create_qcomflash_pkg() {
...
+    else
+        install -m 0644 ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config.elf .
```
kraghava-qti: Same as above.

classes-recipe/image_types_qcom.bbclass
quaresmajose: We could move it to after the find and make the move in the current directory. That way keeping the sstate intact

classes-recipe/image_types_qcom.bbclass
```diff
@@ -104,16 +103,22 @@ create_qcomflash_pkg() {
...
             -name '*.fv' -o \
             -name 'sec.dat'` ; do
         install -m 0644 ${bfw} .
     done
 
-    # abl2esp
-    if [ -e "${DEPLOY_DIR_IMAGE}/abl2esp-${ABL_SIGNATURE_VERSION}.elf" ]; then
-        install -m 0644 "${DEPLOY_DIR_IMAGE}/abl2esp-${ABL_SIGNATURE_VERSION}.elf" .
+    xbl_config="${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config.elf"
+    if [ "${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}" = "true" ]; then
```
ricardosalveti: You can use `if ${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}; then` as I said in the other comment, no need to check with = "true".

classes-recipe/image_types_qcom.bbclass
```diff
@@ -104,16 +103,22 @@ create_qcomflash_pkg() {
...
             -name '*.mbn*' -o \
             -name '*.fv' -o \
             -name 'sec.dat'` ; do
         install -m 0644 ${bfw} .
     done
 
-    # abl2esp
-    if [ -e "${DEPLOY_DIR_IMAGE}/abl2esp-${ABL_SIGNATURE_VERSION}.elf" ]; then
-        install -m 0644 "${DEPLOY_DIR_IMAGE}/abl2esp-${ABL_SIGNATURE_VERSION}.elf" .
+    xbl_config="${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config.elf"
```
ricardosalveti: I would prefer having a XBL_CONFIG variable that you just set to xbl_config.elf or xbl_config_kvm.elf, and then use the complete path later on.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -104,16 +103,22 @@ create_qcomflash_pkg() {
...
+    if [ "${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}" = "true" ]; then
+        xbl_config="${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config_kvm.elf"
+    fi
+
+    if [ -f "${xbl_config}" ]; then
+        install -m 0644 "${xbl_config}" xbl_config.elf
+    else
+        bbnote "File ${xbl_config} does NOT exist!"
```
ricardosalveti: No need for this note, just copy if available and move on if not.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -104,16 +103,22 @@ create_qcomflash_pkg() {
...
 
-    # abl2esp
-    if [ -e "${DEPLOY_DIR_IMAGE}/abl2esp-${ABL_SIGNATURE_VERSION}.elf" ]; then
-        install -m 0644 "${DEPLOY_DIR_IMAGE}/abl2esp-${ABL_SIGNATURE_VERSION}.elf" .
+    xbl_config="${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config.elf"
+    if [ "${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}" = "true" ]; then
+        xbl_config="${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/xbl_config_kvm.elf"
+    fi
+
+    if [ -f "${xbl_config}" ]; then
```
ricardosalveti: Then use `${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${XBL_CONFIG}` instead.

kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

classes-recipe/image_types_qcom.bbclass
kraghava-qti: done

ricardosalveti:
There is now a conflict due https://github.com/qualcomm-linux/meta-qcom/pull/1199 (which just got merged), please rebase your changes.

kraghava-qti:
> There is now a conflict due #1199 (which just got merged), please rebase your changes.

Rebased.

koenkooi (approved):

ricardosalveti (dismissed):

classes-recipe/image_types_qcom.bbclass
```diff
@@ -104,7 +104,7 @@ create_qcomflash_pkg() {
 
     # boot firmware
     for bfw in `find ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR} -maxdepth 1 -type f \
-            \( -name '*.elf' ! -name 'abl2esp*.elf' \) -o \
+            \( -name '*.elf' ! -name 'abl2esp*.elf' -name 'xbl_config*.elf' \) -o \
```
kraghava-qti: forgot the !

quic-khuzuri:
2025-11-13 20:53:26,143 [( 32:3056   ) MetaCli:1092             ERROR     ] MetaCli: unexpected number of raw partition files: 0
2025-11-13 20:53:26,144 [( 32:3056   ) MetaBuild:2545           ERROR     ] Failed to get raw partition XML for storage type UFS
2025-11-13 20:53:26,145 [( 32:3056   ) BuildDownloadController:1700 INFO      ] [UI] Error - [ZT3BBX100019]: No download files found or files missing for device


build failed to load with this PR

PR #1204, opened 2025-11-12
workflows: Improve build summary

mwasilew:
Turn build summary into a table. Summary takes less space this way and action page is easier to read.


mwasilew:
Screenshot from 2025-11-12 15-16-19
This is how it looks in the action summary page. Example here: https://github.com/qualcomm-linux/meta-qcom/actions/runs/19276551703

PR #1206, opened 2025-11-13
BACKPORT: ofono: enable D-Bus use with PipeWire

rbujala-quic:
Added configuration rule in ofono.conf to establish correct D-Bus communication with PipeWire, when
pipewire running as a systemd service.

Link:
https://git.kernel.org/pub/scm/network/ofono/ofono.git/commit/?id=1a414d4bf3404117e952ec2ab942c0288af4d341

This change is part of the ofono master branch and will be available in later versions (starting from 2.19). Therefore, we have added this change to the specific version of ofono used on the QLI mainline.

### DIFF
--- a/recipes-connectivity/ofono/ofono_2.18.bbappend
+++ b/recipes-connectivity/ofono/ofono_2.18.bbappend
@@ -1,0 +1,6 @@
+do_install:append:qcom () {
+  sed -i '/<\/busconfig>/i\
+  <policy user="\pipewire\">\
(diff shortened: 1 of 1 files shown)


lumag (changes requested):
Please submit this fix to OE-Core instead using a .patch file.

ricardosalveti:
Yeah, please send it as a patch to oe-core, nothing qcom specific here.

PR #1212, opened 2025-11-14
Bluetooth: Add Bluetooth BD address setup service and script

quicjathot:
- If controller returns Bluetooth Device Address (BDA) which is not unique then device will set to unconfigured state, Hence add `qca_set_bdaddr.service` and `qca_set_bdaddr.sh` to configure a unique BDA on boot.
 - Add `bluez5_%.bbappend` to:
   - Include the new service and script in the BlueZ5 recipe.
   - Include obex to support obex related profiles like OPP, FTP, MAP, PBAP
   - Include only desired tools that are conditional on READLINE in bluez
   - Install the script to `/etc/initscripts` and the systemd unit to `bluetooth.target.wants`.

### DIFF
--- a/recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.service
+++ b/recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=start set BDA script for bluetooth
+ConditionPathExists=/sys/devices/soc0/serial_number
+Requires=bluetooth.service
+After=bluetooth.service
+
+[Service]
+Type=simple
+ExecStart=+/etc/initscripts/qca_set_bdaddr.sh
+
+[Install]
+WantedBy=bluetooth.target
--- a/recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
+++ b/recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
@@ -1,0 +1,67 @@
+#!/bin/sh
+# Copyright (c) 2025 Qualcomm Innovation Center, Inc. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# Function to format BD-Address and Set it
+set_bda() {
+    # Read the serial number
+    serial_number=$(cat /sys/devices/soc0/serial_number)
+
+    # Convert to hexadecimal
+    hex_serial_number=$(printf '%x\n' $serial_number)
+
+    # Format the serial number
+    formatted_serial_number=$(echo "${hex_serial_number}" | sed 's/../&:/g;s/:$//')
+
+    # Append qualcomm general prefix BDA value 22:22
+    BDA=22:22:${formatted_serial_number:0:14}
+
+    #After FW download completed wait for some time
+    sleep 1
+
+    # Pass the formatted serial number to btmgmt tool
+    # Start btmgmt and send commands
+    {
+        echo "public-addr $BDA"
+        sleep 1
+    } | btmgmt
+}
+
+
+# Function to check BD Address and wait until it becomes configured or unconfigured
+validate_and_set_bda() {
+    while true; do
+        # Get the BD Address
+        hciconfig_output=$(hciconfig)
+
+        bd_address=$(echo "$hciconfig_output" | grep 'BD Address' | awk '{print $3}')
+        unconfigured=$(echo "$hciconfig_output" | grep -o 'DOWN RAW')
+        configured=$(echo "$hciconfig_output" | grep -o 'DOWN')
+
+        # Check if the BD Address is 00:00:00:00:00:00
+        if [[ "$bd_address" == "00:00:00:00:00:00" ]]; then
+            sleep 1
+        elif  [[ "$unconfigured" == "DOWN RAW" ]]; then
+           break
+        elif [[ "$configured" == "DOWN" ]]; then
+           echo "BD-Address already configured!!"
+           return 0
+        else
+           break
+        fi
+    done
+
+    # Call the set_bda function
+    set_bda
+}
+
+
+
+# Run bluetoothctl show and capture the output
+bluetoothctl_output=$(bluetoothctl show)
+
+# Check if the BD Address is valid
+if echo "$bluetoothctl_output" | grep -q "No default controller available"; then
+    # Call the validate_and_set_bda function
+    validate_and_set_bda
+fi
--- a/recipes-connectivity/bluez5/bluez5_%.bbappend
+++ b/recipes-connectivity/bluez5/bluez5_%.bbappend
@@ -1,0 +1,52 @@
+inherit systemd
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
+
+RRECOMMENDS:${PN} += " \
+    glibc-gconv-utf-16 \
+    glibc-gconv-utf-32 \
+"
+
+SRC_URI:append = " file://qca_set_bdaddr.service \
+                   file://qca_set_bdaddr.sh \
+"
+
+#Include obex to support obex related profiles like OPP, FTP, MAP, PBAP
+RDEPENDS:${PN} += "${PN}-obex"
+
+#Include only desired tools that are conditional on READLINE in bluez
+INST_TOOLS_READLINE = " \
+    tools/bluetooth-player \
+    tools/obexctl \
+    tools/btmgmt \
+"
+
+#Remove desired tools from noinst-tools
+NOINST_TOOLS:remove = " \
+    ${@bb.utils.contains('PACKAGECONFIG', 'readline', '${INST_TOOLS_READLINE}', '', d)} \
+"
+
+do_install:append:qcom() {
+
+    #Install desired tools that upstream leaves in build area
+    for f in ${INST_TOOLS_READLINE} ; do
+        install -m 755 ${B}/$f ${D}/${bindir}
+    done
+
+    #Create below directory which is used by obex service
+    install -d ${D}${localstatedir}/bluetooth
+
+    # Install script to set unique BDA
+    install -d ${D}${sysconfdir}/initscripts
+    install -m 755 ${UNPACKDIR}/qca_set_bdaddr.sh ${D}${sysconfdir}/initscripts/
+
+    # Install service that will run qca_set_bdaddr.sh script on boot
+    install -m 0644 ${UNPACKDIR}/qca_set_bdaddr.service -D ${D}${systemd_system_unitdir}/qca_set_bdaddr.service
+    install -d ${D}${systemd_system_unitdir}/bluetooth.target.wants/
+    ln -sf ${systemd_system_unitdir}/qca_set_bdaddr.service ${D}${systemd_system_unitdir}/bluetooth.target.wants/qca_set_bdaddr.service
+}
+
+SYSTEMD_SERVICE:${PN} += "qca_set_bdaddr.service"
+FILES:${PN}:append = "${systemd_system_unitdir}/qca_set_bdaddr.service \
+                           ${sysconfdir}/initscripts/qca_set_bdaddr.sh \
+"


recipes-connectivity/bluez5/bluez5_%.bbappend
```diff
@@ -0,0 +1,52 @@
...
+    done
+
+    #Create below directory which is used by obex service
+    install -d ${D}${localstatedir}/bluetooth
+
+    # Install script to set unique BDA
+    install -d ${D}${sysconfdir}/initscripts
+    install -m 755 ${UNPACKDIR}/qca_set_bdaddr.sh ${D}${sysconfdir}/initscripts/
+
+    # Install service that will run qca_set_bdaddr.sh script on boot
```
koenkooi, line 43: systemd.bbclass should be enabling this by default already:
```
# Whether to enable or disable the services on installation.
SYSTEMD_AUTO_ENABLE ??= "enable"
[...]
        if [ "${SYSTEMD_AUTO_ENABLE}" = "enable" ]; then
                for service in ${@systemd_filter_services("${SYSTEMD_SERVICE_ESCAPED}", False, d)}; do
                        systemctl ${OPTS} enable "$service"
                done

                for service in ${@systemd_filter_services("${SYSTEMD_SERVICE_ESCAPED}", True, d)}; do
                        systemctl --global ${OPTS} enable "$service"
                done
        fi
```

recipes-connectivity/bluez5/bluez5_%.bbappend
```diff
@@ -0,0 +1,52 @@
+inherit systemd
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
+
+RRECOMMENDS:${PN} += " \
```
koenkooi, line 5: Please keep variables sorted by usage, so R* goes below do_install, SRC* goes above `do_unpack`, configure vars go right above `do_configure` and installation vars go right above `do_install`.

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/sh
+# Copyright (c) 2025 Qualcomm Innovation Center, Inc. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+# Function to format BD-Address and Set it
+set_bda() {
```
koenkooi, line 6: There's zero error checking and handling in this method

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
```diff
@@ -0,0 +1,67 @@
...
+    serial_number=$(cat /sys/devices/soc0/serial_number)
+
+    # Convert to hexadecimal
+    hex_serial_number=$(printf '%x\n' $serial_number)
+
+    # Format the serial number
+    formatted_serial_number=$(echo "${hex_serial_number}" | sed 's/../&:/g;s/:$//')
+
+    # Append qualcomm general prefix BDA value 22:22
+    BDA=22:22:${formatted_serial_number:0:14}
```
lumag, line 17: Bashism, will not work with posix shells

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
```diff
@@ -0,0 +1,67 @@
...
+
+    #After FW download completed wait for some time
+    sleep 1
+
+    # Pass the formatted serial number to btmgmt tool
+    # Start btmgmt and send commands
+    {
+        echo "public-addr $BDA"
+        sleep 1
+    } | btmgmt
```
lumag, line 27: Why is it so complex? `btmgmt --index 0 public-addr ${addr}`

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
```diff
@@ -0,0 +1,67 @@
...
+        unconfigured=$(echo "$hciconfig_output" | grep -o 'DOWN RAW')
+        configured=$(echo "$hciconfig_output" | grep -o 'DOWN')
+
+        # Check if the BD Address is 00:00:00:00:00:00
+        if [[ "$bd_address" == "00:00:00:00:00:00" ]]; then
+            sleep 1
+        elif  [[ "$unconfigured" == "DOWN RAW" ]]; then
+           break
+        elif [[ "$configured" == "DOWN" ]]; then
+           echo "BD-Address already configured!!"
```
lumag, line 47: No need to say anything if it works as expected.

recipes-connectivity/bluez5/bluez5_%.bbappend
```diff
@@ -0,0 +1,52 @@
...
+    glibc-gconv-utf-16 \
+    glibc-gconv-utf-32 \
```
lumag, line 7: Please add a comment, why do you need these

recipes-connectivity/bluez5/bluez5_%.bbappend
```diff
@@ -0,0 +1,52 @@
...
+"
+
+SRC_URI:append = " file://qca_set_bdaddr.service \
+                   file://qca_set_bdaddr.sh \
+"
+
+#Include obex to support obex related profiles like OPP, FTP, MAP, PBAP
+RDEPENDS:${PN} += "${PN}-obex"
```
lumag, line 15: This is unrelated. In general, such things should be split into separate commits. In this particular case it should be solved by the distro layer instead by pulling the right set of packages.

recipes-connectivity/bluez5/bluez5_%.bbappend
```diff
@@ -0,0 +1,52 @@
...
+INST_TOOLS_READLINE = " \
+    tools/bluetooth-player \
+    tools/obexctl \
+    tools/btmgmt \
+"
+
+#Remove desired tools from noinst-tools
+NOINST_TOOLS:remove = " \
+    ${@bb.utils.contains('PACKAGECONFIG', 'readline', '${INST_TOOLS_READLINE}', '', d)} \
+"
```
lumag, line 27: Separate commit with a proper explanation

recipes-connectivity/bluez5/bluez5_%.bbappend
```diff
@@ -0,0 +1,52 @@
...
+
+do_install:append:qcom() {
+
+    #Install desired tools that upstream leaves in build area
+    for f in ${INST_TOOLS_READLINE} ; do
+        install -m 755 ${B}/$f ${D}/${bindir}
+    done
```
lumag, line 34: This should be patched in the upstream or in OE-Core

lumag (changes requested):
Also please don't describe patch contents in the commit message. Describe oly rationale and what is done to solve that.

recipes-connectivity/bluez5/bluez5_%.bbappend
quicjathot: Thanks for the guidance. will reorder variables accordingly

quicjathot:
> This seems an invasive enough change to be its own recipe and perhaps a bluez.bbappend to add the stand alone recipe to `RRECOMMENDS`.

Thanks! Agreed, will move the script and unit into a new recipe (qca-bt-bdaddr) and keep bluez5_%.bbappend separate. This keeps BlueZ packaging untouched and scopes the change properly.

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
quicjathot: sure will will error checking for each command.

recipes-connectivity/bluez5/bluez5_%.bbappend
quicjathot: sure, will handle in different PR

recipes-connectivity/bluez5/bluez5_%.bbappend
quicjathot: Sure, will pull the commit changes related to obex tools to separate commit and PR

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
quicjathot: noisy success messages Understood. will remove success prints

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=start set BDA script for bluetooth
```
ricardosalveti, line 2: Please improve the description.

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.service
```diff
@@ -0,0 +1,12 @@
...
+ConditionPathExists=/sys/devices/soc0/serial_number
+Requires=bluetooth.service
+After=bluetooth.service
+
+[Service]
+Type=simple
+ExecStart=+/etc/initscripts/qca_set_bdaddr.sh
```
ricardosalveti, line 9: nack, script should be an executable available in /usr/bin

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
```diff
@@ -0,0 +1,67 @@
...
+    {
+        echo "public-addr $BDA"
+        sleep 1
+    } | btmgmt
+}
+
+
+# Function to check BD Address and wait until it becomes configured or unconfigured
+validate_and_set_bda() {
+    while true; do
```
ricardosalveti, line 33: I would prefer a fixed set of runs here, probably trying the loop for up to 5 times with a sleep 1 or similar between runs, no need for while true.

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
ricardosalveti: the entire script actually

recipes-connectivity/bluez5/bluez5_%.bbappend
```diff
@@ -0,0 +1,52 @@
+inherit systemd
```
ricardosalveti, line 1: Please create a different recipe for this, it needs bluez but there is no need to be part of it.

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.service
lumag: Or ${sbindir}.

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
lumag: Also, where does 22:22 come from?

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.service
quicjathot: sure, noted.

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
quicjathot: Since btmgmt is interactive it is actually stuck If we directly use  the command during boot in script . With above approach everytime it is working and it is not  getting stuck.

recipes-connectivity/bluez5/bluez5_%.bbappend
quicjathot: will reuse these SYSTEMD_AUTO_ENABLE.

recipes-connectivity/bluez5/bluez5_%.bbappend
quicjathot: will handle in different recipe and PR with updated the comment

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
quicjathot: sure , will consider and update the iteration based on the results

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
quicjathot: we took android HAL code as reference for this

lumag: Don't take random unreliable sources. There are standards, owners, etc.

recipes-connectivity/bluez5/bluez5/qca_set_bdaddr.sh
lumag: Please read carefully what I wrote.

PR #1214, opened 2025-11-14
linux-qcom-next: Added security hardening.config to kernel config

abhilash-manna:
This PR introduces support for conditionally including `hardening.config` during kernel configuration for linux-qcom-next.
The inclusion is based on DISTRO_FEATURES containing **hardened**.
Uses merge_config.sh to combine the base .config, hardening.config, and any additional fragments.
Default it's not enabled for all build and only for production builds with "hardened" is enabled.

Changes:
Added logic in `do_configure:prepend` to merge `hardening.config` when `hardened` is present in DISTRO_FEATURES.
No changes for distros without this feature.

Verified build on qcs9100-ride-sx with DISTRO_FEATURES += "hardened".
Kernel config includes hardening options as expected.

### DIFF
--- a/recipes-kernel/linux/linux-qcom-next_git.bb
+++ b/recipes-kernel/linux/linux-qcom-next_git.bb
@@ -36,10 +36,13 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
+CONFIG_LIST =  "${@" ".join(find_cfgs(d))}"
+CONFIG_LIST += "${@bb.utils.contains('DISTRO_FEATURES', 'hardened', '${S}/kernel/configs/hardening.config', '', d)}"
+
 do_configure:prepend() {
     # Use a copy of the 'defconfig' from the actual repo to merge fragments
     cp ${S}/arch/${ARCH}/configs/${KBUILD_DEFCONFIG} ${B}/.config
 
     # Merge fragment for QCOM value add features
-    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${@" ".join(find_cfgs(d))}
+    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${CONFIG_LIST}
 }


recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,10 +36,12 @@ S = "${UNPACKDIR}/${BP}"
...
 
+HARDENING_CFG = "${@bb.utils.contains('DISTRO_FEATURES', 'sec-hardening', '${S}/kernel/configs/hardening.config', '', d)}"
+
 do_configure:prepend() {
     # Use a copy of the 'defconfig' from the actual repo to merge fragments
     cp ${S}/arch/${ARCH}/configs/${KBUILD_DEFCONFIG} ${B}/.config
 
     # Merge fragment for QCOM value add features
-    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${@" ".join(find_cfgs(d))}
+    ${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${HARDENING_CFG} ${@" ".join(find_cfgs(d))}
```
lumag: Move `${@" ".join(find_cfgs(d))}` to a new variable. Append it depending on the distro features

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,10 +36,12 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
+HARDENING_CFG = "${@bb.utils.contains('DISTRO_FEATURES', 'sec-hardening', '${S}/kernel/configs/hardening.config', '', d)}"
```
quaresmajose: KBUILD_CONFIG_EXTRA would be more apropriate in my opinion. Having a more generic name will simplify if we need to add other snippets in the future.

The DISTRO_FEATURES sec-hardening doesn't seem to exist anywhere in upstream meta layers. If the KBUILD_CONFIG_EXTRA were declared empty, it could later be modified to a specific targt image in other distro layers, for example. That way we wouldn't introduce a DISTRO_FEATURES that nobody knows into the community.

```
KBUILD_CONFIG_EXTRA ??= ""
```

recipes-kernel/linux/linux-qcom-next_git.bb
abhilash-manna: Updated.

recipes-kernel/linux/linux-qcom-next_git.bb
lumag: This was left uncommented

abhilash-manna: Thanks for the suggestion.
I understand your point about KBUILD_CONFIG_EXTRA being generic and flexible.
However, BitBake doesn’t pass this variable to the kernel during do_configure; it uses its own fragment merging via merge_config.sh.

So, setting KBUILD_CONFIG_EXTRA ??= "" won’t have any effect unless we modify the kernel class or tasks to support it.
Also, linux-qcom-next doesn’t include linux-yocto.inc or inherit kernel-yocto, so Yocto’s native KERNEL_CONFIG_FRAGMENTS mechanism isn’t available by default.
Adding kernel-yocto might not be compatible with the vendor kernel.

lumag (approved):

recipes-kernel/linux/linux-qcom-next_git.bb
quaresmajose: My sugestion would basically adding a placeholder named KBUILD_CONFIG_EXTRA. I understand that having it clean by default won't have any effect but it need to be injected in merge_config.sh as well. You CONFIG_LIST is a better name and also serves the propose I wanted.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,10 +36,13 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
+CONFIG_LIST =  "${@" ".join(find_cfgs(d))}"
```
ricardosalveti, line 39: @abhilash-manna extra whitespace.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,10 +36,13 @@ S = "${UNPACKDIR}/${BP}"
...
+CONFIG_LIST += "${@bb.utils.contains('DISTRO_FEATURES', 'hardened', '${S}/kernel/configs/hardening.config', '', d)}"
```
ricardosalveti, line 40: Why not enabled it by default @abhilash-manna?

recipes-kernel/linux/linux-qcom-next_git.bb
abhilash-manna: Will update in the next change that makes hardening the default.

recipes-kernel/linux/linux-qcom-next_git.bb
abhilash-manna: Raising a change that makes hardening the default.

PR #1217, opened 2025-11-17
packagegroup-qcs8300-ride: correct WLAN firmware packages

WeiZhang-stone:
Based on the subversion ID of the HSP card on the qcs8300-ride platform,
the HSP card should correspond to the QCA6698AQ firmware. Therefore,
update the WLAN firmware package by replacing
`linux-firmware-ath11k-wcn6855` with `linux-firmware-ath11k-qca6698aq`.


lumag (changes requested):
No Gerrit tags. Also please describe the reason for the change, not just the patch contents.

WeiZhang-stone:
Rebased with latest master.

Issue #1201, opened 2025-11-10
Including Kernel Hardening Flags in linux-qcom-next


abhilash-manna:
Working on enabling several kernel hardening flags in our custom Yocto-based kernel recipe.

After consulting with upstream maintainers, we were advised to use: 

**KERNEL_FEATURES:append = " features/security/security.scc"** which pulls scc files form yocto-kernel-cache.

However, we want to ensure our approach aligns with Yocto mainline practices, where yocto-kernel-cache holds feature-based configuration files.
These are typically enabled using the "KERNEL_FEATURES:append" variable, which pulls in .scc files that reference relevant config fragments.

e.g. of security flags that we are looking @

- CONFIG_SLAB_FREELIST_RANDOM=y
- CONFIG_SLAB_FREELIST_HARDENED=y
- CONFIG_HARDENED_USERCOPY=y
- CONFIG_FORTIFY_SOURCE=y

....

I don't see .scc file being used in linux-qcom-next — or am I missing something?  
What would be the recommended approach to enable these configs: **should we go with security.cfg under configs/, or is there a better way suggested?**

ricardosalveti:
> [@shashim-quic](https://github.com/shashim-quic) , thanks we are going to re-use haredening.config and now tracked with PR [#1214](https://github.com/qualcomm-linux/meta-qcom/pull/1214)

It is bound to a distro feature, shouldn't we just include it by default? I don't see why create a different distro feature and yet another image permutation for things that shouldn't really break the system.

abhilash-manna:
Enabled hardened feature default to build and now tracked with PR #1223

PR #1223, opened 2025-11-19
Enabled hardening configs default to build.

abhilash-manna:
This PR introduces support for merging `hardening.config` into the kernel configuration for linux-qcom-next.
The goal is to ensure that security hardening options are applied consistently across all builds.

What’s Changed
Added logic to include `hardening.config` during kernel configuration using merge_config.sh.
Override `CONFIG_KSTACK_ERASE=n`

Why This Change
Kernel builds previously lacked default hardening options, which could lead to reduced security.
This update ensures that compiler-based mitigations and other hardening features are applied by default.

Override `CONFIG_KSTACK_ERASE=n` because enabling it introduces absolute workspace paths into out‑of‑tree (OOT) kernel module builds, which cause Yocto's package QA to flag this as build error.

Verification
verified build on QCS9100-ride-sx. Kernel configuration includes hardening options as expected.

Ref: 
[1201#issuecomment-3543906617](https://github.com/qualcomm-linux/meta-qcom/issues/1201#issuecomment-3543906617)
https://github.com/qualcomm-linux/meta-qcom/issues/1628#issuecomment-3951962801

### DIFF
--- a/recipes-kernel/linux/linux-qcom-6.18/configs/bsp-additions.cfg
+++ b/recipes-kernel/linux/linux-qcom-6.18/configs/bsp-additions.cfg
@@ -314,3 +314,5 @@ CONFIG_NFT_TPROXY=m
 CONFIG_NFT_TUNNEL=m
 CONFIG_PACKET_DIAG=y
 CONFIG_VETH=m
+# Disable stack erase plugin to avoid buildpath leakage in out-of-tree modules
+CONFIG_KSTACK_ERASE=n
--- a/recipes-kernel/linux/linux-qcom_6.18.bb
+++ b/recipes-kernel/linux/linux-qcom_6.18.bb
@@ -39,7 +39,7 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
-KBUILD_CONFIG_EXTRA = "${@bb.utils.contains('DISTRO_FEATURES', 'hardened', '${S}/kernel/configs/hardening.config', '', d)}"
+KBUILD_CONFIG_EXTRA = "${S}/kernel/configs/hardening.config"
 KBUILD_CONFIG_EXTRA:append:aarch64 = " ${S}/arch/arm64/configs/prune.config"
 KBUILD_CONFIG_EXTRA:append:aarch64 = " ${S}/arch/arm64/configs/qcom.config"
 KBUILD_CONFIG_EXTRA:append = " ${@oe.utils.vartrue('DEBUG_BUILD', '${S}/kernel/configs/debug.config', '', d)}"


ci/base.yml
```diff
@@ -39,7 +39,7 @@ local_conf_header:
     IMAGE_CLASSES += "image_types_qcom"
     IMAGE_FSTYPES += "qcomflash"
   extra: |
-    DISTRO_FEATURES:append = " efi pni-names"
+    DISTRO_FEATURES:append = " efi pni-names hardened"
```
vkraleti: If we are enabling by default why distro feature is needed? Can't CONFIG_LIST be populated unconditionally?

abhilash-manna: Opt-Out Support: The hardened feature remains configurable, allowing users to explicitly remove it from DISTRO_FEATURES if they choose to disable hardening.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,7 +36,7 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
-CONFIG_LIST =  "${@" ".join(find_cfgs(d))}"
+CONFIG_LIST = "${@" ".join(find_cfgs(d))}"
```
lumag: unrelated

ci/base.yml
lumag: So, any user not using Kas will not get hardened features? Why do we have them here then? We are not publishing binary images.

abhilash-manna: Acknowledged. updated and remove dependency on the `hardened` feature.

recipes-kernel/linux/linux-qcom-next_git.bb
abhilash-manna: Removed extra whitespace.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,8 +36,8 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
-CONFIG_LIST =  "${@" ".join(find_cfgs(d))}"
-CONFIG_LIST += "${@bb.utils.contains('DISTRO_FEATURES', 'hardened', '${S}/kernel/configs/hardening.config', '', d)}"
+CONFIG_LIST = "${@" ".join(find_cfgs(d))}"
+CONFIG_LIST += "${S}/kernel/configs/hardening.config"
```
quaresmajose: Any variable can be a list if they have more than one string. I tink is better to follow KBUILD_DEFCONFIG naming, with something like KBUILD_CONFIG_EXTRA and pass it bellow to merge_config.sh.
```
KBUILD_CONFIG_EXTRA = "${S}/kernel/configs/hardening.config"
```
Keep only internal kernel config files in this new variable will be better because the others *.cfg are already controlled in the SRC_URI and so they remain somehow separated.

recipes-kernel/linux/linux-qcom-next_git.bb
quaresmajose: Here we use both configs, extarnal and internal
```
${S}/scripts/kconfig/merge_config.sh -m -O ${B} ${B}/.config ${@" ".join(find_cfgs(d))} ${KBUILD_CONFIG_EXTRA}
```

recipes-kernel/linux/linux-qcom-next_git.bb
abhilash-manna: understood.

recipes-kernel/linux/linux-qcom-next_git.bb
abhilash-manna: updated as suggested.

quaresmajose (dismissed):
LGTM. thanks

lumag:
> > `2025-11-19 16:37:57 - ERROR - ERROR: kgsl-dlkm-0.0+git-r0 do_package_qa: QA Issue: File /usr/lib/modules/6.18.0-rc5-00249-g53919c1f684a/updates/.debug/msm_kgsl.ko in package kgsl-dlkm-dbg contains reference to TMPDIR [buildpaths] `
> > Interesting, wonder which new option is causing that for this module in particular.
> 
> `CONFIG_KSTACK_ERASE` is causing package_qa errors in the kgsl & camx module (similar errors observed). This configuration is critical for security and must remain enabled. To address QA failures without disabling this option, we are considering adding the following directive to the kgsl-dlkm and camx-dlkm recipes :
> 
> ```
> INSANE_SKIP:${PN}-dbg += "buildpaths"
> ```

No. This would mean that the builds are not binary-reproducible. So, NAK.

> 
> This would apply only to the debug package. Discussions are ongoing with the respective technical teams to review this approach.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,7 +36,7 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
-KBUILD_CONFIG_EXTRA = "${@bb.utils.contains('DISTRO_FEATURES', 'hardened', '${S}/kernel/configs/hardening.config', '', d)}"
+KBUILD_CONFIG_EXTRA = "${S}/kernel/configs/hardening.config "
```
vkraleti: Remove extra white space at the end.

vkraleti (approved):

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -36,7 +36,7 @@ S = "${UNPACKDIR}/${BP}"
 KBUILD_DEFCONFIG ?= "defconfig"
 KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 
-KBUILD_CONFIG_EXTRA = "${@bb.utils.contains('DISTRO_FEATURES', 'hardened', '${S}/kernel/configs/hardening.config', '', d)}"
+KBUILD_CONFIG_EXTRA = "${S}/kernel/configs/hardening.config"
```
anujm1, line 39: Doesn't this need to be changed in the other kernel recipe as well?

jaihindy: once this change is merged , we'll make the change in other kernel recipe as well.

abhilash-manna: Yes, this is required for all kernel recipes, as we plan to enable hardening across all of them.

lumag: Please perform this as a part of this PR

abhilash-manna: @lumag, team is actively working on this and will raise separate PR soon for remaining.

lumag: Please submit all of the changes at once via this PR

abhilash-manna: Acknowledged and updated as requested.

PR #1232, opened 2025-11-21
Drop custom BSP EFI variable settings

ricardosalveti:
Rely instead on the default EFI_PREFIX value (/boot, as this variable sets the expected path where ESP is mounted in the rootfs), and adapt esp-qcom-image to move files from /boot/EFI.

### DIFF
--- a/conf/machine/include/qcom-common.inc
+++ b/conf/machine/include/qcom-common.inc
@@ -44,14 +44,5 @@ INITRAMFS_MAXSIZE:pn-initramfs-kerneltest-full-image = "983040"
 # Use systemd-boot as the EFI bootloader
 EFI_PROVIDER = "systemd-boot"
 
-# Install packages at root of ESP
-EFI_PREFIX = ""
-
-# Location of dtb inside ESP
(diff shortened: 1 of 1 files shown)


recipes-kernel/images/esp-qcom-image.bb
```diff
@@ -33,6 +33,10 @@ IMAGE_LINGUAS = ""
 IMAGE_FEATURES = ""
 
 remove_unused_files() {
+    # Move EFI content from packages expecting /boot to be the ESP location
```
lumag, line 36: Nit: this functions now does more than removing unused files.

lumag (approved):

ricardosalveti:
> It took me a lot longer than needed to parse the commit log from "esp-qcom-image: support content from /boot/EFI". The name of the image wasn't a big enough clue for me, could you move the "This is the ESP image" forward in the description?

Updated, see if it is better now.

recipes-kernel/images/esp-qcom-image.bb
ricardosalveti: Renamed to setup_efi_folder.

PR #1237, opened 2025-11-22
Drop extra SoC prefix from machine names

lumag:
Shift towards using just the commercial name for the machines, keeping
SoC prefix only for machines which have it explicitly (like different
RIDE, HDK or MTP devices).


mwasilew:
This requires changes in LAVA as well. I'll need to add aliases for the LAVA devices. I'll do that now. We can drop them if this PR is not accepted

mwasilew:
I think the list of required checks needs to be updated to reflect new machine names.

PR #1255, opened 2025-11-27
qcom-qcs6490.inc: add msm.dpu_use_virtual_planes=1 to kernel cmdline

quicmahap:
Add msm.dpu_use_virtual_planes=1 to KERNEL_CMDLINE_EXTRA for QCS6490 platform to enable virtual planes in the MSM DPU driver.


lumag:
> Why isn't it enabled by default? I would really like to avoid having custom kernel command line arguments here, as this won't really scale with other distributions and we should be validating and testing the same stack.

Because it is an experimental feature. I plan to enable it for 6.20 and then to drop the old implementation hopefully from 6.21

lumag:
For the reference: https://patchwork.freedesktop.org/series/148983/

PR #1258, opened 2025-11-27
ci: add kas fragment and ci jobs for qcom-distro-sota

ricardosalveti:
Add kas fragment and extend build-yocto to also cover qcom-distro-sota as another distro variant.


codeshravan:
LGTM according to design, I have verified on rb3-gen2.

ricardosalveti:
> Breaks on armv7 because there we use zImage

Disabled uki for armv7 on meta-updater, should be good now.

vkraleti (approved):

PR #1270, opened 2025-12-02
QCLINUX: Wireplumber: Enable bluez plugin.

rbujala-quic:
- The BlueZ monitors are integrated with logind to ensure that only one user at a time can use the Bluetooth audio devices.
For more information:
https://pipewire.pages.freedesktop.org/wireplumber/daemon/configuration/bluetooth.html

- on QLI devices, there is no user login sessions enabled. so disable this feature will allow bluez plugin to launch without looking for logind session.

- Add bluetooth configuration file in wireplumber delta conf folder for user preferred configuration.
- To alter supported sampling rate, codecs, profiles and etc.,

### DIFF
--- a/recipes-multimedia/audio/wireplumber_%.bbappend
+++ b/recipes-multimedia/audio/wireplumber_%.bbappend
@@ -1,0 +1,11 @@
+do_install:append:qcom () {
+
+  install -v -m 0644 \
+      ${S}/src/config/wireplumber.conf.d.examples/bluetooth.conf \
+      ${D}${datadir}/wireplumber/wireplumber.conf.d/
+
+  sed -i '/^[[:space:]]*main[[:space:]]*=[[:space:]]*{/,/^[[:space:]]*}/{
+    /policy\.standard[[:space:]]*=/a\
+  monitor.bluez.seat-monitoring = disabled
+  }' ${D}${datadir}/wireplumber/wireplumber.conf
+}


recipes-multimedia/audio/wireplumber_%.bbappend
```diff
@@ -0,0 +1,11 @@
+do_install:append:qcom () {
+
+  install -v -m 0644 \
+      ${S}/src/config/wireplumber.conf.d.examples/bluetooth.conf \
+      ${D}${datadir}/wireplumber/wireplumber.conf.d/
+
+  sed -i '/^[[:space:]]*main[[:space:]]*=[[:space:]]*{/,/^[[:space:]]*}/{
```
lumag, line 7: This is unreadable. Please use . patch to change the source code.

lumag (changes requested):
Also please fix the commit subject and message. The message should be an English test, describing the issue you are trying to solve, not a bullet list. The subject should be "recipe: what". Don't use extra useless prefixes like QCLINUX.

recipes-multimedia/audio/wireplumber_%.bbappend
rbujala-quic: I have modified changes as per to the suggestions and the change is raised on meta-qcom-distro
as this change has dependency with audio module. so moved it to the appropriate distro.
PR details:
https://github.com/qualcomm-linux/meta-qcom-distro/pull/100

PR #1274, opened 2025-12-02
Multi-DTB Fitimage implementation

Kavinaya99:
Add support for DTB-only FIT image generation

Extend OE-Core FIT image script to handle Qualcomm-specific needs:
- Generate DTB-only FIT images
- Add compatibility string handling
- Support custom .its for Qualcomm platforms

This enables packaging multiple DTBs in one image for different SoCs and boards, allowing UEFI firmware to 
select the correct DTB dynamically during boot.

### DIFF
--- a/classes/linux-qcom-dtbbin.bbclass
+++ b/classes/linux-qcom-dtbbin.bbclass
@@ -4,6 +4,8 @@
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
 
+inherit dtb-fit-image
+
 DTBBIN_DEPLOYDIR = "${WORKDIR}/qcom_dtbbin_deploy-${PN}"
 DTBBIN_SIZE ?= "4096"
 
@@ -21,6 +23,10 @@ do_qcom_dtbbin_deploy() {
         mcopy -i "${DTBBIN_DEPLOYDIR}/dtb-${dtb_base_name}-image.vfat" -vsmpQ ${DTBBIN_DEPLOYDIR}/$dtb_base_name/* ::/
         rm -rf ${DTBBIN_DEPLOYDIR}/$dtb_base_name
     done
+
+    # Generate qclinux_fit.img along side combined-dtb.dtb
+    mkfs.vfat -S ${QCOM_VFAT_SECTOR_SIZE} -C ${DTBBIN_DEPLOYDIR}/multi-dtb.vfat ${DTBBIN_SIZE}
+    mcopy -i "${DTBBIN_DEPLOYDIR}/multi-dtb.vfat" -vsmpQ ${DEPLOY_DIR_IMAGE}/qclinuxfitImage ::/qclinux_fit.img
 }
 addtask qcom_dtbbin_deploy after do_populate_sysroot do_packagedata before do_deploy
 


lib/qcom/dtb_only_fitimage.py
```diff
@@ -0,0 +1,131 @@
...
+
+import os
+import shlex
+import subprocess
+import bb
+from typing import Tuple
+import oe.fitimage
+from oe.fitimage import ItsNodeRootKernel, get_compatible_from_dtb
+
+class QcomItsNodeRoot(ItsNodeRootKernel):
```
ndechesne, line 21: We need to document the Qualcomm-specific requirements for DTB-only FIT images to be able to properly review this series of patches.

ndechesne: actually, it seems to be documented here https://github.com/qualcomm-linux/qcom-dtb-metadata/blob/main/Documentation.md. it should have been clear in the PR :)

lib/qcom/dtb_only_fitimage.py
```diff
@@ -0,0 +1,131 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+#
+# This file contains functions for Qualcomm specific DTB-only fitimage generation,
+# which imports classes from OE-Core fitimage.py and enhances to meet Qualcomm FIT
+# specifications.
```
ricardosalveti, line 8: Add a link to https://github.com/qualcomm-linux/qcom-dtb-metadata/blob/main/Documentation.md here.

lib/qcom/dtb_only_fitimage.py
```diff
@@ -0,0 +1,131 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
ricardosalveti, line 1: I would rather rename this file to be specific to DTB (e.g. qcom_dtb_fitimage or similar), as this will later confuse people since we do also expect u-boot users to leverage the traditional fit image from oe-core (making it just qcom_fit gives the impression that this is just a qcom implementation of the generic fit).

lib/qcom_fitimage.py
```diff
@@ -0,0 +1,131 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
```
ricardosalveti: We are extending fitimage.py from OE, which is GPL-2.0-only, we should align the license here.

lib/qcom/dtb_only_fitimage.py
ricardosalveti: Add a comment to this class explaining why it needs to extend ItsNodeRootKernel.

lib/qcom_fitimage.py
```diff
@@ -0,0 +1,131 @@
...
+                         mkimage_sign, mkimage_sign_args, hash_algo, sign_algo,
+                         pad_algo, sign_keyname_conf, sign_individual, sign_keyname_img)
+
+        self._mkimage_extra_opts = []
+        self._dtbs = []
+
+    def set_extra_opts(self, mkimage_extra_opts):
+        self._mkimage_extra_opts = shlex.split(mkimage_extra_opts) if mkimage_extra_opts else []
+
+    # Override DTB section to allow compatible_override
```
ricardosalveti: Make this a python-based function specific comment instead.

lib/qcom_dtb_only_fitimage.py
```diff
@@ -0,0 +1,131 @@
...
+                        counter += 1
+                        self._fitimage_emit_one_section_config(conf_name, dtb_node)
+                else:
+                    dtb_node = entry
+                    conf_name = f"{self._conf_prefix}{counter}"
+                    counter += 1
+                    self._fitimage_emit_one_section_config(conf_name, dtb_node)
+        else:
+            # Currently exactly one kernel is supported.
+            self._fitimage_emit_one_section_config(self._conf_prefix + "1")
```
ricardosalveti: Kernel?

lib/qcom_fitimage.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def get_compatible_from_dtb(dtb_path, fdtget_path="fdtget"):
+        compatible = None
+        cmd = [fdtget_path, "-t", "s", dtb_path, "/", "compatible"]
+        try:
+            ret = subprocess.run(cmd, check=True, capture_output=True, text=True)
+            compatible = ret.stdout.strip().split()
+        except subprocess.CalledProcessError:
+            compatible = None
+        return compatible
```
ricardosalveti: Looks like you just duplicated this function?

conf/machine/include/qcom-common.inc
```diff
@@ -46,3 +46,16 @@ EFI_PROVIDER = "systemd-boot"
 
 # Unified Kernel Image (UKI) name
 EFI_LINUX_IMG ?= "linux-${MACHINE}.efi"
+
+# Variables to generate ITS file
+UBOOT_CONFIG ??= "qcom"
+UBOOT_CONFIG[qcom] ?= "qcom_defconfig"
+
+# Pass additional options to mkimage command
```
ricardosalveti, line 51: You should instead explain why the extra options are needed here.

lib/qcom/dtb_only_fitimage.py
```diff
@@ -0,0 +1,131 @@
...
+            # Currently exactly one kernel is supported.
+            self._fitimage_emit_one_section_config(self._conf_prefix + "1")
+
+        default_conf = self.configurations.sub_nodes[0].name
+        if default_dtb_image and self._dtbs:
+            default_conf = self._conf_prefix + default_dtb_image
+        self.configurations.add_property('default', default_conf)
+
+    # Override mkimage assemble to inject extra opts
+    def run_mkimage_assemble(self, itsfile, fitfile):
```
ricardosalveti, line 80: This change can probably go directly to oe-core upstream, you are just adding support for extra mkimage arguments.

recipes-kernel/linux/qcom-dtb-metadata_git.bb
```diff
@@ -0,0 +1,25 @@
...
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+DEPENDS = "dtc-native"
+
+SRC_URI = "git://github.com/qualcomm-linux/qcom-dtb-metadata.git;branch=main;protocol=https"
+
+SRCREV = "6b9e4b9c093cba36a80035af103015f0a7d3f9fc"
+
+PV = "0.1+git"
```
ricardosalveti: This revision is exactly the v0.1 tag, no need for +git.

ricardosalveti: Also rename the recipe to be specific to the version used, which is v0.1 (instead of _git).

recipes-kernel/linux/qcom-dtb-metadata_git.bb
```diff
@@ -0,0 +1,25 @@
...
+
+inherit deploy
+
+do_compile() {
+    oe_runmake
+}
```
ricardosalveti: No need, already covered by base.bbclass.

recipes-kernel/linux/qcom-dtb-metadata_0.1.bb
```diff
@@ -0,0 +1,25 @@
...
+DEPENDS = "dtc-native"
+
+SRC_URI = "git://github.com/qualcomm-linux/qcom-dtb-metadata.git;branch=main;protocol=https"
+
+SRCREV = "6b9e4b9c093cba36a80035af103015f0a7d3f9fc"
+
+PV = "0.1+git"
+
+inherit deploy
+
```
ricardosalveti, line 18: `do_configure[noexec] = "1"`

recipes-kernel/linux/qcom-dtb-metadata_git.bb
```diff
@@ -0,0 +1,25 @@
...
+
+do_deploy() {
+    install -d ${DEPLOY_DIR_IMAGE}
+    install -m 0644 ${S}/qcom-metadata.dtb ${DEPLOY_DIR_IMAGE}/qcom-metadata.dtb
```
ricardosalveti: `${B} ` instead

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,111 @@
+inherit kernel-arch uboot-config
```
ricardosalveti: uboot-config?

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,111 @@
...
+DEPENDS += "\
+    u-boot-tools-native dtc-native \
+    qcom-dtb-metadata \
+"
+
+# Initialize root node
+python fitimage_init_rootnode() {
+    import sys, os
+    file_path = d.getVar('FILE')
+    customfit = os.path.join(os.path.dirname(os.path.dirname(os.path.dirname(file_path))), 'lib')
```
ricardosalveti: Use pathlib instead of this nested dirname calls.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,111 @@
...
+    if os.path.isdir(customfit) and customfit not in sys.path:
+        sys.path.insert(0, customfit)
+        bb.note("Added to sys.path (parse-time): %s" % customfit)
+    import oe.types
+    from qcom_fitimage import QcomItsNodeRoot
```
ricardosalveti: Move import lines together to the initial ones, add an empty line after.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,111 @@
...
+        d.getVar("FIT_DESC"), d.getVar("FIT_ADDRESS_CELLS"),
+        d.getVar('HOST_PREFIX'), d.getVar('UBOOT_ARCH'), d.getVar("FIT_CONF_PREFIX"),
+        oe.types.boolean(d.getVar('UBOOT_SIGN_ENABLE')), d.getVar("UBOOT_SIGN_KEYDIR"),
+        d.getVar("UBOOT_MKIMAGE"), d.getVar("UBOOT_MKIMAGE_DTCOPTS"),
+        d.getVar("UBOOT_MKIMAGE_SIGN"), d.getVar("UBOOT_MKIMAGE_SIGN_ARGS"),
+        d.getVar('FIT_HASH_ALG'), d.getVar('FIT_SIGN_ALG'), d.getVar('FIT_PAD_ALG'),
+        d.getVar('UBOOT_SIGN_KEYNAME'),
+        oe.types.boolean(d.getVar('FIT_SIGN_INDIVIDUAL')), d.getVar('UBOOT_SIGN_IMG_KEYNAME')
+    )
+    d.setVar("__fit_root_node", root_node)
```
ricardosalveti: Use a more standard OE var, use caps here.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,111 @@
...
+    bb.build.exec_func('fitimage_modify_its_file', d)
+    root_node.run_mkimage_assemble(itsfile, fitname)
+    root_node.run_mkimage_sign(fitname)
+}
+
+do_compile_qcom_fitimage[depends] += "qcom-dtb-metadata:do_deploy"
+
+python do_compile_qcom_fitimage() {
+    bb.build.exec_func('fitimage_generate_its', d)
+}
```
ricardosalveti, line 78: Do we really need to create all these functions here? It is fine to have it all under a single do_compile_qcom_fitimage function, similar to the way it is done at meta/classes-recipe/kernel-fit-image.bbclass.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,111 @@
...
+
+
+do_deploy_qcom_fitimage() {
+    install -m 0644 "${DEPLOYDIR}/fitImage" "${DEPLOY_DIR_IMAGE}/fitImage"
+    install -m 0644 "${DEPLOYDIR}/fit-image.its" "${DEPLOY_DIR_IMAGE}/fit-image.its"
```
ricardosalveti: That's how the normal kernel-fit is also deployed, pick a different name for both the binary and its.

conf/machine/include/qcom-common.inc
```diff
@@ -46,3 +46,16 @@ EFI_PROVIDER = "systemd-boot"
 
 # Unified Kernel Image (UKI) name
 EFI_LINUX_IMG ?= "linux-${MACHINE}.efi"
+
+# Variables to generate ITS file
+UBOOT_CONFIG ??= "qcom"
+UBOOT_CONFIG[qcom] ?= "qcom_defconfig"
```
ricardosalveti: How is this related to fit at all?

The only thing we're using from u-boot is mkimage.

conf/machine/include/qcom-common.inc
```diff
@@ -46,3 +46,16 @@ EFI_PROVIDER = "systemd-boot"
...
+
+# Pass additional options to mkimage command
+UBOOT_MKIMAGE_EXTRA_OPTS = "-E -B 0x0001000"
```
ricardosalveti: Why?

conf/machine/include/qcom-common.inc
```diff
@@ -46,3 +46,16 @@ EFI_PROVIDER = "systemd-boot"
...
+
+# Set default configuration
+FIT_CONF_DEFAULT_DTB = "1"
+
+# Set hash algorithm
+FIT_HASH_ALG = ""
```
ricardosalveti: These options will also affect users using kernel-fit (e.g. u-boot users).

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -4,8 +4,10 @@
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
 
+inherit dtb-fit-image
+
 DTBBIN_DEPLOYDIR = "${WORKDIR}/qcom_dtbbin_deploy-${PN}"
-DTBBIN_SIZE ?= "4096"
+DTBBIN_SIZE ?= "65536"
```
ricardosalveti: This will also impact the size of every dtb vfat image we generate.

lib/qcom/dtb_only_fitimage.py
vkraleti: Yes. Waiting for https://patchwork.yoctoproject.org/project/oe-core/patch/20251011095022.950420-1-kavinaya@qti.qualcomm.com/ to be accepted.

ricardosalveti: Thanks, replied asking for clarification for the reason why we need to customize the options here, it is not clear even in this PR.

conf/machine/include/qcom-common.inc
amrianan: It appears that @Kavinaya99 has referred to the kmake-image Docker script:
https://github.com/qualcomm-linux/kmake-image/blob/main/make_fitimage.sh#L86

The -E option is used to place binary data outside the FIT structure. However, since we are using /incbin/ in the .its file, all binaries are appended into the same image. The FIT structure then contains the offset and size for each appended binary.

The -B 8 option enforces 8‑byte alignment for the image. We want this standard to be applied across all QCOM targets. For Android specifically, Google requires 8‑byte alignment, so we have standardized on 8‑byte alignment for all cases. This also improves the robustness of the parsing logic.

lib/qcom/dtb_only_fitimage.py
Kavinaya99: Added the link to the new patch.

lib/qcom/dtb_only_fitimage.py
Kavinaya99: Renamed the patch file to lib/qcom_dtb_only_fitimage.py

lib/qcom_fitimage.py
Kavinaya99: Updated the License to GPL=2.0-only.

lib/qcom_fitimage.py
Kavinaya99: Updated in the patch.

lib/qcom_dtb_only_fitimage.py
Kavinaya99: Removed the comment.

lib/qcom_fitimage.py
Kavinaya99: Removed the duplicated function.

conf/machine/include/qcom-common.inc
Kavinaya99: Without setting these variables, Iam geting the below build error:
linux-qcom-next PROVIDES virtual/kernel but was skipped: Either UBOOT_MACHINE or UBOOT_CONFIG must be set in the rb3gen2-core-kit machine configuration

recipes-kernel/linux/qcom-dtb-metadata_git.bb
Kavinaya99: Updated the version in the new patch.

recipes-kernel/linux/qcom-dtb-metadata_git.bb
Kavinaya99: Removed the do_compile

recipes-kernel/linux/qcom-dtb-metadata_0.1.bb
Kavinaya99: Added do_configure in new patch.

recipes-kernel/linux/qcom-dtb-metadata_git.bb
Kavinaya99: Added ${B} in new patch.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: uboot-config is needed for uboot-mkimage which is used by run_mkimage_assemble

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Updated the new patch with Pathlib.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Added import statements to the beginning.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Added all the functions into one task.

classes/linux-qcom-dtbbin.bbclass
Kavinaya99: I have modified the size back to 4096

conf/machine/include/qcom-common.inc
Kavinaya99: I have moved these to dtb-fit-image.bbclass so that it will not affect users using kernel-fit.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Modified the filenames to qclinux-fit-image.its and qclinuxfitImage.

conf/machine/include/qcom-common.inc
Kavinaya99: Without setting these variables, Iam geting the below build error:
linux-qcom-next PROVIDES virtual/kernel but was skipped: Either UBOOT_MACHINE or UBOOT_CONFIG must be set in the rb3gen2-core-kit machine configuration

lib/qcom/dtb_only_fitimage.py
Kavinaya99: Added the comment.

classes-recipe/dtb-fit-image.bbclass
lumag: It's obviously incorrect way to get it, as you can see from adding dummy UBOOT_CONFIG variables.

lib/qcom/dtb_only_fitimage.py
```diff
@@ -0,0 +1,127 @@
...
+# For details on Qualcomm DTB metadata and FIT requirements, see:
+# https://github.com/qualcomm-linux/qcom-dtb-metadata/blob/main/Documentation.md
+
+import os
+import shlex
+import subprocess
+import bb
+from typing import Tuple
+import oe.fitimage
+from oe.fitimage import ItsNodeRootKernel, get_compatible_from_dtb
```
lumag, line 18: You've just imported a module, now you are importing parts of the module again. Why?

lib/qcom_dtb_only_fitimage.py
```diff
@@ -0,0 +1,127 @@
...
+        }
+        if load:
+            opt_props["load"] = f"<{load}>"
+
+        compatibles = None
+        if add_compatible:
+            if compatible_str:
+                compatibles = str(compatible_str).split()
+            elif dtb_ext != ".dtbo":
+                compatibles = get_compatible_from_dtb(dtb_abspath)
```
lumag: What for? Our standard does not allow using compatible entries from the DTB.

lib/qcom_dtb_only_fitimage.py
```diff
@@ -0,0 +1,127 @@
...
+            "none",
+            opt_props,
+            compatibles
+        )
+        self._dtbs.append((dtb_node, compatibles or []))
+
+    def fitimage_emit_section_config(self, default_dtb_image=None):
+        if self._dtbs:
+            counter = 0
+            for entry in self._dtbs:
```
lumag: `for counter, entry in enumerate(self._dtbs):`

lib/qcom_dtb_only_fitimage.py
```diff
@@ -0,0 +1,127 @@
...
+                if isinstance(entry, tuple) and len(entry) == 2:
```
lumag: Is there a way for this to be not true?

lib/qcom_dtb_only_fitimage.py
```diff
@@ -0,0 +1,127 @@
...
+                            self._fitimage_emit_one_section_config(conf_name, dtb_node)
+                            conf_node = self.configurations.sub_nodes[-1]
+                            conf_node.add_property('compatible', comp)
+                    else:
+                        conf_name = f"{self._conf_prefix}{counter}"
+                        counter += 1
+                        self._fitimage_emit_one_section_config(conf_name, dtb_node)
+                else:
+                    dtb_node = entry
+                    conf_name = f"{self._conf_prefix}{counter}"
```
lumag: There are 3 strings which do the same. Care to pull them up?

recipes-kernel/linux/qcom-dtb-metadata_0.1.bb
```diff
@@ -0,0 +1,21 @@
+SUMMARY = "Build qcom-metadata.dtb from vendored qcom-dtb-metadata"
+HOMEPAGE = "https://github.com/qualcomm-linux/qcom-dtb-metadata"
+
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+DEPENDS = "dtc-native"
+
+SRC_URI = "git://github.com/qualcomm-linux/qcom-dtb-metadata.git;branch=main;protocol=https"
```
lumag: ;tag=v${PV}

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,116 @@
...
+
+python do_compile_qcom_fitimage() {
+    import sys, os, shutil, re
+    import oe.types
+    from pathlib import Path
+
+    itsfile = os.path.join(d.getVar('DEPLOYDIR'), "qclinux-fit-image.its")
+    fitname = os.path.join(d.getVar('DEPLOYDIR'), "qclinuxfitImage")
+
+    customfit = Path(d.getVar('FILE')).parents[2] / 'lib'
```
lumag: FILE seems to be too generic.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,116 @@
...
+    if customfit.is_dir() and str(customfit) not in sys.path:
+        sys.path.insert(0, str(customfit))
```
lumag: Please add a comment, describing what is going on here.

conf/machine/include/qcom-common.inc
lumag: NAK

conf/machine/include/qcom-common.inc
lumag: Your comment doesn't match what is written in the file.

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -21,8 +23,13 @@ do_qcom_dtbbin_deploy() {
...
         rm -rf ${DTBBIN_DEPLOYDIR}/$dtb_base_name
     done
+    # Generate qclinux_fit.img along side combined-dtb.dtb
+    if [ -f "${DEPLOY_DIR_IMAGE}/qclinuxfitImage" ]; then
+        mkfs.vfat -S ${QCOM_VFAT_SECTOR_SIZE} -C ${DTBBIN_DEPLOYDIR}/multi-dtb.vfat ${DTBBIN_SIZE}
+        mcopy -i "${DTBBIN_DEPLOYDIR}/multi-dtb.vfat" -vsmpQ ${DEPLOY_DIR_IMAGE}/qclinuxfitImage ::/qclinux_fit.img
+    fi
 }
-addtask qcom_dtbbin_deploy after do_populate_sysroot do_packagedata before do_deploy
+addtask qcom_dtbbin_deploy after do_populate_sysroot do_packagedata do_deploy do_deploy_qcom_fitimage before do_build
```
lumag: Nope. This means that the files you deploy wouldn't be a part of the deploy sstate.

conf/machine/iq-9075-evk.conf
```diff
@@ -28,3 +28,5 @@ QCOM_BOOT_FILES_SUBDIR = "qcs9100"
 QCOM_PARTITION_FILES_SUBDIR ?= "partitions/iq-9075-evk/ufs"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-qcs9100"
+
+FIT_DTB_COMPATIBLE[lemans-evk] = "qcom,qcs9075-iot qcom,qcs9075v2-iot"
```
lumag: Why do we need both?

conf/machine/qcs9100-ride-sx.conf
```diff
@@ -33,3 +33,8 @@ QCOM_BOOT_FILES_SUBDIR = "qcs9100"
 QCOM_PARTITION_FILES_SUBDIR ?= "partitions/qcs9100-ride-sx/ufs"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-qcs9100"
+
+FIT_DTB_COMPATIBLE[qcs9100-ride] = "qcom,qcs9100-qam qcom,qcs9100v2-qam"
+FIT_DTB_COMPATIBLE[qcs9100-ride-r3] = "qcom,qcs9100-qamr2 qcom,qcs9100v2-qamr2"
+FIT_DTB_COMPATIBLE[sa8775p-ride] = "qcom,sa8775p-qam qcom,sa8775pv2-qam"
+FIT_DTB_COMPATIBLE[sa8775p-ride-r3] = "qcom,sa8775p-qamr2 qcom,sa8775pv2-qamr2"
```
lumag: Are they really identified as different SKUs?

conf/machine/rb3gen2-core-kit.conf
```diff
@@ -28,3 +28,6 @@ QCOM_BOOT_FILES_SUBDIR = "qcm6490"
 QCOM_PARTITION_FILES_SUBDIR ?= "partitions/qcs6490-rb3gen2/ufs"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-qcs6490"
+
+FIT_DTB_COMPATIBLE[qcs6490-rb3gen2-vision-mezzanine] = "qcom,qcs6490-iot-subtype2"
```
lumag: This is really sad. How do we define combinations of mezzanines?

lumag (changes requested):

conf/machine/include/qcom-common.inc
vkraleti: @lumag Could you please elaborate a bit on the mismatch you are referring to? Both "-B 0x0001000" and "-B 8" represent the same value, if that is what you meant.

lumag: How comes that 0x1000 = 8? (no I haven't looked into the docs)

classes-recipe/dtb-fit-image.bbclass
ricardosalveti: You should find a way to depend only on the mkimage tool (u-boot-mkimage-native).

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,116 @@
...
+
+    from qcom_dtb_only_fitimage import QcomItsNodeRoot
+
+    root_node = QcomItsNodeRoot(
+        d.getVar("FIT_DESC"),
+        d.getVar("FIT_ADDRESS_CELLS"),
+        d.getVar('HOST_PREFIX'),
+        d.getVar('UBOOT_ARCH'),
+        d.getVar("FIT_CONF_PREFIX"),
+        oe.types.boolean(d.getVar('UBOOT_SIGN_ENABLE')),
```
ricardosalveti: This is a bit overloaded, we're not expecting to use the same keys and related variables from u-boot. How will fit-dtb be signed and validated, via UEFI secure boot, no?

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,116 @@
...
+        sys.path.insert(0, str(customfit))
+        bb.note("Added to sys.path (build-time): %s" % customfit)
+
+    from qcom_dtb_only_fitimage import QcomItsNodeRoot
+
+    root_node = QcomItsNodeRoot(
+        d.getVar("FIT_DESC"),
+        d.getVar("FIT_ADDRESS_CELLS"),
+        d.getVar('HOST_PREFIX'),
+        d.getVar('UBOOT_ARCH'),
```
ricardosalveti: This won't cause any issue because ARCH will match, but again, abusing the u-boot integration.

classes/linux-qcom-dtbbin.bbclass
ricardosalveti: And the task is already added by dtb-fit-image.bbclass, why are you duplicating it here as well?

conf/machine/include/qcom-common.inc
ricardosalveti: That's because you are using uboot-config.bbclass, which should only be included if the indent is to build u-boot, which is not our use case here.

Think that your new class should also work correctly with users that are defining and building u-boot at the same time (not necessarily a use case we will use, but technically they should be able to be built together without interfering on each other).

lib/qcom_dtb_only_fitimage.py
```diff
@@ -0,0 +1,127 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+#
```
ricardosalveti: Remove this extra line, copyright lines should be together.

lib/qcom/dtb_only_fitimage.py
ricardosalveti: @Kavinaya99 please send a v2 explaining in more details why these extra arguments are required (explain why there is a need to generate a fit a place its data out of the description, the alignment, etc.

conf/machine/include/qcom-common.inc
Kavinaya99: Removed UBBOT_CONFIG and UBOOT_CONFIG[qcom] .

lib/qcom/dtb_only_fitimage.py
Kavinaya99: Sure

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Sure, updated the new patch after removing UBOOT_CONFIG and UBOOT_CONFIG[qcom] .

conf/machine/include/qcom-common.inc
Kavinaya99: Updated the new patch with the changes to remove dependency on these variables.

lib/qcom/dtb_only_fitimage.py
Kavinaya99: Will update in the new patch.

conf/machine/include/qcom-common.inc
Kavinaya99: @amrianan added the reason.

lib/qcom_dtb_only_fitimage.py
Kavinaya99: UEFI expects compatible strings for all DTBs to be present in the ITS file mandatorily. So, for some DTBs which does not have compatible strings passed through the variable "FIT_DTB_COMPATIBLE", we have a fallback mechanism to get the compatible string from the DTB itself.

conf/machine/include/qcom-common.inc
vkraleti: Documentation is not very clear if hex values / decimals need to be passed to '-B'. But from code this option is getting converted to _unsigned long int_. 

https://github.com/u-boot/u-boot/blob/master/tools/mkimage.c#L117 
https://github.com/u-boot/u-boot/blob/master/tools/mkimage.c#L240

conf/machine/qcs9100-ride-sx.conf
vkraleti: Yes, ride & ride-r3 are different SKUs.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: I used FILE because it was the bitbake variable that pointed to the current class file during parsing, allowing me to compute a relative path to lib/directory in meta-qcom.

conf/machine/iq-9075-evk.conf
Kavinaya99: We have multiple boards with same CDT but different SOC IDs.
To accommodate this, we need multiple compatible strings for a DTB.

conf/machine/rb3gen2-core-kit.conf
vkraleti: AFAIK, UEFI doesn't support dynamic switching of mezzanines. Each combination is considered as a different board and should have a separate entry in .its.

 @quic-kaushalk @amrianan please confirm.

classes-recipe/dtb-fit-image.bbclass
vkraleti: Most of these parameters are not really put into use in qcom_dtb_only_fitimage.py. They are passed just to satisfy the base class constructor. May be unwanted parameters can be simply passed as _None_ and just limit defining only required ones here.

classes/linux-qcom-dtbbin.bbclass
vkraleti: Agree. Problem here is the code is looking for .dtb files under DEPLOYDIR. Instead, if they are referenced from KERNEL_DTBDEST, there is no need for this complex task dependency order.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Added the comments.

classes/linux-qcom-dtbbin.bbclass
Kavinaya99: Will update the changes in the next patch

lib/qcom_dtb_only_fitimage.py
Kavinaya99: Updated in the new patch

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Removed the u-boot related variables in the new patch.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Removed the u-boot related patches in the new patch.

lib/qcom_dtb_only_fitimage.py
Kavinaya99: Done

lib/qcom_dtb_only_fitimage.py
Kavinaya99: In many instances, we have DTBs having multiple compatible strings. In such cases,the switch to enumerate(self._dtbs) ties the conf-* index to the DTB position rather than to the number of configurations emitted. 
We need a single DTB to emit multiple configuration blocks when it carries multiple compatible strings. 
With enumerate, all those blocks reuse the same index (e.g., conf-5), producing duplicate node names causing mkimage to fail with the error: duplicate conf entries
                conf-5 {
                        description = "0 FDT blob";
                        fdt = "fdt-lemans-evk.dtb";
                        compatible = "qcom,qcs9075v2-iot";
                };
                conf-5 {
                        description = "0 FDT blob";
                        fdt = "fdt-qcs9100-ride.dtb";
                        compatible = "qcom,qcs9100-qam";

lib/qcom_dtb_only_fitimage.py
lumag: This is lame. If there is no crafted compat string, then error out instead of making it look like everything works, while it doesn't.

lib/qcom_dtb_only_fitimage.py
lumag: I don't see a DT generating multiple entries in your example. Also please use MD formatting when pasting code.

conf/machine/iq-9075-evk.conf
lumag: `qcom,qcs9075-iot` should match _all_ QCS9075 SKUs, not matter which version is it.

conf/machine/qcs9100-ride-sx.conf
lumag: I was thinking about `qcom,sa8775p` vs `qcom,sa8775pv2`. It's the same issue as the one few lines above: it lists an unspecified version (which should match all versions) and an exact version.

conf/machine/rb3gen2-core-kit.conf
lumag: I am really, really leaning towards using UEFI variable for defining mezzanines. Otherwise on every plug or unplug of any mezz one would have to reflash the CDT.

conf/machine/include/qcom-common.inc
lumag: Hex 0x1000 is 4096 decimal, not 8.

classes-recipe/dtb-fit-image.bbclass
lumag: Just define a variable in the `layer.conf` and then reference it here. Also there is no need to check that it is a dir, you _know_ that.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,110 @@
...
+    sys.dont_write_bytecode = True
+    os.environ['PYTHONDONTWRITEBYTECODE'] = '1'
+
+    # Add the custom 'lib' directory (two levels up from this recipe file) to sys.path
+    # so that Python can import our custom FIT image helper module (qcom_dtb_only_fitimage.py)
+    # during BitBake build time.
+    customfit = Path(d.getVar('FILE')).parents[2] / 'lib'
+    if customfit.is_dir() and str(customfit) not in sys.path:
+        sys.path.insert(0, str(customfit))
+        bb.note("Added to sys.path (build-time): %s" % customfit)
```
lumag: no need to spam the logs.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,110 @@
...
+    import sys, os, shutil, re
+    import oe.types
+    from pathlib import Path
+
+    itsfile = os.path.join(d.getVar('DEPLOYDIR'), "qclinux-fit-image.its")
+    fitname = os.path.join(d.getVar('DEPLOYDIR'), "qclinuxfitImage")
+
+    # Prevent Python from writing .pyc/.pyo files during build.
+    sys.dont_write_bytecode = True
+    os.environ['PYTHONDONTWRITEBYTECODE'] = '1'
```
ricardosalveti: What is the problem with .pyc/.pyo files?

conf/machine/include/qcom-common.inc
```diff
@@ -46,3 +46,6 @@ EFI_PROVIDER = "systemd-boot"
 
 # Unified Kernel Image (UKI) name
 EFI_LINUX_IMG ?= "linux-${MACHINE}.efi"
+
+# Pass additional options to mkimage command
+MKIMAGE_EXTRA_OPTS = "-E -B 8"
```
ricardosalveti: Make this variable specific to your needs, like FIT_DTB_MKIMAGE_EXTRA_OPTS or similar.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Sure

conf/machine/include/qcom-common.inc
Kavinaya99: Added -B 8

lib/qcom_dtb_only_fitimage.py
Kavinaya99: Updated new patch with the suggestion

lib/qcom_dtb_only_fitimage.py
Kavinaya99: Added a warning if a dtb is not having compatible string.

classes/linux-qcom-dtbbin.bbclass
Kavinaya99: Will upload a new patch to resolve this.

conf/machine/iq-9075-evk.conf
Kavinaya99: Added FIT_DTB_COMPATIBLE[lemans-evk] = "qcom,qcs9075-iot"

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Keeps the source directory free from extra files like __pycache__.
Improves build reproducibility by eliminating timestamp-dependent artifacts.
Avoids host-specific bytecode

conf/machine/include/qcom-common.inc
Kavinaya99: Updated in the new patch.

conf/machine/qcs9100-ride-sx.conf
Kavinaya99: Updated the variable for all the DTBs.

classes-recipe/dtb-fit-image.bbclass
ricardosalveti: Sure, but this is not really required here, so unless really done to fix an issue within the class, remove it.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,109 @@
...
+
+    from qcom_dtb_only_fitimage import QcomItsNodeRoot
+
+    root_node = QcomItsNodeRoot(
+        d.getVar("FIT_DESC"),
+        d.getVar("FIT_ADDRESS_CELLS"),
+        d.getVar("FIT_CONF_PREFIX"),
+        d.getVar("MKIMAGE"),
+    )
+    bb.debug(1, "Global root_node initialized")
```
ricardosalveti: This debug is not really useful I would say, better remove it.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,109 @@
...
+}
+
+do_deploy_qcom_fitimage() {
+    install -m 0644 "${DEPLOYDIR}/qclinuxfitImage" "${DEPLOY_DIR_IMAGE}/qclinuxfitImage"
+    install -m 0644 "${DEPLOYDIR}/qclinux-fit-image.its" "${DEPLOY_DIR_IMAGE}/qclinux-fit-image.its"
+}
+
+
+addtask compile_qcom_fitimage after do_deploy
+addtask deploy_qcom_fitimage after do_compile_qcom_fitimage do_deploy
```
ricardosalveti: Rename the function to remove "compile" from it, as it can be confusing since this is not really related to the compile task. 'generate_qcom_fitimage' or similar would work better.

conf/machine/include/qcom-common.inc
ricardosalveti: Comment is still not explaining the needs for the extra arguments.

classes/linux-qcom-dtbbin.bbclass
Kavinaya99: Uploaded a new patch to address this.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Removed this in the new patch.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Removed the line.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Updated the name in the new patch.

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -21,9 +24,14 @@ do_qcom_dtbbin_deploy() {
         mcopy -i "${DTBBIN_DEPLOYDIR}/dtb-${dtb_base_name}-image.vfat" -vsmpQ ${DTBBIN_DEPLOYDIR}/$dtb_base_name/* ::/
         rm -rf ${DTBBIN_DEPLOYDIR}/$dtb_base_name
     done
+    # Generate qclinux_fit.img along side combined-dtb.dtb
+    if [ -f "${D}/${KERNEL_DTBDEST}/qclinuxfitImage" ]; then
+        mkfs.vfat -S ${QCOM_VFAT_SECTOR_SIZE} -C ${DTBBIN_DEPLOYDIR}/multi-dtb.vfat ${DTBBIN_SIZE}
+        mcopy -i "${DTBBIN_DEPLOYDIR}/multi-dtb.vfat" -vsmpQ ${D}/${KERNEL_DTBDEST}/qclinuxfitImage ::/qclinux_fit.img
```
vkraleti: This  isn’t correct. At this stage, the build is already in the deploy phase. Updating ${D}, which is expected to be populated during do_install, is not desirable. This may not even work if do_install is skipped due to a matching in sstate cache.

lumag:
> @ricardosalveti @lumag I am having a second thought about specifying compatible strings in each machine configuration individually. Since there’s also a requirement to support a generic qcom-armv8 machine, and given the possibility of sharing DTBs across machines, would it be better to consolidate all compatible strings into a single file and include it in qcom-common.inc? Are you aligned with this approach?

Yes, please.

lib/qcom/dtb_only_fitimage.py
Kavinaya99: https://lists.openembedded.org/g/openembedded-core/message/227471

conf/machine/include/qcom-common.inc
Kavinaya99: - `-E` enables external data in FIT images, which is essential for
  modular boot setups, secure boot workflows, and reducing image size.
- `-B 8` enforces 8-byte alignment, ensuring compatibility with boot
  ROM requirements, improving memory access efficiency, and supporting
  predictable offsets for multi-component FIT images.

classes/linux-qcom-dtbbin.bbclass
Kavinaya99: Updated the new patch.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,108 @@
...
+    for dtb in kernel_devicetree.split():
+        dtb_name = os.path.basename(dtb)
+        dtb_base = os.path.splitext(dtb_name)[0]
+        compatible = d.getVarFlag("FIT_DTB_COMPATIBLE", dtb_base) or ""
+        dtb_path = os.path.join(dtb_dir, f"{dtb_base}.dtb")
+        if not compatible:
+            bb.warn(f"FIT_DTB_COMPATIBLE[{dtb_base}] is not set. ")
+        root_node.fitimage_emit_section_dtb(
+            dtb_name, dtb_path,
+            False,
```
lumag: What does `dtb_loadaddress = False` mean? Or `dtbo_loadaddress = False`?

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,108 @@
...
+        bb.fatal("Missing qcom-metadata.dtb at %s (ensure qcom-dtb-metadata:do_deploy ran)" % qcom_meta_src)
+    shutil.copy(qcom_meta_src, qcom_meta_dst)
+
+    for dtb in kernel_devicetree.split():
+        dtb_name = os.path.basename(dtb)
+        dtb_base = os.path.splitext(dtb_name)[0]
+        compatible = d.getVarFlag("FIT_DTB_COMPATIBLE", dtb_base) or ""
+        dtb_path = os.path.join(dtb_dir, f"{dtb_base}.dtb")
+        if not compatible:
+            bb.warn(f"FIT_DTB_COMPATIBLE[{dtb_base}] is not set. ")
```
lumag: bb.fatal() ? Or bb.error()?

lib/qcom_dtb_only_fitimage.py
```diff
@@ -0,0 +1,109 @@
...
+                         None, None, False, None)
+
+        self._mkimage_extra_opts = []
+        self._dtbs = []
+
+    def set_extra_opts(self, mkimage_extra_opts):
+        self._mkimage_extra_opts = shlex.split(mkimage_extra_opts) if mkimage_extra_opts else []
+
+    # Emit the DTB section for the FIT image
+    def fitimage_emit_section_dtb(self, dtb_id, dtb_path, dtb_loadaddress=None,
```
lumag: Load addresses are always 0, add_compatible is always True, can we drop them?

lib/qcom_dtb_only_fitimage.py
```diff
@@ -0,0 +1,109 @@
...
+        opt_props = {
+            "data": '/incbin/("' + dtb_path + '")',
+            "arch": self._arch
+        }
+        if load:
+            opt_props["load"] = f"<{load}>"
+
+        compatibles = None
+        if add_compatible and compatible_str:
+            compatibles = str(compatible_str).split()
```
lumag: Why do we need extra str() here?

lib/qcom_dtb_only_fitimage.py
```diff
@@ -0,0 +1,109 @@
...
+
+        dtb_node = self.its_add_node_dtb(
+            "fdt-" + dtb_id,
+            "Flattened Device Tree blob",
+            "flat_dt",
+            "none",
+            opt_props,
+            compatibles
+        )
+        self._dtbs.append((dtb_node, compatibles or []))
```
lumag: compatibles can't be None here, we know that for sure.

lib/qcom_dtb_only_fitimage.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def fitimage_emit_section_config(self, default_dtb_image=None):
+        counter = 0
+        for counter, (dtb_node, compatibles) in enumerate(self._dtbs):
+            if compatibles:
```
lumag: compatibles can't be None.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,108 @@
...
+        root_node.fitimage_emit_section_dtb(
+            dtb_name, dtb_path,
+            False,
+            False,
+            True,
+            compatible_str=compatible,
+        )
+
+    root_node.fitimage_emit_section_config(d.getVar("FIT_CONF_DEFAULT_DTB"))
```
lumag: What does default DTB mean in this Fit Image?

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,108 @@
...
+    # Prepend qcom-metadata.dtb to KERNEL_DEVICETREE
+    kernel_devicetree = d.getVar('KERNEL_DEVICETREE') or ""
+    kernel_devicetree = ('qcom-metadata.dtb ' + kernel_devicetree).strip()
+
+    deploy_dir_image = d.getVar('DEPLOY_DIR_IMAGE') 
+    dtb_dir = os.path.join(d.getVar('B'), "arch", d.getVar('ARCH'), "boot", "dts", "qcom")
+    qcom_meta_src = os.path.join(deploy_dir_image, 'qcom-metadata.dtb')
+    qcom_meta_dst = os.path.join(dtb_dir, 'qcom-metadata.dtb')
+    if not os.path.exists(qcom_meta_src):
+        bb.fatal("Missing qcom-metadata.dtb at %s (ensure qcom-dtb-metadata:do_deploy ran)" % qcom_meta_src)
```
lumag: If the recipe depends on it, how could it not run?

conf/machine/include/qcom-common.inc
ricardosalveti: Please update the comment in the file saying what you described here.

Kavinaya99: Added comments in the new patch.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,108 @@
+inherit kernel-arch
```
vkraleti, line 7:
```
#
# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
#
# SPDX-License-Identifier: BSD-3-Clause-Clear
#
```
Add above at the beginning.

Kavinaya99: Sure, added in the new patch.

classes-recipe/dtb-fit-image.bbclass
lumag: From the _platform_'s point of view.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,114 @@
...
+    # Generate custom ITS file
+    with open(itsfile, 'r') as f:
+        content = f.read()
+
+    # Replace type for qcom metadata node
+    content = re.sub(
+        r'(fdt-qcom-metadata\.dtb\s*\{[^}]*?)type\s*=\s*"flat_dt";',
+        r'\1type = "qcom_metadata";',
+        content, flags=re.DOTALL
+    )
```
lumag: Why do we need to patch it if we just generated the file?

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,114 @@
...
+
+    # Remove conf-0 entry which corresponds to fdt-0
+    content = re.sub(r'conf-0\s*\{[^}]*\};', '', content, flags=re.DOTALL)
```
lumag: Why??

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,114 @@
...
+SSTATETASKS += "do_generate_qcom_fitimage"
+do_generate_qcom_fitimage[sstate-inputdirs] = "${QCOMFIT_DEPLOYDIR}"
+do_generate_qcom_fitimage[sstate-outputdirs] = "${DEPLOY_DIR_IMAGE}"
+
+python do_generate_qcom_fitimage_setscene () {
+    sstate_setscene(d)
+}
+addtask do_generate_qcom_fitimage_setscene
+
+do_generate_qcom_fitimage[stamp-extra-info] = "${MACHINE_ARCH}"
```
lumag, line 81: Why?

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -21,6 +23,12 @@ do_qcom_dtbbin_deploy() {
         mcopy -i "${DTBBIN_DEPLOYDIR}/dtb-${dtb_base_name}-image.vfat" -vsmpQ ${DTBBIN_DEPLOYDIR}/$dtb_base_name/* ::/
         rm -rf ${DTBBIN_DEPLOYDIR}/$dtb_base_name
     done
+
+    # Generate qclinux_fit.img along side combined-dtb.dtb
+    if [ -f "${DEPLOY_DIR_IMAGE}/qclinuxfitImage" ]; then
```
lumag: Under which conditions there will be no `qclinuxfitImage` file?

conf/machine/include/qcom-common.inc
```diff
@@ -46,3 +47,11 @@ EFI_PROVIDER = "systemd-boot"
 
 # Unified Kernel Image (UKI) name
 EFI_LINUX_IMG ?= "linux-${MACHINE}.efi"
+
+# Pass additional options to mkimage command
+# The -E option enables external data in FIT images, which is essential for
+# modular boot setups, secure boot workflows, and reducing image size.
+# The -B 8 option enforces 8-byte alignment, ensuring compatibility with boot
+# ROM requirements, improving memory access efficiency, and supporting
```
lumag: Is the **boot ROM** going to read the DTB partition?

recipes-kernel/linux/qcom-dtb-metadata_0.1.bb
```diff
@@ -0,0 +1,21 @@
+SUMMARY = "Build qcom-metadata.dtb from vendored qcom-dtb-metadata"
```
lumag: The summary is going to be embedded into the generated package. The "build something" isn't going to be a correct description of a package: it doesn't build anything.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Updated the changes in the new patch

lib/qcom_dtb_only_fitimage.py
Kavinaya99: Sure, Updated the changes in the new patch

lib/qcom_dtb_only_fitimage.py
Kavinaya99: Updated the changes in the new patch

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: The defaul field is not needed. Updated the changes in the new patch

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Updated the changes in the new patch.

recipes-kernel/linux/qcom-dtb-metadata_0.1.bb
Kavinaya99: Updated the changes in the new patch

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Updated the changes in the new patch.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Updated the changes in the new patch

lib/qcom_dtb_only_fitimage.py
Kavinaya99: Sure. Updated the changes in the new patch

lib/qcom_dtb_only_fitimage.py
Kavinaya99: Updated the changes in the new patch

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Changed it to bb.fatal.

classes/linux-qcom-dtbbin.bbclass
Kavinaya99: Removed this check and updated in the new patch.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: The FIT image and DTB/DTBO selection are machine‑specific and deployed under deploy/images/${MACHINE}. Setting do_generate_qcom_fitimage[stamp-extra-info] = ${MACHINE_ARCH} keys the sstate per machine, preventing cross‑machine reuse and ensuring correct artifacts are restored.

conf/machine/include/qcom-common.inc
Kavinaya99: Updated the comment. The -B 8 is a requirement from the UEFI. Without this, device will not bootup.

recipes-kernel/linux/qcom-dtb-metadata_0.1.bb
```diff
@@ -0,0 +1,21 @@
+SUMMARY = "Device Tree metadata blob for QCOM platforms"
```
quaresmajose, line 1: `qcom-dtb-metadata: add recipe for FIT image generation` is more aligned with the oe-core commit pattern.

recipes-kernel/linux/qcom-dtb-metadata_0.1.bb
```diff
@@ -0,0 +1,21 @@
...
+
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+DEPENDS += "dtc-native"
+
+SRC_URI = "git://github.com/qualcomm-linux/qcom-dtb-metadata.git;branch=main;protocol=https;tag=v${PV}"
+
+SRCREV = "6b9e4b9c093cba36a80035af103015f0a7d3f9fc"
+
```
quaresmajose, line 12: can you please set the `INHIBIT_DEFAULT_DEPS = "1"` to not pull in the default dependencies on the build sysroot.

conf/layer.conf
```diff
@@ -27,3 +27,6 @@ BBFILES_DYNAMIC += " \
 # Set default provider for virtual-diag-router to avoid warnings
 # when both diag and diag-router are available
 PREFERRED_RPROVIDER_virtual-diag-router ?= "diag"
+
+# Location for layer-specific Python helpers
+QCOM_PYTHON_LIB = "${LAYERDIR}/lib"
```
quaresmajose: can you please try to use the `addpylib` directive? something like: `addpylib ${LAYERDIR}/lib oe` will do the same.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -0,0 +1,89 @@
...
+
+    fit_dir = d.getVar('QCOMFIT_DEPLOYDIR')
+
+    itsfile = os.path.join(fit_dir, "qclinux-fit-image.its")
+    fitname = os.path.join(fit_dir, "qclinuxfitImage")
+
+    # Add the custom 'lib' directory to sys.path
+    # so that Python can import our custom FIT image helper module (qcom_dtb_only_fitimage.py)
+    # during BitBake build time.
+    libdir = Path(d.getVar('QCOM_PYTHON_LIB'))
```
quaresmajose: if the addpylib works we can drop this one.

conf/machine/include/qcom-common.inc
```diff
@@ -46,3 +47,9 @@ EFI_PROVIDER = "systemd-boot"
 
 # Unified Kernel Image (UKI) name
 EFI_LINUX_IMG ?= "linux-${MACHINE}.efi"
+
+# Pass additional options to mkimage command
+# The -E option enables external data in FIT images, which is essential for
+# modular boot setups, secure boot workflows, and reducing image size.
+# The -B 8 option enforces 8-byte alignment, as per QCOM UEFI requirements.
+FIT_DTB_MKIMAGE_EXTRA_OPTS = "-E -B 8"
```
quaresmajose: can you set it with a weak assign `=?` so we can change it if needed?

Kavinaya99: Sure.Updated in the new patch.

classes-recipe/dtb-fit-image.bbclass
Kavinaya99: Updated with the use of addpylib

conf/layer.conf
Kavinaya99: Added addpylib ${LAYERDIR}/lib qcom

recipes-kernel/linux/qcom-dtb-metadata_0.1.bb
Kavinaya99: Sure.Updated in the new patch.

recipes-kernel/linux/qcom-dtb-metadata_0.1.bb
Kavinaya99: Updated the commit message

quaresmajose: it still starts with `recipes-kernel: add recipe for FIT image generation`, the coomom pattern starts with the recipe or class name

conf/layer.conf
quaresmajose: thanks

quaresmajose (approved):
LGTM

PR #1283, opened 2025-12-04
refpolicy: Add SELinux policy for qrtr-ns.service

sasikuma-qti:


sasikuma-qti:
> Are there any cons of managing these as part of the original project itself? The project recipe can compile and package policies that can't be upstreamed so this doesn't need to be in a `bbappend` unless these are very Yocto/distro specific which isn't mentioned here.

qrtr-ns.service is Qualcomm specific service. Upstream refpolicy community accept policies for generic services only.

PR #1288, opened 2025-12-05
Add CI jobs for qcom-distro-selinux

vkraleti:
Add kas fragment and extend build-yocto.yml to also cover qcom-distro-selinux as another variant for prop image.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -76,9 +76,10 @@ jobs:
             yamlfile: ""
           - name: qcom-distro
             yamlfile: ':ci/qcom-distro-prop-image.yml'
-          # kvm is compatible with all the compile warm up machines
           - name: qcom-distro-kvm
             yamlfile: ':ci/qcom-distro-kvm.yml'
+          - name: qcom-distro-selinux
+            yamlfile: ':ci/qcom-distro-selinux.yml'
           - name: qcom-distro-sota
             yamlfile: ':ci/qcom-distro-sota.yml'
         kernel:
@@ -128,6 +129,8 @@ jobs:
             yamlfile: ""
           - name: qcom-distro
             yamlfile: ':ci/qcom-distro-prop-image.yml'
+          - name: qcom-distro-selinux
+            yamlfile: ':ci/qcom-distro-selinux.yml'
           - name: qcom-distro-sota
             yamlfile: ':ci/qcom-distro-sota.yml'
         kernel:


.github/workflows/build-yocto.yml
```diff
@@ -81,6 +81,8 @@ jobs:
           # kvm is compatible with all the compile warm up machines
           - name: qcom-distro-kvm
             yamlfile: ':ci/qcom-distro-kvm.yml'
+          - name: qcom-distro-selinux-prop-image
+	    yamlfile: ':ci/qcom-distro-selinux-prop-image.yml'
```
ricardosalveti: tab x spaces

.github/workflows/build-yocto.yml
```diff
@@ -81,6 +81,8 @@ jobs:
           # kvm is compatible with all the compile warm up machines
           - name: qcom-distro-kvm
             yamlfile: ':ci/qcom-distro-kvm.yml'
+          - name: qcom-distro-selinux-prop-image
+            yamlfile: ':ci/qcom-distro-selinux-prop-image.yml'
```
lumag: No, this should not be done in the 'cache warm up' stage. I'll send corresponding changes to drop other jobs from this part.

jaihindy: Thanks for clarifying that this shouldn’t be part of the cache warm-up stage. Hope you are working on this and will  share a change for this.

ndechesne: @lumag , i think it makes sense to kvm and sota in the warmup section since they are generic distro built for several machines, so the warmup seems 'useful' in this case (though we might benefit to split into 2 warmup sections for 'base' and slightly modified distro.. but it's pushing too much). 

However we should not have any of the 'prop' distro or images in the warmup, since there is very little benefit since the prop content is typically machine specific or proprietary blob. 

do you agree?

lumag: I think, the main purpose of the warmup is to push armv8a  / armv8-2a builds into the sstate-cache. Please correct me if I'm wrong, but neither kvm nor sota will result in a big number of additions to the sstate-cache (on the other hand, building them at this stage will take more time as the packages are not a part of the cache).

Performing separate prop build at this stage doesn't make sense. However I still hope to land #1106, removing a separate build step.

ndechesne: > I think, the main purpose of the warmup is to push armv8a / armv8-2a builds into the sstate-cache. Please correct me if I'm wrong, but neither kvm nor sota will result in a big number of additions to the sstate-cache (on the other hand, building them at this stage will take more time as the packages are not a part of the cache).
> 

there are a few changes for kvm and sota, but maybe not much, so perhaps it's indeed easier to keep the warmup simple, so let's move them in the compile section.

> Performing separate prop build at this stage doesn't make sense. However I still hope to land #1106, removing a separate build step.

It had been a while since I checked #1106, it does look really good now. we could split compile into compile_armv8 and compile_armv7 to simplify a bit the matrix definition, but it can be done separately.

rsiddoji: @lumag , @ndechesne   hope we are still tracking and working on this request.

PR #1294, opened 2025-12-09
linux-qcom-next: enable I2C_QCOM_GENI as built-in

vdadhani:
On Lemans and Monaco EVK platforms the boot-from-SD-card feature requires an IO expander to be enabled at an early stage to detect the SD card. This IO expander is connected over I2C.

Currently the I2C driver is built as a loadable module which resides on the SD card along with other modules. This creates a circular dependency where to detect the SD card the IO expander must be initialized the IO expander depends on the I2C driver and the I2C driver is on the SD card which cannot be accessed until detection succeeds.

To break this dependency enable I2C_QCOM_GENI as built-in so that the driver is available during early boot.


vkraleti:
@vdadhani Having an SD card slot is a board-level design decision, not a distribution choice. The purpose of qcom.cfg is to manage kernel configurations required by the distribution. In my opinion, the right place for this configuration would be:
https://github.com/qualcomm-linux/kernel/blob/qcom-next/arch/arm64/configs/qcom.config

Could you check if this configuration can be accepted there?

PR #1297, opened 2025-12-09
Create commit-check.yml

ynancher:
Adds the commit-msg-check GitHub Action
Checks commit messages added in pull requests
Valid for any PRs raised on this repository
Enforces these checks:

Commit Subject: Must exist and be ≤ 50 characters
Commit Body: Each line must wrap at 72 characters
Blank Line Check: Ensures a blank line between subject and body Also checks for blank line before Signed-off-by

### DIFF
--- a/.github/workflows/commit-check.yml
+++ b/.github/workflows/commit-check.yml
@@ -1,0 +1,19 @@
+name: Commit Msg Check Action
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  check-commits:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


lumag:
I prefer not to have written-in-stone limits and use common sense instead.

ynancher:
> I prefer not to have written-in-stone limits and use common sense instead.

I see your point about not using strict limits. The reason we suggest having configurable checks is to maintain consistency across multiple contributors. While these checks are optional, they help ensure clarity and readability in commit history. If you’d prefer not to enable them, that’s completely fine. Thank you for your time!

PR #1311, opened 2025-12-17
efivar: Add patch to support runtime variable persistence with file-based storage

codeshravan:
This patch adds support for persisting EFI variable changes on embedded boards that store variables in a file on the ESP partition rather than in dedicated hardware storage.

Many embedded boards lack secure hardware storage (like RPMB) for EFI variables and instead store them in a file on the ESP. When using systemd-boot with file-based variable storage, runtime variable changes are lost on reboot since they only exist in memory.

The upstream patch (pr-282.patch) implements automatic persistence by:
- Detecting RTStorageVolatile and VarToFile EFI variables
- Automatically writing variable changes back to the ESP file
- Ensuring EFI variable modifications survive reboots

This enables proper support for:
- systemd-boot operations that rely on SetVariable at runtime
- Boot loader variables (LoaderEntryDefault, LoaderEntryOneShot, etc.)
- UEFI capsule updates that modify variables
- Any runtime EFI variable changes that need to persist

The patch is currently pending upstream review.

Upstream-Status: Pending

### DIFF
--- a/recipes-bsp/efivar/efivar/pr-282.patch
+++ b/recipes-bsp/efivar/efivar/pr-282.patch
@@ -1,0 +1,239 @@
+From fb90073d51dd8a9f726fbbfcf4fab4aa7781eea3 Mon Sep 17 00:00:00 2001
+From: Ilias Apalodimas <ilias.apalodimas@linaro.org>
+Date: Wed, 18 Jun 2025 22:37:04 +0300
+Subject: [PATCH] efivarfs: Update a file variable store On SetVariable RT
+
+Upstream-Status: Pending
+
+Embedded boards have hardware limitations when storing and managing EFI
+variables. Some hardware comes with an eMMC & an RPMB partition which they
+use to store the EFI variables securely. However, the vast majority of
+boards (using U-Boot), stores the EFI variables in a file in the ESP.
+
+This has a few limitations
+- UEFI secure boot cannot be enabled as it can be very easily
+  overridden
+- SetVariable at runtime is impossible to support
+
+Distros and capsule updates on-disk do rely on the that service though
+and U-Boot does implement a workaround.
+
+U-Boot enables SetVariableRT in the RTPROP table and creates a memory backend,
+so the linux kernel can naturally read and write variables via the efivarfs
+filesystem. Those reads and writes end up in memory though. So they are visible
+while the OS is live and are lost in the event of a reboot.
+
+At the same time it also creates two EFI RO variables.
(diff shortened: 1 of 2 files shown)


recipes-bsp/efivar/efivar/pr-282.patch
```diff
@@ -0,0 +1,239 @@
+From fb90073d51dd8a9f726fbbfcf4fab4aa7781eea3 Mon Sep 17 00:00:00 2001
+From: Ilias Apalodimas <ilias.apalodimas@linaro.org>
+Date: Wed, 18 Jun 2025 22:37:04 +0300
+Subject: [PATCH] efivarfs: Update a file variable store On SetVariable RT
+
+Upstream-Status: Pending
```
lumag, line 6: The tags should be at the end of the message. Also please replace Pending with a proper one.

recipes-bsp/efivar/efivar_%.bbappend
```diff
@@ -0,0 +1,2 @@
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}:"
+SRC_URI:append:qcom = " file://pr-282.patch "
```
lumag, line 2: I would prefer if this goes to OE-Core, there is nothing Qualcomm-specific in it. If it is rejected there, we will take it into this layer.

sbanerjeeqc: @codeshravan the pr is merged in upstream efivars repo.
I have created an issue enquiring about v40 release for efivars.
Once maintainer responds we can decide either to bump oe-core efivars to v40 or otherwise backport.

codeshravan: @sbanerjee-quic @lumag based upon reply on issue:https://github.com/rhboot/efivar/issues/291 we will take it forward

recipes-bsp/efivar/efivar/pr-282.patch
quaresmajose: In this case the upstream status is backport 
```
Upstream-Status: Backport [https://github.com/rhboot/efivar/commit/68daa04654acbe1bbaa17ebfc23c371b39e69c6b]
```

codeshravan:
@lumag I realized it later, I have sent the patch through email, and close the pull request on email.

lool:
@codeshravan could you share a list of boards you'd like to use this on? I'm curious which boards are affected by this as to add efivars to the Debian builds too.

PR #1312, opened 2025-12-18
Workflows/lava test plans

mwasilew:
Replace LAVA test job generation mechanism with lava-test-plans. This change allows to manage the test job generation outside of meta-qcom and reuse the same tests and test jobs in different projects.

### DIFF
--- a/.github/actions/lava-test-plans/action.yml
+++ b/.github/actions/lava-test-plans/action.yml
@@ -26,6 +26,10 @@ runs:
           path: lava-test-plans
           ref: cdb8221de2a8f78533ef6b812fc4f5a2461c5525
 
+      - uses: actions/setup-python@v6
+        with:
+          python-version: 3.11
+
       - name: 'Download build URLs'
         uses: actions/download-artifact@v6
         with:
@@ -42,6 +46,7 @@ runs:
           OUT_PATH="${GITHUB_WORKSPACE}/out"
           VARS_OUT_PATH="${OUT_PATH}/vars"
           JOBS_OUT_PATH="${OUT_PATH}/jobs"
+          echo "JOBS_OUT_PATH=${JOBS_OUT_PATH}" >> $GITHUB_ENV
           mkdir -p "${VARS_OUT_PATH}"
           mkdir -p "${JOBS_OUT_PATH}"
           echo "PROJECT_NAME=${GITHUB_REPOSITORY#*/}" >> "${VARS_OUT_PATH}/gh-variables.ini"
@@ -100,12 +105,12 @@ runs:
       - name: "List files"
         shell: bash
         run: |
-          echo $GITHUB_WORKSPACE
-          ls -R $GITHUB_WORKSPACE
+          echo "${GITHUB_WORKSPACE}"
+          ls -R "${GITHUB_WORKSPACE}"
 
       - name: 'Upload test jobs'
         uses: actions/upload-artifact@v4
         with:
           name: ${{ inputs.prefix }}-${{ inputs.machine }}-${{ inputs.distro_name }}
-          path: ${{ env.MACHINE}}*-${{ inputs.distro_name }}*
+          path: ${{ env.JOBS_OUT_PATH }}/**
 


.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: Run lava-test-plans
+        shell: bash
+        run: |
+          cd lava-test-plans && pip install .
+          lava-test-plans --help
+          echo "PROJECT_NAME=${GITHUB_REPOSITORY#*/}" >> gh-variables.ini
+          echo "PROJECT=projects/${GITHUB_REPOSITORY#*/}/" >> gh-variables.ini
+          #echo "PROJECT_NAME=meta-qcom" >> gh-variables.ini
+          #echo "PROJECT=projects/meta-qcom/" >> gh-variables.ini
+          echo "LAVA_JOB_PRIORTIY=50" >> gh-variables.ini
```
smuppand: If lava-test-plans expects LAVA_JOB_PRIORITY, this will silently do nothing. Please fix the spelling.

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,109 @@
...
+          echo "IMAGE_FILE_NAME=${IMAGE_TYPE}-${{ inputs.machine }}.rootfs.qcomflash.tar.gz" >> gh-variables.ini
+          export BUILD_URL=""
+          export BUILD_URL_FILE="$GITHUB_WORKSPACE/build-url_${{ inputs.machine }}_${{ inputs.distro_name }}/build-url_${{ inputs.machine }}_${{ inputs.distro_name }}"
+          echo "${BUILD_URL_FILE}"
+          if [ -f "$BUILD_URL_FILE" ]; then
+              export BUILD_URL=$(cat "${BUILD_URL_FILE}")
+          fi
+          if [ -z "${BUILD_URL}" ]; then
+             echo "BUILD_URL missing"
+             exit 0
```
smuppand: This makes the step succeed even if artifact download / build-url generation broke.

Fail fast: exit 1 so the workflow catches infra regressions.
Controlled skip: add an input like allow_missing_build_url (default false). If true, echo "Skipping..." and exit 0 otherwise exit 1.

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,109 @@
...
+        uses: actions/download-artifact@v6
+        with:
+          github-token: ${{ inputs.gh_token }}
+          run-id: ${{ inputs.build_id }}
+          pattern: build-url_*
+
+      - name: Run lava-test-plans
+        shell: bash
+        run: |
+          cd lava-test-plans && pip install .
```
smuppand, line 44: Add actions/setup-python@v5 in the composite action, pinned version (e.g., 3.11/3.12).
Use python -m pip install -U pip then python -m pip install .
This avoids works on some runners, fails on others issues.

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,109 @@
...
+      - name: 'Download build URLs'
+        uses: actions/download-artifact@v6
+        with:
+          github-token: ${{ inputs.gh_token }}
+          run-id: ${{ inputs.build_id }}
+          pattern: build-url_*
+
+      - name: Run lava-test-plans
+        shell: bash
+        run: |
```
smuppand, line 42: The bash script can continue after failures in unexpected places. Recommend adding at the start of the 
run: 
script:
set -euo pipefail
Keep || true only where you intentionally want best-effort behavior.
Should-fix / quality

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,109 @@
...
+          cd lava-test-plans && pip install .
+          lava-test-plans --help
```
smuppand: lava-test-plans --help
This clutters logs. If this is only a sanity check, better
lava-test-plans --version (or) lava-test-plans --help >/dev/null

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,109 @@
...
+          echo "PROJECT_NAME=${GITHUB_REPOSITORY#*/}" >> gh-variables.ini
```
smuppand: You’re creating gh-variables.ini, dragonboard-410c.ini, etc. inside lava-test-plans/ after cd. This is workable, but it mixes checked-out tool repo with generated artifacts.
Recommend writing them to a dedicated output folder, e.g.:
$GITHUB_WORKSPACE/out/vars/gh-variables.ini
$GITHUB_WORKSPACE/out/jobs/... and refer to them via absolute paths.
This also makes the upload step easier and less error-prone.

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,109 @@
...
+        shell: bash
+        run: |
+          echo $GITHUB_WORKSPACE
+          ls -R $GITHUB_WORKSPACE
+
+      - name: 'Upload test jobs'
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ inputs.prefix }}-${{ inputs.machine }}-${{ inputs.distro_name }}
+          path: ${{ env.MACHINE}}*-${{ inputs.distro_name }}*
```
smuppand: This can match unintended files/dirs. Prefer uploading a known folder (e.g., out/jobs/**) created by the step.

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,109 @@
...
+              #lava-test-plans --dry-run --variables gh-variables.ini --test-plan "meta-qcom/${{ inputs.distro_name }}/${{ inputs.testplan }}" --device-type "${{ inputs.machine }}" --dry-run-path ../${{ inputs.machine }}-${{ inputs.distro_name }}-${{ inputs.testplan }} || true
+              lava-test-plans --dry-run --variables gh-variables.ini --test-plan "${GITHUB_REPOSITORY#*/}/${{ inputs.distro_name }}/${{ inputs.testplan }}" --device-type "${{ inputs.machine }}" --dry-run-path ../${{ inputs.machine }}-${{ inputs.distro_name }}-${{ inputs.testplan }} || true
+              echo "MACHINE=${{ inputs.machine }}" >> $GITHUB_ENV
+          fi
+
+      - name: "List files"
+        shell: bash
+        run: |
+          echo $GITHUB_WORKSPACE
+          ls -R $GITHUB_WORKSPACE
```
smuppand: Here as well

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,109 @@
...
+              cat gh-variables.ini
+              #lava-test-plans --dry-run --variables gh-variables.ini --test-plan "meta-qcom/${{ inputs.distro_name }}/${{ inputs.testplan }}" --device-type "${{ inputs.machine }}" --dry-run-path ../${{ inputs.machine }}-${{ inputs.distro_name }}-${{ inputs.testplan }} || true
+              lava-test-plans --dry-run --variables gh-variables.ini --test-plan "${GITHUB_REPOSITORY#*/}/${{ inputs.distro_name }}/${{ inputs.testplan }}" --device-type "${{ inputs.machine }}" --dry-run-path ../${{ inputs.machine }}-${{ inputs.distro_name }}-${{ inputs.testplan }} || true
+              echo "MACHINE=${{ inputs.machine }}" >> $GITHUB_ENV
+          fi
+
+      - name: "List files"
+        shell: bash
+        run: |
+          echo $GITHUB_WORKSPACE
```
smuppand: The "List files" step does ls -R $GITHUB_WORKSPACE — quote it: ls -R "$GITHUB_WORKSPACE"

smuppand (changes requested):

.github/actions/lava-test-plans/action.yml
mwasilew: Because of armv8 build which is used by multiple machines this step is tricky to get right. I don't think it's likely that the pattern is matched by random files. IIUC the end goal is to use a single build for all machines, so this issue will remain.

.github/actions/lava-test-plans/action.yml
mwasilew: This is debug step. I'd like to keep it as it helps in diagnosing failed jobs. We had issues with github API limitations a few times already. Listing files helps debug such problems.

.github/actions/lava-test-plans/action.yml
mwasilew: Good catch. It defaults to 50 in lava-test-plans, so it's not a big issue. But in case we want to change it in the future the variable must be fixed.

.github/actions/lava-test-plans/action.yml
mwasilew: OK. Changed to `exit 1`. I don't remember why I made it this way. Maybe it's just a mistake.

.github/actions/lava-test-plans/action.yml
mwasilew: I think I misread the comment. I'll add the quotes.

.github/actions/lava-test-plans/action.yml
mwasilew: OK, now I remember why I set this to fail silently. When machine list in test workflow doesn't match the list in build workflow, some URLs will be missing. This means there won't be any test jobs for the mismatched machines. I lean toward the idea that such setup should not fail the whole github job. @smuppand @lumag any opinion on that?

smuppand: > OK, now I remember why I set this to fail silently. When machine list in test workflow doesn't match the list in build workflow, some URLs will be missing. This means there won't be any test jobs for the mismatched machines. I lean toward the idea that such setup should not fail the whole github job. @smuppand @lumag any opinion on that?

My opinion
Keep the job green when a single machine is missing a URL, but make it very visible:
Emit a clear warning in the logs (and ideally in the GitHub Actions summary).
Ensure the later “Upload test jobs” step doesn’t fail when nothing was generated (or skip it).

.github/actions/lava-test-plans/action.yml
mwasilew: Fixed.

.github/actions/lava-test-plans/action.yml
mwasilew: Fixed.

.github/actions/lava-test-plans/action.yml
mwasilew: Fixed. Uploading from separate dirs works here even better. Thanks for suggestion.

.github/actions/lava-test-plans/action.yml
mwasilew: Fixed.

.github/actions/lava-test-plans/action.yml
mwasilew: Fixed.

.github/actions/lava-test-plans/action.yml
mwasilew: Fixed.

.github/actions/lava-test-plans/action.yml
mwasilew: Fixed.

.github/actions/lava-test-plans/action.yml
mwasilew: I left `exit 1` in the code. If it proves to be a problem we can revisit.

.github/actions/list-jobs/action.yml
```diff
@@ -0,0 +1,54 @@
...
+      shell: bash
+      run: |
+        echo $GITHUB_WORKSPACE
+        ls -R $GITHUB_WORKSPACE
+
+    - name: "List jobs"
+      id: listjobs
+      shell: bash
+      run: |
+        JOBFILES=$(find . -name "*.yaml" | grep "${{ inputs.prefix }}")
```
smuppand, line 37: This can match unintended paths and breaks on space/newlines. Instead search under the artifacts directories you downloaded and use safer loop (find ... -print0 | while IFS= read -r -d '' ...)

.github/actions/list-jobs/action.yml
```diff
@@ -0,0 +1,54 @@
...
+        RESULT_JSON=$(jq -n '{target: []}')
+        for J in $JOBFILES
+          do
+              echo $J
+              NAME=$(echo "$J" | cut --delimiter "/" --field 3)
```
smuppand, line 42: if the artifact directory structure changes, NAME becomes wrong. Better NAME=$(basename "$(dirname "$J")") or exact from known artifact root.

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,116 @@
...
+          OUT_PATH="${GITHUB_WORKSPACE}/out"
+          VARS_OUT_PATH="${OUT_PATH}/vars"
+          JOBS_OUT_PATH="${OUT_PATH}/jobs"
+          echo "JOBS_OUT_PATH=${JOBS_OUT_PATH}" >> $GITHUB_ENV
+          mkdir -p "${VARS_OUT_PATH}"
+          mkdir -p "${JOBS_OUT_PATH}"
+          echo "PROJECT_NAME=${GITHUB_REPOSITORY#*/}" >> "${VARS_OUT_PATH}/gh-variables.ini"
+          echo "PROJECT=projects/${GITHUB_REPOSITORY#*/}/" >> "${VARS_OUT_PATH}/gh-variables.ini"
+          echo "LAVA_JOB_PRIORITY=50" >> "${VARS_OUT_PATH}/gh-variables.ini"
+          echo "OS_INFO=qcom-distro" >> "${VARS_OUT_PATH}/gh-variables.ini"
```
smuppand, line 55: this is hardcoded. If this workflow runs for other distros (qcom-distro-kvm, qcom-distro-prop-image etc.,)  this is very likely wrong. I suggest OS_INFO=${{ inputs.distro_name }} or map specific values if needed.

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,116 @@
...
+
+          echo "ROOTFS_URL=${BUILD_URL}/${{ inputs.distro_name }}/${{ inputs.machine }}/${IMAGE_TYPE}-${{ inputs.machine }}.rootfs.qcomflash.tar.gz" >> "${VARS_OUT_PATH}/gh-variables.ini"
+
+          if [ "${{ inputs.machine }}" = "qcom-armv8a" ]; then
+              echo "ROOTFS_IMG_FILE=core-image-base-qcom-armv8a.rootfs.ext4" >> "${VARS_OUT_PATH}/gh-variables.ini"
+              cp "${VARS_OUT_PATH}/gh-variables.ini" dragonboard-410c.ini
+              echo "DEVICE_TYPE=dragonboard-410c" >> dragonboard-410c.ini
+              echo "BOOT_IMG_FILE=boot-apq8016-sbc-qcom-armv8a.img" >> dragonboard-410c.ini
+              cat dragonboard-410c.ini
+              lava-test-plans --dry-run --variables dragonboard-410c.ini --test-plan "${GITHUB_REPOSITORY#*/}/${{ inputs.distro_name }}/${{ inputs.testplan }}" --device-type "dragonboard-410c" --dry-run-path "${JOBS_OUT_PATH}/dragonboard-410c-${{ inputs.distro_name }}-${{ inputs.testplan }}" || true
```
smuppand, line 90: this means the action can succeed while producing incomplete/ empty jobs, and later steps will still upload atifcats. If this is meant for gating, then consider Remove || true or add an input like allow_empty=true/flase and fail when no job files were created.

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,116 @@
...
+          echo "FLASHER_DEVICE_TYPE=qcs" >> "${VARS_OUT_PATH}/gh-variables.ini"
+          IMAGE_TYPE="core-image-base"
+          if [ "${{ inputs.distro_name }}" = "qcom-distro" ]; then
+              IMAGE_TYPE="qcom-multimedia-image"
+              echo "AUTO_LOGIN_PASSWORD_PROMPT=Password" >> "${VARS_OUT_PATH}/gh-variables.ini"
+              echo "AUTO_LOGIN_PASSWORD=oelinux123" >> "${VARS_OUT_PATH}/gh-variables.ini"
+          fi
+          echo "IMAGE_FILE_NAME=${IMAGE_TYPE}-${{ inputs.machine }}.rootfs.qcomflash.tar.gz" >> "${VARS_OUT_PATH}/gh-variables.ini"
+          export BUILD_URL=""
+          export BUILD_URL_FILE="$GITHUB_WORKSPACE/build-url_${{ inputs.machine }}_${{ inputs.distro_name }}/build-url_${{ inputs.machine }}_${{ inputs.distro_name }}"
```
smuppand, line 70: this is very specific. If the downloaded artifact name or internal filename changes, it breaks. better consider adding a debug print only when missing and/or a fallback `find "$GITHUB_WORKSPACE" -name "build-url_*" ... to locate the file reliably.

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,116 @@
...
+              export BUILD_URL=$(cat "${BUILD_URL_FILE}")
+          fi
+          if [ -z "${BUILD_URL}" ]; then
+             echo "BUILD_URL missing"
+             # fail action if BUILD_URL is empty. At this point the variable should be filled in
+             exit 1
+          fi
+          echo "BUILD_DOWNLOAD_URL=${BUILD_URL}/" >> "${VARS_OUT_PATH}/gh-variables.ini"
+
+          echo "ROOTFS_URL=${BUILD_URL}/${{ inputs.distro_name }}/${{ inputs.machine }}/${IMAGE_TYPE}-${{ inputs.machine }}.rootfs.qcomflash.tar.gz" >> "${VARS_OUT_PATH}/gh-variables.ini"
```
smuppand, line 82: this might be correct, but if BUILD_URL already points inside a distro/machine subtree, it will duplicate path segments. if possible, document the expected BUILD_URL format or normalize it.

.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,116 @@
...
+              cat "${VARS_OUT_PATH}/gh-variables.ini"
+              lava-test-plans --dry-run --variables "${VARS_OUT_PATH}/gh-variables.ini" --test-plan "${GITHUB_REPOSITORY#*/}/${{ inputs.distro_name }}/${{ inputs.testplan }}" --device-type "${{ inputs.machine }}" --dry-run-path "${JOBS_OUT_PATH}/${{ inputs.machine }}-${{ inputs.distro_name }}-${{ inputs.testplan }}" || true
+              echo "MACHINE=${{ inputs.machine }}" >> $GITHUB_ENV
+          fi
+
+      - name: "List files"
+        shell: bash
+        run: |
+          echo "${GITHUB_WORKSPACE}"
+          ls -R "${GITHUB_WORKSPACE}"
```
smuppand, line 109: This will dump the entire workspace every run, better restrict to out/ and the downloaded build-url artifacts directory only (or guard with an evn like DEBUG=1)

.github/actions/lava-test-plans/action.yml
mwasilew: this is intentional. We don't test all builds that are produced (as we don't have all HW in the lab).

.github/actions/lava-test-plans/action.yml
mwasilew: The file is produced in the build step, that's why it's specific: https://github.com/qualcomm-linux/meta-qcom/blob/master/.github/actions/compile/action.yml#L105

.github/actions/lava-test-plans/action.yml
smuppand: > this is intentional. We don't test all builds that are produced (as we don't have all HW in the lab).

ok, got it.

.github/actions/lava-test-plans/action.yml
mwasilew: Again, this is produced in build step: https://github.com/qualcomm-linux/meta-qcom/blob/master/.github/actions/compile/action.yml#L105. I'm not sure where do you want the documentation to live?

.github/actions/lava-test-plans/action.yml
mwasilew: I prefer to leave it as is. There were instances that I spend days trying to figure out the failure in the workflow without this kind of debug. I think it's useful.

.github/actions/lava-test-plans/action.yml
mwasilew: OK, this can be done. For now the tests run only on qcom-distro. However this needs to match existing file in lava-test-plans: https://github.com/qualcomm-linux/lava-test-plans/blob/cdb8221de2a8f78533ef6b812fc4f5a2461c5525/lava_test_plans/projects/meta-qcom/flasher.jinja2#L11
Otherwise job generation will fail. I would leave it as is for now.

.github/actions/lava-test-plans/action.yml
mwasilew: I'll add a comment to avoid confusion

.github/actions/list-jobs/action.yml
mwasilew: Is it OK to leave it for now and address in the next PR?

.github/actions/list-jobs/action.yml
mwasilew: Is it OK to leave it for now and address in the next PR?

.github/actions/list-jobs/action.yml
smuppand: > Is it OK to leave it for now and address in the next PR?

Yes, since it is not currently blocking, it can resolve with a new PR.

.github/actions/list-jobs/action.yml
smuppand: > Is it OK to leave it for now and address in the next PR?

Yes, since it is not currently blocking, it can resolve with a new PR.

PR #1313, opened 2025-12-18
Disable PCIe ASPM through kernel command line for iq-8275-evk

krishnachaitanya-linux:
PCIe ASPM is not working with the current boards, so disable ASPM with kernel command line.

### DIFF
--- a/conf/machine/iq-8275-evk.conf
+++ b/conf/machine/iq-8275-evk.conf
@@ -8,6 +8,8 @@ MACHINE_FEATURES += "efi pci"
 
 QCOM_DTB_DEFAULT ?= "monaco-evk"
 
+KERNEL_CMDLINE_EXTRA ?= "pcie_aspm=off"
+
 KERNEL_DEVICETREE ?= " \
                       qcom/monaco-evk.dtb \
                       qcom/monaco-evk-camera-imx577.dtb \


krishnachaitanya-linux:
Instead of creating new one, reopened the same PR with new patch.

krishnachaitanya-linux:
> Also beware that this cmdline param disables all ASPM states, not just L1ss where the issue is currently. So there is power penalty.

There is no other way currently to disable L1ss unless a driver change, like I said in the previous reply, we are hoping to close this by next week, then we will come with proper solution, untill then to unblock test teams we can use this.

vkraleti:
@krishnachaitanya-linux IMHO using the kernel command line for this purpose is not appropriate. command line is for user configurations, not for masking underlying issues. The correct approach here is to root cause and fix the issues that are causing WLAN failures.

If disabling PCIe ASPM is indeed the ultimate fix, please apply it through the defconfig rather than via the kernel command line.

ricardosalveti:
> @krishnachaitanya-linux IMHO using the kernel command line for this purpose is not appropriate. command line is for user configurations, not for masking underlying issues. The correct approach here is to root cause and fix the issues that are causing WLAN failures.

+1, we should work to avoid any custom kernel command line here, as we want the kernel to work as it would work on any other generic distribution (e.g. with systemready we are not supposed to add custom kernel command line arguments).

So NAK from my side.

lumag:
Per @Mani-Sadhasivam , this should be handled in the kernel driver. I totally agree with that conclusion. Closing.

PR #1321, opened 2025-12-22
libdiag: Add runtime dependency on diag-router

Jairaj-quic:
Add RDEPENDS on diag-router package as libdiag operates in a client-server architecture where the library connects to the diag-router daemon at runtime.


recipes-test/libdiag/libdiag_1.0.3.bb
```diff
@@ -11,6 +11,7 @@ SRC_URI[sha256sum] = "e6780a404aaa89ffa7f20e659efae4f7b45beaeda8ca1fc05b5a1229ae
 S = "${UNPACKDIR}"
 
 DEPENDS += "glib-2.0"
+RDEPENDS:${PN} += "diag-router"
```
lumag, line 14: Does it work only with diag-router or will it work with the open-source diag too?

Jairaj-quic: Lib won't be functional with open source diag, requires diag-router.

PR #1327, opened 2025-12-23
linux-yocto: upgrade to the latest LTS

lumag:
OE-Core has upgrade linux-yocto recipe to 6.18, the latest LTS, and dropped 6.17, breaking meta-qcom.

Upgrade linux-yocto bbappend to 6.18 and drop 6.16 (EOL) and 6.12 (old LTS).

The major drawback is that 6.12 is promised to be supported until 2036, while 6.18 is only going to be supported until December, 2027.

Note: alsa-ucm-conf changes don't logically belong to the PR, but they are included in order to unbreak the build.


ricardosalveti (approved):

lumag:
@ricardosalveti "build-pr / qcom-armv8a/qcom-distro+linux-yocto-lts" needs to be disabled, it's removed as a part of this PR.

PR #1335, opened 2025-12-29
sensinghub: upgrade 1.0.3 -> 1.0.4

bmasta018:
Upgraded sensinghub from version 1.0.3 to 1.0.4. Updated the header 
documentation of sensinghub to enhance the understanding of APIs.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-support/sensors/sensinghub_1.0.4.bb
+++ b/dynamic-layers/openembedded-layer/recipes-support/sensors/sensinghub_1.0.4.bb
@@ -4,8 +4,7 @@ HOMEPAGE = "https://github.com/qualcomm/sensinghub"
 LICENSE = "BSD-3-Clause-Clear"
 LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=9701d0ef17353f1d05d7b74c8712ebbd"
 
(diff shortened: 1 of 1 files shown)


lumag:
Nit: you can not 'enhance better understanding'.

dynamic-layers/openembedded-layer/recipes-support/sensors/sensinghub_1.0.4.bb
```diff
@@ -4,8 +4,8 @@ HOMEPAGE = "https://github.com/qualcomm/sensinghub"
 LICENSE = "BSD-3-Clause-Clear"
 LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=9701d0ef17353f1d05d7b74c8712ebbd"
 
-SRCREV = "15eaa4651923da584f8d9f3417c3e92d0b61de35"
-PV = "1.0.3"
+SRCREV = "5f07e63c010b4080e2a0bebdbed514eea9145b88"
+PV = "1.0.4"
```
lumag: I didn't notice that before. Please drop manual PV setting. It should not be necessary.

bmasta018: Addressed it in the latest patch.

PR #1337, opened 2025-12-30
firmware-boot: correct the license file and hash values

shoudil:
According to the latest legal guideline, correct the license information used for non-login boot binaries.

### DIFF
--- a/recipes-bsp/firmware-boot/firmware-qcom-boot-qcs615.inc
+++ b/recipes-bsp/firmware-boot/firmware-qcom-boot-qcs615.inc
@@ -1,6 +1,6 @@
 DESCRIPTION = "QCOM NHLOS Firmware for Qualcomm QCS615 platform"
-LICENSE = "Proprietary"
-LIC_FILES_CHKSUM = "file://${UNPACKDIR}/${BOOTBINARIES}/Qualcomm-Technologies-Inc.-Proprietary;md5=58d50a3d36f27f1a1e6089308a49b403"
+LICENSE = "LICENSE.qcom-2"
+LIC_FILES_CHKSUM = "file://${UNPACKDIR}/LICENSE.${BPN};md5=6e1bae7ef13289c332a27b917fb49764"
 
 FW_ARTIFACTORY = "softwarecenter.qualcomm.com/download/software/chip/qualcomm_linux-spf-1-0/qualcomm-linux-spf-1-0_test_device_public"
 FW_BUILD_ID = "r1.0_${PV}/qcs615-le-1-0"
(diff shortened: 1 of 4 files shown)


recipes-bsp/firmware-boot/firmware-qcom-boot-qcs615.inc
```diff
@@ -9,9 +9,11 @@ BOOTBINARIES = "QCS615_bootbinaries"
 
 SRC_URI = " \
     https://${FW_ARTIFACTORY}/${FW_BUILD_ID}/${FW_BIN_PATH}/${BOOTBINARIES}.zip;downloadfilename=${BOOTBINARIES}_r1.0_${PV}.zip;name=bootbinaries \
+    https://${FW_ARTIFACTORY}/${FW_BUILD_ID}/LICENSE.txt;downloadfilename=LICENSE.${BPN};name=license \
```
lumag, line 12: Nope. Please don't download licences separately, they should be moved into the tarballs.

recipes-bsp/firmware-boot/firmware-qcom-boot-qrb2210-rb1_23.12.bb
```diff
@@ -1,6 +1,6 @@
 SUMMARY = "Prebuilt bootlader images for Qualcomm RB1"
 
-LICENSE = "Proprietary"
+LICENSE = "LICENSE.qcom"
```
lumag, line 3: Please submit this change separately.

recipes-bsp/firmware-boot/firmware-qcom-boot-qcs615.inc
shoudil: Currently all the LICENSE files for these QLI targets are present outside of tarballs

shoudil: On QCLI1.0, the initial intention should be re-use the same LICENSE for boot binaries tarball and firmware libs tarball.

lumag: So, it should be pushed in. I'm fine to upgrade the LICENSE variable in the recipes, but also please make sure that platform teams get licences sorted out.

recipes-bsp/firmware-boot/firmware-qcom-boot-qrb2210-rb1_23.12.bb
shoudil: > Please submit this change separately.

Raised separate PR for this https://github.com/qualcomm-linux/meta-qcom/pull/1338

recipes-bsp/firmware-boot/firmware-qcom-boot-qcs615.inc
shoudil: The firmware-qcom-boot-xxx  recipes also take care of CDT tarball, then need to push LICENSE in it as well ?

lumag: Yes.

vkraleti (approved):

PR #1343, opened 2025-12-31
camx-dlkm: Update to the latest revision

gkhose-qipl:
This change includes the following updates:

- Add kernel‑compatible hard IRQ helper. Linux v6.19 removes in_irq() in favor of in_hardirq(), while ≤ v6.18 still provides in_irq(), causing camera‑kernel builds to fail on v6.19.

- Fix potential use‑after‑free on scratch buffers. User‑mode drivers provide scratch buffers for internal use by HW/kernel drivers.

- Fix CPAS DT node parsing with FIT runtime overlays. of_find_node_by_name() searches recursively and can return nodes outside the expected subtree, leading to incorrect references and cleanup issues when overlays are applied at runtime.

- Backports from 1.0 camera kernel component Pull in assorted fixes and cleanups from the 1.0 camera kernel component to keep behavior consistent across camera-kernel versions.


cjha-alt:
> Your commit message missed dots in several places, which turned PR description into the garbled text. Please always end up the sentence with the dot (full stop).

We have updated the commit message to make it more concise and included proper closing dots. Please review it again and let us know if any further updates are required.

lumag (approved):

PR #1344, opened 2025-12-31
sensors: upgrade 1.0.3 -> 1.0.5

bmasta018:
Upgraded sensor binaries from 1.0.3 to 1.0.5. Updated the target-specific runtime files to enable the sensors features on the device.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.5.bb
+++ b/dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.5.bb
@@ -5,9 +5,9 @@ validate sensor services functionality through the Sensinghub Interface."
 LICENSE = "LICENSE.qcom-2"
 LIC_FILES_CHKSUM = "file://LICENSE.qcom-2;md5=f33ba334514c4dfabc6ab7377babb377"
 
-PBT_BUILD_DATE = "251216"
+PBT_BUILD_DATE = "260110"
 SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/sensors.lnx.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qcom-sensors-prebuilts_${PV}_armv8a.tar.gz"
-SRC_URI[sha256sum] = "f0daac75a263a6931211aa1f2eb2a6b52ffef73c784d0b784f9277371a037d40"
+SRC_URI[sha256sum] = "8b8ab3ccfd286682257bb732b899248f49a0526631a257888df7f90e1930c64d"
 
 S = "${UNPACKDIR}"
 
@@ -25,9 +25,9 @@ do_install() {
     install -d ${D}${libdir}
     install -d ${D}${libdir}/pkgconfig
     install -d ${D}${includedir}
-    install -d ${D}/etc/sensors/config
+    install -d ${D}${sysconfdir}/sensors/config
+    install -d ${D}${sysconfdir}/sensors/registry
     install -d ${D}${systemd_system_unitdir}
-    install -d ${D}/etc/sensors/registry/registry
 
     # Install binaries
     install -m 0755 ${S}/usr/bin/* ${D}${bindir}/
@@ -44,7 +44,9 @@ do_install() {
     oe_libinstall -C ${S}/usr/lib -so libSEESalt ${D}${libdir}
 
     # Install registry and Service files
-    install -m 0644 ${S}/etc/sensors/config/* ${D}/etc/sensors/config/
+    install -m 0644 ${S}/etc/sensors/sns_reg_config ${D}${sysconfdir}/sensors/
+    install -m 0644 ${S}/etc/sensors/config/* ${D}${sysconfdir}/sensors/config/
+    install -m 0644 ${S}/etc/sensors/registry/sns_reg_version ${D}${sysconfdir}/sensors/registry/
     install -m 0644 ${S}${systemd_system_unitdir}/sscrpcd.service ${D}${systemd_system_unitdir}/sscrpcd.service
 
     # Install pkgconfig


dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.4.bb
```diff
@@ -54,4 +56,9 @@ do_install() {
     install -m 0644 ${S}/usr/include/*.h ${D}${includedir}/
 }
 
+do_install:append:qcm6490() {
```
lumag: No, this makes the recipe machine-specific. Please avoid that.

dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.4.bb
```diff
@@ -44,7 +44,9 @@ do_install() {
     oe_libinstall -C ${S}/usr/lib -so libSEESalt ${D}${libdir}
 
     # Install registry and Service files
-    install -m 0644 ${S}/etc/sensors/config/* ${D}/etc/sensors/config/
+    install -m 0644 ${S}/etc/sensors/sns_reg_config ${D}/etc/sensors/
+    install -m 0644 ${S}/etc/sensors/config/sns_*.json ${D}/etc/sensors/config/
+    install -m 0644 ${S}/etc/sensors/registry/sns_reg_version ${D}/etc/sensors/registry/
```
lumag: I see that the registry is empty. Is it expected? You can't be generating it at the runtime.

Also, are the config files generic enough? Consider a rootfs that should work on several different kinds of devices (e.g. RB3 Gen2 and Lemans EVK). Is it possible with the single set of sensors configuration files?

bmasta018: The registry is expected to be empty. The input files required for the registry are installed under etc/sensors/config and are parsed at runtime. To support different targets, we maintain two sets of configuration files:

Common files – shared across all supported targets

Target‑specific files – tailored for individual platforms (e.g., RB3 Gen2).

Sensors are not supported on Lemans, so no target‑specific files are provided for that device.

dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.4.bb
```diff
@@ -44,7 +44,9 @@ do_install() {
     oe_libinstall -C ${S}/usr/lib -so libSEESalt ${D}${libdir}
 
     # Install registry and Service files
-    install -m 0644 ${S}/etc/sensors/config/* ${D}/etc/sensors/config/
+    install -m 0644 ${S}/etc/sensors/sns_reg_config ${D}/etc/sensors/
```
anujm1: Please also use `${sysconfdir}` with `${D}` instead of /etc.

dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.4.bb
lumag: Let me rephrase it differently: you can't be installing files, depending on your target. All targets must be able to work with a single, common, read-only rootfs.

dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.4.bb
bmasta018: As mentioned below, we have some machine‑specific files that need to be installed on the target. Could you please guide us on what would be the best approach here?

lumag: Always install them and then at runtime determine if they need to be parsed or not. I think, you apps are doing that anyway, as config files have the `soc_id`, `hw_platform` and `platform_subtype_id` fields.

dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.4.bb
lumag: Also, is registry completely generated in RAM or something being written to `/etc/sensors/registry`?

bmasta018: During device boot-up, the input registry files are read from /etc/sensors/config/, and the parsed output is written to /etc/sensors/registry/.

dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.4.bb
bmasta018: Addressed it in the latest patch

dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.4.bb
bmasta018: All registry files are installed through this recipe, and we will determine on the target whether they should be parsed based on the soc_id.

dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.4.bb
lumag: You can't be writing anything to /etc during bootup. This will break if somebody wants to run with the r/o rootfs. Registry must be compiled in advance, during the package building time.

lumag: Another option would be to compile registry files to `/var`.

Last option, select the registry files which are applicable to each device and ship them via hexagon-dsp-binaries. I'd prefer this option since it would allow sharing those files between this implementation and libssc.

bmasta018: Discussed internally with our team. Below are our conclusions:

- Few years back, all registry config files were placed in a single folder and selected at firmware based on SoC ID. Since the number of files was huge and it delayed sensor firmware initialization, we optimized by segregating registry config files into common and target-specific files, which made parsing much quicker.

- Earlier, registry was parsed under /var/cache/sensors/registry/, but with OSTREE implemented, /var will be empty at build time, causing runtime issues.

- As discussed regarding the hexagon-dsp-binaries approach in another PR, we will consider this input as part of moving our files into public git. This needs evaluation from both firmware and hexagon-dsp-binaries perspectives.

Looking forward to your feedback.

lumag: - Please use /var if registry is generated at runtime and /etc if registry is compiled at the build time.
- We have a clear request from some of the customers to enable rootfs which supports several devices. Thus machine-specific versions of packages are prohibited. You can split necessary files to a machine-specific package, but the packages should be co-installable. In other words, it should be possible to have qcom-armv8a rootfs, which works on _all_ devices we support.

dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.5.bb
```diff
@@ -44,7 +44,9 @@ do_install() {
     oe_libinstall -C ${S}/usr/lib -so libSEESalt ${D}${libdir}
 
     # Install registry and Service files
-    install -m 0644 ${S}/etc/sensors/config/* ${D}/etc/sensors/config/
+    install -m 0644 ${S}/etc/sensors/sns_reg_config ${D}${sysconfdir}/sensors/
+    install -m 0644 ${S}/etc/sensors/config/* ${D}${sysconfdir}/sensors/config/
+    install -m 0644 ${S}/etc/sensors/registry/sns_reg_version ${D}${sysconfdir}/sensors/registry/
```
lumag, line 49: I think, my last question is the registry dir. `/etc` definitely should not be written at runtime

bmasta018: Addressed the comments in the latest patch. We have started using /var for runtime registry parsing.

lumag: Great!

dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.5.bb
```diff
@@ -5,9 +5,9 @@ validate sensor services functionality through the Sensinghub Interface."
 LICENSE = "LICENSE.qcom-2"
 LIC_FILES_CHKSUM = "file://LICENSE.qcom-2;md5=f33ba334514c4dfabc6ab7377babb377"
 
-PBT_BUILD_DATE = "251216"
+PBT_BUILD_DATE = "260110"
 SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/sensors.lnx.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qcom-sensors-prebuilts_${PV}_armv8a.tar.gz"
-SRC_URI[sha256sum] = "f0daac75a263a6931211aa1f2eb2a6b52ffef73c784d0b784f9277371a037d40"
+SRC_URI[sha256sum] = "8b8ab3ccfd286682257bb732b899248f49a0526631a257888df7f90e1930c64d"
```
lumag, line 10: Commit message needs to be updated, it still mentions 1.0.4

bmasta018: Updated the commit message

PR #1347, opened 2026-01-02
Workflows/test result reporting

mwasilew:
Improve test result reporting in both push and PR generated builds.

### DIFF
--- a/.github/actions/list-jobs/action.yml
+++ b/.github/actions/list-jobs/action.yml
@@ -34,12 +34,12 @@ runs:
       id: listjobs
       shell: bash
       run: |
-        JOBFILES=$(find . -name "*.yaml" | grep "${{ inputs.prefix }}")
+        JOBFILES=$(find . -name "${{ input.prefix }}*.yaml")
         RESULT_JSON=$(jq -n '{target: []}')
         for J in $JOBFILES
           do
               echo $J
-              NAME=$(echo "$J" | cut --delimiter "/" --field 3)
+              NAME=$(basename "$J")
               echo $NAME
               F_TMP="${J#*/}"
               RESULT_NAME="${F_TMP//\//-}"


.github/actions/lava-test-plans/action.yml
```diff
@@ -0,0 +1,116 @@
...
+
+          echo "ROOTFS_URL=${BUILD_URL}/${{ inputs.distro_name }}/${{ inputs.machine }}/${IMAGE_TYPE}-${{ inputs.machine }}.rootfs.qcomflash.tar.gz" >> "${VARS_OUT_PATH}/gh-variables.ini"
+
+          if [ "${{ inputs.machine }}" = "qcom-armv8a" ]; then
+              echo "ROOTFS_IMG_FILE=core-image-base-qcom-armv8a.rootfs.ext4" >> "${VARS_OUT_PATH}/gh-variables.ini"
+              cp "${VARS_OUT_PATH}/gh-variables.ini" dragonboard-410c.ini
+              echo "DEVICE_TYPE=dragonboard-410c" >> dragonboard-410c.ini
+              echo "BOOT_IMG_FILE=boot-apq8016-sbc-qcom-armv8a.img" >> dragonboard-410c.ini
+              cat dragonboard-410c.ini
+              lava-test-plans --dry-run --variables dragonboard-410c.ini --test-plan "${GITHUB_REPOSITORY#*/}/${{ inputs.distro_name }}/${{ inputs.testplan }}" --device-type "dragonboard-410c" --dry-run-path "${JOBS_OUT_PATH}/dragonboard-410c-${{ inputs.distro_name }}-${{ inputs.testplan }}" || true
```
smuppand, line 90: the || true on lava-test-plans --dry-run, you intentionally ignore failures. If generation fails, still it will upload something (maybe empty)? and downstream will fail in confusing ways. Better to fail fast if generation fails unless you are explicitly trying to allow missing device templates.

.github/actions/list-jobs/action.yml
```diff
@@ -0,0 +1,54 @@
...
+      shell: bash
+      run: |
+        echo $GITHUB_WORKSPACE
+        ls -R $GITHUB_WORKSPACE
+
+    - name: "List jobs"
+      id: listjobs
+      shell: bash
+      run: |
+        JOBFILES=$(find . -name "*.yaml" | grep "${{ inputs.prefix }}")
```
smuppand: better to use `find . -name "${{ inputs.prefix }} *.yaml"`

.github/actions/list-jobs/action.yml
```diff
@@ -0,0 +1,54 @@
...
+        RESULT_JSON=$(jq -n '{target: []}')
+        for J in $JOBFILES
+          do
+              echo $J
+              NAME=$(echo "$J" | cut --delimiter "/" --field 3)
```
smuppand: it assumes a fixed folder depth, this breaks easily. 
`NAME="$(basename "$(dirname "$J")")"` or simply `NAME="$(basename "$J")"` (whatever you want logically).

.github/actions/test-job-summary/action.yml
```diff
@@ -0,0 +1,137 @@
...
+              JOB_URL="https://lava.infra.foundries.io/results/$JOB_ID"
+              JOB_DETAILS=$(curl -s "https://lava.infra.foundries.io/api/v0.2/jobs/$JOB_ID/")
+              JOB_HEALTH=$(echo "$JOB_DETAILS" | jq -r ".health")
+              JOB_STATE=$(echo "$JOB_DETAILS" | jq -r ".state")
+              JOB_DEVICE_TYPE=$(echo "$JOB_DETAILS" | jq -r ".requested_device_type")
+              echo "| [$JOB_ID]($JOB_URL) | $JOB_DEVICE_TYPE | $JOB_STATE | $JOB_HEALTH |" >> ${{inputs.summary_file_name}}
+          done
+      - name: Upload summary
+        id: upload-summary
+        uses: actions/upload-artifact@v6
```
smuppand, line 131: upload-artifact is v4 and it does not accept github-token/ run-id. v6 supports?

.github/actions/test-job-summary/action.yml
```diff
@@ -0,0 +1,137 @@
...
+inputs:
+  build_id:
+    description: "ID of the workflow from which build URL will be downloaded"
+    required: true
+  gh_token:
+    description: "Github token. It's required to download artifacts from the workflow"
+    required: true
+  prefix:
+    description: "Prefix for the compressed file that is uploaded as workflow artifact"
+    default: test-job*
```
smuppand, line 11: Here if default to prefix includes *, and later `find ... -name "${{ inputs.prefix }}.json"` this becomes test-job*.json.json logic confusion. 

prefix = test-job- and search -name "${prefix}*.json", OR
prefix = test-job-* and search -name "${prefix}.json" (but then prefix already includes wildcard)
Right now it’s inconsistent.

smuppand (changes requested):

.github/actions/lava-test-plans/action.yml
mwasilew: this is the same commit as in #1312. I don't think it's a good idea to change it before it's merged.

.github/actions/list-jobs/action.yml
mwasilew: this is the same commit as in #1312. I don't think it's a good idea to change it before it's merged.

.github/actions/list-jobs/action.yml
mwasilew: this is the same commit as in #1312. I don't think it's a good idea to change it before it's merged.

.github/actions/test-job-summary/action.yml
mwasilew: Could you explain what you mean?
https://github.com/actions/upload-artifact/releases/tag/v6.0.0

.github/actions/test-job-summary/action.yml
mwasilew: I don't understand this comment. The file names are in form `test-job-12345.json`. I'm not sure what is wrong with `test-job*.json` pattern?

.github/actions/lava-test-plans/action.yml
smuppand: > this is the same commit as in #1312. I don't think it's a good idea to change it before it's merged.

OK—if this PR is intentionally identical to #1312 , let’s keep it aligned and we can address the review items in a follow-up PR right after merge

.github/actions/list-jobs/action.yml
smuppand: > this is the same commit as in #1312. I don't think it's a good idea to change it before it's merged.

OK—if this PR is intentionally identical to #1312, let’s keep it aligned and we can address the review items in a follow-up PR right after merge

.github/actions/list-jobs/action.yml
smuppand: > this is the same commit as in #1312. I don't think it's a good idea to change it before it's merged.

OK—if this PR is intentionally identical to #1312, let’s keep it aligned and we can address the review items in a follow-up PR right after merge

.github/actions/test-job-summary/action.yml
smuppand: > Could you explain what you mean? https://github.com/actions/upload-artifact/releases/tag/v6.0.0

Re upload-artifact@v6: my question was about major-version consistency/compatibility. I’m OK with it.

.github/actions/test-job-summary/action.yml
smuppand: > I don't understand this comment. The file names are in form `test-job-12345.json`. I'm not sure what is wrong with `test-job*.json` pattern?

default test-job*: it does work (pattern becomes test-job*.json). My point is it’s broader than needed and could match unrelated test-job*.json files in future, test-job-*.json is a tighter/safe default aligned with the current naming.

.github/actions/lava-test-plans/action.yml
mwasilew: I can add another commit on top. We discussed fixing this in #1312. I'll try that now.

.github/actions/test-job-summary/action.yml
mwasilew: IMHO we should switch to v6 in all workflows. I'll make this change.

.github/actions/test-job-summary/action.yml
mwasilew: OK, got it.

.github/actions/list-jobs/action.yml
mwasilew: Fixed. I also rebased on top of master as #1312 was merged.

.github/actions/list-jobs/action.yml
mwasilew: Fixed.

.github/actions/test-job-summary/action.yml
mwasilew: Fixed. I added commit changing to v6 in all workflows.

.github/actions/test-job-summary/action.yml
mwasilew: Fixed.

.github/actions/lava-test-plans/action.yml
mwasilew: This is a bit more complicated.In general lava-test-plans will only fail if it can't render a template. This means no jobs will be generated in case of failure. However this only works in case when there is a single test job file produced. Test plan can contain more than one test job. In this case a failure may result in some test job files, but not all that were expected. It should not happen as the workflows/tests in lava-test-plans repository renders all possible jobs for all existing machines.

 We have 2 options to address the issue:
 - only run workflow on machines included in lava-test-plans (I prefer this option). In this case we can remove `||true` and assume all matrix workflows will produce something
 - keep `|| true` but restrict all test plans to only a single job (I don't think it's a good idea).

I'll implement the 1st option and we can discuss further.

mwasilew: This is what happens if I include device that doesn't exist: https://github.com/qualcomm-linux/meta-qcom/actions/runs/20713210403/job/59469534291

smuppand: You're right. Remove || true from lava-test-plans --dry-run.
Add pre-check that device template exists.
Add post-check that at least one job file was generated (and non-empty).

ricardosalveti (approved):

lumag (approved):

.github/actions/test-job-summary/action.yml
```diff
@@ -0,0 +1,137 @@
...
+                          fi
+                      else
+                          TEST_NAME="boot"
+                          TEST_URL="${JOB_URL}"
+                          TEST_RESULT_OBJ=$(jq --arg url "$TEST_URL" --arg result "pass" -n -c '$ARGS.named')
+                          jq -n -c --arg key "$TEST_NAME" --argjson value "$TEST_RESULT_OBJ" '$ARGS.named'
+                      fi
+                  done | jq -s -c --sort-keys 'from_entries')
+              fi
+              echo "{\"key\": \"$JOB_DEVICE_TYPE\", \"value\": $TEST_RESULTS}" | jq
```
smuppand, line 71: If the job is not complete, TEST_RESULTS will be empty - you emit invalid JSON ("value": ) - jq fails - step fails -workflow fails.

So IMO initialize TEST_RESULTS='{}'

JOB_ID=$(jq -r '.id' "$TESTJOB")
JOB_DETAILS=$(curl -fsS "https://lava.infra.foundries.io/api/v0.2/jobs/$JOB_ID/")
 
JOB_HEALTH=$(echo "$JOB_DETAILS" | jq -r ".health // empty")
JOB_STATE=$(echo "$JOB_DETAILS" | jq -r ".state // empty")
JOB_DEVICE_TYPE=$(echo "$JOB_DETAILS" | jq -r ".requested_device_type // empty")
 
TEST_RESULTS='{}'
if [ "$JOB_STATE" = "Finished" ] && [ "$JOB_HEALTH" = "Complete" ]; then
   # compute TEST_RESULTS...
fi
 
echo "{\"key\": \"$JOB_DEVICE_TYPE\", \"value\": $TEST_RESULTS}" | jq

.github/actions/test-job-summary/action.yml
```diff
@@ -0,0 +1,137 @@
...
+          done
+          echo "### All jobs summary" >> ${{inputs.summary_file_name}}
+          echo "" >> ${{inputs.summary_file_name}}
+          echo "| Job ID | Device | State | Health |" >> ${{inputs.summary_file_name}}
+          echo "| ---- | ---- | ---- | ---- |" >> ${{inputs.summary_file_name}}
+          for TESTJOB in $(find ${{ github.workspace }} -name "${{ inputs.prefix }}.json")
+          do
+              JOB_ID=$(cat "${TESTJOB}" | jq ".id")
+              JOB_URL="https://lava.infra.foundries.io/results/$JOB_ID"
+              JOB_DETAILS=$(curl -s "https://lava.infra.foundries.io/api/v0.2/jobs/$JOB_ID/")
```
smuppand, line 123: Recommend switching from `curl -s ` to `curl -fsS` so API failures fail clearly (instead of generating empty strings and weird jq errors).

.github/actions/list-jobs/action.yml
```diff
@@ -34,12 +34,12 @@ runs:
       id: listjobs
       shell: bash
       run: |
-        JOBFILES=$(find . -name "*.yaml" | grep "${{ inputs.prefix }}")
+        JOBFILES=$(find . -name "${{ input.prefix }}*.yaml")
```
smuppand, line 37: composite actions the context is inputs, not input. right?

.github/actions/list-jobs/action.yml
```diff
@@ -34,12 +34,12 @@ runs:
...
         RESULT_JSON=$(jq -n '{target: []}')
         for J in $JOBFILES
           do
               echo $J
-              NAME=$(echo "$J" | cut --delimiter "/" --field 3)
+              NAME=$(basename "$J")
```
smuppand, line 42: Right now NAME becomes e.g. test-job-rb3gen2-qcom-distro.yaml. If downstream expects a “device type” or “job name” without extension, this will be noisy.

`NAME=$(basename "$J" .yaml)`

.github/actions/test-job-summary/action.yml
mwasilew: I don't think you're correct. Have you tried this in on a real example?

smuppand: > I don't think you're correct. Have you tried this in on a real example?

I've shared my thoughts based on my understanding, but I haven't tried it myself. Please let me know if my understanding is incorrect.

mwasilew: This is the result you get for incomplete job: https://lava.infra.foundries.io/api/v0.2/jobs/20513/. All the fields are there. There are also some results: https://lava.infra.foundries.io/api/v0.2/jobs/20513/suites/70466/tests/. So unless  there an issue with LAVA server, there will be values in the fields you listed. If LAVA server is malfunctioning, no jobs will run anyway, so the workflow will fail before this step.

smuppand: > This is the result you get for incomplete job: https://lava.infra.foundries.io/api/v0.2/jobs/20513/. All the fields are there. There are also some results: https://lava.infra.foundries.io/api/v0.2/jobs/20513/suites/70466/tests/. So unless there an issue with LAVA server, there will be values in the fields you listed. If LAVA server is malfunctioning, no jobs will run anyway, so the workflow will fail before this step.

Thanks, — agreed that those endpoints do return fields for that specific incomplete job and suites/tests URLs (the examples you shared confirm it).

PR #1349, opened 2026-01-02
gstreamer1.0-plugins-base: Add bbappend to apply Qualcomm patches

Raja-Ganapathi-Busam:
### Summary
This PR adds `bbappend` file to the `gstreamer1.0-plugins-base` recipe to append additional patches from an external series file after the base recipe patches. The patches are imported using `quilt` in the correct order and applied after existing patches.

### Key Changes
- Fetched additional patches and `series` file from external Git repository
- Skipped lines starting with `#` in the series file (treated as comments).
- Imported and applied patches in series file order using quilt import and push

### Temporary Setup
There is a separate pull request raised on the **patches repository** which is pending review and merge.  
Until that PR is merged, `SRC_URI` in this recipe points to **my forked repository** to fetch the patches.

### Merge Condition
**Please merge this PR only after the patches PR is merged** in the patches repository, `gst-qti-oss-patches`,  to ensure consistency.

### Testing
- Verified patches apply in correct order mentioned in `series` file.
- Confirmed that comment lines (`#`) in the series file, if any, are skipped.
- Build completed successfully with updated patches.

### DIFF
--- a/recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-video-Add-support-for-NV12_Q08C-compressed-8-bit-for.patch
+++ b/recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-video-Add-support-for-NV12_Q08C-compressed-8-bit-for.patch
@@ -1,0 +1,179 @@
+From 3f1b805c6104b7e109bd6f0061f3fe14aa5dda0e Mon Sep 17 00:00:00 2001
+From: Pratik Pachange <ppachang@qti.qualcomm.com>
+Date: Fri, 14 Feb 2025 15:05:28 +0530
+Subject: [PATCH] video: Add support for NV12_Q08C (compressed 8-bit) format
+
+Upstream-Status: Denied [https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/9712]
+
+Signed-off-by: Pratik Pachange <ppachang@qti.qualcomm.com>
+---
+ gst-libs/gst/video/video-format.c | 23 ++++++++++++++++++++++
+ gst-libs/gst/video/video-format.h | 10 +++++++++-
+ gst-libs/gst/video/video-info.c   | 32 +++++++++++++++++++++++++++++++
+ gst-libs/gst/video/video-scaler.c |  1 +
+ tests/check/libs/video.c          |  4 ++++
+ 5 files changed, 69 insertions(+), 1 deletion(-)
+
+diff --git a/gst-libs/gst/video/video-format.c b/gst-libs/gst/video/video-format.c
+index 9bd9c69..b28abc3 100644
+--- a/gst-libs/gst/video/video-format.c
++++ b/gst-libs/gst/video/video-format.c
+@@ -1723,6 +1723,24 @@ unpack_NV21 (const GstVideoFormatInfo * info, GstVideoPackFlags flags,
+   }
+ }
+ 
++#define PACK_NV12_Q08C GST_VIDEO_FORMAT_NV12, unpack_NV12_Q08C, 1, pack_NV12_Q08C
++static void
++unpack_NV12_Q08C (const GstVideoFormatInfo * info, GstVideoPackFlags flags,
++    gpointer dest, const gpointer data[GST_VIDEO_MAX_PLANES],
++    const gint stride[GST_VIDEO_MAX_PLANES], gint x, gint y, gint width)
++{
++  GST_FIXME ("Unpack NV12_Q08C is not yet supported.");
++}
++
++static void
++pack_NV12_Q08C (const GstVideoFormatInfo * info, GstVideoPackFlags flags,
++    const gpointer src, gint sstride, gpointer data[GST_VIDEO_MAX_PLANES],
++    const gint stride[GST_VIDEO_MAX_PLANES], GstVideoChromaSite chroma_site,
++    gint y, gint width)
++{
++  GST_FIXME ("Pack NV12_Q08C is not yet supported.");
++}
++
+ #define PACK_AV12 GST_VIDEO_FORMAT_AYUV, unpack_AV12, 1, pack_AV12
+ static void
+ unpack_AV12 (const GstVideoFormatInfo * info, GstVideoPackFlags flags,
+@@ -8108,6 +8126,9 @@ static const VideoFormat formats[] = {
+       PSTR8888, PLANE0, OFFS0, SUB4444, PACK_Y416_BE),
+   MAKE_GRAY_LE_FORMAT (GRAY10_LE16, "raw video", DPTH10, PSTR2, PLANE0, OFFS0,
+       SUB4, PACK_GRAY10_LE16),
++  MAKE_YUV_FORMAT (NV12_Q08C, "raw video",
++      GST_MAKE_FOURCC ('Q', '0', '8', 'C'), DPTH888, PSTR122, PLANE011,
++      OFFS001, SUB420, PACK_NV12_Q08C),
+ };
+ 
+ G_GNUC_END_IGNORE_DEPRECATIONS;
+@@ -8368,6 +8389,8 @@ gst_video_format_from_fourcc (guint32 fourcc)
+       return GST_VIDEO_FORMAT_VUYA;
+     case GST_MAKE_FOURCC ('A', 'R', '3', '0'):
+       return GST_VIDEO_FORMAT_BGR10A2_LE;
++    case GST_MAKE_FOURCC ('Q', '0', '8', 'C'):
++      return GST_VIDEO_FORMAT_NV12_Q08C;
+ 
+     default:
+       return GST_VIDEO_FORMAT_UNKNOWN;
+diff --git a/gst-libs/gst/video/video-format.h b/gst-libs/gst/video/video-format.h
+index b59ec38..6593f8a 100644
+--- a/gst-libs/gst/video/video-format.h
++++ b/gst-libs/gst/video/video-format.h
+@@ -174,6 +174,7 @@ G_BEGIN_DECLS
+  * @GST_VIDEO_FORMAT_Y416_LE: packed 4:4:4:4 YUV, 16 bits per channel(U-Y-V-A) (Since: 1.26)
+  * @GST_VIDEO_FORMAT_Y416_BE: packed 4:4:4:4 YUV, 16 bits per channel(U-Y-V-A) (Since: 1.26)
+  * @GST_VIDEO_FORMAT_GRAY10_LE16: 10-bit grayscale, packed into 16bit words (6 bits padding) (Since: 1.26)
++ * @GST_VIDEO_FORMAT_NV12_Q08C: Qualcomm NV12 8bit compressed format
+  *
+  * Enum value describing the most common video formats.
+  *
+@@ -670,6 +671,13 @@ typedef enum {
+    */
+   GST_VIDEO_FORMAT_GRAY10_LE16,
+ 
++  /**
++   * GST_VIDEO_FORMAT_NV12_Q08C:
++   *
++   * Qualcomm NV12 8-bit compressed format.
++   */
++  GST_VIDEO_FORMAT_NV12_Q08C,
++
+   /* Update GST_VIDEO_FORMAT_LAST below when adding more formats here */
+ } GstVideoFormat;
+ 
+@@ -680,7 +688,7 @@ typedef enum {
+  *
+  * Since: 1.26
+  */
+-#define GST_VIDEO_FORMAT_LAST (GST_VIDEO_FORMAT_GRAY10_LE16 + 1)
++#define GST_VIDEO_FORMAT_LAST (GST_VIDEO_FORMAT_NV12_Q08C + 1)
+ 
+ #define GST_VIDEO_MAX_PLANES 4
+ #define GST_VIDEO_MAX_COMPONENTS 4
+diff --git a/gst-libs/gst/video/video-info.c b/gst-libs/gst/video/video-info.c
+index 04d52b0..f86b9b3 100644
+--- a/gst-libs/gst/video/video-info.c
++++ b/gst-libs/gst/video/video-info.c
+@@ -1389,6 +1389,38 @@ fill_planes (GstVideoInfo * info, gsize plane_size[GST_VIDEO_MAX_PLANES])
+       info->size = info->offset[1] + info->offset[1] / 2;
+       break;
+     }
++    case GST_VIDEO_FORMAT_NV12_Q08C:
++    {
++      gint y_plane_stride = 0, y_meta_stride = 0, y_meta_scanline = 0;
++      gint uv_plane_stride = 0, uv_meta_stride = 0, uv_meta_scanline = 0;
++      gint y_plane_scanline = 0, y_plane_size = 0, y_meta_size = 0;
++      gint uv_plane_scanline = 0, uv_plane_size = 0, uv_meta_size = 0;
++
++      y_plane_stride = GST_ROUND_UP_128 (width);
++      y_plane_scanline = GST_ROUND_UP_32 (height);
++      uv_plane_stride = GST_ROUND_UP_128 (width);
++      uv_plane_scanline = GST_ROUND_UP_32 (GST_ROUND_UP_2 (height) / 2);
++
++      y_meta_stride = GST_ROUND_UP_64 (GST_ROUND_UP_32 (width) / 32);
++      y_meta_scanline = GST_ROUND_UP_16 (GST_ROUND_UP_8 (height) / 8);
++      uv_meta_stride = GST_ROUND_UP_64 (
++          GST_ROUND_UP_16 (GST_ROUND_UP_2 (width) / 2) / 16);
++      uv_meta_scanline = GST_ROUND_UP_16 (
++          GST_ROUND_UP_8 (GST_ROUND_UP_2 (height) / 2) / 8);
++
++      y_plane_size = GST_ROUND_UP_N (y_plane_stride * y_plane_scanline, 4096);
++      uv_plane_size = GST_ROUND_UP_N (uv_plane_stride * uv_plane_scanline, 4096);
++
++      y_meta_size = GST_ROUND_UP_N (y_meta_stride * y_meta_scanline, 4096);
++      uv_meta_size = GST_ROUND_UP_N (uv_meta_stride * uv_meta_scanline, 4096);
++
++      info->stride[0] = y_plane_stride;
++      info->stride[1] = uv_plane_stride;
++      info->offset[0] = 0;
++      info->offset[1] = y_plane_size + y_meta_size;
++      info->size = info->offset[1] + uv_plane_size + uv_meta_size;
++      break;
++    }
+     case GST_VIDEO_FORMAT_ENCODED:
+     case GST_VIDEO_FORMAT_DMA_DRM:
+       break;
+diff --git a/gst-libs/gst/video/video-scaler.c b/gst-libs/gst/video/video-scaler.c
+index d645087..0fa9e3d 100644
+--- a/gst-libs/gst/video/video-scaler.c
++++ b/gst-libs/gst/video/video-scaler.c
+@@ -1262,6 +1262,7 @@ get_functions (GstVideoScaler * hscale, GstVideoScaler * vscale,
+     case GST_VIDEO_FORMAT_NV21:
+     case GST_VIDEO_FORMAT_NV24:
+     case GST_VIDEO_FORMAT_NV61:
++    case GST_VIDEO_FORMAT_NV12_Q08C:
+       *bits = 8;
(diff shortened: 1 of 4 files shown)


CherukuriNaveenQC:
@lumag As canonical suggesting hosting in separate repo instead of putting patches along with distribution. Our team had initial discussion with Nico and explained the reason to host in Github

@ndechesne could you please comment your feedback here.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
```diff
@@ -0,0 +1,37 @@
+# To apply patches to support Qualcomm specific formats and fixes
+SRC_URI:append:qcom = " \
+    git://github.com/Raja-Ganapathi-Busam/gst-qti-oss-patches.git;protocol=ssh;user=git;branch=${PV};name=qti_patches;destsuffix=patchrepo;subpath=${BPN} \
+"
+
+SRCREV_qti_patches = "${AUTOREV}"
+
+qcom_do_patch() {
+    PATCHREPO="${WORKDIR}/sources/patchrepo"
```
sbanerjeeqc: Start as do_patch_<purpose>
Can you name this more appropriately in the purpose?

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
```diff
@@ -0,0 +1,37 @@
...
+    if [ -f "${PATCHREPO}/series" ]; then
+        while read p; do
+            bbnote "Iterating ${PATCHREPO}/${p}"
+            # Skip empty/comment lines
+            case "$p" in
+                ""|\#*) continue ;;
+            esac
+
+            if [ -f "${PATCHREPO}/${p}" ]; then
+                quilt import "${PATCHREPO}/${p}"
```
sbanerjeeqc: you need to depend on quit-native

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
```diff
@@ -0,0 +1,37 @@
+# To apply patches to support Qualcomm specific formats and fixes
+SRC_URI:append:qcom = " \
+    git://github.com/Raja-Ganapathi-Busam/gst-qti-oss-patches.git;protocol=ssh;user=git;branch=${PV};name=qti_patches;destsuffix=patchrepo;subpath=${BPN} \
+"
+
+SRCREV_qti_patches = "${AUTOREV}"
```
ndechesne: no AUTOREV, we only rely on fixed content/metadata in meta-qcom.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
```diff
@@ -0,0 +1,37 @@
+# To apply patches to support Qualcomm specific formats and fixes
+SRC_URI:append:qcom = " \
+    git://github.com/Raja-Ganapathi-Busam/gst-qti-oss-patches.git;protocol=ssh;user=git;branch=${PV};name=qti_patches;destsuffix=patchrepo;subpath=${BPN} \
```
ndechesne: if the patches are available in a public http link, perhaps we could list them here directly. that might be a better approach that strictly depending on an external git tree.

a link like this one (this is just an example)

https://raw.githubusercontent.com/qualcomm-linux/meta-qcom/e74ef28e7df313edf77e6aafe2f789ef324b09d5/conf/layer.conf

At least you wouldn't need to have a custom do_patch implementation.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
Raja-Ganapathi-Busam: In my understanding, Yocto uses **quilt as the default PATCHTOOL**.
Since this function runs inside do_patch:append, the environment already ensures quilt-native is available.
Therefore, we **do not need to explicitly add DEPENDS += "quilt-native"** in the .bbappend file.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
Raja-Ganapathi-Busam: Our intention is to avoid making changes in meta-qcom repo if there is any update in patches repo.
Please suggest.

ndechesne: there is no other way around. You need to make changes to meta-qcom when you change the version of the software we build in meta-qcom.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
CherukuriNaveenQC: We are in process of making git://github.com/qualcomm/gst-qti-oss-patches public. Meanwhile, we would like to have a review of our approach

Raja-Ganapathi-Busam: We are in the process of making the patches repository public.
Once it becomes publicly accessible, I will update the link accordingly.

ndechesne: I understand, and I made the review in this context. I am not saying that the link is not public, but suggesting a different approach. Instead of downloading the whole tree and manually applying the patches yourself in the qcom_do_patch() function, you can list all the patches as http:// links SRC_URI.

That way, you still preserver your tree with standalone patches which is convenient for what we discussed, and we still integrate patches 'normally' (almost) from an OE metadata perspective

koenkooi: > I understand, and I made the review in this context. I am not saying that the link is not public, but suggesting a different approach. Instead of downloading the whole tree and manually applying the patches yourself in the qcom_do_patch() function, you can list all the patches as http:// links SRC_URI.
> 
> That way, you still preserver your tree with standalone patches which is convenient for what we discussed, and we still integrate patches 'normally' (almost) from an OE metadata perspective

I think that is also what as suggested by me in the meetings before the solstice break. Having a custom do_patch is very fragile and rarely required.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
lumag: Of course you need to do it. If something is changed in OE-Core then your recipe gets broken.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
lumag: You can still point out the raw URIs inside your repo (and change them later to the official one once it's public).

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
lumag: +1 here. If you make changes to your repo, it is a _different_ version, a different recipe revision and a different set of packages. The item will go away once you drop the custom do_patch implementation and use OE's one (together with using raw URIs)

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
```diff
@@ -0,0 +1,37 @@
...
+                bbfatal "Patch '${p}' listed in series not found at ${PATCHREPO}/${p}"
+            fi
+        done < "${PATCHREPO}/series"
+    else
+        bbnote "No series file found at ${PATCHREPO}/series"
+    fi
+}
+
+python do_patch:append:qcom() {
+    bb.build.exec_func('qcom_do_patch', d)
```
lumag: Any reason for not extending the function directly?

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
Raja-Ganapathi-Busam: Removed dependency on quilt and added the raw URIs directly in the SRC_URI.
Comment not applicable now.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
Raja-Ganapathi-Busam: As suggested, removed dependency on quilt and added the raw URIs directly in the SRC_URI.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
Raja-Ganapathi-Busam: Removed dependency on quilt and added the raw URIs directly in the SRC_URI.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
Raja-Ganapathi-Busam: Removed dependency on quilt and added the raw URIs directly in the SRC_URI.
Comment not applicable now.

anujm1:
> Okay.
> 
> > @lumag As canonical suggesting hosting in separate repo instead of putting patches along with distribution and we would like to maintain patches in one central repo for all distributions and our team had initial discussion with Nico and explained the reason to host in Github
> 
> This will not work in a long term. Ubuntu, Debian, Fedora and OE will have different versions of the GStreamer. At some point your patches won't be unuquelly applicable to those versions, so you will have to have separate branches or revisions. Having that in mind it is much easier to incorporate those patches into the _distro_ package as it is usually done. I definitely don't expect Debian / Ubuntu packages fetching additional repo during the build stage.
> 

In addition to this, since we now have meta-qcom running on YP autobuilder, every gstreamer version update will break master-next build.

CherukuriNaveenQC:
When we branch off QLI 2.x then both Yocto and Ubuntu uses same version of GStreamer in my opinion. So, instead of maintaining patches in multiple places, we would like to maintain in one place. Also, I got to know that Canonical does not want attach patches in distribution and instead they asking us to host patches in a repo.

ndechesne:
> When we branch off QLI 2.x then both Yocto and Ubuntu uses same version of GStreamer in my opinion. So, instead of maintaining patches in multiple places, we would like to maintain in one place. Also, I got to know that Canonical does not want attach patches in distribution and instead they asking us to host patches in a repo.

There is no guarantee that Yocto and Ubuntu will use the same version of Gstreamer. It might happen, but maybe not.
What you say about Canonical (or Ubuntu) is also not right, since Ubuntu is hosting patches into its distribution (on launchpad). What they are asking is a convenient place to find our patches, but they will duplicate them and make a copy into Ubuntu.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
```diff
@@ -0,0 +1,15 @@
...
+    ${PATCH_BASE}/0001-video-Add-support-for-NV12_Q08C-compressed-8-bit-for.patch;name=q08c_patch \
+    ${PATCH_BASE}/0002-video-Add-support-for-NV12_Q10LE32C-compressed-10-bi.patch;name=q10c_patch \
+    ${PATCH_BASE}/0003-videometa-Update-the-aggregation-logic-for-stride-al.patch;name=vmeta_patch \
+    ${PATCH_BASE}/0004-videorate-Add-option-to-set-passthrough-mode.patch;name=vrate_patch \
+"
+
+SRC_URI[q08c_patch.sha256sum] = "7455df8252fc1655184a4659bdc8ce1479595ed3bd788578c3e909582af04803"
+SRC_URI[q10c_patch.sha256sum] = "2d7f448abc9da7843959b193861d1266a11a71365908900d7b4a31d8e20c19e6"
+SRC_URI[vmeta_patch.sha256sum] = "76ba2ae3e40d1cbb016561ad7fb8027cacd120f67a6fde4c2a8d992be3ae27ba"
+SRC_URI[vrate_patch.sha256sum] = "393cdd2ede68080db6e6f969373358d66dc92e6025cddc4307595aca650d6ef5"
```
lumag: This doesn't play well with the Yocto Compatibility requirements. While we probably can come up with a different scheme, I would ask to vendor in all patches here rather than pulling them from the external repo.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
```diff
@@ -0,0 +1,15 @@
+# To apply patches to support Qualcomm specific formats and fixes
+
+PATCH_BASE = "https://raw.githubusercontent.com/Raja-Ganapathi-Busam/gst-qti-oss-patches/refs/heads/${PV}/${BPN}"
```
ndechesne: can you open the tree so that we can have a look? at least share with the meta-qcom maintainers?

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
```diff
@@ -0,0 +1,15 @@
...
+
+SRC_URI:append:qcom = " \
```
ndechesne, line 5: I understand it is more work on your side to manage the patches, but it really look much better than the first patch, thanks!

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
lumag: This still allows updating the repo in a way that the patches will be changed, breaking the builds. The names of the patches don't incorporate version of the recipe (nor the revision of the patch), so when the patch is updated in the repo, it will not be redownloaded, triggering the checksum errors for the builders.

ndechesne: this is always a problem. if 'upstream' changes the tarball we use, it breaks the build.. here if we do crazy things on gst-qti-oss-patches and rewrite commits, then yes, it's bad and it will break things. i don't see how it's different here. we have checksums for the files, so we check for such issues.

lumag: There is a difference between checking for upstream misbehaviour and knowingly introducing a change that will report an error once the repo gets updated.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
lumag: You can observe the issue as a failure of the yocto-run-checks step.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
CherukuriNaveenQC: @ndechesne we have shared patches via email as patches repo still not public due to legal review

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0004-videorate-Add-option-to-set-passthrough-mode.patch
```diff
@@ -0,0 +1,101 @@
+From 93d38aa1d311dfccfb5bb1169e7e6192f8db1fee Mon Sep 17 00:00:00 2001
+From: Kemal Rasim Sh <kshakir@qti.qualcomm.com>
+Date: Tue, 21 Oct 2025 00:43:10 -0700
+Subject: [PATCH 6/6] videorate: Add option to set passthrough mode
+
+Add option to set passthrough mode in order to prevent videorate to copy
+the buffer when the buffer is not writable.
+
+Upstream-Status: Pending
```
ricardosalveti: Did you try to upstream this patch already? It would be nice to have it at least submitted upstream for review before we include it here.

ricardosalveti:
Please add more details in your git commit message explaining why these are "qualcomm patches", why they are relevant, why they are not going to be proposed upstream, etc.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0004-videorate-Add-option-to-set-passthrough-mode.patch
Raja-Ganapathi-Busam: We started a discussion in upstream.
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/merge_requests/10513

lumag: And it seems it got rejected.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-video-Add-support-for-NV12_Q08C-compressed-8-bit-for.patch
```diff
@@ -0,0 +1,179 @@
+From 3f1b805c6104b7e109bd6f0061f3fe14aa5dda0e Mon Sep 17 00:00:00 2001
+From: Pratik Pachange <ppachang@qti.qualcomm.com>
+Date: Fri, 14 Feb 2025 15:05:28 +0530
+Subject: [PATCH] video: Add support for NV12_Q08C (compressed 8-bit) format
+
+Upstream-Status: Inappropriate [Qualcomm specific]
```
lumag: No, it's not Inappropriate. It is `Denied`

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0004-videorate-Add-option-to-set-passthrough-mode.patch
Raja-Ganapathi-Busam: We are still in discussion with community and till we get conclusion we will remove this patch. We will create new PR based on conclusion.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0001-video-Add-support-for-NV12_Q08C-compressed-8-bit-for.patch
Raja-Ganapathi-Busam: Updated.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0004-videorate-Add-option-to-set-passthrough-mode.patch
Raja-Ganapathi-Busam: This patch is removed from this PR.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
```diff
@@ -0,0 +1,10 @@
+# To apply patches to support Qualcomm specific formats and fixes
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
+
+SRC_URI:append:qcom = " \
+    file://0001-video-Add-support-for-NV12_Q08C-compressed-8-bit-for.patch \
+    file://0002-video-Add-support-for-NV12_Q10LE32C-compressed-10-bi.patch \
+    file://0003-videometa-Update-the-aggregation-logic-for-stride-al.patch \
+"
+
```
lumag: Extra empty line

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
```diff
@@ -0,0 +1,10 @@
+# To apply patches to support Qualcomm specific formats and fixes
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
```
lumag: `:qcom`

Raja-Ganapathi-Busam: Done

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base_%.bbappend
Raja-Ganapathi-Busam: Removed

lumag (approved):

recipes-multimedia/gstreamer/gstreamer1.0-plugins-base/0004-videorate-Add-option-to-set-passthrough-mode.patch
ricardosalveti: Next time please add the link to the pr sent upstream as well, helps reviewing the pr, and state would be 'Submitted'.

PR #1353, opened 2026-01-05
ci: Optimizing costs for AWS self-hosted runners

sampra2025:
With this update, the AWS self-hosted runner instance type will change from "c5ad.8xlarge" to "m8a.4xlarge". This adjustment will improve build performance by at least 20% and lower the hourly cost by 30%.

The runner instance has been tested in the Stage environment. For reference, here is a sample build workflow:
https://github.com/qualcomm-linux-stg/aws-meta-qcom/actions/runs/20283368465

image

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -15,7 +15,7 @@ env:
 jobs:
   kas-setup:
     if: github.repository == 'qualcomm-linux/meta-qcom'
-    runs-on: [self-hosted, qcom-u2404, amd64-ssd]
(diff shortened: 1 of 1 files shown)


sampra2025:
> > Thank you for your feedback. With the recent build, we are seeing similar performance. I agree that running a clean build is a good idea. Please let me know the best approach for running a clean build.
> 
> In your staging environment you can just erase the sstate cache entirely before doing a new build with both worker types, then can easily compare.

Great.. I will do it in stage environment and share the result with you.

PR #1354, opened 2026-01-06
ci: qcom-distro-prop-image: add container orchestration image

gyenugul:
Include qcom-container-orchestration-image to target list.

Purpose:
Enable validation of container orchestration and related technologies against the qcom-next kernel and ensure the Qualcomm reference distro includes the packages needed to support Kubernetes and container runtimes out of the box.

Related: https://github.com/qualcomm-linux/meta-qcom-distro/pull/127

### DIFF
--- a/ci/qcom-distro-prop-image.yml
+++ b/ci/qcom-distro-prop-image.yml
@@ -8,3 +8,4 @@ header:
 target:
   - qcom-multimedia-image
   - qcom-multimedia-proprietary-image
+  - qcom-container-orchestration-image


ci/qcom-distro-container-orchestration.yml
```diff
@@ -0,0 +1,9 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+  includes:
+   - ci/qcom-distro.yml
+
+distro: qcom-distro-container-orchestration
```
lumag: Why do we need a special distro here?

lumag:
PR message is invalid, the mentioned PR is not related at all.

vkraleti:
> Add qcom-distro-container-orchestration.yml:
> 
> * includes ci/qcom-distro.yml to reuse distro defaults
> * sets target=qcom-container-orchestration-image
> 
> Purpose: Enable validation of container orchestration and related technologies against the qcom-next kernel and ensure the Qualcomm reference distro includes the packages needed to support Kubernetes and container runtimes out of the box.
> 
> Related: #127

@gyenugul are you referring https://github.com/qualcomm-linux/meta-qcom-distro/pull/127 here? #[number] links to numbered PR in current repo. So what you mention is not right.

ci/qcom-distro-container-orchestration.yml
```diff
@@ -3,6 +3,7 @@
 header:
   version: 14
   includes:
-  - ci/base.yml
+   - ci/qcom-distro.yml
 
-machine: iq-x7181-evk
+distro: qcom-distro-container-orchestration
```
vkraleti: https://github.com/qualcomm-linux/meta-qcom-distro/pull/127/changes/fbde27e48997407b03488259002abefd3db60bf6 silently introduced this DISTRO without mentioning anything in commit.  During review I completely missed this file. :(

ci/qcom-distro-container-orchestration.yml
gyenugul: The image qcom-container-orchestration-image requires virtualization in REQUIRED_DISTRO_FEATURES. Since the existing qcom-distro does not include virtualization by default, I took qcom-distro-kvm as reference and introduced a new distro configuration.

vkraleti: Before introducing, is it really deliberated and agreed by all stake holders that a new DISTRO is really required? What is stopping from adding 'virtualization' to qcom-distro DISTRO?

gyenugul: For qcom-distro-kvm, virtualization was added in a separate distro to keep KVM-specific features isolated. I followed the same approach here to ensure orchestration-specific features remain separate and do not mix into the base qcom-distro, which is used for multiple builds.

vkraleti: There is a need to make builds with and without KVM. Hence a new DISTRO is introduced. That may not be the case with 'virtualization' DISTRO_FEATURE. It can coexist with other features. Isn't it?

vkraleti (changes requested):

ci/qcom-distro-container-orchestration.yml
gyenugul: I agree with your point. Since virtualization can coexist with other features, there’s no need for a separate distro. I’ve updated the commit by removing the new distro inclusion. Will wait for https://github.com/qualcomm-linux/meta-qcom-distro/pull/131 to merge and then resume this change

vkraleti:
@gyenugul With #1106 there  is no need of another config file now. Please rebase on tip and add `qcom-container-orchestration-image` as a target in [qcom-distro-prop-image.yml](https://github.com/qualcomm-linux/meta-qcom/blob/master/ci/qcom-distro-prop-image.yml)

vkraleti:
@gyenugul with #1106 there is no need of a additional ci yml file.  Please rebase on tip and add `qcom-container-orchestration-image` as a target in [qcom-distro-prop-image.yml](https://github.com/qualcomm-linux/meta-qcom/blob/master/ci/qcom-distro-prop-image.yml)

ricardosalveti (approved):

PR #1358, opened 2026-01-07
Properly compress Adreno firmware

lumag:
Due to the way it is handled, Adreno firmware ends up being uncompressed in the generated packages. Make sure that it gets compressed.


koenkooi (approved):
LGTM.

PR #1359, opened 2026-01-07
android-tools-conf-configfs: Additional adb-serial sources

loicpoulain:
If the androidboot.serialno parameter is not specified in the kernel command line, we should first fall back to the device’s serial attributes. These can be obtained from the generic SMBIOS/DMI product_serial ID, or alternatively from the platform’s SoC serial attribute.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf-configfs/qcom/android-gadget-setup.machine
+++ b/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf-configfs/qcom/android-gadget-setup.machine
@@ -1,6 +1,7 @@
 manufacturer=Qualcomm
 model=`hostname`
 androidserial="$(sed -n -e '/androidboot.serialno/  s/.*androidboot.serialno=\([^ ]*\).*/\1/gp ' /proc/cmdline)"
+[ -z "$androidserial" ] && [ -e /sys/class/dmi/id/product_serial ] && androidserial=$(cat /sys/class/dmi/id/product_serial)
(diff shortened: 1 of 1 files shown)


dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf-configfs/qcom/android-gadget-setup.machine
```diff
@@ -1,6 +1,8 @@
 manufacturer=Qualcomm
 model=`hostname`
 androidserial="$(sed -n -e '/androidboot.serialno/  s/.*androidboot.serialno=\([^ ]*\).*/\1/gp ' /proc/cmdline)"
+[ -z "$androidserial" ] && [ -e /sys/class/dmi/id/product_serial ] && androidserial=$(cat /sys/class/dmi/id/product_serial)
```
lumag, line 4: No, the sda / emmc probing matches the way ABL / UEFI calculates serial.

loicpoulain: It appears that the serial from DMI also matches this value, at least in the Agatti/U-Boot case. This seems to provide a more generic and reliable approach (when available), rather than depending on a specific block whose index or path cannot be guaranteed.

However, this does not apply to the `/sys/devices/soc0/serial_number ` value below. Would you recommend removing it and preferring the default 0123... serial instead? or should I simply assign it the lowest priority?.

lumag: I'd recommend dropping SoC serial number, yes.

ricardosalveti:
Please fix your git commit body message (single long line), and your email as well (should probably be the oss one).

PR #1363, opened 2026-01-08
workflows/pr.yml: skip on documentation changes

anujm1:
Documentation changes do not need full builds to be run. Skip to save resources.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -2,6 +2,10 @@ name: Build on PR
 
 on:
   pull_request:
+    paths-ignore:
+      - 'README.md'
+      - 'README'
+      - 'SECURITY.md'
 
 permissions:
   checks: write


lumag (dismissed):

.github/workflows/pr.yml
```diff
@@ -2,6 +2,11 @@ name: Build on PR
 
 on:
   pull_request:
+    paths-ignore:
+      - 'README.md'
+      - 'README'
+      - 'SECURITY.md'
+      - 'licenses/**'
```
vkraleti: Although the possibility is very low, what would happen if an existing license removed without updating the recipes that reference it? Aren't we supposed to build? Can a Linter check catch this case, instead of build?

anujm1: > Although the possibility is very low, what would happen if an existing license removed without updating the recipes that reference it? Aren't we supposed to build? Can a Linter check catch this case, instead of build?

Yes, theoretically possible. If a commit only removes a license from `licenses/` without a change in recipe, a full build cycle would still be an overkill I think. It should not be ignored either so I will remove it from here for now.

PR #1366, opened 2026-01-09
minkipc: add REDEPENDS for minkipc-qteesupplicant

jiaxli-QC:
Ensure the minkipc-qteesupplicant package declares a runtime dependency on minkipc.

The minkipc-qteesupplicant component requires the minkipc related libraries at runtime. Without an explicit runtime dependency the minkipc package may not be pulled into images, which can lead to runtime failures for qtee_supplicant. Add a RDEPENDS entry to the minkipc_git.bb recipe so package managers and image assembly pull in minkipc automatically.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_git.bb
+++ b/dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_git.bb
@@ -33,6 +33,8 @@ FILES:${PN}-qteesupplicant = "${bindir}/qtee_supplicant \
                               ${nonarch_libdir}/udev/rules.d/99-qcomtee-udev.rules \
 "
 
+RDEPENDS:${PN}-qteesupplicant = "${PN}"
+
 # Currently, this recipe only builds and installs for ARMv8 (aarch64) machine.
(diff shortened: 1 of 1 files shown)


dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_git.bb
```diff
@@ -33,6 +33,7 @@ FILES:${PN}-qteesupplicant = "${bindir}/qtee_supplicant \
                               ${nonarch_libdir}/udev/rules.d/99-qcomtee-udev.rules \
 "
 
+RDEPENDS:${PN}-qteesupplicant = " ${PN}"
```
harshaldev27: Can we please avoid unnecessary spaces? `"${PN}"`

jiaxli-QC: In the .bb file, different targets are separated by spaces. 
The advantage of this approach is that it prevents missing spaces during variable concatenation at compile time.

jiaxli-QC: Updated

Issue #1071, opened 2025-09-12
Need inputs on adding   SELinux  support  in meta-qcom


rsiddoji:
**Proposal: SELinux Integration with Optional Enablement**

Objective
To integrate SELinux (Security-Enhanced Linux) support into the project tree, with the feature disabled by default. Users can opt-in to enable SELinux through distribution-level configuration. This approach ensures flexibility while laying the groundwork for enhanced security capabilities.
Goals and Expectations
When SELinux is enabled:
•	 The system should compile successfully with SELinux in enforcing mode.
•	 The system must boot to the console with BusyBox and essential shell utilities functioning as expected.
•	 A placeholder SEPolicy will be included to support future enhancements and community contributions.
•	 The project should be open to support requests and contributions related to SEPolicy, including use case coverage and general improvements.
Scope of Changes
1. Yocto Layers
•	 Adding layers : meta-selinux, meta-openembedded  (for meta-oe, meta-python)
2. Configuration Control
•	SELinux enablement will be controlled via:  DISTRO_FEATURES_append = " pam selinux"
•	Utilities like   runtime disablement, moving to permissive and standard startup scripts and services and standard utilities like chcon , restroecon   will be part of the build.
3. Library Dependencies
•	 Some applications/bins may require linking with libselinux We will monitor and aim to minimize unnecessary dependencies ( eg . busybox)
4. BusyBox Enhancements
•	 Add support for  SELinux labels in busybox  ( ls -Z / and ps -Z.)
•	- Extended process listing support (currently missing in mainline BusyBox).
5. Kernel Fragment Support
•	 Introduce new kernel fragments: security+selinux.cfg (in addition to existing qcom.cfg).
•	 Selinux.cfg will enable the SELinux LSM (Linux Security Module) in the kernel.
•	 Security hardening features will be mandatory for all product deployments controlled by Security  which might not be enabled by default but  can be used as reference for production process. 
So above will add 2 more kernel fragments so want you comments if we can add  inline to qcom.cfg.
6. SEPolicy Files
•	Include required policy files for Qualcomm services  ( .te, .if, .fc  files  )  mostly  adding this as dynamic layer  1st  approch with update to [meta-qcom](https://github.com/qualcomm-linux/meta-qcom/tree/master)/[conf](https://github.com/qualcomm-linux/meta-qcom/tree/master/conf)/layer.conf   , 2nd approch to add new  recipes-security  So want to know  which is workable  . 

[...]

PR #1368, opened 2026-01-09
refpolicy: Add SELinux policy for qcom location daemons

challako:
Adds necessary permissions for qcom_loc_hal_daemon, qcom_loc_launcher and test app (running in initrc_t or unconfined_t context) for socket creation, communication.

### DIFF
--- a/dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted_git.bbappend
+++ b/dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted_git.bbappend
@@ -1,0 +1,5 @@
+FILESEXTRAPATHS:append := "${THISDIR}/${PN}:"
+
+SRC_URI += " \
+        file://0001-refpolicy-Add-SELinux-policy-for-qcom-location-daemo.patch \
+        "
--- a/dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted/0001-refpolicy-Add-SELinux-policy-for-qcom-location-daemo.patch
+++ b/dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted/0001-refpolicy-Add-SELinux-policy-for-qcom-location-daemo.patch
@@ -1,0 +1,498 @@
+From 4cddc38cab1f1e60540861528786d7926f9b5e5c Mon Sep 17 00:00:00 2001
+From: Challa Koti Reddy <challako@qti.qualcomm.com>
+Date: Thu, 8 Jan 2026 09:27:49 +0530
+Subject: [PATCH] refpolicy: Add SELinux policy for qcom location daemons
+
+Adds necessary permissions for qcom_loc_hal_daemon,
+qcom_loc_launcher and test app (running in initrc_t or
+unconfined_t context) for socket creation, communication.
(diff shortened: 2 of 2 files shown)


challako:
The qcom location daemons will soon be included in meta-qcom-extras.

dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted_git.bbappend
```diff
@@ -0,0 +1,5 @@
+FILESEXTRAPATHS:append := "${THISDIR}/${PN}:"
+
+SRC_URI += " \
```
anujm1, line 3: Needs `:qcom` override along with :append

dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted_git.bbappend
```diff
@@ -0,0 +1,5 @@
+FILESEXTRAPATHS:append := "${THISDIR}/${PN}:"
```
anujm1, line 1: `BPN`

anujm1:
Instead of appending refpolicy, should these be compiled along with the application/product that needs them in their recipe but possibly packaged separately?

That way, we won't need these policy details in meta-qcom and this bbappend won't be needed for people who use meta-selinux with meta-qcom but might not care about this particular application. The policy itself can move/change along with the application when required and will not need patch updates here.

Are there advantages to appending refpolicy?

sasikuma-qti:
> Instead of appending refpolicy, should these be compiled along with the application/product that needs them in their recipe but possibly packaged separately?
> 
> That way, we won't need these policy details in meta-qcom and this bbappend won't be needed for people who use meta-selinux with meta-qcom but might not care about this particular application. The policy itself can move/change along with the application when required and will not need patch updates here.
> 
> Are there advantages to appending refpolicy?

@anujm1 - A discussion happened regarding where to host the policies https://github.com/qualcomm-linux/meta-qcom/issues/1071

**Based on the discussion, the summary is:**
generic policies, should be contributed to refpolicy (chrony, rpcbind, etc.)
Yocto-specific policies which should be contributed to meta-selinux (mount points, etc.)
Qualcomm BSP-specific policies, which should be a part of meta-qcom (tqftpserv, rmtfs, etc.)
Other policies, specific for our distro. They should be a part of meta-qcom-distro.

anujm1:
> @anujm1 - A discussion happened regarding where to host the policies #1071
> 
> **Based on the discussion, the summary is:** generic policies, should be contributed to refpolicy (chrony, rpcbind, etc.) Yocto-specific policies which should be contributed to meta-selinux (mount points, etc.) Qualcomm BSP-specific policies, which should be a part of meta-qcom (tqftpserv, rmtfs, etc.) Other policies, specific for our distro. They should be a part of meta-qcom-distro.

Thank you for this background. I wasn't suggesting that we move these out of meta-qcom. My question is more about whether these policies should be compiled as part of the component recipe that they are needed for. Does it make sense and whether it is possible to do something like:

```
make -f ${datadir}/selinux/devel/Makefile ${modulename}.pp
```

and then package it separately.

PR #1372, opened 2026-01-09
Introduce linux-qcom-next-rt recipe

vivpuar:
Introduce a new RT recipe from the qcom-next branch of qualcomm-linux/kernel tree.

The kernel rt recipe is used to build a fully preemptible Linux kernel and used to ensure deterministic, low-latency performance for time-sensitive applications.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -156,6 +156,14 @@ jobs:
                 type: additional
                 dirname: "+linux-yocto-dev"
                 yamlfile: ":ci/linux-yocto-dev.yml"
+          - machine: iq-9075-evk
+            distro:
+                name: qcom-distro
+                yamlfile: ':ci/qcom-distro-prop-image.yml'
+            kernel:
+                type: qcom-next-rt
+                dirname: "+linux-qcom-next-rt"
+                yamlfile: ":ci/linux-qcom-next-rt.yml"
           # include kvm compatible machines for the builds
           - machine: qcs9100-ride-sx
             distro:


recipes-kernel/linux/linux-qcom-next-rt_git.bb
```diff
@@ -0,0 +1,5 @@
+require linux-qcom-next_git.bb
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/linux-qcom-next:"
```
lumag, line 3: What for?

.github/workflows/build-yocto.yml
```diff
@@ -88,6 +88,9 @@ jobs:
           - type: 6.18
             dirname: "+linux-qcom-6.18"
             yamlfile: ":ci/linux-qcom-6.18.yml"
+          - type: qcom-next-rt
+            dirname: "+linux-qcom-next-rt"
+            yamlfile: ":ci/linux-qcom-next-rt.yml"
```
lumag: There should be no need to do this at the warm-up stage.

recipes-kernel/linux/linux-qcom-next-rt_git.bb
vkraleti: Since this linux-qcom-next-rt_git.bb recipe is including linux-qcom-next_git.bb, to resolve paths of [linux-qcom-next/configs/bsp-additions.cfg](https://github.com/qualcomm-linux/meta-qcom/blob/master/recipes-kernel/linux/linux-qcom-next/configs/bsp-additions.cfg) this is added.

lumag: Then I'd rather modify the original recipe.

vivpuar: oe-core has separate recipe for rt, so referred that and create a separate recipe. 
Are you suggesting to add the config for rt in linux-qcom-next recipe based on some condition?

lumag: No, I was thinking about keeping FILESEXTRAPATH in the original recipe, but I didn't realize it's unnecessary. So, this looks fine.

.github/workflows/build-yocto.yml
vivpuar: Will remove from workflow for now

lumag (approved):

sbanerjee-quic (approved):

lumag (changes requested):
Please rebase on tip. Also please enable testing of the RT kernel.

vivpuar:
> Please rebase on tip. Also please enable testing of the RT kernel.

Sure will rebase.
To enable testing, is updating "build-yocto.yml" enough, the commit which I made earlier?

vivpuar:
> > > Please rebase on tip. Also please enable testing of the RT kernel.
> > 
> > 
> > Sure will rebase. To enable testing, is updating "build-yocto.yml" enough, the commit which I made earlier?
> 
> Yes, you need to add the rt kernel on the kernel matrix.

Updated.

lumag (dismissed):

vkraleti (dismissed):

ricardosalveti (dismissed):

.github/workflows/build-yocto.yml
```diff
@@ -137,6 +140,9 @@ jobs:
...
```
ricardosalveti: This will create several new CI runs, do we want to build and test the rt flavor for every target? Fine if the answer is yes, just wanted to confirm first.

vivpuar:
> > Could we just test -rt and default to it if it's working on all targets? I assume that everything is supposed to work as well on -rt as on non-rt nowadays (perhaps with a minor performance hit that might even be runtime configurable).
> 
> @vivpuar @sbanerjee-quic @vkraleti do we need to build and validate rt on multiple targets for mainline? Maybe we could just pick one or two reference targets and use that for validation instead.
> 
> Then in the qclinux 2.0 branch we could have a wider permutation if needed, as we don't expect the same amount of changes there, since the baseline will be the LTS branches.

Sure, for now we can enable only for lemans for validation and later on qclinux 2.0 we will enable for all the targets.

vkraleti (approved):

ricardosalveti (approved):

PR #1376, opened 2026-01-11
Use FIT-based multi-DTB on applicable boards

vkraleti:
Bootbins v00114.0 release supports FIT-based multi DTB selection feature,
enabling a single dtb.bin for multiple boards. Update `QCOM_DTB_DEFAULT` to
`multi-dtb` for applicable boards to use same dtb.bin on multiple board variants.

### DIFF
--- a/conf/machine/qcs615-ride.conf
+++ b/conf/machine/qcs615-ride.conf
@@ -6,8 +6,6 @@ require conf/machine/include/qcom-qcs615.inc
 
 MACHINE_FEATURES += "efi pci"
 
-QCOM_DTB_DEFAULT ?= "qcs615-ride"
-
 KERNEL_DEVICETREE ?= " \
                       qcom/qcs615-ride.dtb \
                       "


sbanerjee-quic (dismissed):

conf/machine/rb3gen2-core-kit.conf
```diff
@@ -6,7 +6,7 @@ require conf/machine/include/qcom-qcs6490.inc
 
 MACHINE_FEATURES += "efi pci"
 
-QCOM_DTB_DEFAULT ?= "qcs6490-rb3gen2"
+QCOM_DTB_DEFAULT ?= "multi-dtb"
```
lumag: Should we turn this into a default setting, dropping it from individual board configs?

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -8,6 +8,7 @@ inherit dtb-fit-image
 
 DTBBIN_DEPLOYDIR = "${WORKDIR}/qcom_dtbbin_deploy-${PN}"
 DTBBIN_SIZE ?= "4096"
+MULTIDTB_NAME ?= "multi-dtb"
```
ricardosalveti: Do we need / want this value to be changed? Is there any value?

Otherwise just use it directly later.

conf/machine/rb3gen2-core-kit.conf
ricardosalveti: Yeah, since all the qclinux supported targets (current and newer ones) will all support multi-dtb (based on the same edk2 codebase), we can probably set by default.

sbanerjeeqc: Some targets , especially the ones that have been through Android programs will come with ABL and AOSP boot.img style support. For such targets we will need to start with abl2esp and dtb in uki, like we have done for a couple now in meta-qcom.

lumag: That has nothing to do with the `multi-dtb` being a default value.

vkraleti: Updated `QCOM_DTB_DEFAULT ?= "multi-dtb"` in image_types_qcom.bbclass. For machines whose UEFI firmware does not yet support multi-DTB, `QCOM_DTB_DEFAULT ` is explicitly overridden in the machine configuration.

classes/linux-qcom-dtbbin.bbclass
vkraleti: Dropped `MULTIDTB_NAME `.

lumag (approved):

sbanerjee-quic (approved):

PR #1378, opened 2026-01-12
workflows: lava-test-plans: Add pvt tag for iq-9075-evk target

qcom-anilyada:
workflows: lava-test-plans: Fix device selection for iq-9075-evk

LAVA jobs for iq-9075-evk fail when scheduled on incompatible devices.
Lab devices are tagged with "pvt" but job definitions lack this tag.

Add "pvt" tag to iq-9075-evk job definitions to ensure correct device
selection by LAVA scheduler.


mwasilew:
I think this commit is in a wrong place. The change should be done in lava-test-plans. This patch forces all jobs (boot and pre-merge) to be run on PVT devices. This reduces the pool of devices to 2. IMHO this is wrong. Only the tests that require PVT should ask for the tag and this should be done in lava-test-plans. In other words, this commit doesn't belong here.

PR #1379, opened 2026-01-12
optee-client: Add support for libckteec from optee_client

harshaldev27:
Added support for the ckteec library v4.0.0 via an optee-client recipe.

libckteec provides the PKCS#11 platform-agnostic interface for talking to cryptographic hardware such as smart cards and HSMs. It calls into the Global Platform defined standard TEE APIs for talking to the PKCS#11 Trusted Application (TA) running on the platform's TEE.

On Qualcomm platforms, these Global Platform APIs are implemented via the minkteec library and so the ckteec library must link to it instead of the teec library. This is due to Qualcomm TEE only understanding the MINK IPC protocol.

The optee-client recipe is written to only install the ckteec library from the optee-client package since the other libraries from this package cannot be currently used on Qualcomm platforms which only support QTEE. The recipe also patches the CMake file to alter it's linkage from the default teec library to minkteec.

The ckteec library interface is tested by running the pkcs11 test suite from the xtest client v4.0.0 hosted at github.com/OP-TEE/optee_test on the RB3Gen2 board with a compatible PKCS#11 TA.

### DIFF
--- a/classes-recipe/dtb-fit-image.bbclass
+++ b/classes-recipe/dtb-fit-image.bbclass
@@ -19,9 +19,7 @@ QCOMFIT_DEPLOYDIR = "${WORKDIR}/qcom_fitimage_deploy-${PN}"
 do_generate_qcom_fitimage[depends] += "qcom-dtb-metadata:do_deploy"
 do_generate_qcom_fitimage[cleandirs] += "${QCOMFIT_DEPLOYDIR}"
 python do_generate_qcom_fitimage() {
-    import sys, os, shutil, re
-    import oe.types
-    from pathlib import Path
+    import os, shutil
     from qcom.dtb_only_fitimage import QcomItsNodeRoot
 
     fit_dir = d.getVar('QCOMFIT_DEPLOYDIR')
@@ -38,29 +36,68 @@ python do_generate_qcom_fitimage() {
 
     root_node.set_extra_opts(d.getVar("FIT_DTB_MKIMAGE_EXTRA_OPTS") or "")
 
-    # Prepend qcom-metadata.dtb to KERNEL_DEVICETREE
-    kernel_devicetree = d.getVar('KERNEL_DEVICETREE') or ""
-    kernel_devicetree = ('qcom-metadata.dtb ' + kernel_devicetree).strip()
-
-    deploy_dir_image = d.getVar('DEPLOY_DIR_IMAGE') 
+    deploy_dir_image = d.getVar('DEPLOY_DIR_IMAGE')
     dtb_dir = os.path.join(d.getVar('B'), "arch", d.getVar('ARCH'), "boot", "dts", "qcom")
+    os.makedirs(fit_dir, exist_ok=True)
+
+    # Always include QCOM metadata first
     qcom_meta_src = os.path.join(deploy_dir_image, 'qcom-metadata.dtb')
     qcom_meta_dst = os.path.join(dtb_dir, 'qcom-metadata.dtb')
     shutil.copy(qcom_meta_src, qcom_meta_dst)
+    root_node.fitimage_emit_section_dtb("qcom-metadata.dtb", qcom_meta_dst, compatible_str=None, dtb_type="qcom_metadata")
 
-    for dtb in kernel_devicetree.split():
-        dtb_name = os.path.basename(dtb)
-        dtb_base = os.path.splitext(dtb_name)[0]
-        compatible = d.getVarFlag("FIT_DTB_COMPATIBLE", dtb_base.replace(',', '_')) or ""
-        dtb_path = os.path.join(dtb_dir, f"{dtb_base}.dtb")
-        if not compatible and dtb_name != "qcom-metadata.dtb":
-            bb.fatal(f"FIT_DTB_COMPATIBLE[{dtb_base}] is not set. ")
-        root_node.fitimage_emit_section_dtb(
-            dtb_name, dtb_path,
-            compatible_str=compatible,
-        )
+    # KERNEL_DEVICETREE contains both .dtb and .dtbo
+    files_set = {os.path.basename(x) for x in (d.getVar('KERNEL_DEVICETREE') or "").split()}
 
-    root_node.fitimage_emit_section_config()
+    # Collect DTB/DTBO names selected in KERNEL_DEVICETREE to validate declarative FIT_DTB_COMPATIBLE combinations
+    base_dtb_list  = {os.path.splitext(f)[0] for f in files_set if f.endswith(".dtb")}
+    overlay_dtbo_list = {os.path.splitext(f)[0] for f in files_set if f.endswith(".dtbo")}
+
+    # Parse composite compatible keys :
+    # FIT_DTB_COMPATIBLE[base+ovl1+ovl2] = "..."
+    overlay_groups  = {}
+    overlay_compats = {}
+
+    compat_flags = d.getVarFlags("FIT_DTB_COMPATIBLE") or {}
+    for key, compat_val in compat_flags.items():
+        if '+' not in key:
+            continue
+
+        parts = [os.path.basename(p) for p in key.split('+')]
+        if not parts:
+            continue
+
+        base_stem = parts[0]
+        ovl_stems = parts[1:]
+
+        # Skip base+overlay combinations not present in KERNEL_DEVICETREE to avoid generating invalid FIT configs
+        # from declarative FIT_DTB_COMPATIBLE metadata
+        if not (base_stem in base_dtb_list and all(ovl in overlay_dtbo_list for ovl in ovl_stems)):
+            continue
+
+        base = base_stem + ".dtb"
+        overlays = [ovl + ".dtbo" for ovl in ovl_stems]
+
+        overlay_groups.setdefault(base, []).append(overlays)
+        overlay_compats[key] = compat_val
+
+    # Emit DTB/DTBO sections for every entry from KERNEL_DEVICETREE
+    for fname in files_set:
+        dtb_path = os.path.join(dtb_dir, fname)
+        if not os.path.exists(dtb_path):
+            bb.fatal(f"Required file '{fname}' not found at '{dtb_path}'.")
+
+        compatible = ""
+        if fname.endswith(".dtb"):
+            dtb_key = os.path.splitext(fname)[0].replace(',', '_')
+            compatible = d.getVarFlag("FIT_DTB_COMPATIBLE", dtb_key) or ""
+            if not compatible:
+                bb.fatal(f"FIT_DTB_COMPATIBLE[{dtb_key}] is not set for base DTB '{fname}'.")
+
+        root_node.fitimage_emit_section_dtb(fname, dtb_path, compatible_str=compatible, dtb_type="flat_dt")
+
+    # Emit configuration sections
+    root_node.fitimage_emit_section_qcomconfig(overlay_groups, overlay_compats)
 
     root_node.write_its_file(itsfile)
 
--- a/classes/linux-qcom-dtbbin.bbclass
+++ b/classes/linux-qcom-dtbbin.bbclass
@@ -16,6 +16,8 @@ do_qcom_dtbbin_deploy() {
         bbdebug 1 " combining: $dtbf"
         dtb=`normalize_dtb "$dtbf"`
         dtb_ext=${dtb##*.}
+        # Skip DTBOs
+        [ "$dtb_ext" = "dtbo" ] && continue
         dtb_base_name=`basename $dtb .$dtb_ext`
         mkdir -p ${DTBBIN_DEPLOYDIR}/$dtb_base_name
         cp ${D}/${KERNEL_DTBDEST}/$dtb_base_name.dtb ${DTBBIN_DEPLOYDIR}/$dtb_base_name/combined-dtb.dtb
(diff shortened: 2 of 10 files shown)


dynamic-layers/openembedded-layer/recipes-security/optee-client/optee-client/0001-libckteec-Add-tee-trace-files-until-minkteec-exports.patch
```diff
@@ -0,0 +1,331 @@
+From 0c716be7de1609e3f78d3052a007181f8a69d3a2 Mon Sep 17 00:00:00 2001
+From: Harshal Dev <harshal.dev@oss.qualcomm.com>
+Date: Mon, 12 Jan 2026 18:04:09 +0530
+Subject: [PATCH] libckteec: Add tee trace files until minkteec exports them
+
+Added teec_trace.h and teec_trace.c to libckteec library since minkteec
+library does not export and install them unlike the optee teec library.
+
+Upstream-Status: Pending
```
lumag: Please submit upstream

dynamic-layers/openembedded-layer/recipes-security/optee-client/optee-client/0002-libckteec-Link-ckteec-library-to-minkteec-for-Qualco.patch
```diff
@@ -0,0 +1,128 @@
...
+minkteec library which exposes the Global Platform (GP) interface for
+talking to the Qualcomm Trusted Execution Environment (QTEE) on Qualcomm
+platforms.
+
+Currently, QTEE only supports Object-based MINK IPC protocol for
+communication and so, minkteec must be used instead of the teec library
+on Qualcomm platforms to talk to the PKCS#11 Trusted Application via
+conversion between the GP interface and the MINK protocol.
+
+Upstream-Status: Pending
```
lumag: Likewise

dynamic-layers/openembedded-layer/recipes-security/optee-client/optee-client_4.0.0.bb
```diff
@@ -0,0 +1,48 @@
...
+LIC_FILES_CHKSUM = "file://LICENSE;md5=69663ab153298557a59c67a60a743e5b"
+
+inherit systemd cmake pkgconfig
+
+DEPENDS += "minkipc util-linux"
+EXTRA_OEMAKE += "PKG_CONFIG=pkg-config"
+
+#v4.0.0
+SRCREV = "acb0885c117e73cb6c5c9b1dd9054cb3f93507ee"
+SRC_URI = "git://github.com/OP-TEE/optee_client.git;branch=master;protocol=https \
```
lumag: Why are we not using optee-client from meta-arm?

dynamic-layers/openembedded-layer/recipes-security/optee-client/optee-client_4.0.0.bb
```diff
@@ -0,0 +1,48 @@
...
+DESCRIPTION = "Open Portable Trusted Execution Environment - Normal World Client side of the TEE"
+HOMEPAGE = "https://www.op-tee.org/"
+
+LICENSE = "BSD-2-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=69663ab153298557a59c67a60a743e5b"
+
+inherit systemd cmake pkgconfig
+
+DEPENDS += "minkipc util-linux"
+EXTRA_OEMAKE += "PKG_CONFIG=pkg-config"
```
ricardosalveti: This will break other layers that are also building optee-client, since nothing here is isolated to the qcom override.

anujm1: I wonder if this is even needed as the recipe is using cmake.

b49020:
@harshaldev27 IIRC, last time we talked about this feature was to add support for `libckteec` in the `minkipc` library itself [1] as an immediate solution. For sure directly patching `optee-client` like this will break OP-TEE use-cases. The long-term solution as @ricardosalveti mentioned and we discussed earlier as well was to contribute directly to OP-TEE projects to allow for QTEE trusted OS as an alternative backend to OP-TEE OS.

[1] https://github.com/qualcomm/minkipc

harshaldev27:
Hi @ricardosalveti , thank you so much for your comments and providing useful pointers.

> We are expecting to integrate OP-TEE (not just this TA) soon (see #1172), so NAK for including a custom optee-client recipe that is only suitable for minkipc.

I think there is a slight confusion here. This PR is not for integrating the PKCS#11 Trusted Application (TA). This PR is for fetching and building the `libckteec` Linux library from the optee_client repo with patches to change its linkage from `libteec` to `libminkteec` provided as part of the `minkipc` package already integrated into meta-qcom.

Also, this PR is orthogonal to the OP-TEE enablement PR. I can see that there is a discussion on PR #1172 between @lumag and @b49020 regarding whether `meta-arm` should be a dynamic layer or a default layer dependency of `meta-qcom`. And I agree with @lumag on that discussion that right now, Kodiak currently boots with proprietary firmware blobs, QTEE (`tz.mbn` image) being one of them.
And so, this PR is targeting Kodiak which boots with QTEE, and not OP-TEE. As such, it appears to me that I cannot as-is leverage the recipes from `meta-arm`, since the `optee-client` recipe provided by it is building and installing all libraries/daemons from the `optee_client` repo, out of which only `libckteec` is useful when we are booting with QTEE. None of the other libraries, such as `libteec`, `libseteec` or the `tee-supplicant` daemon are applicable to us currently.
See: https://git.yoctoproject.org/meta-arm/tree/meta-arm/recipes-security/optee/optee-client.inc

When OP-TEE support comes, I have no doubt that these will become relevant again, and must be included as part of an alternate OPTEE-based distro configuration perhaps? (I don't know what is the plan there, please let me know)

> 
> What should be done here:
> 
> * Work with upstream to support building with minkipc (and remove the need for extra patches)

I agree with this, I believe I can try to send the patch for providing a build-time option for linking `libckteec` with either `libteec` (default) or `libminkteec` (for QC boards booting with QTEE). @b49020  I would appreciate your input on this.

> * Leverage the optee recipes from meta-arm, no need to duplicate any recipe, and should be aligned with the latest optee-client revision

Like I mentioned, I am fine with leveraging the `meta-arm` recipes, but I realize that I will have to find some way to modify its behavior such that it builds `libckteec` with linkage to `libminkteec` followed by uninstallation of all the other irrelevant libraries provided by it.
Also, we cannot currently support the latest revision of `optee-client` from that layer, because the PKCS#11 TA currently supported by QTEE is only aligned with v4.0.0 of the libckteec library. Work for upgrading to newer versions is planned by QTEE team internally.

Is there some recommendation for writing an override for the `optee-client` recipe leveraged from meta-arm to achieve these objectives? If so, I'd be happy to pursue this path.

> * Supporting specific minkipc options based on the qcom override (and here should also need to find a way to support building traditional optee and just optee-client when qtee is used)

I could not understand this comment fully @ricardosalveti , could you please elaborate a bit further what are you requesting here?

Thanks,
Harshal

harshaldev27:
Hi Sumit,

> @harshaldev27 IIRC, last time we talked about this feature was to add support for `libckteec` in the `minkipc` library itself >  [1] as an immediate solution. For sure directly patching `optee-client` like this will break OP-TEE use-cases. 

We did not want to pursue the temporary solution of adding `libckteec` code directly in `minkipc` for the following reasons which came to light a bit later for me:
1. The PKCS#11 TA for Kodiak is currently only supporting v4.0.0 of the `libckteec` library. However, work for upgrading the versions is constantly underway. Going this path would make it a repeated effort for us to support newer and newer versions of `libckteec` in the `minkipc` repo.
2. Since `libckteec` is not exactly related to MINK IPC. We would need some additional approvals from legal to get this going, which anyways means that this is not a immediate solution.
3. I believe being able to directly sync from the `optee-client` repo and support in `meta-qcom` is the most convenient way forward in the long-term. Allowing us to leverage both new releases and bug fixes seamlessly overtime.

Also, my intent wasn't to break any OPTEE-use cases. The `.patch` was meant only for this repo. Now I see based on your PR work is underway to support OP-TEE boot as well, in that case we do need a way to not affect the `optee-client` binaries you plan to install via recipe overrides.

> The long-term solution as @ricardosalveti mentioned and we discussed earlier as well was to contribute directly to OP-TEE > projects to allow for QTEE trusted OS as an alternative backend to OP-TEE OS.

I agree, I think better to allow a build-time option which enables alternatively linking `libckteec` with `libminkteec`?

> [1] https://github.com/qualcomm/minkipc

Thanks,
Harshal

b49020:
@harshaldev27 

> The PKCS#11 TA for Kodiak is currently only supporting v4.0.0 of the libckteec library. However, work for upgrading the versions is constantly underway. Going this path would make it a repeated effort for us to support newer and newer versions of libckteec in the minkipc repo.

This effort has to be done irrespective of that since `libminkteec` and `libckteec` aren't developed in the same repo. The suggestion to import `libckteec` in minkipc repo gives you control over supporting future updates of `libckteec`.

> Since libckteec is not exactly related to MINK IPC. We would need some additional approvals from legal to get this going, which anyways means that this is not a immediate solution.

Open-source projects usually have third party dependencies. Both projects follow the BSD license.

> I believe being able to directly sync from the optee-client repo and support in meta-qcom is the most convenient way forward in the long-term. Allowing us to leverage both new releases and bug fixes seamlessly overtime.

Nope `meta-qcom` shouldn't be used to manage inter-dependencies via patching different components. It simply becomes un-manageable in the long term and especially since we want to support both QTEE and OP-TEE based stacks.

ricardosalveti:
> I completely agree, we should have a top-level distro/machine variable to select the specific TEE we want to use for the platform along with the Linux side packages we want to install at runtime. @b49020 Do let me know if you are planning to introduce any such variable with your OP-TEE PR. 😄

We can use qtee or optee in MACHINE_FEATURES, for example.

b49020:
> > `optee-client` added support for it since it exposes OP-TEE based TA as the PKCS#11 token with backend GP TEE APIs implementation supported via `libteec`. In case you want to support QTEE based GP interfaces with `optee-client` then you need to port `libminkteec` to `optee-client` which doesn't make sense to me.
> 
> The advantage of having minkteec support via libteec is that we could also easily reuse the same TAs and libs available in optee-client later on, if needed. Importing will make it pkcs11 specific (which is fine for this scenario as we don't yet have ways to build the TAs that will later be used by QTEE as part of the yocto build process).

The minkteec library itself is based on MINK object invoke protocol (different IOCTL) which is mostly different from GP TEE protocol IOCTLs used by OP-TEE. The other major difference is how the use-space supplicants are supported. So, I am not sure how much library code reuse will be there but if we want to unify the user-space libraries then I think that would be a much bigger effort. BTW, I would encourage QTEE folks to explore that possibility as it can make the TEEC library common which can provide same GP interface to clients while dynamically adjusting to whichever TEE is supported underneath based on kernel interface.

b49020:
> > Thinking out loud here, if we have a build-time flag in `optee-client` repo Makefile which changes `libckteec` linkage to `libminkteec`, we could have a `.bbappend` for `optee-client` recipe, which passes this flag, and then finally, in the `.bbappend` we could use the top-level distro/machine variable to conditionally install only libckteec from the optee-client package?
> > This seems like a cleaner and do-able approach to target in the long-run.
> 
> This could work, if all you need is linking to a different teec library.

It's not just different library linkage but there are other artifacts too like tee-supplicant, other libs etc. So, we need all of them to be not packaged. As I mentioned earlier, try to initiate discussions upstream first to see if such a flag would be acceptable upstream.

b49020:
> > I completely agree, we should have a top-level distro/machine variable to select the specific TEE we want to use for the platform along with the Linux side packages we want to install at runtime. @b49020 Do let me know if you are planning to introduce any such variable with your OP-TEE PR. 😄
> 
> We can use qtee or optee in MACHINE_FEATURES, for example.

I am more inclined to having a separate machine for TF-A/OP-TEE based stack given all the discussions I had on corresponding PR. So, feel free to add QTEE related features for the current machines available.

harshaldev27:
> > > `optee-client` added support for it since it exposes OP-TEE based TA as the PKCS#11 token with backend GP TEE APIs implementation supported via `libteec`. In case you want to support QTEE based GP interfaces with `optee-client` then you need to port `libminkteec` to `optee-client` which doesn't make sense to me.
> > 
> > 
> > The advantage of having minkteec support via libteec is that we could also easily reuse the same TAs and libs available in optee-client later on, if needed. Importing will make it pkcs11 specific (which is fine for this scenario as we don't yet have ways to build the TAs that will later be used by QTEE as part of the yocto build process).
> 
> The minkteec library itself is based on MINK object invoke protocol (different IOCTL) which is mostly different from GP TEE protocol IOCTLs used by OP-TEE. The other major difference is how the use-space supplicants are supported. So, I am not sure how much library code reuse will be there but if we want to unify the user-space libraries then I think that would be a much bigger effort. BTW, I would encourage QTEE folks to explore that possibility as it can make the TEEC library common which can provide same GP interface to clients while dynamically adjusting to whichever TEE is supported underneath based on kernel interface.

I'll take this feedback. Thank you.

ricardosalveti:
> I am more inclined to having a separate machine for TF-A/OP-TEE based stack given all the discussions I had on corresponding PR. So, feel free to add QTEE related features for the current machines available.

Yeah, in the end the final representation could be a different machine, and what would indeed change is the machine features content (optee x qtee).

PR #1382, opened 2026-01-12
Disable uninative

quaresmajose:
Uninative is an attempts to isolate the build system from the host distribution's
C library in order to make re-use of native shared state artifacts across different
host distributions practical.

We are running on some build issue with clang/rust and we belive which are caused
by the uninative glibc. To unlock the CI, we'll temporarily disable the uninative.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -54,6 +54,8 @@ local_conf_header:
     EXTRA_IMAGE_FEATURES = "allow-empty-password empty-root-password allow-root-login"
     IMAGE_ROOTFS_EXTRA_SPACE = "307200"
     WATCHDOG_RUNTIME_SEC:pn-systemd = "30"
+  uninative: |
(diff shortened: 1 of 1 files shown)


lumag:
> THanks! Do you know, what exactly fixes the build issue - is it a diskmon or uninative?

And #1381 answers that question.

ricardosalveti:
Let's see if the opkg issue will keep showing up with this, at least it didn't fail during this run.

ricardosalveti (approved):

koenkooi:
> Investigation still on going upstream:
> 
> fedora 43 sstate contamination issues https://bugzilla.yoctoproject.org/show_bug.cgi?id=16087

That bug report makes me loose the litte remaining faith I had in uninative and by extension, the published sstate. We should stop using it and continue relying on kas-container. It shouldn't be needed, but it sadly is :(

PR #1383, opened 2026-01-12
initramfs-firmware-mega-image: remove empty linux-firmware package

ricardosalveti:
linux-firmware is just an empty package since oe-core [1], remove from initramfs-firmware-mega-image as an attempt to avoid bug https://bugzilla.yoctoproject.org/show_bug.cgi?id=16010, which started to happen after the above oe-core commit.

[1]
https://git.openembedded.org/openembedded-core/commit/?id=9b883aa6f3cf881bfd60442b9ec193ae191b4cbe

### DIFF
--- a/recipes-bsp/images/initramfs-firmware-mega-image.bb
+++ b/recipes-bsp/images/initramfs-firmware-mega-image.bb
@@ -1,7 +1,6 @@
 DESCRIPTION = "Huge image with all firmware files. This is intended to check for possible conflicts, etc."
 
 PACKAGE_INSTALL = " \
-    linux-firmware \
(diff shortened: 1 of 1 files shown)


ricardosalveti:
> Well... It's not empty. It pulls _all_ other firmware packages. My main goal with initramfs-firmware-mega-image is to ensure that there are no conflicts between generic and device-specific packages, that different firmware packages / installation places look consistent, etc. E.g. it was the mega image (or my review of its contents) which prompted me to change locations of audio topologies (otherwise it's nearly impossible to notice inconsistencies).

Oh, ok, I had the impression it was needed because we had relevant firmware which was not yet packaged in a separated package, and not to use it to identify conflicts (which is valid).

I wonder if the long recommends list is causing this, but at the same time disabling uninative seems to get things to work again.

PR #1390, opened 2026-01-14
Use `single-dtb` on QCS(8300|9100)-Ride Boards

vkraleti:
Multi-DTB support for QCS (8300|9100) Ride boards is not available.
Revert to single-dtb on these boards to avoid boot failures.


lumag (approved):
I took the liberty to fix the commit message: `s/combined/single/`, we don't use combined-dtb.

PR #1391, opened 2026-01-14
Revert temporarily disable uninative

quaresmajose:


quaresmajose:
> Right, what are the actual benefits we will have here?

We can better reuse the Yocto upstream sstate-cache mirror. One of the major things of the uninative is to re-use the sstate-cache in the various supported Linux distributions and not make sstate-cache dependent on the host.

ricardosalveti (approved):
As it is now enabled by default in oe-core.

PR #1396, opened 2026-01-15
workflows: prevent creating commit check with no results

mwasilew:
When there are no test results publish-results workflow will create a "pass" check on the commit. This is a limitation of the result publishing action. This patch prevents publishing action from running when there are no test results available. The check is not added. Since the check is on the list of expected, missing check will raise alarm and prevent auto-merging of patches.

### DIFF
--- a/.github/workflows/publish-results.yml
+++ b/.github/workflows/publish-results.yml
@@ -42,9 +42,14 @@ jobs:
             github-token: ${{ github.token }}
 
       - name: "List files"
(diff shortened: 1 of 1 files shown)


mwasilew:
I tested this in my repository with download-action enabled and disabled. It looks like this:
- downloads disabled: 
Screenshot from 2026-01-15 11-05-38
- downloads enabled:
Screenshot from 2026-01-15 11-05-11

Issue #1276, opened 2025-12-02
Perf builds not available in Mainline


ctheegal:
Existing mainline builds does not have all the required perf flavor/build specific optimizations like ttynull console and disabling DEBUG kernel-defconfigs as it does not have any “perf build”.

Maintaining perf flavor builds helps us to give the right picture about chipset capabilities (especially w.r.t power and performance) to external world. Also, note that this perf flavour is the final production build used by customers and we can't afford to miss this critical part.

1. Following DEBUG defconfigs are enabled. These need to be disabled for perf build.

CONFIG_CMA_DEBUG=y
CONFIG_CMA_DEBUGFS=y
CONFIG_ATH11K_DEBUG=y
CONFIG_ATH12K_DEBUG=y
CONFIG_DYNAMIC_DEBUG=y
CONFIG_DEBUG_PAGEALLOC=y
CONFIG_DEBUG_OBJECTS=y
CONFIG_DEBUG_OBJECTS_ENABLE_DEFAULT=1
CONFIG_DEBUG_SPINLOCK=y
CONFIG_DEBUG_LIST=y

2. "console=ttynull" need to be part of kernel cmdline.

3."Default cpu-freq governor as Schedutil" needs to be enabled for perf build. 
CONFIG_CPU_FREQ_GOV_SCHEDUTIL=y

ricardosalveti:
This is a distro-level decision, you could add a meta-qcom-distro distro feature that could optimize the build to the level you want.

As @ndechesne said, we should also get concrete data showing the actual 'cost' of every change suggested here.

vkraleti:
Since the performance optimizations and their effects are known, wouldn’t it be sufficient to document them, validate periodically, and allow users to enable or disable them based on their preferences? Not every optimization will suit every user. For example, I personally wouldn’t want the console disabled on my device, while others might prefer more extensive logging.

rsiddoji:
how are perf build going to be enabled  , Via a distro feature ? , Can this be used as production flag  where we can enable security  hardening , based on this perf flag .  How many variant are planned to be validated as part of PR-build process ?  is this just going to be kernel defconfig changes or user space #if DEBUG macro code discard also ?

quic-bjorande:
> 2. "console=ttynull" need to be part of kernel cmdline.
> 

Fully disabling the console has usability and debuggability implications, can we instead put "quiet" on the command line? This removes the vast majority of the overhead, while still providing a functional console.

[...]

PR #1400, opened 2026-01-16
Include `debug.config` in KBUILD_CONFIG_EXTRA

vkraleti:
Include `debug.config` in KBUILD_CONFIG_EXTRA to enable additional
kernel debug options during build only when `DEBUG_BUILD` is enabled.
This enables additional kernel debug options for builds that require 
extended debugging, while leaving normal builds unaffected.

### DIFF
--- a/recipes-kernel/linux/linux-qcom_6.18.bb
+++ b/recipes-kernel/linux/linux-qcom_6.18.bb
@@ -40,6 +40,7 @@ KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 KBUILD_CONFIG_EXTRA = "${@bb.utils.contains('DISTRO_FEATURES', 'hardened', '${S}/kernel/configs/hardening.config', '', d)}"
 KBUILD_CONFIG_EXTRA:append:aarch64 = " ${S}/arch/arm64/configs/prune.config"
 KBUILD_CONFIG_EXTRA:append:aarch64 = " ${S}/arch/arm64/configs/qcom.config"
+KBUILD_CONFIG_EXTRA:append = " ${@oe.utils.vartrue('DEBUG_BUILD', '${S}/kernel/configs/debug.config', '', d)}"
 
 do_configure:prepend() {
     # Use a copy of the 'defconfig' from the actual repo to merge fragments


recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -39,6 +39,7 @@ KBUILD_DEFCONFIG:qcom-armv7a = "qcom_defconfig"
 KBUILD_CONFIG_EXTRA = "${@bb.utils.contains('DISTRO_FEATURES', 'hardened', '${S}/kernel/configs/hardening.config', '', d)}"
 KBUILD_CONFIG_EXTRA:append:aarch64 = " ${S}/arch/arm64/configs/prune.config"
 KBUILD_CONFIG_EXTRA:append:aarch64 = " ${S}/arch/arm64/configs/qcom.config"
+KBUILD_CONFIG_EXTRA:append = " ${S}/kernel/configs/debug.config"
```
lumag: No. These options are not suitable for the kernel that is used for anything else than debugging

```
...
CONFIG_DEBUG_KMEMLEAK=y
CONFIG_DEBUG_KMEMLEAK_AUTO_SCAN=y
CONFIG_DEBUG_OBJECTS=y
CONFIG_DEBUG_OBJECTS_ENABLE_DEFAULT=1
CONFIG_DEBUG_OBJECTS_FREE=y
CONFIG_DEBUG_OBJECTS_PERCPU_COUNTER=y
CONFIG_DEBUG_OBJECTS_RCU_HEAD=y
CONFIG_DEBUG_OBJECTS_TIMERS=y
CONFIG_DEBUG_OBJECTS_WORK=y
CONFIG_DEBUG_PER_CPU_MAPS=y
CONFIG_DEBUG_STACK_USAGE=y
CONFIG_DEBUG_VIRTUAL=y
CONFIG_DEBUG_VM=y
CONFIG_DEBUG_VM_PGFLAGS=y
CONFIG_DEBUG_VM_RB=y
CONFIG_DEBUG_VM_VMACACHE=y
CONFIG_KASAN=y
....
```

vkraleti: meta-qcom nightly builds are used for long‑duration tests to uncover stability and memory issues. These issues are difficult to reproduce and often require multiple iterations over several hours or even days to replicate and capture logs. Therefore, it is essential to always collect the right logs during the run, which requires enabling the appropriate debug configurations.
Kernel configuration options known to impact general performance have already been removed from debug.config in https://github.com/qualcomm-linux/kernel/commit/a3d23553f04a1a564da87113089224e2a7e9c58b. 

To name few disabled configs from above list.
```
CONFIG_DEBUG_KMEMLEAK=y
CONFIG_DEBUG_KMEMLEAK_AUTO_SCAN=y
CONFIG_KASAN=y

```
If there are additional options that are not needed, they can also be considered for removal. However, retaining the non‑problematic debug options is necessary for these tests.

lumag: No. The config still contains a lot of useless debug options which however will impact the behaviour and the performance of the running system (thus triggereing behaviour and bugs different from the ones which are visible on the production system). For example, what kind of issues do you want to uncover with `CONFIG_DEBUG_OBJECTS`?  What is the impact of `CONFIG_DEBUG_VM` or `CONFIG_DEBUG_VIRTUAL`? On the other hand, the mentioned commit has disabled (among others) one of the few really necessary and sensible Kconfig debug options.

Overall, default meta-qcom builds should be suitable for the _production_. If one needs a debug build, they can create it on their own. Feel free to provide support for it via e.g. DISTRO_FEATURES or any other available way.

vkraleti: In https://github.com/qualcomm-linux/meta-qcom/issues/1276#issuecomment-3604189278, it is suggested that builds be tuned for `performance` using a DISTRO_FEATURE, which is for production requirements. However, the recommendation here to keep debug options under another DISTRO_FEATURE. This creates a conflict: one feature aims for maximum performance, while the other aims for maximum debuggability.

I believe there should be a single, mutually exclusive feature. Either `performance `or `debuggability` but not both at the same time.

lumag: As I wrote, I think the default should be production-ready. Which means, no extra optimisations, but also no extra debugs. I don't see a problem with enabling debugfs, but options from `debug.config` are definitely an overkill. I would not run such a kernel anywhere.

vkraleti: So there are going to be 3 variants of builds 
1) Default/production
2) Performance optimized
3) Extended debug enabled

Added two new distros via https://github.com/qualcomm-linux/meta-qcom-distro/pull/140 to facilitate this discussion.

@ndechesne @idlethread @ricardosalveti are you aligned with this approach?

ctheegal: Maintaining 3 variants which I think, is an overkill. Why not club #1 (default / production) & #2 (performance optimized) ?

vkraleti: Unless the defconfig is aligned between the current qcom-next build and the requirement mentioned in https://github.com/qualcomm-linux/meta-qcom/issues/1276, it won’t be possible to club items 1 and 2 together.
If we all agree to handle this in qualcomm-linux/kernel instead of meta-qcom, then item 1 can be dropped.

Also, please note that adding `quiet `to the kernel command line reduces the number of logs but does not completely disable the console. For performance builds, the expectation is to eliminate the console entirely to ensure more accurate benchmarking results. If that is indeed the requirement, then we will need a 3rd build variant with the console fully disabled.

lumag: The default image should be _production_. Always. There should be no need for a separate "performance-optimized" config.
Next, if we need, we can add the "debug" variant. But I would rather not build it through meta-qcom CI. Its usage should be very, very limited. By default all the testing should be preformed using the production builds.

vkraleti: Updated the change to include `debug.config` only when user builds with `DEBUG_BUILD = 1` option. This should enable limited usage of debug.config for users on need basis.

PR #1406, opened 2026-01-19
ci: actions/compile: : retain OSTree commit artifacts

codeshravan:
Include *.rootfs.ostreecommit.tar.xz files in the artifact staging and upload process for nightly builds. This ensures OSTree commit archives are preserved alongside other build artifacts.

The find command pattern is added to the symbolic link copy operation, which will gracefully handle cases where ostreecommit files are not present for certain machine configurations.

Files matching this pattern will be:
- Copied to the uploads directory
- Uploaded as private artifacts
- Published to the S3 bucket for build artifact storage

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -89,7 +89,7 @@ runs:
         uploads_dir=./uploads/${{ inputs.distro_name }}${{ inputs.kernel_dirname }}/${{ inputs.machine }}
         mkdir -p $uploads_dir
         find $deploy_dir/ -maxdepth 1 -type f -exec cp {} $uploads_dir/ \;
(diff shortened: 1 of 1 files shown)


quaresmajose:
The change is on the compile actions and they will deployed whenever the file `*.rootfs.ostreecommit.tar.xz` exists and not only on the nightly builds. it's best to remove the nightly builds mentions from the commit subject and message.

`ci: actions/compile: : retain OSTree commit artifacts` describes the commit better in my opinion.

codeshravan:
> `s/were not preserved/are not preserved/`. LGTM otherwise.

Done

lumag (approved):

PR #1412, opened 2026-01-20
recipes-ml: Add recipe for QAIRT SDK

tjamula:
This PR adds a new bitbake recipe for integrating the QAIRT SDK (v2.41.0.251128) into the build system.

These are pre built binaries and more information about the SDK can be found at : [QAIRT overview](https://docs.qualcomm.com/doc/80-63442-10/topic/general_overview.html)

### DIFF
--- a/conf/machine/iq-8275-evk.conf
+++ b/conf/machine/iq-8275-evk.conf
@@ -18,6 +18,7 @@ KERNEL_DEVICETREE:append:pn-linux-qcom-next = " \
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-iq-8275-evk-firmware \
     packagegroup-iq-8275-evk-hexagon-dsp-binaries \
+    qairt-sdk-hexagon-v75 \
     qcom-fastcv-binaries-qcs8300-ride-dsp \
 "
 
--- a/conf/machine/iq-9075-evk.conf
+++ b/conf/machine/iq-9075-evk.conf
@@ -18,6 +18,7 @@ KERNEL_DEVICETREE:append:pn-linux-qcom-next = " \
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-iq-9075-evk-firmware \
     packagegroup-iq-9075-evk-hexagon-dsp-binaries \
+    qairt-sdk-hexagon-v73 \
     qcom-fastcv-binaries-sa8775p-ride-dsp \
 "
 
--- a/conf/machine/qcm6490-idp.conf
+++ b/conf/machine/qcm6490-idp.conf
@@ -13,6 +13,7 @@ KERNEL_DEVICETREE ?= " \
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-rb3gen2-firmware \
     packagegroup-rb3gen2-hexagon-dsp-binaries \
+    qairt-sdk-hexagon-v68 \
     qcom-fastcv-binaries-thundercomm-rb3gen2-dsp \
 "
 
--- a/conf/machine/qcs8300-ride-sx.conf
+++ b/conf/machine/qcs8300-ride-sx.conf
@@ -19,6 +19,7 @@ KERNEL_DEVICETREE:append:pn-linux-qcom-next = " \
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-qcs8300-ride-firmware \
     packagegroup-qcs8300-ride-hexagon-dsp-binaries \
+    qairt-sdk-hexagon-v75 \
     qcom-fastcv-binaries-qcs8300-ride-dsp \
 "
 
--- a/conf/machine/qcs9100-ride-sx.conf
+++ b/conf/machine/qcs9100-ride-sx.conf
@@ -25,6 +25,7 @@ KERNEL_DEVICETREE:append:pn-linux-qcom-next = " \
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-sa8775p-ride-firmware \
     packagegroup-sa8775p-ride-hexagon-dsp-binaries \
+    qairt-sdk-hexagon-v73 \
     qcom-fastcv-binaries-sa8775p-ride-dsp \
 "
 
--- a/conf/machine/rb3gen2-core-kit.conf
+++ b/conf/machine/rb3gen2-core-kit.conf
@@ -18,6 +18,7 @@ KERNEL_DEVICETREE:append:pn-linux-qcom-next = " \
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-rb3gen2-firmware \
     packagegroup-rb3gen2-hexagon-dsp-binaries \
+    qairt-sdk-hexagon-v68 \
     qcom-fastcv-binaries-thundercomm-rb3gen2-dsp \
 "
 


recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,66 @@
...
+
+SRC_URI = "https://softwarecenter.qualcomm.com/api/download/software/sdks/Qualcomm_AI_Runtime_Community/All/${PV}/v${PV}.zip"
+SRC_URI[sha256sum] = "c6306f7d70457e4d8736416bb406d8670f5186fb648dd6808db24e9b2176c28d"
+
+S = "${UNPACKDIR}/qairt/${PV}"
+
+def platform_dir(d):
+    sdk_lib_dir = d.getVar("S", True) + "/lib/"
+    if os.path.exists(sdk_lib_dir) and os.path.isdir(sdk_lib_dir):
+        dir_prefix = "aarch64-oe-linux-gcc"
```
koenkooi, line 27: Don't hardcode TARGET_VENDOR  to '-oe' here, it will break on any distro that sets it.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,66 @@
+LICENSE = "Proprietary"
+LIC_FILES_CHKSUM = "file://LICENSE.pdf;md5=878b885995f453e328edbcd5a1302306"
+
+SUMMARY          = "QAIRT-SDK"
```
koenkooi: No fancy alignment please, single spaces only for these variables

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,66 @@
...
+    install -d ${D}/${libdir}/rfsa/adsp
+    install -d ${D}/${bindir}
+
+    cp -r ${S}/include/* ${D}/${includedir}
+    cp -r ${S}/lib/${PLATFORM_DIR}/* ${D}/${libdir}
+    cp -r ${S}/lib/${HEXAGON_DIR}/unsigned/* ${D}/${libdir}/rfsa/adsp
+    cp -r ${S}/bin/${PLATFORM_DIR}/* ${D}/${bindir}
+}
+
+INHIBIT_PACKAGE_STRIP = "1"
```
koenkooi, line 77: Add a comment why these are needed.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,66 @@
...
+INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
+
+INSANE_SKIP:${PN} += "already-stripped"
+INSANE_SKIP:${PN}-dev += "already-stripped"
+INSANE_SKIP:${PN}-dbg += "already-stripped"
+
+INSANE_SKIP_${PN} += "arch"
+
+SOLIBS = ".so"
```
koenkooi, line 83: This and all the lines below need extensive comments in the recipe itself explaining why the defaults don't work.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,66 @@
+LICENSE = "Proprietary"
```
lumag: No. Please define and/or provide more exact licence.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,66 @@
...
+            for folder in os.listdir(sdk_lib_dir):
+                if folder.startswith(pf_dir):
+                    pf_dir += "*"
+                    return pf_dir
+
+PLATFORM_DIR = "${@platform_dir(d)}"
+
+HEXAGON_VERSION = "v68"
+HEXAGON_VERSION:qcs9100 = "v73"
+HEXAGON_VERSION:qcs8300 = "v75"
```
lumag: No. Please see how it's done for FastRPC. No target-specific packages are allowed.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,66 @@
...
+HEXAGON_DIR = "hexagon-${HEXAGON_VERSION}"
+
+do_compile[noexec] = "1"
+do_package_qa[noexec] = "1"
+
+do_install() {
+    install -d ${D}/${includedir}
+    install -d ${D}/${libdir}/rfsa/adsp
```
lumag: This dir is going to be ignored by FastRPC daemon

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,66 @@
...
+}
+
+INHIBIT_PACKAGE_STRIP = "1"
+INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
+
+INSANE_SKIP:${PN} += "already-stripped"
+INSANE_SKIP:${PN}-dev += "already-stripped"
+INSANE_SKIP:${PN}-dbg += "already-stripped"
+
+INSANE_SKIP_${PN} += "arch"
```
lumag: Split Hexagon to separate packages. Moreover this line won't work, if I'm not mistaken.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: The folder names such as aarch64-oe-linux-gcc* come directly from the upstream SDK ZIP.
These directories are not generated by us and are not derived from TARGET_VENDOR.
They are fixed names in the binary SDK delivered externally.
The SDK ships multiple target‑specific folders, for example:

aarch64-android
aarch64-oe-linux-gcc8.2
aarch64-oe-linux-gcc9.3
aarch64-oe-linux-gcc11.2
aarch64-ubuntu-gcc9.4
aarch64-windows-msvc
android
arm64x-windows-msvc
x86_64-linux-clang
x86_64-windows-msvc

Since these names are SDK‑defined, the recipe must select and install the correct library folder based on the target we are building for.

What approach would you recommend for selecting the appropriate SDK directory based on the target?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: In our case, the SDK we are integrating ships multiple prebuilt, mutually exclusive, target‑specific library directories, such as:
hexagon-v66
hexagon-v68
hexagon-v69
hexagon-v73
hexagon-v75
hexagon-v79
hexagon-v81

Since only one of these is valid for the target we are building for, the recipe does need some logic to select the correct directory otherwise we install the wrong binaries.

Given this, could you please suggest what the recommended approach should be for selecting the correct SDK directory?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: The SDK can be downloaded from:
https://softwarecenter.qualcomm.com/api/download/software/sdks/Qualcomm_AI_Runtime_Community/All/2.41.0.251128/v2.41.0.251128.zip

The corresponding license file is located at:
${UNPACKDIR}/qairt/${PV}/LICENSE.pdf

Given that the license is explicitly provided within the SDK, what would you recommend using as the LICENSE value instead of Proprietary?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
lumag: As I wrote, please see how it's handled by the FastCV

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
lumag: Define new non-generic licence type, see `NO_GENERIC_LICENSE`. Also please add licence in the text format.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
lumag: Ugh, In my first message I wrote FastRPC instead of FastCV. Sorry for the confusion.

tjamula: https://github.com/qualcomm-linux/meta-qcom/blob/3cfd3029b83c7bf9e32a24ea89373e700e8c40ea/recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.4.bb#L55

In the FastCV recipe, they create several target‑specific packages and assign the corresponding files to each one. Can you confirm whether this same approach is expected for the qairt hexagon libraries as well?

lumag: Yes. We are working with FastRPC to implement loading of system-specific Hexagon libraries in a generic way, but until that's implemented the set of small packages is a correct way to handle QAIRT.

tjamula: Each target has its [corresponding hexagon architecture](https://docs.qualcomm.com/doc/80-63442-10/topic/QNN_general_overview.html#supported-snapdragon-devices) and the libs should be installed based on that.

So, i was planning to package these libs based on the Hexagon architecture and have the install directories and package names based on the hexagon architecture.

But i’ve noticed that in fastcv and the [hexagon-dsp-binaries](https://github.com/qualcomm-linux/meta-qcom/blob/3cfd3029b83c7bf9e32a24ea89373e700e8c40ea/recipes-bsp/hexagon-dsp-binaries/hexagon-dsp-binaries_20260110.bb) the package names and installation paths include the target device names.

Should I follow the exact same convention of using the target‑based install paths and package naming, or is it preferable to adopt the Hexagon‑architecture‑based packaging approach instead?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: The current SDK does not include the license in a plain text format. I’ve raised this issue with the QAIRT team, but even if they decide to add a .txt version, it will only come in their next release, which may take 1–2 months.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: Updated with the comments.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: Updated with the comments.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: Will update this based on the packaging of the hexagon libs.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
ricardosalveti: This license seems to be AI specific, so won't be LICENSE.qcom and LICENSE.qcom-2, needs to be yet another license for qualcomm, specific to AI.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,69 @@
+LICENSE = "Proprietary"
+LIC_FILES_CHKSUM = "file://LICENSE.pdf;md5=878b885995f453e328edbcd5a1302306"
+
+SUMMARY = "QAIRT-SDK"
```
ricardosalveti: This is not really a summary, please extend.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
lumag: So, what happens with your approach next time the customer asks to have a single rootfs which works on a variety of devices (they do)?
So, it's not possible to have just a single set of Hexagon libs in a rootfs. We are trying to solidify a good solution for the co-installing device-independent libs. However for now the best course of action is to install those libs into a device-specific location (as it will be looked into by the FastRPC anyway).

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,69 @@
...
+
+INSANE_SKIP_${PN} += "arch"
+
+# The SDK ships unversioned .so files, so keep them in the main package
+# and treat plain .so as the library (avoid -dev splitting).
+SOLIBS = ".so"
+FILES_SOLIBSDEV = ""
+
+# Include non standard library paths.
+FILES:${PN} += "${libdir}/*"
```
anujm1: This isn't non-standard library path.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,69 @@
...
+    cp -r ${S}/bin/${PLATFORM_DIR}/* ${D}/${bindir}
+}
+
+# SDK ships already-stripped proprietary binaries
+# which need not be re-striped or split into debug symbols.
+# So, disable strip, debug split and mute 'already-stripped' warnings.
+INHIBIT_PACKAGE_STRIP = "1"
+INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
+INSANE_SKIP:${PN} += "already-stripped"
+INSANE_SKIP:${PN}-dev += "already-stripped"
```
anujm1: What is in -dev that is already stripped?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
anujm1: I think the description needs to move to summary and that part needs to be more descriptive ...

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: This ensures that the Hexagon libraries installed under ${libdir}/rfsa/adsp are packaged.

The final install path for these libraries is still under discussion. If we later move to
device‑specific packages with their own Hexagon library sets, the corresponding FILES
entries will be added there and this entry can be removed.

lumag: I'm sorry, it's not. `${libdir}/rfsa/adsp` is unsupported and must be dropped.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: I have updated the summary and description.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: dev package only contains the header files and the dbg is empty. Removed the suppression for dev and dbg packages.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: I’ve updated the recipe so that each Hexagon architecture has its own package. My suggestion was to follow the same approach - provide one package per architecture instead of one per target, since multiple targets can share the same Hexagon architecture. The DSP could then simply pick the correct directory based on the architecture version. This also makes the machine configuration cleaner, allowing us to recommend the appropriate architecture-specific package.

For example, in this:
[iq-9075-evk.conf](https://github.com/qualcomm-linux/meta-qcom/blob/18c6e3f42ddb07f4b4a87f44324b3fb60e6c3a2e/conf/machine/iq-9075-evk.conf#L18)
the package we would add could be qairt-sdk-hexagon-v73.

The open question now is whether we can add the required architecture‑specific path to the relevant PATH variable and if the recommended locations are already fixed, which specific paths should be used for the specific device.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: Could you suggest a suitable non-generic license to use here? Let me know if you need more details about the license.

tjamula: Are you suggesting that I add the SDK’s license file to meta-qcom/licenses as well?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,86 @@
...
+    ${PN}-hexagon-v75 \
+"
+
+FILES:${PN}-hexagon-v68 += "${datadir}/qcom/dsp/cdsp/v68"
+FILES:${PN}-hexagon-v73 += "${datadir}/qcom/dsp/cdsp/v73"
+FILES:${PN}-hexagon-v75 += "${datadir}/qcom/dsp/cdsp/v75"
+
+RDEPENDS:{PN}-hexagon-v68 = "{PN}"
+RDEPENDS:{PN}-hexagon-v73 = "{PN}"
+RDEPENDS:{PN}-hexagon-v75 = "{PN}"
```
anujm1: Missing `$`

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: Discussed with @ekanshibu from fastrpc team and updated the install paths accordingly.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
lumag: No. See how others do it.

tjamula: https://github.com/qualcomm-linux/meta-qcom/tree/18c6e3f42ddb07f4b4a87f44324b3fb60e6c3a2e/licenses

LICENSE.qcom and LICENSE.qcom-2 do not match with the LICENSE used in the SDK.

Please share the relevant recipe to refer to, if available.

ricardosalveti: Since this is the first recipe using this new qcom AI license, use NO_GENERIC_LICENSE as part of this recipe for now. Later, if we have other recipes reusing the same license, we could add the license (in text format) in meta-qcom/licenses.

tjamula: Added the license as a NO_GENERIC_LICENSE

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,88 @@
...
+            pf_dir = dir_prefix + version
+            for folder in os.listdir(sdk_lib_dir):
+                if folder.startswith(pf_dir):
+                    pf_dir += "*"
+                    return pf_dir
+
+PLATFORM_DIR = "${@platform_dir(d)}"
+
+do_compile[noexec] = "1"
+do_package_qa[noexec] = "1"
```
anujm1: why?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,88 @@
...
+    cp -r ${S}/lib/hexagon-v75/unsigned/* ${D}${datadir}/qcom/qcs8300/Qualcomm/QCS8300-RIDE/dsp/cdsp
+    cp -r ${S}/bin/${PLATFORM_DIR}/* ${D}${bindir}
+}
+
+# SDK ships already-stripped proprietary binaries
+# which need not be re-striped or split into debug symbols.
+# So, disable strip, debug split and mute 'already-stripped' warnings.
+INHIBIT_PACKAGE_STRIP = "1"
+INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
+INSANE_SKIP:${PN} += "already-stripped"
```
anujm1, line 79: Is this needed even after setting `INHIBIT_PACKAGE_STRIP`?

tjamula: Yes, otherwise I encounter numerous warnings while building the recipe.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: Would you suggest having the following instead?

(Facing a QA error:
ERROR: qairt-sdk-2.41.0.251128-r0 do_package_qa: QA Issue: /usr/lib/libhta_hexagon_runtime_snpe.so contained in package qairt-sdk requires libcdsprpc.so()(64bit), but no providers found in RDEPENDS:qairt-sdk? [file-rdeps]
)
libcdsprpc.so is being packed as part of fastrpc package but this is not being resolved even when i add fastrpc to the RDEPENDS:${PN} += "fastrpc". Hence the below skip.
+INSANE_SKIP:${PN} += "file-rdeps"

+# Skip QA checks that don’t apply to prebuilt Hexagon DSP/HTP libraries.
+INSANE_SKIP:${PN}-hexagon-v68 += "arch libdir ldflags file-rdeps"
+INSANE_SKIP:${PN}-hexagon-v73 += "arch libdir ldflags file-rdeps"
+INSANE_SKIP:${PN}-hexagon-v75 += "arch libdir ldflags file-rdeps"

anujm1: Yeah, they should be separately explained.

`RDEPENDS` isn't working in this case because `libhta_hexagon_runtime_snpe.so` has:
```
0x0000000000000001 (NEEDED)             Shared library: [libcdsprpc.so]
```

while the soname for fastrpc library is `libcdsprpc.so.1`. 
```
 0x000000000000000e (SONAME)             Library soname: [libcdsprpc.so.1]
```

The libraries also leak build host path in `RPATH` that should also be fixed.

tjamula: The QAIRT SDK is maintained by another team, so I’ll forward this feedback to them. In the meantime, is there a recommended workaround we can rely on until they address these issues in a future release?

lumag: Add necessary INSANE_SKIP:foo entries, add comments, drop them in the next release.

tjamula: Added the entries with comments.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,97 @@
+LICENSE = "qcom-ai-stack"
+LIC_FILES_CHKSUM = "file://LICENSE.pdf;md5=878b885995f453e328edbcd5a1302306"
+NO_GENERIC_LICENSE[qcom-ai-stack] = "LICENSE.pdf"
+
+
```
ricardosalveti: Please move the license information after HOMEPAGE.

conf/machine/iq-8275-evk.conf
```diff
@@ -19,6 +19,7 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-iq-8275-evk-firmware \
     packagegroup-iq-8275-evk-hexagon-dsp-binaries \
     qcom-fastcv-binaries-qcs8300-ride-dsp \
+    qairt-sdk-hexagon-v75 \
```
ricardosalveti, line 21: SDK runtime as MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS? Shouldn't this be selected by the distro instead?

lumag: It's not the full runtime, just the Hexagon skel libs. We don't have a good way to pull them otherwise.

lumag: It will be (hopefully) sorted out once we sort out system lib dirs for the FastRPC.

koenkooi: Should we have a `HEXAGON_FAMILY` in our includes, e.g. `HEXAGON_FAMILY = "v75"` to make things like this easier and less repetitive?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,96 @@
...
+LICENSE = "qcom-ai-stack"
+LIC_FILES_CHKSUM = "file://LICENSE.pdf;md5=878b885995f453e328edbcd5a1302306"
+NO_GENERIC_LICENSE[qcom-ai-stack] = "LICENSE.pdf"
+
+SRC_URI = "https://softwarecenter.qualcomm.com/api/download/software/sdks/Qualcomm_AI_Runtime_Community/All/${PV}/v${PV}.zip"
+SRC_URI[sha256sum] = "c6306f7d70457e4d8736416bb406d8670f5186fb648dd6808db24e9b2176c28d"
+
+S = "${UNPACKDIR}/qairt/${PV}"
+
+RDEPENDS:${PN} += "fastrpc"
```
koenkooi: R* variables go below do_install since they are packaging variables, not build variables.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,96 @@
...
+
+def platform_dir(d):
```
koenkooi, line 24: This needs a comment explaining that the archive has a bunch or folders labeled after the compiler and we need to pick one.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,96 @@
...
+    install -d ${D}${includedir}
+    install -d ${D}${libdir}
+    install -d ${D}${datadir}/qcom/qcm6490/Thundercomm/RB3gen2/dsp/cdsp
+    install -d ${D}${datadir}/qcom/sa8775p/Qualcomm/SA8775P-RIDE/dsp/cdsp
+    install -d ${D}${datadir}/qcom/qcs8300/Qualcomm/QCS8300-RIDE/dsp/cdsp
+    install -d ${D}${bindir}
+
+    cp -r ${S}/include/* ${D}${includedir}
+    cp -r ${S}/lib/${PLATFORM_DIR}/* ${D}${libdir}
+    cp -r ${S}/lib/hexagon-v68/unsigned/* ${D}${datadir}/qcom/qcm6490/Thundercomm/RB3gen2/dsp/cdsp
```
koenkooi, line 64: This also needs a comment explaining why the files go from generic to very specific. For example, where does the rubikpi3 board need to get the files from?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,96 @@
...
+            for folder in os.listdir(sdk_lib_dir):
+                if folder.startswith(pf_dir):
+                    pf_dir += "*"
+                    return pf_dir
+
+PLATFORM_DIR = "${@platform_dir(d)}"
+
+do_compile[noexec] = "1"
+
+# Within our builds we currently use and test it only on ARMv8 (aarch64) machines.
```
koenkooi: This comment heavily implies this recipes actually builds something, which it doesn't. Please state clearly that the selection for PLATFORM_DIR is aarch64 only and hence the recipe needs this COMPATIBLE_MACHINE>

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,96 @@
...
+INHIBIT_PACKAGE_STRIP = "1"
+INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
+INSANE_SKIP:${PN} += "already-stripped"
+
+# The SDK ships unversioned .so files, so keep them in the main package
+# and treat plain .so as the library (avoid -dev splitting).
+SOLIBS = ".so"
+FILES_SOLIBSDEV = ""
+
+PACKAGES += "\
```
koenkooi, line 86: Coming back to my earlier question about moving from generic to specific, why do these packages have such a generic name when their contents are very much not?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,96 @@
...
+RDEPENDS:${PN}-hexagon-v68 = "${PN}"
+RDEPENDS:${PN}-hexagon-v73 = "${PN}"
+RDEPENDS:${PN}-hexagon-v75 = "${PN}"
+
+# Skip QA checks that don’t apply to prebuilt Hexagon DSP/HTP libraries.
+INSANE_SKIP:${PN}-hexagon-v68 += "arch libdir ldflags file-rdeps"
+INSANE_SKIP:${PN}-hexagon-v73 += "arch libdir ldflags file-rdeps"
+INSANE_SKIP:${PN}-hexagon-v75 += "arch libdir ldflags file-rdeps"
+
+FILES:${PN}-hexagon-v68 += "${datadir}/qcom/qcm6490/Thundercomm/RB3gen2/dsp/cdsp"
```
koenkooi: Move this between the PACKAGES and RDEPENDS sections.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: These installation paths were decided based on the recommendations from fastrpc team and taking [fastcv](https://github.com/qualcomm-linux/meta-qcom/blob/master/recipes-multimedia/fastcv/qcom-fastcv-binaries_1.8.4.bb) as a reference. These are additional references that were shared with me:
https://github.com/linux-msm/hexagon-dsp-binaries/blob/trunk/config.txt
https://github.com/linux-msm/hexagon-dsp-binaries/blob/trunk/00-hexagon-dsp-binaries.yaml

What would you suggest having as a comment?

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: Please refer to this conversation on this for more details: https://github.com/qualcomm-linux/meta-qcom/pull/1412#discussion_r2712313959

These packages use a generic name because their contents are generic - they only depend on the Hexagon architecture version, not on the specific target device. Keeping the naming generic also makes the machine configuration easier to read.

Yes, the installation paths themselves are not generic. The FastRPC team is considering moving these paths to be architecture‑specific instead of target‑specific, but that change is still under discussion.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: Updated the comment.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: I removed the old comment inside the function and added a clearer one here.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
koenkooi: Something like:

> These installation paths were decided based on the recommendations from fastrpc team and taking fastcv as a reference. They might change in the future, so keep the entries in PACKAGES below generic.

tjamula: Added a similar comment before the installation.

koenkooi (dismissed):
All my feedback has been either addressed or rendered moot, so PR approved.

lumag (dismissed):

conf/machine/iq-8275-evk.conf
ndechesne, line 21: please keep list sorted.

tjamula:
> Hmm why no -dev package and versioned SONAMEs?

 Currently, the SDK ships only unversioned .so libraries and only the header files are packed in the -dev package.

tjamula:
> Please rebase once #1443 is merged.

Is it required? The changes are independent and touch different files.

vkraleti:
> > Please rebase once #1443 is merged.
> 
> Is it required? The changes are independent and touch different files.

GitHub PR workflows run against the pull_request branch’s head commit, not the latest tip of the base branch. If a PR is created on top of an unstable base, the workflows will continue to fail because they do not automatically pick recent fixes from the base branch. In this case, the RPM package build issues already exist on tip. Rebasing the PR on top of https://github.com/qualcomm-linux/meta-qcom/pull/1443 is required to pick up those fixes and resolve the build failures.

tjamula:
> > > Please rebase once #1443 is merged.
> > 
> > 
> > Is it required? The changes are independent and touch different files.
> 
> GitHub PR workflows run against the pull_request branch’s head commit, not the latest tip of the base branch. If a PR is created on top of an unstable base, the workflows will continue to fail because they do not automatically pick recent fixes from the base branch. In this case, the RPM package build issues already exist on tip. Rebasing the PR on top of #1443 is required to pick up those fixes and resolve the build failures.

Rebase done.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -0,0 +1,104 @@
...
+    ${PN}-hexagon-v73 \
+    ${PN}-hexagon-v75 \
+"
+
+FILES:${PN}-hexagon-v68 += "${datadir}/qcom/qcm6490/Thundercomm/RB3gen2/dsp/cdsp"
+FILES:${PN}-hexagon-v73 += "${datadir}/qcom/sa8775p/Qualcomm/SA8775P-RIDE/dsp/cdsp"
+FILES:${PN}-hexagon-v75 += "${datadir}/qcom/qcs8300/Qualcomm/QCS8300-RIDE/dsp/cdsp"
+
+RDEPENDS:${PN} += "fastrpc"
+RDEPENDS:${PN}-hexagon-v68 = "${PN}"
```
lumag: No. This will pull this package into each and every install due to machines pulling the DSP parts. Please drop does from the hexagon packages onto the main package

tjamula: Should I remove it entirely, or should I instead add the Hexagon packages to ${PN}’s RDEPENDS to preserve the dependency, i.e.,
`RDEPENDS:${PN} += "${PN}-hexagon-v68 ${PN}-hexagon-v73 ${PN}-hexagon-v75"`

lumag: You can RRECOMMEND installation of hexagon libs from PN

tjamula: Done

tjamula:
> > nothing provides libc++.so.1 needed by qairt-sdk-hexagon-v68-2.41.0.251128-r0.armv8_2a from oe-repo
> 
> You need to add `SKIP_FILEDEPS:qairt-sdk-hexagon-foo = "1"` for all your Hexagon packages.

Done

lumag (approved):

tjamula:
https://github.com/qualcomm-linux/meta-qcom/actions/runs/21462905552/job/61864160919?pr=1412#step:3:19069

The error does not seem related to the change made.

Issue #1277, opened 2025-12-03
SoC specific kernel cmdline for workarounds


pavankondeti:
The pointer authentication feature (PAuth) is only  supported on
0-3 CPUs but it is not supported on 4-7 CPUS on Monaco (QCS8300).
The ARM64  cpufeature discovery code expects `late` CPUs to have
this feature if `boot` CPU feature has it since PAuth is enabled
early. When a conflict like this is detected, the `late` CPUs are
not allowed to boot.[^1] It is expected that system will continue
to be functional with CPUs with Pauth feature supported and enabled.

`arm64.nopauth` can be passed on kernel cmdline to disable this
feature. This will allow us to boot with all CPUs on Moncao. This has
been our configuration on Qualcomm Linux 1.0 

[^1] : https://elixir.bootlin.com/linux/v6.18-rc6/source/arch/arm64/kernel/cpufeature.c#L3579

pavankondeti:
+ @vkraleti , @shashim-quic

pavankondeti:
> upstream workaround which disables ptr auth for cases such as monaco without any cmdline addition, done by the kernel at runtime

The upstream solution to workaround this problem is `arm64.nopauth` cmdline param. The Pointer Authentication must be enabled early (from a function which never return) i.e before SMP, so the expecation is that boot core decide if the feature can be enabled or not. if `late` cores does not support, they are not allowed to boot. 

if we have to go to upstream for a different solution, I feel we need to give justificaiton on why we can't use kenrel cmdline? Are we saying that UEFI / firmware / Yocto (any distro) can't pass custom kernel cmdline, so this must be handled in kernel? I don't know what would be a solution to this problem? We can't even claim that this is an errata because the boot cluster CPU do support pauth, there is no such thing as SoC errata wrt ARM64 CPU feature detection. It would really be weird to introduce a device tree param (other than bootargs) to specify this quirk.

pavankondeti:
ARM64 maintainers think passing `arm64.nopauth` is the way to fix this problem specific to QCS8300 SoC

ricardosalveti:
> ARM64 maintainers think passing `arm64.nopauth` is the way to fix this problem specific to QCS8300 SoC

[...]

PR #1416, opened 2026-01-21
In CI make KVM builds only on IQ-[8275|9075]-EVK machines

vkraleti:
Due to CI bandwidth constraints, limit KVM builds to IQ-8275-EVK and IQ-9075-EVK
machines, as active testing is currently planned only on these boards. Additional 
machines can be added later as test coverage expands.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -81,8 +81,6 @@ jobs:
             yamlfile: ""
           - name: qcom-distro
             yamlfile: ':ci/qcom-distro-prop-image.yml'
-          - name: qcom-distro-kvm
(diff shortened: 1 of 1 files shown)


lumag:
I see `xbl_config_kvm.elf` in QCS8300 binaries, so KVM is also supported for QCS8300 devices.

vkraleti:
Eventually, KVM will be supported on all boards including QCS8300. However, as of today, only on QCS9100 all multimedia use cases are fully functional. on QCS8300 device won't even boot with KVM ( issue #1277 ) So generating images for the other boards at this stage only adds unnecessary load on CI. If that is not a concern I'll go ahead and make KVM builds for all.

PR #1418, opened 2026-01-21
Introduce weekly build workflow to reduce load on nightly 

vkraleti:
Currently, some of the builds produced by the nightly workflow are not
consumed on a daily basis. Introduce the `weekly-build` workflow to
handle builds that are needed less frequently, reducing the load on
nightly builds.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -3,6 +3,11 @@ name: Build Yocto
 on:
   workflow_call:
     inputs:
+      mode:
+        description: "daily or weekly"
+        required: false
+        type: string
+        default: "daily"
       sdk:
         required: false
         type: string
@@ -22,6 +27,14 @@ env:
   KAS_CLONE_DEPTH: 1
 
 jobs:
+  validate-inputs:
+    runs-on: self-hosted
+    steps:
+      - run: |
+          case "${{ inputs.mode }}" in
+            daily|weekly) echo "Running in ${{ inputs.mode }} mode" ;;
+            *) echo "Invalid mode: ${{ inputs.mode }}. Use daily|weekly." && exit 1 ;;
+          esac
   kas-setup:
     if: github.repository_owner == 'qualcomm-linux'
     runs-on: [self-hosted, qcom-u2404, amd64]
@@ -85,9 +98,9 @@ jobs:
         run: |
           ci/kas-container-shell-helper.sh ci/yocto-patchreview.sh
 
-  compile_warm_up:
+  compile_daily_warm_up:
     needs: [kas-setup, yocto-run-checks]
-    if: github.repository_owner == 'qualcomm-linux'
+    if: ${{ github.repository_owner == 'qualcomm-linux' && inputs.mode == 'daily' }}
     runs-on: [self-hosted, qcom-u2404, amd64]
     strategy:
       fail-fast: true
@@ -102,8 +115,6 @@ jobs:
             yamlfile: ':ci/qcom-distro-prop-image.yml'
           - name: qcom-distro-catchall
             yamlfile: ':ci/qcom-distro-catchall.yml'
-          - name: performance
-            yamlfile: ':ci/qcom-distro-prop-image.yml:ci/performance.yml'
         kernel:
           - type: default
             dirname: ""
@@ -111,14 +122,11 @@ jobs:
           - type: 6.18
             dirname: "+linux-qcom-6.18"
             yamlfile: ":ci/linux-qcom-6.18.yml"
-          - type: rt-6.18
-            dirname: "+linux-qcom-rt-6.18"
-            yamlfile: ":ci/linux-qcom-rt-6.18.yml"
     name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
     steps:
       - uses: actions/checkout@v4
 
-      - name: Run kas build
+      - name: Run daily warmup kas build
         uses: ./.github/actions/compile
         with:
           machine: ${{matrix.machine}}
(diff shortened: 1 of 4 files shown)


.github/workflows/build-yocto-weekly.yml
```diff
@@ -0,0 +1,232 @@
+name: Build Yocto Weekly
```
ndechesne: what are the differences with build-yocto? there is a lot of duplication here and it will cause major pain to maintain I think. Can we not find a better way , maybe with input/params?

vkraleti: Primarily want to move kvm, sota and rt kernel builds out of nightly to a less frequent build cadence. 

Agree there is lot of duplication, hence RFC, will cleanup in next push.

.github/workflows/build-yocto.yml
```diff
@@ -212,7 +248,8 @@ jobs:
           sdk: ${{inputs.sdk}}
 
   publish_summary:
-    needs: compile
+    needs: [compile_daily, compile_weekly]
```
ndechesne, line 336: this is convoluted... can you please explain here and/or in the commit why we don't want the warmup for weekly build?

vkraleti: Added warmup jobs for weekly as well.

.github/workflows/build-yocto.yml
```diff
@@ -17,6 +22,14 @@ env:
   KAS_REPO_REF_DIR: /efs/qli/meta-qcom/kas-mirrors
 
 jobs:
+  validate-inputs:
+    runs-on: [self-hosted, qcom-u2404, amd64]
```
ndechesne: can we use ubuntu-latest instead? no need to spin off one of our machine here.

.github/workflows/build-yocto.yml
```diff
@@ -294,6 +272,61 @@ jobs:
...
+            kernel:
+                type: 6.18
+                dirname: "+linux-qcom-6.18"
+                yamlfile: ":ci/linux-qcom-6.18.yml"
+          - machine: iq-8275-evk
+            distro:
+                name: performance
+                yamlfile: ':ci/qcom-distro-prop-image.yml:ci/performance.yml'
+            kernel:
+                type: 6.18
```
lumag: This part can't get in, we need to be testing 6.18 on PRs that change it.

.github/workflows/build-yocto.yml
```diff
@@ -242,7 +255,50 @@ jobs:
...
+          machine: ${{matrix.machine}}
+          distro_yaml: ${{matrix.distro.yamlfile}}
+          distro_name: ${{matrix.distro.name}}
+          kernel_yaml: ${{matrix.kernel.yamlfile}}
+          kernel_dirname: ${{matrix.kernel.dirname}}
+          cache_dir: ${CACHE_DIR}
+          kas: ${KAS_CONTAINER}
+          sdk: ${{inputs.sdk}}
+
+  compile_weekly_warm_up:
```
ndechesne, line 237: where is the 'strategy' matrix for this job (and compile_weekly)?

.github/workflows/build-yocto.yml
```diff
@@ -294,6 +272,61 @@ jobs:
...
+          - kaanapali-mtp
+          - qcm6490-idp
+          - qcs615-ride
+          - qcs8300-ride-sx
+          - qcs9100-ride-sx
+          - rb1-core-kit
+          - sm8750-mtp
+        distro:
+          - name: qcom-distro-selinux
+            yamlfile: ':ci/qcom-distro-selinux.yml'
```
lumag, line 302: So... How do we make sure that SELinux doesn't regress? Especially on PRs.

.github/workflows/build-yocto.yml
```diff
@@ -272,6 +232,24 @@ jobs:
     needs: [kas-setup, yocto-run-checks]
     if: ${{ github.repository_owner == 'qualcomm-linux' && inputs.mode == 'weekly' }}
     runs-on: [self-hosted, qcom-u2404, amd64]
+    strategy:
```
ndechesne, line 241: i see, you are adding the strategy here, but that means that your initial commit is wrong with non valid job definition. perhaps you need to squash #1 and #3.

lumag (changes requested):
I think the huge build yaml should be split into reusable parts. Then those parts should be selected via some rules (e.g. weekly, when the file changes, etc.). In the end, the SELinux jobs need to be executed when somebody changes the policy. Likewise 6.18 needs to be built (and tested) for PRs which touch 6.18.

.github/workflows/build-yocto.yml
ricardosalveti: Yeah, we should only use the self-hosted when actually building the image.

.github/workflows/build-yocto.yml
ricardosalveti: yeah, add initial strategy as part of the first commit and then have this patch just removing the jobs.

.github/workflows/weekly-build.yml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  schedule:
+  # NOTE - changes to the cron spec should be pushed by https://github.com/quic-yocto-ci
+  # so that build notification emails will be sent out properly.
+  - cron: "13 13 * * 1,4"   # Monday, Thursday job - pick a random "minute"  - top of hour can be busy in github
+
+permissions:
+  checks: write
+  pull-requests: write
```
anujm1: These should ideally not be defined at workflow level. I assume this is needed for `publish-results` only.

Perhaps something that should be done in a separate PR. I can send this once all workflow PRs are merged.

PR #1419, opened 2026-01-21
Add performance build support for IQ-[8275|9075]-EVK machines

vkraleti:
Extend the GitHub workflow to build performance optimized variants for
the IQ-8275-EVK and IQ-9075-EVK machines to support performance benchmark
testing. Additional machines can be added later as test coverage expands.


lumag (changes requested):
Why can't we set it by default?

lumag (changes requested):
My opinion didn't change - `quiet` should be set in meta-qcom-distro for all boards.

PR #1420, opened 2026-01-21
Update camxlib to the latest revision

gkhose-qipl:
camxlib-kodiak: update to the latest revision.
camxlib-lemans: update to the latest revision.
introduce the development package to export necessary headers and shared objects required by GST plugin and QMMF.
create camx-nhx as separate package.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.6.bb
+++ b/dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.6.bb
@@ -67,7 +67,7 @@ do_install() {
 }
 
 RPROVIDES:${PN} = "camxlib-monaco"
-PACKAGE_BEFORE_PN += "camx-lemans chicdk-lemans"
+PACKAGE_BEFORE_PN += "camx-lemans chicdk-lemans camx-nhx"
 RDEPENDS:${PN} += "chicdk-lemans"
 
 FILES:camx-lemans = "\
@@ -90,8 +90,11 @@ FILES:chicdk-lemans = "\
     ${libdir}/camx/lemans/camera/*.bin \
     ${libdir}/camx/lemans/camera/com.qti.sensor*${SOLIBS} \
     ${libdir}/camx/lemans/hw/com.qti.chi.*${SOLIBS} \
+    ${bindir}/camx \
+    "
+FILES:camx-nhx = "\
+    ${bindir}/nhx.sh \
     ${sysconfdir}/camera/test/NHX/ \
-    ${bindir}/ \
     "
 FILES:${PN} = "\
     ${libdir}/libcamera_metadata_lemans*${SOLIBS} \


dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.6.bb
```diff
@@ -57,10 +61,12 @@ do_install() {
     install -m 0644 ${S}/usr/share/doc/${BPN}/NO.LOGIN.BINARY.LICENSE.QTI.pdf ${D}${datadir}/doc/chicdk-kodiak
 }
 
-PACKAGE_BEFORE_PN += "camx-kodiak chicdk-kodiak"
+PACKAGE_BEFORE_PN += "camx-kodiak camx-kodiak-dev chicdk-kodiak"
```
lumag: What is the difference between ${PN}-dev and camx-kodiak-dev?

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.6.bb
```diff
@@ -89,7 +100,20 @@ FILES:chicdk-kodiak = "\
...
     ${libdir}/camx/kodiak/camera/components/*${SOLIBS} \
     "
+FILES:${PN}-dev = "\
+    ${libdir}/libcamera_metadata_kodiak*${SOLIBSDEV} \
+    ${includedir}/camx-api \
+    "
+# By default, when a package contains only one shared librarie at ${libdir},
+# the Yocto/OE 'debian' class may auto-rename runtime and -dev packages based on
+# the library SONAME , instead of keeping the recipe’s ${PN} / ${PN}-dev names.
+# Disable Debian-style automatic library package renaming for this recipe.
```
lumag: You are describing known facts. Why do we want ${PN} not to be renamed?

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.6.bb
```diff
@@ -89,7 +100,20 @@ FILES:chicdk-kodiak = "\
...
+DEBIAN_NOAUTONAME:${PN} = "1"
+DEBIAN_NOAUTONAME:${PN}-dev = "1"
```
lumag: ${PN} is fine. The rest are questionable. Are they really required?

lumag (changes requested):

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.6.bb
gkhose-qipl: We have a few headers that are used from camx, and since camx is shared as a source, its headers are included in camx-kodiak-dev so they can be overridden in meta-qcom-extras.

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.6.bb
gkhose-qipl: just to avoid confusion, add all. so package are created with expected name

ricardosalveti: Should only be added if indeed an issue, don't add all just because it is more convenient.

ricardosalveti:
Can you replace 'latest' with 260120 in your commit message? Looking back at the meta-qcom git history there will be a bunch of "update to latest", where latest is just relevant at a certain point in time.

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.6.bb
lumag: There is no distinction for now. And bbappend in meta-qcom-extras could override headers in the unified `${PN}-dev` package.

gkhose-qipl: One issue that i foresee using bbappend is catering to two types of users:
1) those who receive only chicdk as a source, and
2) those who receive both chicdk and camx as sources.

lumag: I don't see an issue. Even if you use two recipes instead of bbappend, you still have to identify whether your user has access to chicdk or camx sources.

gkhose-qipl: So, if I understand correctly, we only need ${PN}-dev, and everything else can be managed in meta-qcom-extras.

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.6.bb
gkhose-qipl: sure, will keep only ${PN}

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.6.bb
gkhose-qipl: can we add below comment instead of above ?
"Disable Debian-style automatic library package renaming.
 Even though this recipe installs a shared library, we want to keep the
 package names based on ${PN} instead of allowing auto-renaming to lib<name>.
This ensures stable package names for dependencies, feeds, and upgrade paths,
and avoids breaking recipes or components that explicitly RDEPENDS on ${PN}"

anujm1: It should already be doing a `RPROVIDES` on `PN` so this shouldn't be causing failure.

gkhose-qipl: > It should already be doing a `RPROVIDES` on `PN` so this shouldn't be causing failure.

did get, do we need to remove "DEBIAN_NOAUTONAME:${PN} = "1""?

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.6.bb
gkhose-qipl: > I don't see an issue. Even if you use two recipes instead of bbappend, you still have to identify whether your user has access to chicdk or camx sources.

We set BBMASK as the default for the camx recipe.

lumag: And how `BBMASK` is changed? Manually by the user?

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.6.bb
lumag: You don't _need_. But if you want to have it, you have to explain, why.

gkhose-qipl: > You don't _need_. But if you want to have it, you have to explain, why.

added below line in comment
This ensures stable package names for dependencies and avoids breaking recipes or components that explicitly RDEPENDS on ${PN}.

lumag: Could you please _drop_ the obvious parts - the description of package renamings, etc. Please keep only the sensible part of the comment.

anujm1: I read the comment as "did get" .. My point was `RDEPENDS` should not be failing since I thought renaming class adds a `RPROVIDES` on original package name. So, the comment wasn't entirely accurate. So, is this being done because `RDEPENDS` is failing or we want consistent name irrespective of whether it has the library or not?

gkhose-qipl: > I read the comment as "did get" .. My point was `RDEPENDS` should not be failing since I thought renaming class adds a `RPROVIDES` on original package name. So, the comment wasn't entirely accurate. So, is this being done because `RDEPENDS` is failing or we want consistent name irrespective of whether it has the library or not?

It's true that it wasn't failing, but we want to keep the name consistent since we usually only side load camx ipk. The name change caused issues during internal testing (sideloading of camxlib ipk was failed), but after adding the flag, the issue was resolved. will update comment accordingly.

gkhose-qipl: @lumag / @anujm1 can i add below line in comment?
" Preserve ${PN} naming for consistent sideloading, as auto-renaming caused failures."

lumag: Which failures?

gkhose-qipl: > Which failures?
file not found. 

can we update.
" Preserve ${PN} naming for consistent sideloading, as auto-renaming led to file not found errors."

gkhose-qipl: @lumag / @anujm1 ,
can we update below comment
" Preserve ${PN} naming to maintain consistency across customer documentation and avoid confusion with respect to package name."

lumag: You still have obvous description of renaming (why? I asked to drop it). And still doesn't describe why do you want the stable naming. Not to mention that the line is too long.

gkhose-qipl: updated below comment.
Preserve ${PN} naming to avoid ambiguity in package identification.

PR #1421, opened 2026-01-21
Revert "qcs615-ride.conf: stop setting QCOM_DTB_DEFAULT"

lumag:
This reverts commit b6d483f9f6032b9d5e20d608f53251a90bc298ee as the QCS615 Ride boards are not yet supported by the multi-DTB image.

### DIFF
--- a/conf/machine/qcs615-ride.conf
+++ b/conf/machine/qcs615-ride.conf
@@ -6,6 +6,8 @@ require conf/machine/include/qcom-qcs615.inc
 
 MACHINE_FEATURES += "efi pci"
 
+QCOM_DTB_DEFAULT ?= "qcs615-ride"
+
 KERNEL_DEVICETREE ?= " \
                       qcom/qcs615-ride.dtb \
                       "


vkraleti:
I am actively working with @mwasilew and @smuppand to replace the unsupported Talos devices in test bed. Post that this change is not needed.

lumag:
> I am actively working with @mwasilew and @smuppand to replace the unsupported Talos devices in test bed. Post that this change is not needed.

Will QCS615 Ride suddenly get support from the DTB metadata? How can it be "not needed"?

vkraleti:
> > > > I am actively working with @mwasilew and @smuppand to replace the unsupported Talos devices in test bed. Post that this change is not needed.
> > > 
> > > 
> > > Will QCS615 Ride suddenly get support from the DTB metadata? How can it be "not needed"?
> > 
> > 
> > When a board is referred as **QCS615‑Ride**, it does not mean that all such boards have the same SoC. There are several Ride board variants built with different parts like SA6145P, SA4155P, SA6155, and possibly more. These are automotive parts and are non‑POR for the QLI program. Some of these non‑POR boards are currently being used by the test teams. The request is to replace them with the correct boards, whose SoC_ID is 0x2A8, which has support in [qcom-dtb-metadata](https://github.com/qualcomm-linux/qcom-dtb-metadata)
> 
> For all targets we have (or plan to have) two kinds of board: the older Ride and newer IQ devices. Old boards are not supported by the qcom-dtb-metadata, so it was a mistake to make qcs615-ride use multi-dtb image. This PR restores status quo: it makes QCS615 Ride work again. It has (almost) nothing to do with the boards in our lab: even if we had none and we got a similar bug report, we should have reverted the offending commit. I don't understand, what and why we are discussing here.

AFAIK, sticking to a single DTB on older boards was never intended. Multi‑DTB is mandated for all boards to ensure that CAMX, Display, and other planned overlays are exercised as expected. Even on QCS8300/9100 Ride boards, once CI has the correct set of devices, we will need to switch back to multi‑DTB.

@sbanerjee-quic @shashim-quic @quic-kaushalk can you confirm?

PR #1423, opened 2026-01-22
Use CodeLinaro MIRRORS as a backup.

vkraleti:
To reduce git repo fetch failures in builds, add CodeLinaro MIRRORS
as a backup to provide fallback Git sources.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -43,7 +43,12 @@ local_conf_header:
       HALT,${DL_DIR},100M,1K \
       HALT,${SSTATE_DIR},100M,1K \
       HALT,/tmp,10M,1K"
-
+  clo-mirrors: |
+    MIRRORS:append = " \
+    git://github.com git://git.codelinaro.org/clo/yocto-mirrors/github/ \
+    git://.*/.*/ git://git.codelinaro.org/clo/yocto-mirrors/ \
+    https://.*/.*/ https://codelinaro.jfrog.io/artifactory/codelinaro-le/ \
+    "
   cmdline: |
     KERNEL_CMDLINE_EXTRA:append = " qcom_scm.download_mode=1"
   qcomflash: |


ci/base.yml
```diff
@@ -43,7 +43,12 @@ local_conf_header:
       HALT,${DL_DIR},100M,1K \
       HALT,${SSTATE_DIR},100M,1K \
       HALT,/tmp,10M,1K"
-
+  clo-mirrors: |
+    MIRRORS:append = " \
```
ricardosalveti, line 47: We probably don't need append here, += should work.

I was wondering if PREMIRRORS would be better, but I guess the need for the mirrors are not necessarily for everyone.

vkraleti: Accessing CLO mirrors have cost associated. So want to hit them only when needed but not for every build. Hence added to MIRRORS instead of PREMIRRORS.

ci/base.yml
```diff
@@ -43,7 +43,12 @@ local_conf_header:
...
+    git://github.com git://git.codelinaro.org/clo/yocto-mirrors/github/ \
+    git://.*/.*/ git://git.codelinaro.org/clo/yocto-mirrors/ \
+    https://.*/.*/ https://codelinaro.jfrog.io/artifactory/codelinaro-le/ \
```
craigez, line 50: We shouldn't use the jfrog.io URL since it is tied to a specific underlying service. Instead we should use artifacts.codelinaro.org, since that can be redirected if the underlying service changes. (e.g. https://artifacts.codelinaro.org/codelinaro-le/). We will need to update this more broadly in other places we have used the same URL in coming months, final details TBD.

craigez: See https://github.com/qualcomm-linux/meta-qcom/pull/1426

PR #1430, opened 2026-01-23
ci: actions/compile: reduce files duplication on upload

ricardosalveti:
Instead of copying every file available in the deploy folder, copy instead just the final image links created by bitbake plus additional selected files which are not linked.

This should help reducing the load on the upload action as it will also reduce the amount of file duplication we currently see, as we were copying the links and the original files containing the build timestamp.

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -100,8 +100,11 @@ runs:
         deploy_dir=../kas/build/tmp/deploy/images/${{inputs.machine}}
         uploads_dir=./uploads/${{ inputs.distro_name }}${{ inputs.kernel_dirname }}/${{ inputs.machine }}
         mkdir -p $uploads_dir
-        find $deploy_dir/ -maxdepth 1 -type f -exec cp {} $uploads_dir/ \;
-        find $deploy_dir/ -maxdepth 1 -type l \( -name boot-*.img -o -name *.rootfs.ext4.gz -o -name *.rootfs.qcomflash.tar.gz -o -name *.rootfs.ostreecommit.tar.xz \) -exec cp -d {} $uploads_dir/ \;
+        # Publish everything that is linked by bitbake at the end of the build (avoid timestamp and duplication)
+        find $deploy_dir/ -maxdepth 1 -type l -exec cp --dereference {} $uploads_dir/ \;
+        # Files without links: *.vfat, *.elf, *.efi, efi.stub, fit *.its, qcom-metadata.dtb and qclinuxfitImage
+        # Copy *.efi, *.efi and *.vfat as they are useful for debugging
+        find $deploy_dir/ -maxdepth 1 -type f \( -name \*.efi -o -name \*.elf -o -name dtb-\*.vfat \) -exec cp {} $uploads_dir/ \;
         cp buildchart.svg kas-build.yml $uploads_dir/
         if [ -d $deploy_dir/../../sdk ]; then
           cp $deploy_dir/../../sdk/* $uploads_dir/


.github/actions/compile/action.yml
```diff
@@ -88,8 +88,8 @@ runs:
         deploy_dir=../kas/build/tmp/deploy/images/${{inputs.machine}}
         uploads_dir=./uploads/${{ inputs.distro_name }}${{ inputs.kernel_dirname }}/${{ inputs.machine }}
         mkdir -p $uploads_dir
-        find $deploy_dir/ -maxdepth 1 -type f -exec cp {} $uploads_dir/ \;
-        find $deploy_dir/ -maxdepth 1 -type l \( -name boot-*.img -o -name *.rootfs.ext4.gz -o -name *.rootfs.qcomflash.tar.gz -o -name *.rootfs.ostreecommit.tar.xz \) -exec cp -d {} $uploads_dir/ \;
+        find $deploy_dir/ -maxdepth 1 -type l -exec cp {} $uploads_dir/ \;
```
quaresmajose: I belive you will need `cp --dereference` to follow symbolic links when copying from them otherwise you copy  just the symbolic links.

lumag: Should it be `cp -d`?

.github/actions/compile/action.yml
```diff
@@ -88,8 +88,8 @@ runs:
...
+        find $deploy_dir/ -maxdepth 1 -type f \( -name \*.its -o -name qcom-metadata.dtb -o -name \*.vfat \) -exec cp {} $uploads_dir/ \;
```
lumag: Could you please add a comment, showing a list of files that we are going to upload? I doubt we need qcom-metadata.dtb.

ricardosalveti: Yeah, was wondering if metadata was actually useful, but in addition to its and vfat I was just missing the elf files, will extend with a comment.

.github/actions/compile/action.yml
ricardosalveti: It worked without it, but that is probably because the files are upload instead of the links, will update.

ricardosalveti: Updated

.github/actions/compile/action.yml
```diff
@@ -88,8 +88,10 @@ runs:
         deploy_dir=../kas/build/tmp/deploy/images/${{inputs.machine}}
         uploads_dir=./uploads/${{ inputs.distro_name }}${{ inputs.kernel_dirname }}/${{ inputs.machine }}
         mkdir -p $uploads_dir
-        find $deploy_dir/ -maxdepth 1 -type f -exec cp {} $uploads_dir/ \;
-        find $deploy_dir/ -maxdepth 1 -type l \( -name boot-*.img -o -name *.rootfs.ext4.gz -o -name *.rootfs.qcomflash.tar.gz -o -name *.rootfs.ostreecommit.tar.xz \) -exec cp -d {} $uploads_dir/ \;
+        # Publish everything that is linked by bitbake at the end of the build (avoid timestamp and duplication)
+        find $deploy_dir/ -maxdepth 1 -type l -exec cp --dereference {} $uploads_dir/ \;
+        # No links for efi, elf and dtb vfat images, copy those manually
+        find $deploy_dir/ -maxdepth 1 -type f \( -name \*.efi -o -name \*.elf -o -name dtb-\*.vfat \) -exec cp {} $uploads_dir/ \;
```
lumag, line 107: I'd still like to see the list of the files for the upload, please.

ricardosalveti: Yup, waiting the build run and will compare, since some targets might have different files.

ricardosalveti: Files without any deployment link:

* dtb-*.vfat
* abl2esp-v*.elf
* systemd-bootaa64.efi
* linux-$MACHINE-.efi
* linuxaa64.efi.stub
* qclinux-fit-image.its
* qcom-metadata.dtb
* qclinuxfitImage

I believe we need the vfat and elf files, we can also upload the efi files, but they are part of other images.

.github/actions/compile/action.yml
```diff
@@ -88,8 +88,11 @@ runs:
         deploy_dir=../kas/build/tmp/deploy/images/${{inputs.machine}}
         uploads_dir=./uploads/${{ inputs.distro_name }}${{ inputs.kernel_dirname }}/${{ inputs.machine }}
         mkdir -p $uploads_dir
-        find $deploy_dir/ -maxdepth 1 -type f -exec cp {} $uploads_dir/ \;
-        find $deploy_dir/ -maxdepth 1 -type l \( -name boot-*.img -o -name *.rootfs.ext4.gz -o -name *.rootfs.qcomflash.tar.gz -o -name *.rootfs.ostreecommit.tar.xz \) -exec cp -d {} $uploads_dir/ \;
+        # Publish everything that is linked by bitbake at the end of the build (avoid timestamp and duplication)
+        find $deploy_dir/ -maxdepth 1 -type l -exec cp --dereference {} $uploads_dir/ \;
+        # Files without links: *.vfat, *.elf, *.efi, efi.stub, fit its, qcom-metadata.dtb and qclinuxfitImage
```
koenkooi: `fit its` looks like a typo

ricardosalveti: it is actually the its file from fit :-)

ricardosalveti: changed to fit *.its

PR #1432, opened 2026-01-24
gst-plugins-imsdk: Add recipe for IMSDK GStreamer plugin suite

Raja-Ganapathi-Busam:
The recipe fetches IMSDK GStreamer plugins source code from git repository and compiles plugins. The repository contains various GStreamer plugins across several functional domains:
  - Multimedia
  - Computer Vision
  - Machine Learning
  - Messaging and IPC

Plugin enablement and feature selection are fully configurable through PACKAGECONFIG.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-multimedia/imsdk/gst-plugins-imsdk_%.bbappend
+++ b/dynamic-layers/openembedded-layer/recipes-multimedia/imsdk/gst-plugins-imsdk_%.bbappend
@@ -1,0 +1,1 @@
+PACKAGECONFIG:append = " ml messaging"
--- a/recipes-multimedia/imsdk/gst-plugins-imsdk-packaging.inc
+++ b/recipes-multimedia/imsdk/gst-plugins-imsdk-packaging.inc
@@ -1,0 +1,62 @@
+# Extends the GStreamer packaging to handle IMSDK plugins.
+# For plugins that provide runtime modules, all module .so files under
+# <plugin>/modules/ are grouped into a separate per-plugin modules package.
+
+require recipes-multimedia/gstreamer/gstreamer1.0-plugins-packaging.inc
+
+PACKAGESPLITFUNCS =+ "split_imsdk_module_packages"
+PACKAGESPLITFUNCS += "set_imsdk_oss_metapkg_rdepends"
+
+python split_imsdk_module_packages () {
+    imsdk_root = d.expand('${libdir}/imsdk')
+
+    # Qualcomm runtime modules for plugin
+    do_split_packages(d, imsdk_root, r'([^/]+)/modules/[^/]+\.so', d.expand('${PN}-%s-modules'), 'IMSDK runtime modules for %s plugin', recursive=True, extra_depends='', match_path=True)
+}
+
+# List of plugin packages to exclude from OSS meta package
+# Add package names (without version/architecture suffix) to this list
+# Example: to exclude gst-plugins-imsdk-qtimlqnn, add 'qtimlqnn' to the list
+IMSDK_OSS_METAPKG_EXCLUDE ?= "qtimlqnn qtimlsnpe qtismartvencbin"
+
+python set_imsdk_oss_metapkg_rdepends () {
+    import os
+    import oe.utils
+
+    # Go through all generated packages (excluding the main package and
+    # the -meta package itself) and add them to the -meta package as RDEPENDS.
+    # Also exclude packages specified in IMSDK_OSS_METAPKG_EXCLUDE.
+
+    pn = d.getVar('PN')
+    metapkg =  pn + '-oss-meta'
+    d.setVar('ALLOW_EMPTY:' + metapkg, "1")
+    d.setVar('FILES:' + metapkg, "")
+    exclude = [ pn, pn + '-meta', pn + '-oss-meta' ]
+
+    # Add user-defined plugin exclusions to the exclude list
+    exclude_plugins = d.getVar('IMSDK_OSS_METAPKG_EXCLUDE') or ""
+    if exclude_plugins:
+        for plugin in exclude_plugins.split():
+            # Add packages like gst-plugins-imsdk-<plugin>
+            exclude.append(pn + '-' + plugin)
+
+    metapkg_rdepends = []
+    pkgdest = d.getVar('PKGDEST')
+    for pkg in oe.utils.packages_filter_out_system(d):
+        if pkg not in exclude and pkg not in metapkg_rdepends:
+            # See if the package is empty by looking at the contents of its PKGDEST subdirectory.
+            # If this subdirectory is empty, then the package is.
+            # Empty packages do not get added to the meta package's RDEPENDS
+            pkgdir = os.path.join(pkgdest, pkg)
+            if os.path.exists(pkgdir):
+                dir_contents = os.listdir(pkgdir) or []
+            else:
+                dir_contents = []
+            is_empty = len(dir_contents) == 0
+            if not is_empty:
+                metapkg_rdepends.append(pkg)
+    d.setVar('RDEPENDS:' + metapkg, ' '.join(metapkg_rdepends))
+    d.setVar('DESCRIPTION:' + metapkg, pn + ' meta package')
+}
+
+PACKAGES += "${PN}-oss-meta"
--- a/recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
+++ b/recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
@@ -1,0 +1,37 @@
+require gst-plugins-imsdk-packaging.inc
+
+SUMMARY = "Qualcomm IMSDK GStreamer Plugins (QTI OSS)"
+DESCRIPTION = "Open-source Qualcomm IMSDK GStreamer multimedia, CV, ML, and messaging plugins"
+SECTION = "multimedia"
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+inherit cmake features_check pkgconfig
+
+REQUIRED_DISTRO_FEATURES = "opengl"
+
+SRC_URI = "git://github.com/qualcomm/gst-plugins-imsdk;branch=main;protocol=https"
+
+SRCREV = "6ad888f79403ce91092dd49eed3dc3b4242957ca"
+PV = "0.0+git"
+
+DEPENDS += "\
+    gstreamer1.0 \
+    gstreamer1.0-plugins-base \
+    virtual/egl \
+    virtual/libgbm \
+    virtual/libgles2 \
+    virtual/libgles3 \
+"
+
+# This package is currently only used and tested on ARMv8 (aarch64) machines.
+# Therefore, builds for other architectures are not necessary and are explicitly excluded.
+COMPATIBLE_MACHINE = "^$"
+COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
+PACKAGECONFIG ??= "videoproc sw"
+
+PACKAGECONFIG[ml]           = "-DENABLE_GST_ML_PLUGINS=1, -DENABLE_GST_ML_PLUGINS=0, cairo json-glib opencv qairt-sdk, qairt-sdk"
+PACKAGECONFIG[messaging]    = "-DENABLE_GST_MESSAGING_PLUGINS=1, -DENABLE_GST_MESSAGING_PLUGINS=0, librdkafka mosquitto"
+PACKAGECONFIG[sw]           = "-DENABLE_GST_SOFTWARE_PLUGINS=1, -DENABLE_GST_SOFTWARE_PLUGINS=0, gstreamer1.0-rtsp-server smart-venc-ctrl-algo"
+PACKAGECONFIG[videoproc]    = "-DENABLE_GST_VIDEOPROC_PLUGINS=1, -DENABLE_GST_VIDEOPROC_PLUGINS=0, cairo"


recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
+
```
lumag, line 2: Useless empty line

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+SUMMARY = "Qualcomm IMSDK GStreamer Plugins (QTI OSS)"
+DESCRIPTION = "Open-source Qualcomm IMSDK GStreamer multimedia, CV, ML, and messaging plugins"
+SECTION = "multimedia"
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/${LICENSE};md5=7a434440b651f4a472ca93716d01033a"
```
lumag: No, don't checksum files outside of the source tree.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+
+inherit cmake pkgconfig
+
+SRC_URI = "git://github.com/qualcomm/gst-plugins-imsdk;branch=main;protocol=https"
+
+SRCREV = "1a335478f7240906b4b085e8998300005e832fa0"
+PV = "1.0.0"
```
lumag: No, there is no 1.0.0 tag

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+
+# ============================================================================
+# DEPENDENCIES
+# ============================================================================
```
lumag: Drop useless comments

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+DEPENDS = "\
+    gstreamer1.0 \
+    gstreamer1.0-plugins-base \
+    gstreamer1.0-plugins-bad \
+    gstreamer1.0-rtsp-server \
+    json-glib \
+    libsoup-2.4 \
+    qcom-adreno \
+    ${@bb.utils.contains('PREFERRED_PROVIDER_virtual/libgbm', 'gbm', 'gbm', '', d)} \
```
lumag: Why? If it depends on gbm headers, then any headers provider will work.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+# DEPENDENCIES
+# ============================================================================
+DEPENDS = "\
+    gstreamer1.0 \
+    gstreamer1.0-plugins-base \
+    gstreamer1.0-plugins-bad \
+    gstreamer1.0-rtsp-server \
+    json-glib \
+    libsoup-2.4 \
+    qcom-adreno \
```
lumag: Does it depend on extra headers? Will it build against a standard headerset?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+    ${@bb.utils.contains('PREFERRED_PROVIDER_virtual/libgbm', 'gbm', 'gbm', '', d)} \
+"
+
+RDEPENDS:${PN} += "\
+    gstreamer1.0 \
+    gstreamer1.0-plugins-base \
+    ${@bb.utils.contains('PREFERRED_PROVIDER_virtual/libgbm', 'gbm', 'gbm', '', d)} \
```
lumag: Why are these required? Are they not pulled via normal library-generated dependencies?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+PACKAGECONFIG ??= "\
+    base batch smartvencbin socket tools videotemplate \
+    vcomposer voverlay vsplit vtransform \
+    metamux mldemux mlvconverter mlvdiscovery \
+    mlvdetection mlvclassification mlvpose mlpostprocess mlvsegmentation mlvsuperresolution \
+    restrictedzone metatransform mlmetaextractor mlmetaparser mltools \
+    msgbroker redissink rtspbin \
+"
+
+PACKAGECONFIG[base]                 = "-DENABLE_GST_PLUGIN_BASE=1,              -DENABLE_GST_PLUGIN_BASE=0,                     qcom-fastcv-binaries virtual/egl virtual/kernel virtual/libgles2"
```
lumag: What is included in base, so that it requires FastCV, EGL and GL ES?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+PACKAGECONFIG[qmmfsrc]              = "-DENABLE_GST_PLUGIN_QMMFSRC=1,           -DENABLE_GST_PLUGIN_QMMFSRC=0"
```
lumag: No source dependencies?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+PACKAGECONFIG[tools]                = "-DENABLE_GST_PLUGIN_TOOLS=1,             -DENABLE_GST_PLUGIN_TOOLS=0"
+PACKAGECONFIG[batch]                = "-DENABLE_GST_PLUGIN_BATCH=1,             -DENABLE_GST_PLUGIN_BATCH=0"
+PACKAGECONFIG[metamux]              = "-DENABLE_GST_PLUGIN_METAMUX=1,           -DENABLE_GST_PLUGIN_METAMUX=0"
+PACKAGECONFIG[socket]               = "-DENABLE_GST_PLUGIN_SOCKET=1,            -DENABLE_GST_PLUGIN_SOCKET=0"
+PACKAGECONFIG[restrictedzone]       = "-DENABLE_GST_PLUGIN_RESTRICTED_ZONE=1,   -DENABLE_GST_PLUGIN_RESTRICTED_ZONE=0,          opencv"
+PACKAGECONFIG[rtspbin]              = "-DENABLE_GST_PLUGIN_RTSPBIN=1,           -DENABLE_GST_PLUGIN_RTSPBIN=0,                  gstreamer1.0-rtsp-server"
```
lumag: Is RTSP server a build or a runtime dependency?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+# Sample apps
+PACKAGECONFIG[sampleappsutils]      = "-DENABLE_GST_SAMPLE_APPS_UTILS=1,         -DENABLE_GST_SAMPLE_APPS_UTILS=0"
+PACKAGECONFIG[sampleapps]           = "-DENABLE_GST_SAMPLE_APPS=1,               -DENABLE_GST_SAMPLE_APPS=0"
+PACKAGECONFIG[testframework]        = "-DENABLE_GST_TEST_FRAMEWORK=1,            -DENABLE_GST_TEST_FRAMEWORK=0"
+PACKAGECONFIG[pythonexamples]       = "-DENABLE_GST_PYTHON_EXAMPLES=1,           -DENABLE_GST_PYTHON_EXAMPLES=0"
+
+# Messaging
+PACKAGECONFIG[msgbroker]            = "-DENABLE_GST_PLUGIN_MSGBROKER=1,          -DENABLE_GST_PLUGIN_MSGBROKER=0,               librdkafka mosquitto"
+
+CFLAGS:append:base = " -I${S}/gst-plugin-base"
```
lumag: Why?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+CXXFLAGS:append:base = " -I${S}/gst-plugin-base"
+
+# ============================================================================
+# CMAKE OPTIONS
+# ============================================================================
+EXTRA_OECMAKE += "\
+    ${PACKAGECONFIG_CONFARGS} \
```
lumag: Should not be needed

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+    -DGST_VERSION_REQUIRED=1.20.7 \
+    -DGST_PLUGINS_QTI_OSS_INSTALL_INCDIR=${includedir} \
+    -DGST_PLUGINS_QTI_OSS_INSTALL_BINDIR=${bindir} \
+    -DGST_PLUGINS_QTI_OSS_INSTALL_LIBDIR=${libdir} \
+    -DKERNEL_BUILDDIR=${STAGING_INCDIR}/linux-msm \
```
lumag: What does it need from the kernel?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+
+CFLAGS:append:base = " -I${S}/gst-plugin-base"
+CXXFLAGS:append:base = " -I${S}/gst-plugin-base"
+
+# ============================================================================
+# CMAKE OPTIONS
+# ============================================================================
+EXTRA_OECMAKE += "\
+    ${PACKAGECONFIG_CONFARGS} \
+    -DGST_VERSION_REQUIRED=1.20.7 \
```
lumag: ?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+    -DGST_PLUGINS_QTI_OSS_INSTALL_BINDIR=${bindir} \
+    -DGST_PLUGINS_QTI_OSS_INSTALL_LIBDIR=${libdir} \
+    -DKERNEL_BUILDDIR=${STAGING_INCDIR}/linux-msm \
+"
+
+# ============================================================================
+# FILE PACKAGING
+# ============================================================================
+FILES:${PN} += "\
+    ${bindir}/* \
```
lumag: ${PN}-apps

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+    ${libdir}/gstreamer-1.0/* \
```
lumag: Can we use `gstreamer1.0-plugins-packaging.inc` and split the package into per-plugin packages?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+"
+
+FILES:${PN}-dev += "\
+    ${includedir} \
```
lumag: No need to

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+"
+
+FILES:${PN}-dbg += "\
+    ${libdir}/gstreamer-1.0/.debug \
```
lumag: Do we need to specify it?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+"
+
+SOLIBS = ".so*"
+FILES_SOLIBSDEV = ""
+INSANE_SKIP:${PN} += "dev-so"
```
lumag: Comment...

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+
+# ============================================================================
+# CMAKE OPTIONS
+# ============================================================================
+EXTRA_OECMAKE += "\
+    ${PACKAGECONFIG_CONFARGS} \
+    -DGST_VERSION_REQUIRED=1.20.7 \
+    -DGST_PLUGINS_QTI_OSS_INSTALL_INCDIR=${includedir} \
+    -DGST_PLUGINS_QTI_OSS_INSTALL_BINDIR=${bindir} \
+    -DGST_PLUGINS_QTI_OSS_INSTALL_LIBDIR=${libdir} \
```
anujm1: Please consider using `GNUInstallDirs` in your project and pass relative paths to `install`.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,130 @@
...
+PACKAGECONFIG[metatransform]        = "-DENABLE_GST_PLUGIN_METATRANSFORM=1,      -DENABLE_GST_PLUGIN_METATRANSFORM=0"
+PACKAGECONFIG[objtracker]           = "-DENABLE_GST_PLUGIN_OBJTRACKER=1,         -DENABLE_GST_PLUGIN_OBJTRACKER=0,              libeigen"
+PACKAGECONFIG[mlmetaextractor]      = "-DENABLE_GST_PLUGIN_MLMETAEXTRACTOR=1,    -DENABLE_GST_PLUGIN_MLMETAEXTRACTOR=0"
+PACKAGECONFIG[mlpostprocess]        = "-DENABLE_GST_PLUGIN_MLPOSTPROCESS=1,      -DENABLE_GST_PLUGIN_MLPOSTPROCESS=0,           cairo opencv"
+
+# Sample apps
+PACKAGECONFIG[sampleappsutils]      = "-DENABLE_GST_SAMPLE_APPS_UTILS=1,         -DENABLE_GST_SAMPLE_APPS_UTILS=0"
+PACKAGECONFIG[sampleapps]           = "-DENABLE_GST_SAMPLE_APPS=1,               -DENABLE_GST_SAMPLE_APPS=0"
+PACKAGECONFIG[testframework]        = "-DENABLE_GST_TEST_FRAMEWORK=1,            -DENABLE_GST_TEST_FRAMEWORK=0"
+PACKAGECONFIG[pythonexamples]       = "-DENABLE_GST_PYTHON_EXAMPLES=1,           -DENABLE_GST_PYTHON_EXAMPLES=0"
```
anujm1: Python examples do not RDEPEND on Python?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: The `base` directory acts as our central repository for shared utilities and type definitions. It is organized into subdirectories like ml, video, and cv, each offering reusable components for machine learning, GPU‑accelerated video processing on Adreno, and computer vision operations.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Are you referring whether this tag present in original Github repo?
We haven’t started using tags in the GitHub repository yet. As part of our migration to the GitHub Flow process, we are currently working on setting them up.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: `pythonexamples` is currently, not enabled and tested. It would depend on `gstreamer1.0-python` eventually.
Let me revisit the dependency details for disabled packageconfig features.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Since `gst-plugin-base` provides shared utility code, most plugins depend on its headers.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Our idea is to keep the entire IMSDK as single package. So that it is easy to install and maintain.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
ricardosalveti: If you set PV to a proper value, such as 1.0.0, we expect this same tag / release to be available from the upstream repository. Otherwise it would be something like "0.0+git" instead.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
ricardosalveti: Yeah, if not it should just depend on the standard headerset.

ricardosalveti:
`ERROR: Nothing PROVIDES 'librdkafka' (but /work/build-yocto-check-layer-ALU4/meta-qcom/recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb DEPENDS on or otherwise requires it)
`

ndechesne:
> > `ERROR: Nothing PROVIDES 'librdkafka' (but /work/build-yocto-check-layer-ALU4/meta-qcom/recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb DEPENDS on or otherwise requires it) `
> 
> `librdkafka` (which is required for `msgbroker`) is coming from `meta-openembedded` layer. This layer gets included when `qcom-distro` is used.
> 
> We kept it under `qcom` override.
> 
> ```bb
> PACKAGECONFIG:append:qcom
> ```
> 
> Please suggest if this is not the right approach.

It is not the right approach. The 'qcom' override is to include specific features from a recipe when building for a qcom machine (the qcom override is added to all qualcomm machines). 

msgbroker is a recipe in a different layer, and meta-oe is not included 'by default' when we only build meta-qcom alone, and that won't change. 

we rely on 'dynamic-layers' features (see the top level folder in meta-qcom) for that. The way it works is that you can have recipes (or bbappend) which are considered when meta-oe is included in the build. This will impact your build, i suggest you have a look at that, and then we discuss again if it does not work for you

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,81 @@
...
+PACKAGECONFIG[metatransform]        = "-DENABLE_GST_PLUGIN_METATRANSFORM=1,      -DENABLE_GST_PLUGIN_METATRANSFORM=0"
+PACKAGECONFIG[mldemux]              = "-DENABLE_GST_PLUGIN_MLDEMUX=1,            -DENABLE_GST_PLUGIN_MLDEMUX=0"
+PACKAGECONFIG[mlmetaextractor]      = "-DENABLE_GST_PLUGIN_MLMETAEXTRACTOR=1,    -DENABLE_GST_PLUGIN_MLMETAEXTRACTOR=0"
+PACKAGECONFIG[mlmetaparser]         = "-DENABLE_GST_PLUGIN_MLMETAPARSER=1,       -DENABLE_GST_PLUGIN_MLMETAPARSER=0,            json-glib, json-glib"
+PACKAGECONFIG[mlpostprocess]        = "-DENABLE_GST_PLUGIN_MLPOSTPROCESS=1,      -DENABLE_GST_PLUGIN_MLPOSTPROCESS=0,           cairo opencv, cairo opencv"
+PACKAGECONFIG[mltools]              = "-DENABLE_GST_PLUGIN_MLTOOLS=1,            -DENABLE_GST_PLUGIN_MLTOOLS=0"
+PACKAGECONFIG[mlvconverter]         = "-DENABLE_GST_PLUGIN_MLVCONVERTER=1,       -DENABLE_GST_PLUGIN_MLVCONVERTER=0"
+
+# Messaging
+PACKAGECONFIG[msgbroker]            = "-DENABLE_GST_PLUGIN_MSGBROKER=1,          -DENABLE_GST_PLUGIN_MSGBROKER=0,               librdkafka mosquitto, librdkafka mosquitto"
```
ndechesne: since librdkafka  is in meta-oe, the 'right' thing to do is the following:

1. keep your recipe in meta-qcom but with msgbroker disabled by default in PACKAGECONFIG
2. create a bbappend for imsdk in meta-qcom/dynamic-layers/openembedded-layer/xxx
3. in this bbappend you enable msgbroker in PACKAGECONFIG

As a result:
1. when we build meta-qcom in standalone (no meta-qcom-distro, and no meta-oe), images will include ISDK without msgbroker
2. when we include meta-oe in the build, msgbroker will be included in the image.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
ndechesne: does this folder even exist? I doubt it. we don't even depend on the kernel (or headers)

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
ndechesne: I would make the same recommendation as @lumag above. It is better to leverage the OE infrastructure for building gst plugins. you will get individual packages which are good for users, and you will also get a 'meta' package that depends on all of them for ease of use.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,81 @@
...
+PV = "0.0+git"
+
+DEPENDS += "\
+    gstreamer1.0 \
+    gstreamer1.0-plugins-base \
+    qcom-fastcv-binaries \
+    virtual/libgbm \
+"
+RDEPENDS:${PN} += "\
+    gstreamer1.0 \
```
ndechesne: I don't think you need RDEPENDS for gst1.0 and -base. the build system will figure it out automatically.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,81 @@
...
+"
+
+PACKAGECONFIG:append:qcom = "\
+    redissink restrictedzone \
+    mlpostprocess \
+    msgbroker \
+"
+
+# Software plugins -- software
+PACKAGECONFIG[batch]                = "-DENABLE_GST_PLUGIN_BATCH=1,             -DENABLE_GST_PLUGIN_BATCH=0"
```
ndechesne: i am not a big fan of such indentation, i would prefer something like that, which is what we typically find in oe-core recipes

```
PACKAGECONFIG[metamux]     = "-DENABLE_GST_PLUGIN_METAMUX=1, -DENABLE_GST_PLUGIN_METAMUX=0"
PACKAGECONFIG[redissink]   = "-DENABLE_GST_PLUGIN_REDISSINK=1, -DENABLE_GST_PLUGIN_REDISSINK=0, hiredis, hiredis"
```

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: After these changes, how can I verify it in my local build.

ndechesne: You can use :

`kas build meta-qcom/ci/rb3gen2-core-kit.yml
`
to make a build with meta-qcom only, replace with the right machine. In this config we are only using oe-core+meta-qcom

you can use 

`kas build meta-qcom/ci/qcom-distro.yml:meta-qcom/ci/rb3gen2-core-kit.yml
`
it will create a build with meta-qcom+oe+core + meta-qcom-distro , meta-oe, ... 

This is what our CI is testing.

Raja-Ganapathi-Busam: Right now, yocto-check-layer is failing. How to trigger this locally.

ndechesne: Please check https://docs.yoctoproject.org/dev-manual/layers.html#yocto-check-layer-script and let us know if there any question.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,41 @@
...
+SECTION = "multimedia"
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=223037c4be0bfc6cf757035432adf983"
+
+inherit cmake pkgconfig
+
+SRC_URI = "git://github.com/qualcomm/gst-plugins-imsdk;branch=main;protocol=https"
+
+SRCREV = "93610673ea88033c0b0d46ada93a55a24d9f3b75"
+PV = "0.0+git"
```
koenkooi, line 16: Does upstream have a plan to have proper releases? Everything being `0.0+git` makes it hard to spot if a version used by customers has a specific fix or not.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,41 @@
...
+
+DEPENDS += "\
+    gstreamer1.0 \
+    gstreamer1.0-plugins-base \
+    qcom-fastcv-binaries \
+    mesa \
+"
+RDEPENDS:${PN} += "\
```
koenkooi: R* variables and PACKAGES* go below do_install, since they are packaging related. PACKAGE_CONFIG and EXTRA_OECMAKE are using in do_configure(), so they should be above R* and PACKAGES*.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: On master branch, it always points to latest commit.
Once QLI 2.0 branch is created, there we will point to version tags.

lumag: Well. Please use _releases_ for the software released into QLI mainline.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: qmmfsrc is not fully enabled. Removed the option for now. It will be enabled once the camera-service is enabled.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: It has both compile time and runtime dependency.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Removed `base` as separate PACKAGECONFIG option. So, not needed this now.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Removed.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: We are looking for `linux/dma-heap.h` from kernel headers. 
We are addressing this on upstream code to look for standard headers path. So, this is not needed now.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Done

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Removed.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: This is taken care by Yocto. Removed.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: We have made changes in upstream code to make it work without this. Removed.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Thanks @anujm1.
We made changes in upstream code to use `GNUInstallDirs`. Removed this.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: For now, we removed PACKAGECONFIG options which are not enabled yet.
We will enable them in the upcoming PRs.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Verified, this is not needed, hence removed.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Thanks @ndechesne .
Updated with the common indentation style.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,37 @@
+SUMMARY = "Qualcomm IMSDK GStreamer Plugins (QTI OSS)"
+DESCRIPTION = "Open-source Qualcomm IMSDK GStreamer multimedia, CV, ML, and messaging plugins"
+SECTION = "multimedia"
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=223037c4be0bfc6cf757035432adf983"
+
+inherit cmake pkgconfig
+inherit distro_features_check
```
anujm1: Should be on same line and should be `features_check`: https://git.openembedded.org/openembedded-core/tree/meta/classes-recipe/features_check.bbclass

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Removed and handling in source code itself as this is the source code requirement rather than a compile time option.

dynamic-layers/openembedded-layer/recipes-multimedia/imsdk/gst-plugins-imsdk_%.bbappend
```diff
@@ -0,0 +1 @@
+PACKAGECONFIG:append:qcom = " ext-sw ml msgbroker"
```
lumag: Why do you need `:qcom`here?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,37 @@
+SUMMARY = "Qualcomm IMSDK GStreamer Plugins (QTI OSS)"
+DESCRIPTION = "Open-source Qualcomm IMSDK GStreamer multimedia, CV, ML, and messaging plugins"
+SECTION = "multimedia"
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=223037c4be0bfc6cf757035432adf983"
+
+inherit cmake pkgconfig features_check
+
+REQUIRED_DISTRO_FEATURES = "opengl"
```
lumag, line 11: Is OpenGL required to build all plugins? Is it OpenGL and, for example, not CL?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,37 @@
...
+
+SRC_URI = "git://github.com/qualcomm/gst-plugins-imsdk;branch=main;protocol=https"
+
+SRCREV = "93610673ea88033c0b0d46ada93a55a24d9f3b75"
+PV = "0.0+git"
+
+DEPENDS += "\
+    gstreamer1.0 \
+    gstreamer1.0-plugins-base \
+    qcom-fastcv-binaries \
```
lumag: Does it need FastCV to build all plugins?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,37 @@
...
+    virtual/libgbm \
+"
+RDEPENDS:${PN} += "\
+    qcom-fastcv-binaries \
```
lumag: Why?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,37 @@
...
+"
+
+PACKAGES =+ "${PN}-apps"
+
+PACKAGECONFIG ??= "videoproc sw"
```
lumag, line 32: Please provide at least some description. What is the difference between `sw` and `ext-sw`?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,37 @@
...
+
+PACKAGECONFIG[ext-sw]       = "-DENABLE_GST_EXT_SOFTWARE_PLUGINS=1, -DENABLE_GST_EXT_SOFTWARE_PLUGINS=0, hiredis"
+PACKAGECONFIG[ml]           = "-DENABLE_GST_ML_PLUGINS=1, -DENABLE_GST_ML_PLUGINS=0, cairo json-glib opencv"
+PACKAGECONFIG[msgbroker]    = "-DENABLE_GST_PLUGIN_MSGBROKER=1, -DENABLE_GST_PLUGIN_MSGBROKER=0, librdkafka mosquitto"
+PACKAGECONFIG[sw]           = "-DENABLE_GST_SOFTWARE_PLUGINS=1, -DENABLE_GST_SOFTWARE_PLUGINS=0, gstreamer1.0-rtsp-server smart-venc-ctrl-algo"
+PACKAGECONFIG[videoproc]    = "-DENABLE_GST_VIDEOPROC_PLUGINS=1, -DENABLE_GST_VIDEOPROC_PLUGINS=0, cairo"
+
+FILES:${PN} += "${libdir}/gstreamer-1.0/*"
```
lumag: You are stil insisting on having a single package instead of following OE's gstreamer philosophy. Why?

dynamic-layers/openembedded-layer/recipes-multimedia/imsdk/gst-plugins-imsdk_%.bbappend
Raja-Ganapathi-Busam: You are right. Earlier, our understanding was wrong. I will remove.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: We do dynamic load of the fastcv library. So we need to specify it in RDEPENDS.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
ndechesne: +1, again, on that, I would like to request we package plugin one by one. Our users will thank us for that.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
lumag: Well. First of all, I doubt that _all_ plugins do this. So you can be using PACKAGECONFIG to specify this dependency. Next, please add a comment, telling what is happening ("plugin foo dlopen()s FastCV libs").

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: We have a base/common module which requires both opengl and qcom-fastcv-binaries.

lumag: ok

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
ricardosalveti: Mainline should also be released, and then integrated here. These software components will have to be released outside the QLI 2.X context as well, so better to just release mainline directly.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,40 @@
...
+"
+
+PACKAGECONFIG ??= "videoproc sw"
+
+# The name ext below, indicates that the package depends on recipes from external meta layer other than current meta layer or its dependencies.
+PACKAGECONFIG[ext-sw]       = "-DENABLE_GST_EXT_SOFTWARE_PLUGINS=1, -DENABLE_GST_EXT_SOFTWARE_PLUGINS=0, hiredis"
+PACKAGECONFIG[ml]           = "-DENABLE_GST_ML_PLUGINS=1, -DENABLE_GST_ML_PLUGINS=0, cairo json-glib opencv"
+PACKAGECONFIG[msgbroker]    = "-DENABLE_GST_PLUGIN_MSGBROKER=1, -DENABLE_GST_PLUGIN_MSGBROKER=0, librdkafka mosquitto"
+PACKAGECONFIG[sw]           = "-DENABLE_GST_SOFTWARE_PLUGINS=1, -DENABLE_GST_SOFTWARE_PLUGINS=0, gstreamer1.0-rtsp-server smart-venc-ctrl-algo"
+PACKAGECONFIG[videoproc]    = "-DENABLE_GST_VIDEOPROC_PLUGINS=1, -DENABLE_GST_VIDEOPROC_PLUGINS=0, cairo"
```
ricardosalveti, line 37: Why having a similar indentation for all these lines? They are different options anyway, fine to just remove the extra whitespace.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
ricardosalveti: Yeah, please break into multiple packages, that is the expected way to consume gstreamer plugins in OE.

Raja-Ganapathi-Busam: Updated the packaging logic. It is handled in packaging.inc file.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: This is how the GStreamer recipes also follows.
@ndechesne also suggested the same. 
https://github.com/qualcomm-linux/meta-qcom/pull/1432#discussion_r2740157882

This helps readability as the list of options grows in future.

Let me know your opinion.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Added a comment to explain `ext` terminology.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Done

recipes-multimedia/imsdk/gst-plugins-imsdk-packaging.inc
```diff
@@ -0,0 +1,64 @@
+# This .inc file contains functionality for automatically splitting
+# built plugins into individual packages for each plugin. A -meta
+# package is also set up that has no files of its own, but contains
+# the names of all plugin packages in its RDEPENDS list.
```
lumag: Why do you need an .inc file?

recipes-multimedia/imsdk/gst-plugins-imsdk-packaging.inc
```diff
@@ -0,0 +1,64 @@
...
+
+# Dynamically generate packages for all enabled plugins
+PACKAGES_DYNAMIC = "^${PN}-.* ^libgst.*"
```
lumag: Why can't we include gstreamer1.0-plugins-packaging.inc ?

tjamula: The following are the modifications made in the inc file when compared to gstreamer1.0-plugins-packaging.inc

1.  **Plugin Library Name Matching Updated**
    *   The pattern used to detect plugin library names was changed from **`libgst`** to **`libgstqti`**.
    *   The package descriptions were also updated to explicitly include **“Qualcomm”** to reflect the IM SDK–specific plugins.

2.  **Removal of Unused Split Packages**
    *   The split packages for:
        *   **GObject introspection files**, and
        *   **Static GStreamer development libraries**  
            were removed, since the IM SDK packaging does not distribute any such libraries.

3.  **Additional Split Package for Plugin modules**
    *   [split_plugin_modules](https://github.com/Raja-Ganapathi-Busam/meta-qcom/blob/62871d78893fad595fb0e091ca084f13b6f506d0/recipes-multimedia/imsdk/gst-plugins-imsdk-packaging.inc#L23)
    *   Several IM SDK plugins build **multiple module binaries** that the plugin can dynamically use.
    *   To package these module artifacts appropriately, an **additional split package** was introduced to collect and ship these modules.

tjamula: Also, gstreamer1.0-plugins-packaging.inc is not part of the meta-qcom layer, can we reference files from other layers?

ndechesne: yes, you can, see https://docs.yoctoproject.org/bitbake/2.16/bitbake-user-manual/bitbake-user-manual-metadata.html#include-directive

lumag: Why is it `libgstqti-`? It's a GStreamer plugin, so it should be named `libgst-something`. Start plugin names with QTI if that's what you want to have.

tjamula: All Qualcomm GStreamer plugins follow a consistent library naming scheme beginning with libgstqti. This prefix clearly identifies them as Qualcomm‑specific. Because the plugin names themselves already include qti, this prefix was used as the matching string for packaging all Qualcomm GStreamer plugins.

lumag: Okay. Then please update the inc file in OE-Core to allow parametrisation and then use it here.

tjamula: - What exactly are you suggesting should be parameterized? As mentioned earlier, these differences come from the added .inc file compared to gstreamer1.0-plugins-packaging.inc. Can these behaviors be achieved in another way? If so, could you point me to the appropriate reference?

- Also, are you recommending changes to this packaging logic or suggesting that some parts be dropped? The additional split package for plugin modules is required and cannot be removed, because gstreamer1.0-plugins-packaging.inc does not package these module libraries.

> The following are the modifications made in the inc file when compared to gstreamer1.0-plugins-packaging.inc
> 
> 1. **Plugin Library Name Matching Updated**
>    
>    * The pattern used to detect plugin library names was changed from **`libgst`** to **`libgstqti`**.
>    * The package descriptions were also updated to explicitly include **“Qualcomm”** to reflect the IM SDK–specific plugins.
> 2. **Removal of Unused Split Packages**
>    
>    * The split packages for:
>      
>      * **GObject introspection files**, and
>      * **Static GStreamer development libraries**
>        were removed, since the IM SDK packaging does not distribute any such libraries.
> 3. **Additional Split Package for Plugin modules**
>    
>    * [split_plugin_modules](https://github.com/Raja-Ganapathi-Busam/meta-qcom/blob/62871d78893fad595fb0e091ca084f13b6f506d0/recipes-multimedia/imsdk/gst-plugins-imsdk-packaging.inc#L23)
>    * Several IM SDK plugins build **multiple module binaries** that the plugin can dynamically use.
>    * To package these module artifacts appropriately, an **additional split package** was introduced to collect and ship these modules.

Raja-Ganapathi-Busam:
> I get errors while compiling
> 
> ```
>    64 | #include <linux/ion.h>
>       |          ^~~~~~~~~~~~~
> compilation terminated.
> ```
> 
> The project looks at host paths for headers and should be fixed. Looking at the logs, some of them:
> 
> ```
> -I/usr/include/vidc 
> -I/usr/include/uapi
> ```

There are changes in the source code. These are under merge process. Mostly, Today they will merge to github. I will update the SRCREV after that.

recipes-multimedia/imsdk/gst-plugins-imsdk-packaging.inc
tjamula: @lumag / @ndechesne / @anujm1 
Is there any standard variable that points to the root of oe-core layer so that the gstreamer1.0-plugins-packaging.inc can be included as follows:
`<oe-core root>/meta/recipes-multimedia/gstreamer/gstreamer1.0-plugins-packaging.inc`

ndechesne: `include recipes-multimedia/gstreamer/gstreamer1.0-plugins-packaging.inc`

will work, as mentioned in the docs, include files are searched on BBPATH.

lumag: @ndechesne You beat me for the answer :-)

tjamula:
The packaging inc was updated to now include the gstreamer packaging inc and append functionality required for imsdk. Please let us know if this approach looks appropriate.

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,35 @@
...
+DEPENDS += "\
+    gstreamer1.0 \
+    gstreamer1.0-plugins-base \
+    qcom-fastcv-binaries \
+    virtual/libgbm \
+"
+
+PACKAGECONFIG ??= "videoproc sw"
+
+# The name ext below, indicates that the package depends on recipes from external meta layer other than current meta layer or its dependencies.
```
lumag: What would that mean if we build IMSDK plugins on Debian system?

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
```diff
@@ -0,0 +1,35 @@
...
+PACKAGECONFIG[ext-sw]       = "-DENABLE_GST_EXT_SOFTWARE_PLUGINS=1, -DENABLE_GST_EXT_SOFTWARE_PLUGINS=0, hiredis"
+PACKAGECONFIG[ml]           = "-DENABLE_GST_ML_PLUGINS=1, -DENABLE_GST_ML_PLUGINS=0, cairo json-glib opencv qairt-sdk, qairt-sdk"
+PACKAGECONFIG[messaging]    = "-DENABLE_GST_MESSAGING_PLUGINS=1, -DENABLE_GST_MESSAGING_PLUGINS=0, librdkafka mosquitto"
+PACKAGECONFIG[sw]           = "-DENABLE_GST_SOFTWARE_PLUGINS=1, -DENABLE_GST_SOFTWARE_PLUGINS=0, gstreamer1.0-rtsp-server smart-venc-ctrl-algo"
+PACKAGECONFIG[videoproc]    = "-DENABLE_GST_VIDEOPROC_PLUGINS=1, -DENABLE_GST_VIDEOPROC_PLUGINS=0, cairo"
+
+# fcv backend from common library (present in gst-plugin-base) does dlopen()s FastCV libs
```
lumag: Gst-plugin-base or gstqti-plugin-base?

Raja-Ganapathi-Busam: It is gst-plugin-base. I'm referring the source code from IMSDK repo.
https://github.com/qualcomm/gst-plugins-imsdk/tree/main/gst-plugin-base

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
Raja-Ganapathi-Busam: Since we are adding ext-sw to PACKAGECONFIG using bbappend in dynamic-layers, it will only enabled when meta-openembedded layer is present.

ndechesne: What we are saying here, is that you should not make assumptions in your component about which plugins are in oe-core or meta-oe.  you should have PACKAGECONFIG to enable your plugins one by one, or by set. 

looking at IMSDK, GST_EXT_SOFTWARE flag will enable gst-plugin-redissink, which depends in hiredis from meta-oe. So I think you need a PACKAGECONFIG for it, instead of a generic 'ext-sw'. 

I think it's indeed better to enable plugins one by one, and also matches what the gst recipes do in oe-core, no?

Raja-Ganapathi-Busam: If I understand correctly, you are asking to have PACKAGECONFIG option for gst-plugin-redissink instead of a group option named GST_EXT_SOFTWARE.

Our intention was, in future if any such software plugin comes then it will be grouped with GST_EXT_SOFTWARE. Right now, we have only one plugin. By anticipating the future requirements, we created a group option.

We are okay to have individual plugin option for gst-plugin-redissink by keeping other group options as is.
Let us know your opinion.

ndechesne: the main concern is was that "EXT SOFTWARE" does not mean much in the context of your IMSDK project. Since EXT here refers to how your dependencies are managed in OpenEmbedded, so it's not right to make 'design' choices in IMSDK based on that. 

All the IMSDK configs are:

```
option(ENABLE_GST_CAMERA_PLUGINS "Enable Camera Plugins" OFF) # not enabled in opensource mainline
option(ENABLE_GST_EXT_SOFTWARE_PLUGINS "Enable Software plugins depends on external software" ${ENABLE_GST_IMSDK_PLUGINS})
option(ENABLE_GST_MESSAGING_PLUGINS "Enable Messaging plugins" ${ENABLE_GST_IMSDK_PLUGINS})
option(ENABLE_GST_ML_PLUGINS "Enable ML Plugins" ${ENABLE_GST_IMSDK_PLUGINS})
option(ENABLE_GST_VIDEOPROC_PLUGINS "Enable Video processing plugins" ${ENABLE_GST_IMSDK_PLUGINS})
option(ENABLE_GST_SOFTWARE_PLUGINS "Enable Software plugins" ${ENABLE_GST_IMSDK_PLUGINS})
```

You tried to 'group' some plugins in a logical order, but clear SW and EXT_SW do not make sense, and I would like them to be removed. The most natural thing to do would be to have one option per plugin, like Gstreamer does for its plugins set. you can potentially group plugins together if they are inter dependent, and cannot be used otherwise. 

Gstreamer uses 'ext' dependencies to build its plugins, but it's for plugins that have external dependencies to Gstreamer itself (needs a library), that is ok. 

In CAMERA plugins you have jpegenc for example, and I would expect a developer/user should be able to use this plugin without using camera potentially.

Raja-Ganapathi-Busam: 1. I agree with you to remove ext-sw and will replace with for redissink option. 
2. Initially we started with per-plugin option. Since the plugin count is more, also reviewers recommended to group them. Like, camera source plugin and jpegenc plugin depends on camera server, which further depends in camx. 
3. Grouping will help when we have a products like non-GPU SKUs which can disable GPU plugins. Another example, for edge-ai devices those doesn't require camx.

ndechesne: ok. let's at least remove ext-sw for now. we can seek for more improvements later if needed.

lumag: SGTM

ricardosalveti (approved):

recipes-multimedia/imsdk/gst-plugins-imsdk_git.bb
lumag: And now I don't see any of these dependencies, neither on GL nor on FastCV. Where are they hidden?

lumag (changes requested):

lumag:
> > > We're currently using `virtual/libgbm` for the GLES backend.
> > 
> > 
> > What do you mean? `virtual/libgbm` is not guaranteed to provide any GL ES headers.
> 
> For this we are using `REQUIRED_DISTRO_FEATURES = "opengl"`

Again, there is no guarantee that there will be GL ES headers in the build environment. You ensure that GL is enabled on the distro level and that you have GBM headers installed for you. If you need GL ES headers, then you need to list that implicitly in the depdencies list.

> 
> > > FastCV has been removed for now since it's an optional backend. The FastCV backend will only be built when the FastCV headers are available; otherwise, it will be skipped automatically.
> > 
> > 
> > Which means that FastCV backend is never being built in Yocto, since nothing will pull in headers into the sysroot.
> 
> If someone pulls, for example by `opencv` then it will be built. But we are not RDEPENDS on fastcv binaries so it will not pull fastcv binaries to the image. Since the multimedia image will not contain fastcv binaries, our backend will not work, which is expected.

So, are we now depending on the 3rd party packages to enable features in IMSDK?

> 
> Once below change is merged, then fastcv backend will not be built since it will not bring fastcv binaries to dependency list. https://lore.kernel.org/openembedded-devel/20260208211726.2971560-1-dmitry.baryshkov@oss.qualcomm.com/

Please add the depdencency explicitly, where required.

lumag (approved):

ricardosalveti (approved):

PR #1433, opened 2026-01-25
camx-dlkm: Install UAPI headers for camx userspace

gkhose-qipl:
Install the camera UAPI headers from the camx driver to allow camx userspace components to build against them without needing the kernel source trees.

### DIFF
--- a/recipes-multimedia/camx/camx-dlkm_git.bb
+++ b/recipes-multimedia/camx/camx-dlkm_git.bb
@@ -1,6 +1,6 @@
 DESCRIPTION = "Qualcomm Camera driver (CAMX)"
 HOMEPAGE = "https://github.com/qualcomm-linux/camera-driver"
-LICENSE = "GPL-2.0-only"
+LICENSE = "GPL-2.0-with-Linux-syscall-note"
 LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=b234ee4d69f5fce4486a80fdaf4a4263"
 
 SRC_URI = "git://github.com/qualcomm-linux/camera-driver.git;protocol=https;branch=camera-kernel.qclinux.0.0"


recipes-multimedia/camx/camx-dlkm_git.bb
```diff
@@ -13,6 +13,19 @@ inherit module
 MAKE_TARGETS = "modules"
 MODULES_INSTALL_TARGET = "modules_install"
 
+do_install:append() {
+    install -d ${D}${includedir}/camera/media
+    install -d ${D}${includedir}/camera_kt/media
+
+    cp -r --no-preserve=ownership \
```
lumag: Use `install -m 0644` instead.

recipes-multimedia/camx/camx-dlkm_git.bb
```diff
@@ -13,6 +13,17 @@ inherit module
 MAKE_TARGETS = "modules"
 MODULES_INSTALL_TARGET = "modules_install"
 
+do_install:append() {
+    install -d ${D}${includedir}/camera/media
+    install -d ${D}${includedir}/camera_kt/media
+
+    install -m 0644 ${S}/camera_kt/include/uapi/camera/media/* \
+        ${D}${includedir}/camera_kt/media/
```
lumag: What is kt?

recipes-multimedia/camx/camx-dlkm_git.bb
```diff
@@ -13,6 +13,17 @@ inherit module
...
+
+    install -m 0644 ${S}/camera/include/uapi/camera/media/* \
+        ${D}${includedir}/camera/media/
```
lumag: I really don't like this idea. What happens next time, when there is talos. Or amy other _new_ platform?
Will you rename camera to camera-lemans, breaking existing users? Would you add camera-talos?
Or is the lemans/monaco set of headers finally the future-proof uABI?

I took a glance at the differences.

```diff
-#define CAM_ISP_PACKET_META_MODE_SWITCH_CONFIG    16
-#define CAM_ISP_PACKET_META_CSID_LEFT             17
-#define CAM_ISP_PACKET_META_CSID_RIGHT            18
-#define CAM_ISP_PACKET_META_CSID_COMMON           19
+#define CAM_ISP_PACKET_META_CSID_LEFT             16
+#define CAM_ISP_PACKET_META_CSID_RIGHT            17
+#define CAM_ISP_PACKET_META_CSID_COMMON           18
```

Why??

```diff
-#define CAM_ISP_GENERIC_BLOB_TYPE_CSID_CONFIG               10
+#define CAM_ISP_GENERIC_BLOB_TYPE_DISCARD_INITIAL_FRAMES    10
```

The same question. Etc, etc.
So, why? What prevents the team from extending userspace interfaces instead of rewriting them? This is just broken and there is no way to fix it. Please, in future, make sure that such thing won't happen.

recipes-multimedia/camx/camx-dlkm_git.bb
```diff
@@ -13,6 +13,17 @@ inherit module
 MAKE_TARGETS = "modules"
 MODULES_INSTALL_TARGET = "modules_install"
 
+do_install:append() {
+    install -d ${D}${includedir}/camera/media
+    install -d ${D}${includedir}/camera_kt/media
```
ricardosalveti: Should we have both in a common base folder? e.g. ${includedir}/camx/...

recipes-multimedia/camx/camx-dlkm_git.bb
ricardosalveti: Yeah, can't we converge at least the interface here?

ricardosalveti:
Please no merge commits.

recipes-multimedia/camx/camx-dlkm_git.bb
gkhose-qipl: updated to camera-kodiak

recipes-multimedia/camx/camx-dlkm_git.bb
gkhose-qipl: updated to  ${includedir}/camx/...

recipes-multimedia/camx/camx-dlkm_git.bb
cjha01: Currently, camx-dlkm contains two camera sources: one for Kodiak and another for LMT. Correspondingly, on the UAPI side, we also have separate implementations—one for Kodiak and one for LMT.

I agree that there are differences and some ambiguities in the UAPI layer; however, at this moment, we cannot merge them because they are dependent on the camx userspace driver.

Going forward, if a new target is introduced, we will identify it based on its architecture and try to fit it into the existing camera structure.  For now, LMT target, we will continue using the camera UAPI, and for Kodiak, we will use camera-kodiak.

In the future, if an architecture does not fit into these two buckets, then new sources along with the corresponding UAPI will be introduced.

lumag (approved):

ricardosalveti (approved):

PR #1443, opened 2026-01-27
hexagon-dsp-binaries: set SKIP_FILEDEPS

ricardosalveti:
The Hexagon DSP libraries reference libc.so, libgcc.so, and libc++abi.so.1,
but these are provided by the DSP runtime rather than the target rootfs.
Enable SKIP_FILEDEPS on the affected packages to avoid incorrect
file-based dependency errors.


recipes-bsp/hexagon-dsp-binaries/hexagon-dsp-binaries_20260110.bb
```diff
@@ -33,6 +33,8 @@ INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
 INHIBIT_PACKAGE_STRIP = "1"
 INHIBIT_DEFAULT_DEPS = "1"
 
+SKIP_FILEDEPS = "1"
```
lumag: It should be set for each package that generates those deps, `SKIP_FILEDEPS:${PN}-foo = "1"`.

ricardosalveti: I set for all because all these packages are having the same issue, but I can have it manually set per PN as we do have for other variables.

ricardosalveti: Made change package specific, and they are not specific to adsp or cdsp, depends on the implementation done for the target hardware.

PR #1445, opened 2026-01-28
recipes-support: Add new recipe for tftp-server

Pradeep-pvk:
Add new prebuilt recipe for tftp-server to fetch Qualcomm properitary userspace application from artifactory. This application serves the Remote Files System (RFS) requests and provides additional storage space to MPSS and ADSP processors by storing data in the High Level Operating System (HLOS) fileystem.

### DIFF
--- a/conf/machine/include/qcom-common.inc
+++ b/conf/machine/include/qcom-common.inc
@@ -87,3 +87,7 @@ RT_KERNEL_CMDLINE = "${RT_ARGS_ISOL} ${RT_ARGS_IRQ} ${RT_ARGS_NOCBS} \
 KERNEL_CMDLINE_EXTRA:append = "${@bb.utils.contains_any('PREFERRED_PROVIDER_virtual/kernel',\
                               'linux-qcom-rt linux-qcom-next-rt', \
                               ' ${RT_KERNEL_CMDLINE}', '', d)}"
+
+# Set default runtime provider for virtual-tftp-server to avoid
+# warnings when both tqftpserv and tftp-server are available
+PREFERRED_RPROVIDER_virtual-tftp-server ?= "tqftpserv"


Pradeep-pvk:
> > > Please point out the ticket on tqftpserv describing the issue
> > 
> > 
> > [linux-msm/tqftpserv#33](https://github.com/linux-msm/tqftpserv/issues/33) [linux-msm/tqftpserv#34](https://github.com/linux-msm/tqftpserv/issues/34)
> 
> Please respond to Bjorn's questions.

Sure, but please do take a moment to review these changes.

conf/layer.conf
```diff
@@ -30,3 +30,7 @@ PREFERRED_RPROVIDER_virtual-diag-router ?= "diag"
 
 # Location for layer-specific Python helpers
 addpylib ${LAYERDIR}/lib qcom
+
+# Set default provider for virtual-tftp-server to avoid warnings
+# when both tqftpserv and tftp-server are available
+PREFERRED_RPROVIDER_virtual-tftp-server ?= "tftp-server"
```
ricardosalveti: The decision to switch to tftp-server should be done as a separated patch.

recipes-support/tftp-server/tftp-server_1.0.1.bb
```diff
@@ -0,0 +1,40 @@
...
+PBT_BUILD_DATE = "260127"
+SRC_URI[sha256sum] = "6dffae734607ce7bbde40a839b60960558df45796ed66dfd78a74ddf912a8cbd"
+
+S = "${UNPACKDIR}"
+
+DEPENDS += "glib-2.0 virtual/kernel qmi-framework libcap"
+RPROVIDES:${PN} += "virtual-tftp-server"
+RCONFLICTS:${PN} += "tqftpserv"
+
+FILES:${PN} += "/lib/systemd/*"
```
ricardosalveti: Should not be needed (path is also incorrect).

recipes-support/tftp-server/tftp-server_1.0.1.bb
```diff
@@ -0,0 +1,40 @@
...
+FILES:${PN} += "/etc/tmpfiles.d/*"
```
ricardosalveti: Please install the tmpfiles.d at /usr/lib/tmpfiles.d by default (${nonarch_libdir}/tmpfiles.d).

recipes-support/tftp-server/tftp-server_1.0.1.bb
```diff
@@ -0,0 +1,40 @@
...
+
+SYSTEMD_SERVICE:${PN} += "tftp_server.service"
+
+do_install() {
+	# Install all files from the prebuilt tarball
+	cp -r ${S}/* ${D}/
+}
+
+do_install:append() {
```
ricardosalveti: Why adding append if you are also defining do_install above this?

recipes-support/tftp-server/tftp-server_1.0.1.bb
```diff
@@ -13,10 +13,15 @@ SRC_URI[sha256sum] = "6dffae734607ce7bbde40a839b60960558df45796ed66dfd78a74ddf91
...
-DEPENDS += "glib-2.0 virtual/kernel qmi-framework libcap"
+DEPENDS += "glib-2.0 qmi-framework libcap"
 RPROVIDES:${PN} += "virtual-tftp-server"
 RCONFLICTS:${PN} += "tqftpserv"
 
+# This package is currently only used and tested on ARMv8 (aarch64) machines.
+# Therefore, builds for other architectures are not necessary and are explicitly excluded.
+COMPATIBLE_MACHINE = "^$"
+COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
```
ricardosalveti, line 24: Please squash your changes.

recipes-support/tftp-server/tftp-server_1.0.1.bb
anujm1: Yeah. And, if there is a tmpfiles being installed, the recipe should also have a postinst so the tmp file/dir is created when package is installed on device directly without needing a reboot.

recipes-support/tftp-server/tftp-server_1.0.1.bb
```diff
@@ -0,0 +1,45 @@
...
+SYSTEMD_SERVICE:${PN} += "tftp_server.service"
+
+do_install() {
+	# Install all files from the prebuilt tarball
+	cp -r ${S}/* ${D}/
+}
+
+do_install:append() {
+   # Ensure systemd service is properly installed
+   if ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'true', 'false', d)}; then
```
anujm1: Why are we checking for systemd? The recipe has installed the service anyway in `do_install` ..

conf/layer.conf
Pradeep-pvk: sure, i will make it as a separate patch

recipes-support/tftp-server/tftp-server_1.0.1.bb
Pradeep-pvk: sure, i will remove it

recipes-support/tftp-server/tftp-server_1.0.1.bb
Pradeep-pvk: sure, i will move all into do_install()

recipes-support/tftp-server/tftp-server_1.0.1.bb
Pradeep-pvk: sure, i will squash into the main commit

recipes-support/tftp-server/tftp-server_1.0.1.bb
Pradeep-pvk: yes, i will remove the systemd check as it already installed unconditionally.

recipes-support/tftp-server/tftp-server_1.0.1.bb
Pradeep-pvk: sure, i will address both comments

conf/layer.conf
lumag: Definitely no. This will pull in proprietary tftp-server in the open-source image.

lumag (changes requested):

conf/layer.conf
Pradeep-pvk: sure, i will remove this and use DEFAULT_PREFERENCE by setting higher preference to "tqftpserv" than "tftp-server"

recipes-support/tftp-server/tftp-server_1.0.1.bb
```diff
@@ -0,0 +1,51 @@
...
+SRC_URI = "https://softwarecenter.qualcomm.com/nexus/generic/software/chip/component/core-technologies.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/tftp-server_15.0+really${PV}_armv8a.tar.gz"
+
+PBT_BUILD_DATE = "260127"
+SRC_URI[sha256sum] = "6dffae734607ce7bbde40a839b60960558df45796ed66dfd78a74ddf912a8cbd"
+
+S = "${UNPACKDIR}"
+
+DEPENDS += "glib-2.0 qmi-framework libcap"
+RPROVIDES:${PN} += "virtual-tftp-server"
+DEFAULT_PREFERENCE = "0"
```
ricardosalveti: This should probably be -1 instead.

recipes-support/tftp-server/tftp-server_1.0.1.bb
```diff
@@ -0,0 +1,51 @@
+SUMMARY = "Prebuilt Qualcomm tftp-server application"
```
ricardosalveti, line 1: This commit should be broken down into 3 at least.

recipes-support/tftp-server/tftp-server_1.0.1.bb
```diff
@@ -0,0 +1,51 @@
...
+	if [ -f ${D}/etc/tmpfiles.d/tftp-server.conf ]; then
+		install -d ${D}${nonarch_libdir}/tmpfiles.d
+		cp ${D}/etc/tmpfiles.d/tftp-server.conf ${D}${nonarch_libdir}/tmpfiles.d/
+		rm ${D}/etc/tmpfiles.d/tftp-server.conf
+	fi
+}
+
+pkg_postinst:${PN} () {
+	if [ -n "$D" ]; then
+	# During image creation, don't run systemd-tmpfiles
```
ricardosalveti: drop comment.

recipes-support/tftp-server/tftp-server_1.0.1.bb
```diff
@@ -0,0 +1,51 @@
...
+		exit 0
+	fi
+
+        # On target device, create tmpfiles immediately
```
ricardosalveti: drop comment.

conf/layer.conf
```diff
@@ -30,3 +30,7 @@ PREFERRED_RPROVIDER_virtual-diag-router ?= "diag"
 
 # Location for layer-specific Python helpers
 addpylib ${LAYERDIR}/lib qcom
+
+# Set default runtime provider for virtual-tftp-server to avoid
+# warnings when both tqftpserv and tftp-server are available
+PREFERRED_RPROVIDER_virtual-tftp-server ?= "tqftpserv"
```
ricardosalveti: This should probably go to conf/machine/include/qcom-common.inc instead, please don't change conf/layer.conf.

recipes-support/tftp-server/tftp-server_1.0.1.bb
Pradeep-pvk: ok.

recipes-support/tftp-server/tftp-server_1.0.1.bb
Pradeep-pvk: sure, i will split into 3 changes.

recipes-support/tftp-server/tftp-server_1.0.1.bb
Pradeep-pvk: ok

recipes-support/tftp-server/tftp-server_1.0.1.bb
Pradeep-pvk: ok

conf/layer.conf
Pradeep-pvk: ok, i will move to conf/machine/include/qcom-common.inc

conf/machine/include/qcom-common.inc
```diff
@@ -87,3 +87,7 @@ RT_KERNEL_CMDLINE = "${RT_ARGS_ISOL} ${RT_ARGS_IRQ} ${RT_ARGS_NOCBS} \
 KERNEL_CMDLINE_EXTRA:append = "${@bb.utils.contains_any('PREFERRED_PROVIDER_virtual/kernel',\
                               'linux-qcom-rt linux-qcom-next-rt', \
                               ' ${RT_KERNEL_CMDLINE}', '', d)}"
+
+# Set default runtime provider for virtual-tftp-server to avoid
+# warnings when both tqftpserv and tftp-server are available
+PREFERRED_RPROVIDER_virtual-tftp-server ?= "tqftpserv"
```
lumag, line 93: This should not be necessary with the `DEFAULT_PROVIDER`, unless I misunderstand something.

ricardosalveti: It works, but bitbake will issue a note asking the user to consider setting up a PREFERRED_RPROVIDER anyway.

lumag: Then do we need the `DEFAULT_PROVIDER` at all if we define `PREFERRED_PROVIDER`?

lumag:
@Pradeep-pvk what is the path mapping that is used by these binaries?

conf/machine/include/qcom-common.inc
ricardosalveti: That's something to check.

anujm1: `DEFAULT_PREFERENCE` allows to de-select the version provided by a specific provider. So, I think both are not same. In this case, `DEFAULT_PREFERENCE` might not be helpful.

lumag:
> > > > @Pradeep-pvk what is the path mapping that is used by these binaries?
> > > 
> > > 
> > > The current binary uses the below path mapping
> 
> > Why are you providing two paths for each entry?
> 
> The first is the normalized path for each mapping path and the second is a symlink of the first path.

Why do we need a symlink?

> 
> > > For MPSS
> > > /readonly/firmware - /etc/rfs/mdm/mpss/readonly/firmware --> /usr/lib
> 
> > What kind of firmware gets requested? wlanmdsp or something else?
> 
> it would be mostly modem firmware blobs/mbn's, oem mbn's, etc. ex: modem_oem/qdsp6_oemsw.mbn

Okay. So there are two usecases:
- actual modems, requesting extra data
- ath10k WiFi, requesting wlanmdsp.mbn

Loading `/readonly` from the fixed location breaks the latter usecase (wlanmdsp.mbn is SoC and might be machine-specific) and will most likely not work with the former usecase in case of a multi-machine rootfs configuration, which must be supported.

> 
> > > /readwrite/ - /etc/rfs/mdm/mpss/readwrite --> /etc/persist/rfs/mdm/mpss
> 
> > This should be /var at least
> 
> yes, it was under /var initially but due to some new requirement( OSTree for performing OTA) we were asked not to install anything under /var, /opt, /mnt, /home paths.

Rootfs can be read-only. All state files must go to /var  (and anyway, you are not installing anything, it gets stored there, like any other stored file).

> 
> > > /ramdumps/ - /etc/rfs/mdm/mpss/ramdumps --> /etc/tombstones/modem
> 
> > Ramdumps definitely should not go to /etc.
> > yes, it is due to the above reason same comment as above.

Still, nope.

> 
> > > /shared/ - /etc/rfs/mdm/mpss/shared --> /etc/persist/rfs/shared
> > > /hlos - /etc/rfs/mdm/mpss/hlos --> /etc/persist/hlos_rfs/shared
> > 
> > 
> > What is the usecase for those items?
> 
> For /hlos - AFAIR, it will be will be used by location services, xtwifi download use-cases, OIS PD etc. ex: filenames would look something like gdt/RcvdTdpFile.txt

I haven't seen this. Which platforms use this? Are they to be provided by the host or are they stored by the modem?


> 
> For /shared - contents or data that can be shared among multiple subsystems would be placed under this path. ex: handshake information server_info.txt

What is it? Is it stored by the modem or is it to be provided by the host?

> 
> > > For ADSP /readonly/firmware - /etc/rfs/mdm/adsp/readonly/firmware --> /usr/lib
> > 
> > 
> > What kind of firmwre is being loaded by ADSP?
> 
> it would be audio calibration data, sensors registry etc.

calibration data - unsure. Sensors registry should be loaded from /usr/share/qcom/SoC/Vendor/Device/sensors/.

> 
> > > /readwrite/ - /etc/rfs/adsp/mpss/readwrite --> /etc/persist/rfs/mdm/adsp
> > > /ramdumps/ - /etc/rfs/adsp/mpss/ramdumps --> /etc/tombstones/lpass
> > > /shared/ - /etc/rfs/mdm/adsp/shared --> /etc/persist/rfs/shared
> > > /hlos - /etc/rfs/mdm/adsp/hlos --> /etc/persist/hlos_rfs/shared
> > 
> > 
> > Same comments apply.
> > What about CDSP and SLPI?
> 
> yes, it can have similar mappings for CDSP and SLPI as well but currently it is limited to ADSP and MPSS on QLI.

LImited by whom? We support SLPI on MSM8998-SM8450.

Pradeep-pvk:
> > > > > @Pradeep-pvk what is the path mapping that is used by these binaries?
> > > > 
> > > > 
> > > > The current binary uses the below path mapping
> > 
> > 
> > > Why are you providing two paths for each entry?
> > 
> > 
> > The first is the normalized path for each mapping path and the second is a symlink of the first path.
> 
> Why do we need a symlink?
> 
Each HLOS (LA, Windows) have their own way of storing executable-images, data, ramdumps and also their own file-access-permission modes. So the RFS-modules files location in the HLOS will vary from HLOS to HLOS and hence the symlink to the actual path on device for a given normalized rfs path.

> > > > For MPSS
> > > > /readonly/firmware - /etc/rfs/mdm/mpss/readonly/firmware --> /usr/lib
> > 
> > 
> > > What kind of firmware gets requested? wlanmdsp or something else?
> > 
> > 
> > it would be mostly modem firmware blobs/mbn's, oem mbn's, etc. ex: modem_oem/qdsp6_oemsw.mbn
> 
> Okay. So there are two usecases:
> 
> * actual modems, requesting extra data
> * ath10k WiFi, requesting wlanmdsp.mbn
> 
> Loading `/readonly` from the fixed location breaks the latter usecase (wlanmdsp.mbn is SoC and might be machine-specific) and will most likely not work with the former usecase in case of a multi-machine rootfs configuration, which must be supported.
> 

The tftp_server can still support this—it is capable of handling absolute paths requested by the client, as shown in the example above (e.g., wlanmdsp.mbn), where clients prefix the request with /root/<path to wlanmdsp.mbn>/

> > > > /readwrite/ - /etc/rfs/mdm/mpss/readwrite --> /etc/persist/rfs/mdm/mpss
> > 
> > 
> > > This should be /var at least
> > 
> > 
> > yes, it was under /var initially but due to some new requirement( OSTree for performing OTA) we were asked not to install anything under /var, /opt, /mnt, /home paths.
> 
> Rootfs can be read-only. All state files must go to /var (and anyway, you are not installing anything, it gets stored there, like any other stored file).
> 
> > > > /ramdumps/ - /etc/rfs/mdm/mpss/ramdumps --> /etc/tombstones/modem
> > 
> > 
> > > Ramdumps definitely should not go to /etc.
> > > yes, it is due to the above reason same comment as above.
> 
> Still, nope.
> 
> > > > /shared/ - /etc/rfs/mdm/mpss/shared --> /etc/persist/rfs/shared
> > > > /hlos - /etc/rfs/mdm/mpss/hlos --> /etc/persist/hlos_rfs/shared
> > > 
> > > 
> > > What is the usecase for those items?
> > 
> > 
> > For /hlos - AFAIR, it will be will be used by location services, xtwifi download use-cases, OIS PD etc. ex: filenames would look something like gdt/RcvdTdpFile.txt
> 
> I haven't seen this. Which platforms use this? Are they to be provided by the host or are they stored by the modem?
> 
AFAIK, /hlos RFS path will be used on all platforms. The above is just an example (RcvdTdpFile.txt) taken from a Pinnacles MBB target and unsure of this file presence/creation on other targets.

> > For /shared - contents or data that can be shared among multiple subsystems would be placed under this path. ex: handshake information server_info.txt
> 
> What is it? Is it stored by the modem or is it to be provided by the host?
> 
/shared - This folder will be shared by all the qcom-processors (MPSS, ADSP and other peripheral processors). 
i mentioned "server_info.txt" as an example, it is a handshake file created by tftp server during tftp client - server communication and will be seen on all qcom targets.

> > > > For ADSP /readonly/firmware - /etc/rfs/mdm/adsp/readonly/firmware --> /usr/lib
> > > 
> > > 
> > > What kind of firmwre is being loaded by ADSP?
> > 
> > 
> > it would be audio calibration data, sensors registry etc.
> 
> calibration data - unsure. Sensors registry should be loaded from /usr/share/qcom/SoC/Vendor/Device/sensors/.
> 
tftp_server can still support the above path, provided the client requests the actual absolute path prefixed with root.

> > > > /readwrite/ - /etc/rfs/adsp/mpss/readwrite --> /etc/persist/rfs/mdm/adsp
> > > > /ramdumps/ - /etc/rfs/adsp/mpss/ramdumps --> /etc/tombstones/lpass
> > > > /shared/ - /etc/rfs/mdm/adsp/shared --> /etc/persist/rfs/shared
> > > > /hlos - /etc/rfs/mdm/adsp/hlos --> /etc/persist/hlos_rfs/shared
> > > 
> > > 
> > > Same comments apply.
> > > What about CDSP and SLPI?
> > 
> > 
> > yes, it can have similar mappings for CDSP and SLPI as well but currently it is limited to ADSP and MPSS on QLI.
> 
> LImited by whom? We support SLPI on MSM8998-SM8450.

sorry for the confusion, yes it supports SLPI and CDSP paths too.

lumag (changes requested):
After discussion with other meta-qcom maintainers and with the platform maintainers.
- tftpserver doesn't seem to provide benefits for the QLI platforms, which are usually modemless and load sensors registry and sensors configuration through sdsprpcd,
- it will break the usecase of loading ath10k wlanmdsp.mbn firmware file from the existing known locations,
- it seems to provide DSPs with the access to undocumented locations, including possibly complete rootfs access, leading to possible security implications.

As such, I'd kindly ask to continue discussion with @bamse and @quic-kdybcio regarding missing tqftpserv features and spend efforts on improving tqftpserv instead of trying to push and plumb the tftpserver into meta-qcom.

PR #1450, opened 2026-01-28
ci: Use GitHub Actions concurrency to cancel outdated PR workflow runs

sampra2025:
**Background**

By default, GitHub Actions allows multiple workflow runs to execute concurrently, including multiple runs triggered for the same pull request. While this can be beneficial in some cases, our workflows are relatively extensive and resource‑intensive.
We’ve observed that developers often trigger multiple workflow runs on the same PR due to force pushes or rapid successive updates. This results in several outdated runs executing in parallel, even though only the latest commit state is relevant.

**Problem**

Running multiple workflows in parallel for the same PR:

    - Consumes unnecessary CI resources
    - Executes jobs against outdated code that will never be merged
    - Increases noise and wait times without providing additional value

Without context, parallel execution may seem beneficial, but in our case it does not improve feedback quality and instead creates inefficiency.

**Proposed Change**

This PR introduces GitHub Actions concurrency to ensure that only the most recent workflow run for a given PR is executed. When a new commit is pushed to the PR, any in‑progress or queued runs for previous commits are automatically canceled.

**Benefits**

Ensures CI runs are always aligned with the latest PR state
Reduces wasted compute and execution time
Improves signal‑to‑noise ratio for developers and reviewers
Provides faster, more relevant feedback on active changes

This change is intended to make our CI behavior more predictable and efficient, given how our workflows are currently used.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -12,6 +12,11 @@ on:
         description: "URL to retrieve build artifacts"
         value: ${{ jobs.create-output.outputs.url }}
 
+# Concurrency is used to prevent multiple workflows from running for the same PR
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
   CACHE_DIR: /efs/qli/meta-qcom
   KAS_REPO_REF_DIR: /efs/qli/meta-qcom/kas-mirrors


sampra2025:
> > I am sharing the testing results I conducted in the stage environment, which have been documented in the JIRA ticket. 
> 
> This is an open development repo, please don't post results as links to the non-public tickets.

OOps.. I have update my comments and removed link.

.github/workflows/build-yocto.yml
```diff
@@ -12,6 +12,11 @@ on:
         description: "URL to retrieve build artifacts"
         value: ${{ jobs.create-output.outputs.url }}
 
+# Councurrency to prevent running multiple workflow for same PR
```
ricardosalveti: Typo, Concurrency

ricardosalveti:
Please wrap your git commit lines to most 80 columns.

sampra2025:
[Test result for Concurrency.pdf](https://github.com/user-attachments/files/24946280/Test.result.for.Concurrency.pdf)

lumag:
Please improve your commit message: First describe the issue. What and why is wrong. Describe the actions undertaken in the commit only after describing the problem.

ndechesne:
> > Please improve your commit message: First describe the issue. What and why is wrong. Describe the actions undertaken in the commit only after describing the problem.
> 
> I believe the commit message title clearly states the issue and the required fix. The commit body provides additional details about the default GitHub behavior and the proposed solution. Please let me know if I missed anything.

@sampra2025 , I tend to agree with @lumag . I open this PR to check, and I do not understand why you are sending the PR. The commit explains what the changes does, but it does not explain why you are proposing it. This is usually the important information that needs to be captured in the project history. So that 1) people like me who review understand why it is important to merge, 2) people later when looking at the history understand. 

in this specific case, why do you want to prevent multiple workflows in parallel. What problem does it create which needs to be fixed? Without additional information, I would think it's a good thing to run each PR workflow in parallel, to test more things at the same time.

ricardosalveti:
> why do you want to prevent multiple workflows in parallel. What problem does it create which needs to be fixed? Without additional information, I would think it's a good thing to run each PR workflow in parallel, to test more things at the same time.

Right, we can say that our runs are quite extensive and that we are seeing developers triggering multiple runs due force pushes in PRs, so canceling older jobs when there are PR updates is a sane thing to do here.

Basically what caused the proposal of this change.

sampra2025:
> > @ricardosalveti / @ndechesne, thanks for your input, I tried my best to describe on why this change is required. Please, let me know if this description helps or need any further correction, as I am eagerly waiting to get this change merge. This will help to avoid duplicate runs for forced pushes.
> 
> I do not see any chance in the commit message.

I have updated the commit message as well. Let me know if we still need any further corrections.

PR #1456, opened 2026-01-30
Fetch RB3Gen2 Industrial and Vision kit CDTs

vkraleti:
Fetch CDT files for RB3Gen2 Industrial Kit and RB3Gen2 Vision Kit 
boards andretain them in qcomflash directory to extend support 
for additional RB3Gen2 platform variants.

### DIFF
--- a/classes-recipe/image_types_qcom.bbclass
+++ b/classes-recipe/image_types_qcom.bbclass
@@ -100,6 +100,7 @@ create_qcomflash_pkg() {
                 \( -name '*.elf' ! -name 'abl2esp*.elf' ! -name 'xbl_config*.elf' \) -o \
                 -name '*.mbn*' -o \
                 -name '*.fv' -o \
+                -name 'cdt_*.bin' -o \
                 -name 'logfs_*.bin' -o \
                 -name 'sec.dat'` ; do
             install -m 0644 ${bfw} .


lumag:
> > @vkraleti I keep on saying it, but really dislike this idea. Are we going to have CDT for for all 3rd party mezzanines? For combinations of Vision + 3rd-party-mezz? I think, we should use UEFI variable to specify the mezzanines to be selected.
> 
> As of now, Core Kit & Vision/Industrial Mezz Kits are considered as different H/W not as plug-n-play devices.

That doesn't quite match Thundercomm's website, where I can [purchase](https://www.thundercomm.com/product/qualcomm-rb3-gen-2/?add-to-cart=22249) Vision mezz+bracket separately as an addon for the Core kit.

What about someone buying TDK mezzanine and plugging it into RB3 Gen2? Will it get its own CDT? Will there be a separate CDT for Core+TDK and Vision+TDK? Or will TDK be supported in some other way?

And, most important, what about 3rd party mezzanines?

> Hence separate CDTs exist. Moving away from CDT to UEFI variable based checks is a big change and multiple teams need to align and has to be brought in phased manner post alignment. Till then these boards must be serviced from meta-qcom in existing manner. I initiated an internal email thread with the relevant stakeholders for further discussion.

Thanks!

I'm fine with landing this as a temporal measure, but we need to sort out handling of mezzanines.

> 
> @pareekpa @shashim-quic @sbanerjee-quic FYI.

vkraleti:
> What is the RB3Gen2 Industrial Kit? Is it publicly available?

No it isn't for public consumption but for specific customers usage. Support of this board is available in 1.0 and Internally teams are using these boards.

PR #1458, opened 2026-01-30
ci: add KAS configuration for CI

quaresmajose:
We will drop the bitbake rm_work bbclass on ci builds but we'll keep it on default
base.yml config for our users because it significantly facilitates debugging anomalies.

On the CI, instead of removing from the file system what we no longer
need at the end of each bitbake task we can drop everything at the end
of the build when the worker is destroyed.

This will improve the overall build time of [1] by 15% in the best case, when everything
comes from the sstate-cache:

rm_work: yes

real    3m21.962s
user    0m0.204s
sys     0m0.195s

rm_work: no

real    2m52.959s
user    0m0.155s
sys     0m0.159s

[1] time ci/kas-container-qcom.sh build ci/rb3gen2-core-kit.yml:ci/world.yml:ci/ci.yml

### DIFF
--- a/ci/ci.yml
+++ b/ci/ci.yml
@@ -7,3 +7,4 @@ header:
 
 local_conf_header:
   ci: |
+    INHERIT:remove = "rm_work"


ci/base.yml
```diff
@@ -31,7 +31,6 @@ local_conf_header:
   base: |
     CONF_VERSION = "2"
     INHERIT += "buildstats-summary"
-    INHERIT += "buildhistory"
```
ndechesne, line 29: i believe this is an important feature for anyone building yocto, which we are silently removing here. I would prefer not to do that.

quaresmajose: Ok, I'll leave it as it was before.

ricardosalveti:
Curious to see if this will cause any build time difference.

koenkooi (approved):

koenkooi:
I still have reservations about disabling it in the CI, since` rm_work` is very good at uncovering issues, but the speed gain could be worth it.

quaresmajose:
The values ​​measured in the CI are difficult to quantify because they can be altered by various factors. Therefore, I opted to measure the times locally in a more controlled environment and using a specific target.

PR #1459, opened 2026-01-31
CAMX: Update to the 1.0.8 revision

gkhose-qipl:
camxlib-kodiak:  Introduce the development package to export necessary headers required by GST plugin and QMMF.
camxlib-lemans:Generate sensor binary files for Lemans and Monaco based on differences in slotId/cameraId mappings or module‑file XML tag value entries.


lumag (changes requested):
Please switch from dates to PV changes in the commit messages. In the end, all outside world would identify your packages as CamX 1.0.something rather than 20260130.

gkhose-qipl:
> Please switch from dates to PV changes in the commit messages. In the end, all outside world would identify your packages as CamX 1.0.something rather than 20260130.

done

lumag (approved):

vkraleti (approved):

PR #1461, opened 2026-02-01
Support multiple compatible strings per DTB

vkraleti:
A single DTB may be shared across multiple SoC variants and therefore
requires multiple "compatible" strings. Extend the qcom dtb-only FIT
image implementation to handle multiple compatible strings per DTB and
emit a separate FIT config entry for each string.

When multiple compatible strings are defined, they must be delimited
using spaces, for example:
  FIT_DTB_COMPATIBLE[dtb-name] = "compat1 compat2 compat3"

### DIFF
--- a/conf/machine/include/fit-dtb-compatible.inc
+++ b/conf/machine/include/fit-dtb-compatible.inc
@@ -8,7 +8,10 @@ FIT_DTB_COMPATIBLE[apq8016-sbc] = "qcom,apq8016-sbc"
 FIT_DTB_COMPATIBLE[apq8096-db820c] = "qcom,apq8096-sbc"
 FIT_DTB_COMPATIBLE[kaanapali-mtp] = "qcom,kaanapali-mtp"
 FIT_DTB_COMPATIBLE[hamoa-iot-evk] = "qcom,hamoa-evk"
-FIT_DTB_COMPATIBLE[lemans-evk] = "qcom,qcs9075v2-iot"
+FIT_DTB_COMPATIBLE[lemans-evk] = " \
+                                   qcom,qcs9075-iot \
+                                   qcom,qcs9075v2-iot \
+"
 FIT_DTB_COMPATIBLE[lemans-evk-camera-csi1-imx577] = "qcom,lemans-evk"
 FIT_DTB_COMPATIBLE[lemans-evk-camx] = "qcom,qcs9075v2-iot-camx"
 FIT_DTB_COMPATIBLE[monaco-evk] = "qcom,qcs8275-iot"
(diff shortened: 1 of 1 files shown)


conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -1,7 +1,9 @@
-# Add compatible strings for DTBs.
-# FIT_DTB_COMPATIBLE[<dtb_name>] = "<compatible-string>"
-# This variable maps a DTB name (without .dtb extension) to its
-# Device Tree "compatible" string.
+# Mapping from a DTB name (without the .dtb extension) to its
+# Device Tree "compatible" string(s). When present multiple
+# compatible strings must be separated by colons (:).
+#
+# Syntax:
+# FIT_DTB_COMPATIBLE[<dtb-name>] = "<compatible-string>[:<compatible-string>[:<compatible-string>...]]"
```
lumag: Let's keep it more readable (and more in style with Yocto) and use whitespaces to separate items.

lib/qcom/dtb_only_fitimage.py
```diff
@@ -69,12 +69,17 @@ def fitimage_emit_section_dtb(self, dtb_id, dtb_path,
...
+        compatibles_count = 0
         for counter, (dtb_node, compatible_str) in enumerate(self._dtbs):
             # conf-0 corresponds to qcom-metadata.dtb, skipping it as it's metadata-only
             if counter == 0:
                 continue
-            conf_name = f"{self._conf_prefix}{counter}"
-            self._fitimage_emit_one_section_config(conf_name, dtb_node)
+            # add one config for each comaptible string of DTB
+            for compatible in compatible_str.split(":"):
+                compatibles_count += 1
```
lumag: Why do we need a new counter? Can't we continue using the existing one?

conf/machine/include/fit-dtb-compatible.inc
vkraleti: > Let's keep it more readable (and more in style with Yocto) and use whitespaces to separate items. 

Any examples to take a look at?

lumag: > > Let's keep it more readable (and more in style with Yocto) and use whitespaces to separate items.
> 
> Any examples to take a look at?

`KERNEL_DEVICETREE` and any other variable?

I meant a really simple thing:

```FIT_DTB_COMPATIBLE[<dtb-name>] = "<compatible-string>[ <compatible-string>[   <compatible-string>...]]"```

Like:
```
FIT_DTB_COMPATIBLE[lemans-evk] = " \
    qcom,qcs9075-iot \
   qcom,qcs9075v2-iot \
"
```

File: lib/qcom/dtb_only_fitimage.py (hunk 2 above)
Comment (vkraleti) [reply lib/qcom/dtb_only_fitimage.py]: There are two loops here. 
1) Outer loop iterates over the list of DTB nodes
2) Inner loop iterates over the list of compatible strings within each node.

Because both loops require independent indices, need two counters.  Do you recommend to generate configs as 
```
conf-1-1 {
...
};
conf-2-1 {
...
};
conf-2-2 {
...
};
conf-2-3 {
...
};
conf-3-1 {
...
};
```

instead of 

```
conf-1 {
...
};
conf-2 {
...
};
conf-3 {
...
};
conf-4 {
...
};
```

Comment (lumag) [reply lib/qcom/dtb_only_fitimage.py]: No, I was suggesting the latter one. But, you don't need to enumerate DTB entries, do you?

Comment (vkraleti) [reply lib/qcom/dtb_only_fitimage.py]: Need to enumerate DTB entries as well to update fdt name in conf For example

```
conf-2 {
	compatible = "qcom,qcs6490-iot";
	fdt = "fdt-qcs6490-rb3gen2.dtb";   <--- to get this value need to loop over list of DTB nodes.
}; 
```

Sample its: https://github.com/qualcomm-linux/qcom-dtb-metadata/blob/main/qcom-fitimage.its#L66

File: lib/qcom/dtb_only_fitimage.py [hunk 3]
```diff
@@ -68,13 +68,17 @@ def fitimage_emit_section_dtb(self, dtb_id, dtb_path,
         self._dtbs.append((dtb_node, compatible_str))
 
     def fitimage_emit_section_config(self):
-        counter = 0
+        conf_counter = 0
         for counter, (dtb_node, compatible_str) in enumerate(self._dtbs):
             # conf-0 corresponds to qcom-metadata.dtb, skipping it as it's metadata-only
             if counter == 0:
```
lumag: Please replace this with the node check, then you can drop `counter`.

vkraleti: Done. Updated code to use only one counter to increment config name.

lumag (dismissed):

lib/qcom/dtb_only_fitimage.py
```diff
@@ -69,12 +69,16 @@ def fitimage_emit_section_dtb(self, dtb_id, dtb_path,
...
-        for counter, (dtb_node, compatible_str) in enumerate(self._dtbs):
-            # conf-0 corresponds to qcom-metadata.dtb, skipping it as it's metadata-only
-            if counter == 0:
+        for dtb_node, compatible_str in self._dtbs:
+            # qcom-metadata don't need any config entry
+            if dtb_node.properties.get("type") == "qcom_metadata":
                 continue
-            conf_name = f"{self._conf_prefix}{counter}"
-            self._fitimage_emit_one_section_config(conf_name, dtb_node)
+            # add one config for each comaptible string of DTB
```
ricardosalveti: typo, compatible

ricardosalveti (dismissed):

sbanerjee-quic (approved):

lib/qcom/dtb_only_fitimage.py
vkraleti: @ricardosalveti Corrected the typo. Can you please review one more time?

ricardosalveti (approved):

PR #1467, opened 2026-02-02
linux-qcom-next: update to tag qcom-next-6.19-rc6-20260130

fangwu1:
Move to the latest linux-qcom-next release tag qcom-next-6.19-rc6-20260205

### DIFF
--- a/recipes-kernel/linux/linux-qcom-next_git.bb
+++ b/recipes-kernel/linux/linux-qcom-next_git.bb
@@ -8,12 +8,12 @@ inherit kernel cml1
 
 COMPATIBLE_MACHINE = "(qcom)"
 
-LINUX_VERSION ?= "6.18+6.19-rc4"
+LINUX_VERSION ?= "6.18+6.19-rc6"
 
 PV = "${LINUX_VERSION}+git"
 
-# tag: qcom-next-6.19-rc4-20260112
-SRCREV ?= "65bb05266d477c9f7b06bc275bd622dbed08a53b"
+# tag: qcom-next-6.19-rc6-20260207
+SRCREV ?= "ccc1345fb4742e377d7d89658efe085c62c25164"
 
 SRCBRANCH ?= "nobranch=1"
 SRCBRANCH:class-devupstream ?= "branch=qcom-next"


lumag (changes requested):
NAK, the UFS issue is still not fixed: https://lava.infra.foundries.io/scheduler/job/137538

shoudil:
> NAK, the UFS issue is still not fixed: https://lava.infra.foundries.io/scheduler/job/137538

For the issue specific to target and get pending long time, could we consider to disable CI on it temporally as the kernel tag is consumed by all enabled targets ?  Currently both Talos EVK and Purwa EVK are also waiting to consume the new qcom-next tag.

lumag:
> 
> Disabling console logging or setting loglevel to 6 (currently it is at 7) is helping and not reproducing issue in 30+ reboots.
> 
> @lumag I think loglevel 7 is too verbose and we need to set it to 6.

What if I need to enable debug logging or `CONFIG_DEBUG_DRIVER`? Stop the nonsense. Please fix the driver instead of proposing workarounds.

Mani-Sadhasivam:
@lumag @shashim-quic I don't believe that this issue has anything to do with the UFS driver, even though UFS driver is the victim here. The core issue here is that during boot, all the interrupts are pinned to CPU0 (default affinity) and some driver(s) take the spinlock with local IRQs disabled and keep it so long or the interrupt delivery is getting delayed that it creates starvation for other drivers like UFS. Ofc, we can increase the timeout in the UFS driver to wait a bit more for the completion interrupt to arrive, but that's just a band-aid and I don't want to have such change in place.

We should properly trace the irqs during boot to find out which drivers are holding the lock for so long or why the interrupt delivery is delayed and fix them instead. But it is also surprising to see this issue only on specific platforms. So that could provide some hint about this being a platform specific issue (maybe HYP?).

Re. PCIe fixup (which is also a hack), the rescan from global IRQ thread is already removed for 7.0 in upstream, so once we move to next release, this change will be brought in.

shashim-quic:
> 



> @lumag @shashim-quic I don't believe that this issue has anything to do with the UFS driver, even though UFS driver is the victim here. The core issue here is that during boot, all the interrupts are pinned to CPU0 (default affinity) and some driver(s) take the spinlock with local IRQs disabled and keep it so long or the interrupt delivery is getting delayed that it creates starvation for other drivers like UFS. Ofc, we can increase the timeout in the UFS driver to wait a bit more for the completion interrupt to arrive, but that's just a band-aid and I don't want to have such change in place.
> 
> We should properly trace the irqs during boot to find out which drivers are holding the lock for so long or why the interrupt delivery is delayed and fix them instead. But it is also surprising to see this issue only on specific platforms. So that could provide some hint about this being a platform specific issue (maybe HYP?).
> 
> Re. PCIe fixup (which is also a hack), the rescan from global IRQ thread is already removed for 7.0 in upstream, so once we move to next release, this change will be brought in.

As mentioned before, console log msgs during boot up are causing the delays that misses ufs threaded irq scheduling.
**During ufs time critical path,  cpu call stack shows pcie probe that is printing logs.**

There is no issue with raising of interrupt. Primary ufs handler is called in time, threaded irq is delayed (established from traces).

### Case1: Default log level (7): UFS is timed out
[    0.270279] qnoc-sa8775p interconnect-pcie-anoc: error -EINVAL: invalid resource (null)
[    0.279029] qcom-pcie 1c00000.pcie: supply vdda not found, using dummy regulator
[    0.279073] qcom-pcie 1c10000.pcie: supply vdda not found, using dummy regulator
[    0.279077] qcom-pcie 1c00000.pcie: supply vddpe-3v3 not found, using dummy regulator
[    0.279115] qcom-pcie 1c10000.pcie: supply vddpe-3v3 not found, using dummy regulator
[    0.279195] qcom-pcie 1c00000.pcie: host bridge /pcie@1c00000 ranges:
[    0.279202] qcom-pcie 1c10000.pcie: host bridge /pcie@1c10000 ranges:
[    0.279209] qcom-pcie 1c10000.pcie:       IO 0x0060200000..0x00602fffff -> 0x0000000000
[    0.279213] qcom-pcie 1c10000.pcie:      MEM 0x0060300000..0x007fffffff -> 0x0060300000
[    0.279231] qcom-pcie 1c00000.pcie:       IO 0x0040200000..0x00402fffff -> 0x0000000000
[    0.279235] qcom-pcie 1c00000.pcie:      MEM 0x0040300000..0x005fffffff -> 0x0040300000
[    0.390216] qcom-pcie 1c10000.pcie: iATU: unroll T, 32 ob, 8 ib, align 4K, limit 1024G
[    0.390223] qcom-pcie 1c00000.pcie: iATU: unroll T, 32 ob, 8 ib, align 4K, limit 1024G
[    0.390429] qcom-pcie 1c00000.pcie: PCI host bridge to bus 0000:00
[    0.390481] pci 0000:00:00.0: [17cb:0115] type 01 class 0x060400 PCIe Root Port
[    0.393128] pcieport 0000:00:00.0: PME: Signaling with IRQ 261
[    0.397085] pcieport 0000:00:00.0: AER: enabled with IRQ 256
[    0.600857] qcom-pcie 1c10000.pcie: PCIe Gen.3 x2 link up
[    0.619605] qcom-pcie 1c10000.pcie: PCI host bridge to bus 0001:00
[    0.654606] pci 0001:00:00.0: [17cb:0115] type 01 class 0x060400 PCIe Root Port
[    6.398405] pci 0001:01:00.0: [1d6a:07b1] type 00 class 0x020000 PCIe Endpoint
[    6.449003] pci 0001:01:00.0: 15.752 Gb/s available PCIe bandwidth, limited by 8.0 GT/s PCIe x2 link at 0001:00:00.0 (capable of 31.504 Gb/s with 8.0 GT/s PCIe x4 link)
[    6.545327] pcieport 0001:00:00.0: PME: Signaling with IRQ 268
[    6.571238] pcieport 0001:00:00.0: AER: enabled with IRQ 263
[   11.096368] pci 0000:01:00.0: [17cb:1103] type 00 class 0x028000 PCIe Endpoint
[   11.404384] pcieport 0000:00:00.0: bridge window [mem 0x40400000-0x405fffff]: assigned

### Case2: With loglevel 6: UFS is not timed out
 [    0.133455] qnoc-sa8775p interconnect-pcie-anoc: error -EINVAL: invalid resource (null)
[    0.140292] qcom-pcie 1c00000.pcie: supply vdda not found, using dummy regulator
[    0.140335] qcom-pcie 1c00000.pcie: supply vddpe-3v3 not found, using dummy regulator
[    0.140772] qcom-pcie 1c10000.pcie: supply vdda not found, using dummy regulator
[    0.140807] qcom-pcie 1c10000.pcie: supply vddpe-3v3 not found, using dummy regulator

Mani-Sadhasivam:
> [   15.635707] platform 18591000.cpufreq: deferred probe pending: qcom-cpufreq-hw: Failed to find icc paths

@shashim-quic Looks like CPUFreq driver is not probed during boot. So all the CPUs would be running with the bootloader configured frequency, which might add up to the latency. The cpufreq driver probe failure likely comes from missing CONFIG_INTERCONNECT_QCOM_OSM_L3 driver during boot. Try make it as a built-in and ensure that the cpufreq-hw driver is probed during boot.

sbanerjee-quic (approved):

Mani-Sadhasivam:
@shashim-quic

Can you try this patch and see if it addresses the issue?

```
diff --git a/drivers/ufs/core/ufshcd.c b/drivers/ufs/core/ufshcd.c
index 52ffd0c3aa4c..84ed721ea969 100644
--- a/drivers/ufs/core/ufshcd.c
+++ b/drivers/ufs/core/ufshcd.c
@@ -10810,6 +10810,7 @@ int ufshcd_init(struct ufs_hba *hba, void __iomem *mmio_base, unsigned int irq)
        int err;
        struct Scsi_Host *host = hba->host;
        struct device *dev = hba->dev;
+       struct cpumask mask;
 
        /*
         * dev_set_drvdata() must be called before any callbacks are registered
@@ -10965,6 +10966,17 @@ int ufshcd_init(struct ufs_hba *hba, void __iomem *mmio_base, unsigned int irq)
                hba->is_irq_enabled = true;
        }
 
+       /*
+        * If there are more than one online CPUs, then spread the IRQ affinity
+        * across the available non-boot CPUs. This is done to avoid hogging the
+        * boot CPU which gets overloaded most of the time during boot.
+        */
+       if (num_online_cpus() > 1) {
+               cpumask_copy(&mask, cpu_online_mask);
+               cpumask_clear_cpu(0, &mask);
+               irq_set_affinity(irq, &mask);
+       }
+
        /* Reset the attached device */
        ufshcd_device_reset(hba);
```

This patch effectively spreads the IRQ affinity for UFS interrupt across the non-boot CPUs to avoid hogging the boot CPU (CPU0). I still need to investigate the hogging though.

lumag:
Colleagues, please shift patches and corresponding discussions to the issues on GH rather than this PR.

lumag:
> > Doesn't work on RB3 Gen2. NAK.
> 
> It did boot on the latest run it seems, let me check locally to see how often it fails.

I restarted it twice here without any issues, nor could I reproduce it on my board. I assume it might be one-off on the particular board.

PR #1470, opened 2026-02-02
camx-dlkm: Update to the latest revision

cjha01:
This change includes the following updates:

- Fix CSID, CAMIF, and VFE header handling for QCS615 platforms.
- Add handling for component overwrite error on VFE bus.
- Ignore primary SOF timestamp address read for QCS615.
- Support RDI‑only context configuration on non‑EPOCH targets.
- Correct a conditional branch that was always evaluating false.
- Enable sysfs utility configuration flags for QCS615.
- Fix overflow in I²C write time‑delay variable.
- Correct DMA reset sequencing logic.
- Enable Talos camera support.


vkraleti (approved):

ricardosalveti (approved):
Thanks for providing the changelog.

PR #1473, opened 2026-02-03
uboot: Add U-Boot bootloader support for qcm6490 & qcs615 platforms

aswinm94:
Integrate U-Boot as an alternative bootloader to UEFI with conditional selection via KAS configuration. Changes include:

- Add u-boot-qcom.yml KAS config for bootloader selection
- Add u-boot-qcom-next recipe sourcing uboot from Qualcomm Linux repo
  * Overrides default u-boot recipe from oe-core layer
  * Builds from git://github.com/qualcomm-linux/u-boot.git
  * Integrates qtestsign for signing (u-boot.elf → u-boot.mbn)
- Add qcom-partition-conf bbappend for automatic XML modification
- Conditionally replace uefi.elf with u-boot.mbn in partition XMLs

preferred bootloader can be switched between UEFI and U-Boot by changing the KAS configuration file.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -219,6 +219,22 @@ jobs:
                 type: default
                 dirname: ""
                 yamlfile: ""
+          - machine: qcs615-ride
+            distro:
+                name: poky-altcfg
+                yamlfile: ''
+            kernel:
+                type: u-boot-qcom-6.18
+                dirname: "_linux-qcom-6.18_u-boot-qcom"
+                yamlfile: ":ci/linux-qcom-6.18.yml:ci/u-boot-qcom.yml"
+          - machine: rb3gen2-core-kit
+            distro:
+                name: poky-altcfg
+                yamlfile: ''
+            kernel:
+                type: u-boot-qcom-6.18
+                dirname: "_linux-qcom-6.18_u-boot-qcom"
+                yamlfile: ":ci/linux-qcom-6.18.yml:ci/u-boot-qcom.yml"
     name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
     steps:
       - uses: actions/checkout@v4


ci/u-boot-qcom.yml
```diff
@@ -0,0 +1,7 @@
+header:
+  version: 14
+
+local_conf_header:
+  bootloaderprovider: |
+    PREFERRED_PROVIDER_virtual/bootloader = "u-boot-qcom-next-upstream"
+    IMAGE_INSTALL:append = " u-boot-qcom-next-upstream"
```
lumag: Separate commit

conf/machine/qcs615-ride.conf
```diff
@@ -21,3 +21,5 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
 QCOM_CDT_FILE = "cdt_adp_air_sa6155p"
 QCOM_BOOT_FILES_SUBDIR = "qcs615"
 QCOM_PARTITION_FILES_SUBDIR ?= "partitions/qcs615-ride/ufs"
+
+UBOOT_MACHINE = "qcom_qcs615_defconfig"
```
lumag: UBOOT_CONFIG, separate commit

conf/machine/rb3gen2-core-kit.conf
```diff
@@ -30,3 +30,5 @@ QCOM_BOOT_FILES_SUBDIR = "qcm6490"
 QCOM_PARTITION_FILES_SUBDIR ?= "partitions/qcs6490-rb3gen2/ufs"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-qcs6490"
+
+UBOOT_MACHINE = "qcm6490_defconfig"
```
lumag: U_BOOT_CONFIG, separate commit

recipes-bsp/partition/qcom-partition-conf_git.bbappend
```diff
@@ -0,0 +1,27 @@
+# Modify partition configuration for U-Boot bootloader
+
+do_deploy:append() {
```
lumag: 1. Separate commit
2. Please handle this in `image_types_qcom` in the same way we handle KVM vs non-KVM XBL config.

recipes-bsp/partition/qcom-partition-conf_git.bbappend
```diff
@@ -0,0 +1,27 @@
...
+    # Check if U-Boot is the bootloader
+    BOOTLOADER="${PREFERRED_PROVIDER_virtual/bootloader}"
+
+    if [ "$BOOTLOADER" = "u-boot-qcom-next" ] || [ "$BOOTLOADER" = "u-boot-qcom-next-upstream" ]; then
+        bbnote "U-Boot detected as bootloader, modifying partition XML files"
```
lumag: Don't be noisy. There should be no output by default.

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+
+# Pinned to specific commit from qcom-next branch
+SRCREV ?= "387b3e2f37a06337789531ffaaf3ac3e3c573ce2"
+
+SRCBRANCH ?= "nobranch=1"
+SRCBRANCH:class-devupstream ?= "branch=qcom-next"
+
+SRC_URI = " \
+    git://github.com/qualcomm-linux/u-boot.git;${SRCBRANCH};protocol=https;name=uboot \
+    git://github.com/msm8916-mainline/qtestsign;protocol=https;destsuffix=git/qtestsign;name=qtestsign;branch=main \
```
lumag: Why can't you use existing qtestsign?

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+    ${PYTHON} ${UNPACKDIR}/git/qtestsign/qtestsign.py -v6 aboot -o ${B}/u-boot.mbn ${B}/u-boot.elf
+
+    # Check if u-boot.mbn was created successfully
+    if [ ! -f "${B}/u-boot.mbn" ]; then
+        bbfatal "Failed to create u-boot.mbn"
+    fi
+    bbnote "Successfully created u-boot.mbn"
+}
+
+do_deploy:append() {
```
lumag: uboot_deploy_config:append:qcom

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+PROVIDES += "virtual/bootloader"
+
+# Default config - can be overridden per machine
+UBOOT_MACHINE ?= "qcom_defconfig"
+
+# python3-cryptography needs the legacy provider, so set OPENSSL_MODULES to the
+# right path until this is relocated automatically.
+export OPENSSL_MODULES = "${STAGING_LIBDIR_NATIVE}/ossl-modules"
+
+do_compile:append() {
```
lumag: `uboot_compile_config:append:qcom`

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
+SUMMARY = "U-Boot bootloader for Qualcomm platforms"
+DESCRIPTION = "U-Boot from Qualcomm Linux upstream repository"
```
lumag: These two will be overriden by u-boot.inc and u-boot-common.inc

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+LICENSE = "GPL-2.0-or-later"
+LIC_FILES_CHKSUM = "file://Licenses/README;md5=2ca5f2c35c8cc335f0a19756634782f1"
```
lumag: What's wrong with the values set by `u-boot-common.inc`?

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+
+DEPENDS += "bison-native bc-native dtc-native gnutls-native python3-pyelftools-native xxd-native"
+DEPENDS += "python3-native python3-cryptography-native openssl-native"
```
lumag: `DEPENDS += "bc-native dtc-native qtestsign-native"`
and probably after the requires

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+
+require recipes-bsp/u-boot/u-boot-common.inc
+require recipes-bsp/u-boot/u-boot.inc
+
+COMPATIBLE_MACHINE = "(qcom)"
```
lumag: Even 32-bit?

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+
+PV = "2026.01+git"
+
+# Pinned to specific commit from qcom-next branch
+SRCREV ?= "387b3e2f37a06337789531ffaaf3ac3e3c573ce2"
+
+SRCBRANCH ?= "nobranch=1"
+SRCBRANCH:class-devupstream ?= "branch=qcom-next"
```
lumag: Nice!

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+SRCREV_qtestsign = "${AUTOREV}"
+SRCREV_FORMAT = "uboot_qtestsign"
+
+# To build tip of qcom-next branch set preferred
+# virtual/bootloader provider to 'u-boot-qcom-next-upstream'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "u-boot-qcom-next-upstream"
+SRCREV:class-devupstream ?= "${AUTOREV}"
+
+S = "${UNPACKDIR}/${BP}"
```
lumag: Should not be necessary.

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+
+# Provide virtual/bootloader
+PROVIDES += "virtual/bootloader"
```
lumag: What's wrong with the one already set by `u-boot.inc`?

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+
+# Default config - can be overridden per machine
+UBOOT_MACHINE ?= "qcom_defconfig"
```
lumag: Drop and use UBOOT_CONFIG

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+
+# python3-cryptography needs the legacy provider, so set OPENSSL_MODULES to the
+# right path until this is relocated automatically.
+export OPENSSL_MODULES = "${STAGING_LIBDIR_NATIVE}/ossl-modules"
```
lumag: ` export CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1` in your function

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+
+do_compile:append() {
+    # Run the qtestsign script to sign and convert ELF to MBN
+    bbnote "Signing U-Boot ELF with qtestsign..."
+    ${PYTHON} ${UNPACKDIR}/git/qtestsign/qtestsign.py -v6 aboot -o ${B}/u-boot.mbn ${B}/u-boot.elf
+
+    # Check if u-boot.mbn was created successfully
+    if [ ! -f "${B}/u-boot.mbn" ]; then
```
lumag: Wouldn't qtestsign return an error in this case?

ci/u-boot-qcom.yml
ricardosalveti: Why do we need to install u-boot?

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
ricardosalveti: Why u-boot-qcom-next? Can't we call it u-boot-qcom?

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
ricardosalveti: Right, should use the existing qtestsign from this layer.

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
ricardosalveti: Also, this is based on the released 2026.01 version, so better just name the recipe to align with the actual version we use as base here.

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
```diff
@@ -0,0 +1,70 @@
...
+
+DEPENDS += "bison-native bc-native dtc-native gnutls-native python3-pyelftools-native xxd-native"
+DEPENDS += "python3-native python3-cryptography-native openssl-native"
+
+require recipes-bsp/u-boot/u-boot-common.inc
+require recipes-bsp/u-boot/u-boot.inc
+
+COMPATIBLE_MACHINE = "(qcom)"
+
+PV = "2026.01+git"
```
b49020: The U-Boot version is rather `2026.04-rc1+git`.

recipes-bsp/partition/qcom-partition-conf_git.bbappend
aswinm94: Do we need to copy u-boot.mbn and rename it as uefi.elf, similar to how different XBL configuration files are selected for KVM and non‑KVM builds?

install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/u-boot.mbn" uefi.elf

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
koenkooi: > The U-Boot version is rather`2026.04-rc1+git`

Yes, but that won't sort properly, `2026.04-rc1+git` would evaluate newer than `2026.04`, the proper PV would be something like `2026.01+2026.04-rc1`. You can suffix `+git` to that if you wish.

recipes-bsp/partition/qcom-partition-conf_git.bbappend
b49020: That sounds reasonable to align with what partition layout expects as well.

ci/u-boot-qcom.yml
aswinm94: Created separate commit

conf/machine/qcs615-ride.conf
aswinm94: Created separate commit & used UBOOT_CONFIG for choosing the defconfig

conf/machine/rb3gen2-core-kit.conf
aswinm94: Created separate commit & used UBOOT_CONFIG for choosing the defconfig

recipes-bsp/partition/qcom-partition-conf_git.bbappend
aswinm94: Implemented bootloader selection logic to install the correct bootloader binary
as uefi.elf during image creation.
This avoids the modification of XML files

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Removed PROVIDES += "virtual/bootloader", since it is already set by u-boot.inc

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Dropped UBOOT_MACHINE ?= "qcom_defconfig", always forced to set device specific uboot config for the uboot build.

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Added "export CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1" and removed the export OPENSSL_MODULES = "${STAGING_LIBDIR_NATIVE}/ossl-modules"

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Updated the error handling based on qtestsign return value

ci/u-boot-qcom.yml
aswinm94: Removed the redundant IMAGE_INSTALL.

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Since the uboot version used is "2026.01-rc1", renamed the recipe as **u-boot-qcom_2026.01-rc1.bb**

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Used the existing qtestsign from meta-qcom

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Changed to uboot_deploy_config:append:qcom

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Changed to uboot_compile_config:append:qcom

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Removed SUMMARY & DESCRIPTION in the recipe
it uses the SUMMARY & DESCRIPTION from the include files

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Removed these lines from the recipe
It uses the values set by u-boot-common.inc

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Moved the DEPENDS after requires

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: updated to COMPATIBLE_MACHINE:aarch64, to support only 64-bit machines.

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Removed it

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: The U-Boot version used is "2026.01-rc1", so updated the PV as "2026.01-rc1+git"

lumag: @aswinm94 you've ignored the previous comment. It can't be `2026.01-rc1+git`, because that will trigger the "version going backwards" warning.

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
+require recipes-bsp/u-boot/u-boot-common.inc
+require recipes-bsp/u-boot/u-boot.inc
+
+DEPENDS += "bc-native dtc-native gnutls-native python3-pyelftools-native qtestsign-native xxd-native"
+
+COMPATIBLE_MACHINE:aarch64 = "(qcom)"
+
+PV = "2026.01-rc1+git"
```
lumag: Why are you setting PV manually, if you've it in the recipe filename?

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
...
+# To build tip of qcom-next branch set preferred
+# virtual/bootloader provider to 'u-boot-qcom-upstream'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "u-boot-qcom-upstream"
+SRCREV:class-devupstream ?= "${AUTOREV}"
+
+do_compile:append:qcom() {
+    export CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1
+    
+    elf_file=$(find ${B} -name "u-boot.elf" -type f | head -1)
```
lumag: Nope. Use exact location for u-boot.elf, you should know it at this point

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
...
+    
+    if [ -n "$elf_file" ] && [ -f "$elf_file" ]; then
+        elf_dir=$(dirname "$elf_file")
+        mbn_file="${elf_dir}/u-boot.mbn"
+        
+        if ! qtestsign -v6 aboot -o ${mbn_file} ${elf_file}; then
```
lumag: v6 needs to be overridable from the machine config. Or, better, build all (v3, v5, v6, v7?) versions

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
...
+            bbfatal "qtestsign command failed to sign u-boot.elf"
+        fi
+    else
+        bbfatal "Could not find u-boot.elf in build directory ${B}"
+    fi
+}
+
+do_deploy:append:qcom() {
+    mbn_file=$(find ${B} -name "u-boot.mbn" -type f | head -1)
```
lumag: The same comment. You should not need to find for the file.

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
...
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "u-boot-qcom-upstream"
+SRCREV:class-devupstream ?= "${AUTOREV}"
+
+do_compile:append:qcom() {
+    export CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1
+    
+    elf_file=$(find ${B} -name "u-boot.elf" -type f | head -1)
+    
+    if [ -n "$elf_file" ] && [ -f "$elf_file" ]; then
```
lumag: There is no need to check for it, you know that it exists.

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
...
+        fi
+    else
+        bbfatal "Could not find u-boot.elf in build directory ${B}"
+    fi
+}
+
+do_deploy:append:qcom() {
+    mbn_file=$(find ${B} -name "u-boot.mbn" -type f | head -1)
+    
+    if [ -n "$mbn_file" ] && [ -f "$mbn_file" ]; then
```
lumag: No need to

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
...
+        # Deploy to standard DEPLOYDIR location
+        install -m 0644 ${mbn_file} ${DEPLOYDIR}/u-boot.mbn
+        
+        # Deploy to QCOM_BOOT_FILES_SUBDIR for qcomflash image
+        if [ -n "${QCOM_BOOT_FILES_SUBDIR}" ]; then
+            install -d ${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}
```
lumag: Who allowed you to touch files under ${DEPLOY_DIR_IMAGE}?

classes-recipe/image_types_qcom.bbclass
```diff
@@ -116,6 +116,17 @@ create_qcomflash_pkg() {
             install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${xbl_config}" xbl_config.elf
         fi
 
+        # bootloader selection
+        bootloader_file="uefi.elf"
+        if ${@bb.utils.contains('PREFERRED_PROVIDER_virtual/bootloader', 'u-boot-qcom', 'true', 'false', d)} || \
```
lumag: Either use shell code or use python's `startswith()`.

conf/machine/rb3gen2-core-kit.conf
```diff
@@ -30,3 +30,6 @@ QCOM_BOOT_FILES_SUBDIR = "qcm6490"
 QCOM_PARTITION_FILES_SUBDIR ?= "partitions/qcs6490-rb3gen2/ufs"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-qcs6490"
+
+UBOOT_CONFIG = "qcs6490-rb3gen2"
+UBOOT_CONFIG[qcs6490-rb3gen2] = "qcm6490_defconfig"
```
lumag, line 43: Why can't we use `qcom_defconfig`?

ci/u-boot-qcom.yml
```diff
@@ -0,0 +1,6 @@
+header:
+  version: 14
+
+local_conf_header:
+  bootloaderprovider: |
+    PREFERRED_PROVIDER_virtual/bootloader = "u-boot-qcom-upstream"
```
lumag: Hook this kas file into the CI.

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: @lumag , When using 
UBOOT_CONFIG = "qcs6490-rb3gen2"
UBOOT_CONFIG[qcs6490-rb3gen2] = "qcm6490_defconfig"

it creates a sub dir inside build dir based on the values passed
**build/qcm6490_defconfig-qcs6490-rb3gen2/**
 so this sub dir name changes for each machine
 
 Hence used the find command to get the path of the u-boot.elf

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: @lumag , 
I understand u-boot needs to be in QCOM_BOOT_FILES_SUBDIR so that image_types_qcom can copy the binaries from there to qcomflash dir.
Can we check whether QCOM_BOOT_FILES_SUBDIR is set, and if so, proceed with copying u-boot.mbn?

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
lumag: Don't.

conf/machine/rb3gen2-core-kit.conf
aswinm94: @lumag ,
Platform related configurations for qcm6490 are present in "qcm6490_defconfig", internally qcm6490_defconfig will include qcom_defconfig which has generic configuration for all Qualcomm platforms.
So we can't directly use qcom_defconfig in UBOOT_CONFIG

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
lumag: No, it should go to the deploy directory as usual.

conf/machine/rb3gen2-core-kit.conf
lumag: Why can't `qcom_defconfig` be updated instead of creating the new config?

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: @lumag , any suggestions on how we can handle it?

conf/machine/rb3gen2-core-kit.conf
aswinm94: @lumag , SOC specific initializations should be kept in separate config, upstream community won't allow to add it in generic qcom_defconfig.
qcm6490_defconfig is already part of upstream code base,

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Apologies, I missed addressing it in the previous iteration

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
lumag: Well, you've marked my suggestion as resolved, so I think you've read it. Extend `uboot_compile_config`.

conf/machine/rb3gen2-core-kit.conf
lumag: Ack

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
koenkooi: It's also not a valid version, with the unguared `-rc1`

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
...
+require recipes-bsp/u-boot/u-boot.inc
+
+DEPENDS += "bc-native dtc-native gnutls-native python3-pyelftools-native qtestsign-native xxd-native"
+
+COMPATIBLE_MACHINE:aarch64 = "(qcom)"
+
+PV = "2026.01-rc1+git"
+
+# commit used from qcom-next branch
+SRCREV ?= "387b3e2f37a06337789531ffaaf3ac3e3c573ce2"
```
koenkooi: Useless `?`, remove

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
...
+
+SRCBRANCH ?= "nobranch=1"
```
koenkooi: Useless `?`, remove

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
...
+SRCBRANCH:class-devupstream ?= "branch=qcom-next"
```
koenkooi: Even more useless `?`, remove

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
```diff
@@ -0,0 +1,56 @@
...
+
+SRC_URI = "git://github.com/qualcomm-linux/u-boot.git;${SRCBRANCH};protocol=https;name=uboot"
+
+# To build tip of qcom-next branch set preferred
+# virtual/bootloader provider to 'u-boot-qcom-upstream'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "u-boot-qcom-upstream"
+SRCREV:class-devupstream ?= "${AUTOREV}"
```
koenkooi: Useless `?`, remove

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: > > The U-Boot version is rather`2026.04-rc1+git`
> 
> Yes, but that won't sort properly, `2026.04-rc1+git` would evaluate newer than `2026.04`, the proper PV would be something like `2026.01+2026.04-rc1`. You can suffix `+git` to that if you wish.

@koenkooi , whether we can set the **PV = 2026.01+2026.04~rc1+git${SRCPV}** & keep recipe name as **u-boot-qcom_git.bb**?

koenkooi: Yes, that's exactly what setting PV inside the recipe is for :) Although I would convert the tilde to a hyphen.

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: Extended uboot_compile_config to get exact location for u-boot.elf.

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: UBOOT_MBN_HEADER_VERSION is used to pass the version value. The recipe sets its default value to 6, but it can be overridden in the machine configuration.

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: Extended uboot_deploy_config & removed "find"

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: Removed it.

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: Removed it.

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: Removed copying to QCOM_BOOT_FILES_SUBDIR.

ci/u-boot-qcom.yml
aswinm94: As part of the below commit, added the change
build-yocto: add U-Boot build support for qcs615-ride and rb3gen2-core-kit

classes-recipe/image_types_qcom.bbclass
aswinm94: Updated it to use shell code

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: removed it

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: removed it

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: removed it

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: removed it

recipes-bsp/u-boot/u-boot-qcom-next_git.bb
aswinm94: Updated PV = "2026.01+2026.04-rc1+git${SRCPV}"

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: updated the value
PV = "2026.01+2026.04-rc1+git${SRCPV}"

lumag: Where did you pick up SRCPV? It's long gone.

aswinm94: @lumag, added SRCPV to ensure the increment of package version when the upstream commit changes.
Whether it has to be removed?

lumag: Do you see it being defined, documented or used anywhere?

recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -0,0 +1,33 @@
...
+
+SRC_URI = "git://github.com/qualcomm-linux/u-boot.git;${SRCBRANCH};protocol=https;name=uboot"
+
+# To build tip of qcom-next branch set preferred
+# virtual/bootloader provider to 'u-boot-qcom-upstream'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "u-boot-qcom-upstream"
+SRCREV:class-devupstream = "${AUTOREV}"
+
+UBOOT_MBN_HEADER_VERSION ?= "6"
```
lumag: What is the header version of `qcom-armv8a`?

recipes-bsp/u-boot/u-boot-qcom_2026.01-rc1.bb
aswinm94: @lumag , Its not been used anywhere in meta-qcom 
will set the value as
PV = "2026.01+2026.04-rc1+git"

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag, the header version is SoC‑specific. There is no single, fixed header version for the qcom‑armv8a

lumag: Correct. So, we need to build several versions and then select the most appropriate at the time of building the image.

recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -0,0 +1,33 @@
+require recipes-bsp/u-boot/u-boot-common.inc
+require recipes-bsp/u-boot/u-boot.inc
+
+DEPENDS += "bc-native dtc-native gnutls-native python3-pyelftools-native qtestsign-native xxd-native"
+
+COMPATIBLE_MACHINE:aarch64 = "(qcom)"
+
+PV = "2026.01+2026.04-rc1+git"
+
+# commit used from qcom-next branch
```
ricardosalveti: No need for this comment.

recipes-bsp/u-boot/u-boot-qcom_git.bb
ricardosalveti: This is not resolved, as @lumag said, need to build for all available versions and select the most appropriate when building the final image.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -116,6 +116,18 @@ create_qcomflash_pkg() {
             install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${xbl_config}" xbl_config.elf
         fi
 
+        # bootloader selection
+        if [ "${PREFERRED_PROVIDER_virtual/bootloader}" = "u-boot-qcom" ] || \
+           [ "${PREFERRED_PROVIDER_virtual/bootloader}" = "u-boot-qcom-upstream" ]; then
+            if [ -f "${DEPLOY_DIR_IMAGE}/u-boot.mbn" ]; then
+                install -m 0644 "${DEPLOY_DIR_IMAGE}/u-boot.mbn" uefi.elf
```
ricardosalveti: We shouldn't check just for u-boot-qcom or u-boot-qcom-upstream as we might have users defining their own u-boot recipe / fork.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -116,6 +116,18 @@ create_qcomflash_pkg() {
...
+            fi
+        else
+            if [ -f "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/uefi.elf" ]; then
+                install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/uefi.elf" uefi.elf
```
ricardosalveti: Here you only check for uefi.elf, why not the same for u-boot.mbn?

ci/u-boot-qcom.yml
ricardosalveti: Why u-boot-qcom-upstream and not u-boot-qcom?

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag , @ricardosalveti 
Currently, we are not building multiple U-Boot configurations; instead, we generate a single build based on the configuration defined in the machine configuration file.
The resulting U-Boot binary is specific to the corresponding SoC, and using it across different SoCs would not be appropriate. Given this, generating several header versions may not provide any practical benefit, as they would not be used.
I understand the point regarding building multiple header versions. Based on the current build setup, it seems more practical to select the appropriate header version according to the specific SoC.

classes-recipe/image_types_qcom.bbclass
aswinm94: @ricardosalveti ,
I believe Line 122, checks for the presence of u-boot.mbn as like uefi.elf
            if [ -f "${DEPLOY_DIR_IMAGE}/u-boot.mbn" ]; then
                install -m 0644 "${DEPLOY_DIR_IMAGE}/u-boot.mbn" uefi.elf

ci/u-boot-qcom.yml
aswinm94: @ricardosalveti ,
u-boot-qcom-upstream helps us stay aligned with the upstream‑posted changes that are merged into qualcomm-linux u-boot repo, without requiring manual revision bumps.

recipes-bsp/u-boot/u-boot-qcom_git.bb
lumag: Sounds logical. Then please define it per UBOOT_CONFIG entry via varflags.

classes-recipe/image_types_qcom.bbclass
ricardosalveti: But before you check for PREFERRED_PROVIDER_virtual/bootloader.

ci/u-boot-qcom.yml
ricardosalveti: NAK from my side, this fragment should just point to u-boot-qcom (stable and known revision), if you want to use u-boot-qcom-upstream (or other similar recipes), then we should have that in a separated fragment which we won't really use in our normal CI.

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag , I tried varflags for UBOOT_CONFIG entry but I am facing issues in accessing it inside uboot_compile_config:append:qcom()
Could you please let me know if you have any suggestions I should try?

lumag:
```
UBOOT_CONFIG = "abc def"
BOARD_MBN_HEADER[abc] = "v4"
BOARD_MBN_HEADER[def] = "v8"
```

aswinm94: @lumag , yes, i tried the same
looks like these custom Varflags can be processed during uboot_configure_config() stage
I wasn't able to access it directly in uboot_compile_config()

ci/u-boot-qcom.yml
aswinm94: Changed the PREFERRED_PROVIDER_virtual/bootloader = "u-boot-qcom"
Which now points to the stable known version

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: Removed the comment

classes-recipe/image_types_qcom.bbclass
aswinm94: In the updated version, the logic checks for the pattern u-boot*, ensuring it works with any bootloader provider that begins with u-boot, including custom u‑boot recipes.

classes-recipe/image_types_qcom.bbclass
aswinm94: As per the updated version logic, If the selected bootloader starts with u‑boot, the script installs u‑boot.mbn when that file exists; otherwise, it falls back to installing the uefi.elf file, also only if present.

recipes-bsp/u-boot/u-boot-qcom_git.bb
b49020: @aswinm94 lets rather make the `.mbn` generation machine specific instead. As I would also like to migrate Agatti U-Boot support present in bbapend here: `recipes-bsp/u-boot/u-boot_%.bbappend` to be based on U-Boot qcom-next branch only.

aswinm94: @b49020 , The current implementation already supports machine‑specific .mbn generation. We define
UBOOT_MBN_HEADER_VERSION ?= "6" in the recipe, and each machine can override this in its .conf file as needed.

b49020: What I am saying is Agatti doesn't support .mbn generation but rather chain loaded via ABL in boot.img format.

aswinm94: @lumag , @b49020 

Since using UBOOT_CONFIG varflags path introduced some unexpected issues, a simpler approach would be to remove the default UBOOT_MBN_HEADER_VERSION assignment from the recipe and instead define it only in the respective machine .conf files.

With this change, if UBOOT_MBN_HEADER_VERSION is not set for a machine, we can simply skip running qtestsign. This keeps the signing logic machine‑specific as requested and avoids triggering the signing step on platforms that do not require an .mbn image.

Let me know if this aligns with your expectation.

`uboot_compile_config:append:qcom() {
    if [ -n "${UBOOT_MBN_HEADER_VERSION}" ]; then
        export CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1
        qtestsign -v${UBOOT_MBN_HEADER_VERSION} aboot -o ${B}/${builddir}/u-boot.mbn ${B}/${builddir}/u-boot.elf
    fi
}

uboot_deploy_config:append:qcom() {
    if [ -f ${B}/${builddir}/u-boot.mbn ]; then
        install -m 0644 ${B}/${builddir}/u-boot.mbn ${DEPLOYDIR}/u-boot.mbn
    fi
}
`

lumag: > @lumag , @b49020
> 
> Since using UBOOT_CONFIG varflags path introduced some unexpected issues,

Which issues? I haven't seen any error messages in your comments.

> a simpler approach would be to remove the default UBOOT_MBN_HEADER_VERSION assignment from the recipe and instead define it only in the respective machine .conf files.

Again, what is the `UBOOT_MBN_HEADER_VERSION` for `qcom-armv8a`?

> With this change, if UBOOT_MBN_HEADER_VERSION is not set for a machine, we can simply skip running qtestsign. This keeps the signing logic machine‑specific as requested and avoids triggering the signing step on platforms that do not require an .mbn image.
> 
> Let me know if this aligns with your expectation.

It doesn't. Define it per UBOOT_CONFIG and skip signing if it is not defined.

aswinm94: @lumag 
> Again, what is the UBOOT_MBN_HEADER_VERSION for qcom-armv8a?

As discussed before, the header version is SoC‑specific. There is no single, fixed header version for the qcom‑armv8a

> It doesn't. Define it per UBOOT_CONFIG and skip signing if it is not defined.

As you suggested, able to define mbn header version varflags per UBOOT_CONFIG

recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -0,0 +1,50 @@
...
+SRCBRANCH = "nobranch=1"
+SRCBRANCH:class-devupstream = "branch=qcom-next"
+
+SRC_URI = "git://github.com/qualcomm-linux/u-boot.git;${SRCBRANCH};protocol=https;name=uboot"
+
+# To build tip of qcom-next branch set preferred
+# virtual/bootloader provider to 'u-boot-qcom-upstream'
+BBCLASSEXTEND = "devupstream:target"
+PN:class-devupstream = "u-boot-qcom-upstream"
+SRCREV:class-devupstream = "${AUTOREV}"
```
lumag: No, see https://github.com/qualcomm-linux/meta-qcom/pull/1560

recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -0,0 +1,50 @@
...
+
+    if len(ubootconfig) > 0:
+        for config in ubootconfig:
+            # Get the MBN header version for this specific config
+            mbn_header = d.getVarFlag('BOARD_MBN_HEADER', config)
+
+            if not mbn_header:
+                mbn_header = ""
+
+            d.appendVar('BOARD_MBN_HEADER', mbn_header + " ? ")
```
lumag, line 33: Why?

recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -0,0 +1,50 @@
...
+}
+
+uboot_compile_config:append:qcom() {
+    config_mbn_header=$(uboot_config_get_indexed_value "${BOARD_MBN_HEADER}" $i)
```
lumag, line 37: You are trying to make it too complicated. Get flag for the current UBOOT_CONFIG value by name.

recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -0,0 +1,50 @@
...
+
+    if [ -n "${config_mbn_header}" ]; then
+        export CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1
+        qtestsign -${config_mbn_header} aboot -o ${B}/${builddir}/u-boot.mbn ${B}/${builddir}/u-boot.elf
+    fi
+}
+
+uboot_deploy_config:append:qcom() {
+    if [ -f ${B}/${builddir}/u-boot.mbn ]; then
+        install -m 0644 ${B}/${builddir}/u-boot.mbn ${DEPLOYDIR}/u-boot.mbn
```
lumag: This will not work if there are multiple UBOOT_CONFIGs defined.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -116,6 +116,20 @@ create_qcomflash_pkg() {
...
+                if [ -f "${DEPLOY_DIR_IMAGE}/u-boot.mbn" ]; then
+                    install -m 0644 "${DEPLOY_DIR_IMAGE}/u-boot.mbn" uefi.elf
+                fi
+                ;;
+            *)
+                if [ -f "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/uefi.elf" ]; then
+                    install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/uefi.elf" uefi.elf
+                fi
+                ;;
+        esac
```
lumag, line 126: Merge two install commands by using variable with the filename.

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag ,
This follows the same mechanism used by the U-Boot config system for handling other per‑config varflags, such as UBOOT_CONFIG_BINARY[config].
(_**_Reference_**_: https://git.openembedded.org/openembedded-core/tree/meta/classes-recipe/uboot-config.bbclass#n219)

uboot_config_get_indexed_value() expects the inputs to be in this format
**Input string: "v4 ? v8 ? v10 ? "**
to extract the correct value for each config during shell execution.

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag ,
I tried 
`config_mbn_header="${@d.getVarFlag('BOARD_MBN_HEADER', '${type}') or ''}"`
but it failed with a **bad substitution** error. looks like Inline Python (${@...}) doesn’t work inside shell functions.

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag 
At the moment, a machine is limited to building only one U‑Boot configuration. Multi‑config handling hasn’t been implemented for machines yet. Therefore, the U‑Boot binary corresponding to the active configuration is the one placed in the machine’s deploy directory.

recipes-bsp/u-boot/u-boot-qcom_git.bb
lumag: Of course it does, there are enough examples which follow that approach.

recipes-bsp/u-boot/u-boot-qcom_git.bb
lumag: u-boot.inc supports multiple configurations. Please don't break it.

lumag: As an example, you can enable in `qcom-armv8a.conf` all configurations that you want to support and make sure that the recipe is not broken.

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag,
 I’m not very experienced with Yocto, and I’m running into trouble accessing VARFLAGS for the current UBOOT_CONFIG value by name.
 Could you please suggest the proper way to reference the per‑type VARFLAG here (i.e., the correct call/idiom to get BOARD_MBN_HEADER for the current ${type} inside a shell function)?

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag 
here is my proposed solution
In our machine.conf, we can set the UBOOT_CONFIG as below
```
 UBOOT_CONFIG = "rb3gen2-core-kit" 
 UBOOT_CONFIG[rb3gen2-core-kit] = "qcm6490_defconfig"
```
where **rb3gen2-core-kit** config name should be same as machine name. 
To ensure multi‑config support remains fully functional, we update the deploy step in uboot_deploy_config:append:qcom() to generate configuration‑specific output files, using: 
`install -m 0644 ${B}/${builddir}/u-boot.mbn ${DEPLOYDIR}/u-boot-${type}.mbn. `
Here, ${type} corresponds to the key defined in UBOOT_CONFIG[] i.e **rb3gen2-core-kit**, 

Finally, in **image_types_qcom.bbclass**, we select the correct U‑Boot binary for the active machine when generating the qcomflash directory, using: 
**install -m 0644 "${DEPLOY_DIR_IMAGE}/u-boot-${MACHINE}.mbn" uefi.elf**. 
This ensures that multiple U‑Boot configurations can coexist in the deploy directory while the correct machine‑specific binary is reliably copied into the qcomflash output.

Kindly confirm if this approach aligns with your expectations.

lumag: This won't work for qcom-armv8a.conf. It has several machines in it.

aswinm94: @lumag ,
I would like to understand what the deploy directory structure looks like when building for multiple machines, and how **image_types_qcom.bbclass** selects and copies the correct boot binaries in that case. My understanding is that even if multiple machines are defined, ${MACHINE} always refers to the machine name currently being built, which should allow the class to pick the correct U‑Boot binary from the deploy directory

Please help me understand what the exact shortcoming was in my earlier proposal so I can correct it properly.

lumag: Please make something like this to work:
```
UBOOT_CONFIG = "qcs6490-rb3gen2 sdm845-db845c qrb2210-rb1 qrb4210-rb2"
UBOOT_CONFIG[qcs6490-rb3gen2] = "qcm6490_defconfig"
UBOOT_CONFIG[sdm845-db845c] = "qcom_defconfig"
UBOOT_CONFIG[qrb2210-rb1] = "qcom_defconfig"
UBOOT_MBN_SIGNATURE[qcs6490-rb3gen2] = "v6"
UBOOT_MBN_SIGNATURE[sdm845-db845c] = "v5"
UBOOT_MBN_SIGNATURE[qrb2210-rb1] = "v6"
```

Correspondingly you will have to change your `uboot_deploy_config:append()` snipped to take `${config}` and ${type}` into account.

recipes-bsp/u-boot/u-boot-qcom_git.bb
lumag: Could you please post your code that doesn't work? It's easier to correct existing code.

aswinm94: @lumag , this is my code

```
`
uboot_compile_config:append:qcom() {
    config_mbn_header="${@d.getVarFlag('BOARD_MBN_HEADER', '${type}') or ''}"

    if [ -n "${config_mbn_header}" ]; then
        export CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1
        qtestsign -${config_mbn_header} aboot -o ${B}/${builddir}/u-boot.mbn ${B}/${builddir}/u-boot.elf
    fi
}
`
```
**Error:**
Bad substitution
| WARNING: exit code 2 from a shell command.

**Machine conf**
UBOOT_CONFIG = "qcs6490-rb3gen2"
UBOOT_CONFIG[qcs6490-rb3gen2] = "qcm6490_defconfig"
BOARD_MBN_HEADER[qcs6490-rb3gen2] = "v6"

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag ,
We can update uboot_deploy_config:append() to use ${config} and ${type} so that multiple U‑Boot builds per UBOOT_CONFIG are deployed under deploy dir.

The issue is image_types_qcom.bbclass doesn’t currently receive ${config} or ${type}, so it can’t deterministically select the right U‑Boot binary when assembling the binaries required for flashing in qcomflash directory.

What’s will be the preferred approach to let image_types_qcom.bbclass pick the correct U‑Boot binary from deploy dir?

recipes-bsp/u-boot/u-boot-qcom_git.bb
lumag: Please fix formatting of your message

recipes-bsp/u-boot/u-boot-qcom_git.bb
lumag: In the same way, it's handled for kernels with the bundled DTB. Define a variable with the default u-boot kind in a machine config.

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: Fixed it

lumag: Okay, now it's easy. Please take a look at the existing `uboot_compile_config()`.

Notice that it uses a different approach to get the values. Do you understand why? Hint: notice, what gets expanded at which time. What is a shell variable, what is a python function, what is the bitbake variable. What is the expansion order? You can't be using shell variables inside Python code, because Python is expanded _before_ the shell code is being executed.

Now, check, what existing code is doing and mimic the same approach (`uboot_config_get_indexed_value`).

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: updated to

SRCREV ?= '${@oe.utils.conditional("PREFERRED_PROVIDER_virtual/kernel", "u-boot-qcom-upstream", "${AUTOREV}", "7b9fb537f19db0fa48b824ecc883c9c8512f0e21", d)}'

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: As suggested introduced a UBOOT_CONFIG_DEFAULT variable in machine config specifying the default uboot config, based on which image_types_qcom.bbclass pick the correct U‑Boot binary from deploy dir

classes-recipe/image_types_qcom.bbclass
aswinm94: Merged two install commands to one

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag,
Our current implementation exactly mimics the approach used by the U-Boot config system for handling other per‑config varflags, such as UBOOT_CONFIG_BINARY[config].
(Reference: https://git.openembedded.org/openembedded-core/tree/meta/classes-recipe/uboot-config.bbclass#n219)

**Current implementation:**

```
python __anonymous() {
    for config in ubootconfig:
        mbn_header = d.getVarFlag('BOARD_MBN_HEADER', config)
        d.appendVar('BOARD_MBN_HEADER', mbn_header + " ? ")
}
config_mbn_header=$(uboot_config_get_indexed_value "${BOARD_MBN_HEADER}" $i)
```

Varflags can only be accessed via `d.getVarFlag()` in Python during parse-time. The `python __anonymous()` function processes varflags and builds an indexed string(**v4 ? v8 ? v10**) that `uboot_config_get_indexed_value()` can parse at execution-time using the runtime index `$i`. 
This is the standard BitBake pattern for per-config varflags, used by `UBOOT_CONFIG_BINARY`, `UBOOT_CONFIG_MAKE_OPTS`, and `UBOOT_CONFIG_FRAGMENTS`.

lumag: Ok, I see it now. Sorry for the confusion.

recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -0,0 +1,49 @@
...
+            # Get the MBN header version for this specific config
+            mbn_header = d.getVarFlag('BOARD_MBN_HEADER', config)
+
+            if not mbn_header:
+                mbn_header = ""
+
+            d.appendVar('BOARD_MBN_HEADER', mbn_header + " ? ")
+}
+
+uboot_compile_config:append:qcom() {
```
quaresmajose: Since this recipe is qcom specific, we don't need the `:qcom` override here.

recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -0,0 +1,49 @@
...
+    config_mbn_header=$(uboot_config_get_indexed_value "${BOARD_MBN_HEADER}" $i)
+
+    if [ -n "${config_mbn_header}" ]; then
+        export CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1
+        qtestsign -${config_mbn_header} aboot -o ${B}/${builddir}/u-boot.mbn ${B}/${builddir}/u-boot.elf
+    fi
+}
+
+uboot_deploy_config:append:qcom() {
```
quaresmajose: Same here about the `:qcom` override.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -116,6 +116,18 @@ create_qcomflash_pkg() {
             install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${xbl_config}" xbl_config.elf
         fi
 
+        # bootloader selection
+	bootloader_bin="${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/uefi.elf"
```
quaresmajose: Please fix the identation.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -116,6 +116,18 @@ create_qcomflash_pkg() {
...
+        case "${PREFERRED_PROVIDER_virtual/bootloader}" in
+            u-boot*)
+		bootloader_bin="${DEPLOY_DIR_IMAGE}/u-boot-${UBOOT_CONFIG_DEFAULT}.mbn"
+                ;;
+        esac
+	if [ -f "${bootloader_bin}" ]; then
```
quaresmajose: Please fix the identation.

.github/workflows/build-yocto.yml
```diff
@@ -222,6 +222,22 @@ jobs:
                 type: default
                 dirname: ""
                 yamlfile: ""
+          - machine: qcs615-ride
+            distro:
+                name: poky-altcfg
```
quaresmajose, line 224: Please change this poky-altcfg to nodistro.

.github/workflows/build-yocto.yml
```diff
@@ -222,6 +222,22 @@ jobs:
...
+                yamlfile: ''
+            kernel:
+                type: u-boot-qcom-6.18
+                dirname: "+linux-qcom-6.18+u-boot-qcom"
+                yamlfile: ":ci/linux-qcom-6.18.yml:ci/u-boot-qcom.yml"
+          - machine: rb3gen2-core-kit
+            distro:
+                name: poky-altcfg
```
quaresmajose, line 232: Please change this poky-altcfg to nodistro.

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: removed the :qcom override

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: removed the :qcom override

.github/workflows/build-yocto.yml
aswinm94: Changed to nodistro

.github/workflows/build-yocto.yml
aswinm94: Changed to nodistro

classes-recipe/image_types_qcom.bbclass
aswinm94: Fixed the identation

classes-recipe/image_types_qcom.bbclass
aswinm94: Fixed the identation.

aswinm94:
@quaresmajose , @ricardosalveti 
Rebased to the tip

recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -0,0 +1,49 @@
+require recipes-bsp/u-boot/u-boot-common.inc
+require recipes-bsp/u-boot/u-boot.inc
+
+DEPENDS += "bc-native dtc-native gnutls-native python3-pyelftools-native qtestsign-native xxd-native"
+
+COMPATIBLE_MACHINE:aarch64 = "(qcom)"
+
+PV = "2026.01+2026.04-rc1+git"
+
+SRCREV ?= '${@oe.utils.conditional("PREFERRED_PROVIDER_virtual/kernel", "u-boot-qcom-upstream", "${AUTOREV}", "7b9fb537f19db0fa48b824ecc883c9c8512f0e21", d)}'
```
ricardosalveti: should be virtual/bootloader

aswinm94: Sry missed it, Changed to PREFERRED_PROVIDER_virtual/bootloader

aswinm94:
@koenkooi , all requested changes have been implemented.
Could you please re‑review the PR and change your review status to Approve if everything looks good?

.github/workflows/build-yocto.yml
```diff
@@ -219,6 +219,22 @@ jobs:
                 type: default
                 dirname: ""
                 yamlfile: ""
+          - machine: qcs615-ride
+            distro:
+                name: nodistro
```
ricardosalveti: Change to nodistro is not yet merged: https://github.com/qualcomm-linux/meta-qcom/pull/1671/files

classes-recipe/image_types_qcom.bbclass
```diff
@@ -116,6 +116,17 @@ create_qcomflash_pkg() {
             install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${xbl_config}" xbl_config.elf
         fi
 
+        # bootloader selection
+        bootloader_bin="${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/uefi.elf"
+        case "${PREFERRED_PROVIDER_virtual/bootloader}" in
```
ricardosalveti: This will fail when PREFERRED_PROVIDER_virtual/bootloader is not defined (check the build failure), which is the default when not using u-boot.

Please test on both scenarios.

aswinm94: Resolved the build error when not using u-boot

.github/workflows/build-yocto.yml
aswinm94: @ricardosalveti 
The u-boot build configuration alone specifies distro: name: nodistro and I believe it doesn't depend on the pending change https://github.com/qualcomm-linux/meta-qcom/pull/1671/files.
The nodistro distro is defined in OE-Core and is always available, so our current configuration works independently. Both changes can be merged in any order without conflicts.
@quaresmajose, please correct me if I am wrong

quaresmajose: The nodistro PR #1671 is causing some test errors, so I put it in draft. The best thing you can do is go back to using the poky-alt so you don't get blocked. Meanwhile, I rebase the PR #1671 and then switch any remaning poky-alt.

Sorry for asking you to change it to nodistro above in the review.

mwasilew: The "nodistro" part is _not_ causing issues. The problems are caused bu additional tests that were enabled "in the meantime".

aswinm94: To unblock it, will revert back to poky-alt distro

ricardosalveti: Thanks, while it will be changed after, we should't mix changes from multiple prs here.

.github/workflows/build-yocto.yml
```diff
@@ -219,6 +219,22 @@ jobs:
                 type: default
                 dirname: ""
                 yamlfile: ""
+          - machine: qcs615-ride
+            distro:
+                name: poky-altcfg
+                yamlfile: ''
+            kernel:
+                type: u-boot-qcom-6.18
+                dirname: "+linux-qcom-6.18+u-boot-qcom"
```
ricardosalveti: Please change from + to _, needed after https://github.com/qualcomm-linux/meta-qcom/pull/1661

.github/workflows/build-yocto.yml
```diff
@@ -219,6 +219,22 @@ jobs:
...
+                yamlfile: ":ci/linux-qcom-6.18.yml:ci/u-boot-qcom.yml"
+          - machine: rb3gen2-core-kit
+            distro:
+                name: poky-altcfg
+                yamlfile: ''
+            kernel:
+                type: u-boot-qcom-6.18
+                dirname: "+linux-qcom-6.18+u-boot-qcom"
```
ricardosalveti: Same.

.github/workflows/build-yocto.yml
aswinm94: @ricardosalveti , shall we replace both "+" with "_"?
"_linux-qcom-6.18_u-boot-qcom"

ricardosalveti: Yes, please, we have an issue with + and s3.

aswinm94: Changed to _linux-qcom-6.18_u-boot-qcom

.github/workflows/build-yocto.yml
aswinm94: Changed to "_linux-qcom-6.18_u-boot-qcom"

lumag (approved):

PR #1485, opened 2026-02-04
ci: actions/workflows: set concurrency only for pull requests

sampra2025:
We should have separate builds for each merge, and only cancel PR build requests if they occur during an ongoing build.


ricardosalveti (approved):

lumag (changes requested):
Please improve your commit message: Please start it by describing the issue or the problem that you are facing. What is not working or working not as expected?

PR #1491, opened 2026-02-05
Fix protobuf-related erros

lumag:
Provide local version of the protobuf recipe and use it for CamX libs and sensors recipes.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.6.bb
+++ b/dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.6.bb
@@ -11,7 +11,7 @@ SRC_URI[sha256sum] = "766b31a4382524939ea6cbc9db23f2aebd94f2b1250b5d9a8114765652
 
 S = "${UNPACKDIR}"
 
-DEPENDS = "glib-2.0 protobuf sensinghub qmi-framework libdiag fastrpc"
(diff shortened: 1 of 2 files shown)


dynamic-layers/openembedded-layer/recipes-support/protobuf-camx/protobuf-camx/0001-Fix-build-on-mips-clang.patch
```diff
@@ -0,0 +1,34 @@
+From cd6c5f36c0dc38e4734b74b03923489b7bfb9aba Mon Sep 17 00:00:00 2001
+From: Khem Raj <raj.khem@gmail.com>
+Date: Mon, 10 Mar 2025 19:59:19 -0700
+Subject: [PATCH] Fix build on mips/clang
+
+clang20 crashes on mips, until its fixed upstream disable tailcall on
+mips
```
ricardosalveti: I know we're copying the patch over, but we don't really build this for mips, can't we drop it?

lumag: Done

lumag:
> Having two libraries with different versions installing content is not secure, and the probability of conflict when both install the same content is high. I am testing on my side using with an unmodified copy and with a global PREFERRED_VERSION_protobuf = "6.31.1".

I explicitly don't want to do that, it would make us deviate from the upstream. Note, technically it's not just protobuf. It's libprotobuf31, libprotobuf32, etc.

PR #1494, opened 2026-02-06
kgsl-dlkm: update KGSL tip

quic-ksanadhy:
Refresh the SRCREV to the latest commit in the KGSL source repository. This update integrates the RGMU platform device pointer into gmu_core during the Adreno probe sequence for RGMU based targets.


ricardosalveti:
Please change to 'update to FOOBAR revision' or similar next time, tip is only relevant at a certain point in time.

ricardosalveti (approved):

PR #1502, opened 2026-02-07
sensors: switch to latest protobuf

bmasta018:
Upgraded sensinghub from 1.0.4 to 1.0.5. Updated the code with new protobuf version and open-sourced utils code from qcom-sensors-binaries.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.7.bb
+++ b/dynamic-layers/openembedded-layer/recipes-support/sensors/qcom-sensors-binaries_1.0.7.bb
@@ -5,13 +5,13 @@ validate sensor services functionality through the Sensinghub Interface."
 LICENSE = "LICENSE.qcom-2"
 LIC_FILES_CHKSUM = "file://LICENSE.qcom-2;md5=f33ba334514c4dfabc6ab7377babb377"
 
-PBT_BUILD_DATE = "260117"
+PBT_BUILD_DATE = "260209"
 SRC_URI = "https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/sensors.lnx.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qcom-sensors-prebuilts_${PV}_armv8a.tar.gz"
-SRC_URI[sha256sum] = "766b31a4382524939ea6cbc9db23f2aebd94f2b1250b5d9a8114765652758770"
+SRC_URI[sha256sum] = "02a14811570b30464efb9d0dcdf709ce930bfc24bf95aa9ceedede66484abd61"
 
 S = "${UNPACKDIR}"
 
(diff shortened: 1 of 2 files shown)


lumag:
/draft
The sensinghub, qcom-sensors-binaries and CamX upgrade should go together in a single PR.

lumag:
> > @bmasta018 could you please provide the binaries with just protobuf updates without any extra changes? Your binaries require `libqshUtil.so.1`, which is not provided by the old recipe. Please don't bundle unrelated changes together.
> 
> We’ve been working on moving our implementation from proprietary to open-source as suggested, and part of the codebase has already been open-sourced. During this transition, we were planning to update the SHA in the PRs, but unfortunately the protobuf changes came in at the same time. At this point, we can’t merge only the protobuf updates in isolation since the code has already advanced significantly.

Rebuild _old_ codebase?

bmasta018:
> > > > > @bmasta018 could you please provide the binaries with just protobuf updates without any extra changes? Your binaries require `libqshUtil.so.1`, which is not provided by the old recipe. Please don't bundle unrelated changes together.
> > > > 
> > > > 
> > > > We’ve been working on moving our implementation from proprietary to open-source as suggested, and part of the codebase has already been open-sourced. During this transition, we were planning to update the SHA in the PRs, but unfortunately the protobuf changes came in at the same time. At this point, we can’t merge only the protobuf updates in isolation since the code has already advanced significantly.
> > > 
> > > 
> > > Rebuild _old_ codebase?
> > 
> > 
> > Our codebase has moved ahead significantly and a portion of code has been shifted to github sensinghub and unfortunately we can't build the old codebase now
> 
> Why? If you can't get back to the codebase of one or two months old, it means something must be significantly changed on your side.



> > > > > @bmasta018 could you please provide the binaries with just protobuf updates without any extra changes? Your binaries require `libqshUtil.so.1`, which is not provided by the old recipe. Please don't bundle unrelated changes together.
> > > > 
> > > > 
> > > > We’ve been working on moving our implementation from proprietary to open-source as suggested, and part of the codebase has already been open-sourced. During this transition, we were planning to update the SHA in the PRs, but unfortunately the protobuf changes came in at the same time. At this point, we can’t merge only the protobuf updates in isolation since the code has already advanced significantly.
> > > 
> > > 
> > > Rebuild _old_ codebase?
> > 
> > 
> > Our codebase has moved ahead significantly and a portion of code has been shifted to github sensinghub and unfortunately we can't build the old codebase now
> 
> Why? If you can't get back to the codebase of one or two months old, it means something must be significantly changed on your side.

The proprietary codebase had checkpoints in the form of prebuilts that we tagged and pushed into Artifactory. However, those artifacts are no longer usable because they’ve fallen out of sync due to the protobuf changes. Reverting the codebase isn’t an option at this point, and while we normally could rely on those checkpoints, that’s not possible in this case. Additionally, a significant number of changes have gone in as part of the porting activity, which makes isolating protobuf updates even more challenging.

PR #1509, opened 2026-02-09
ci: use official openembedded project url in kas config files

quaresmajose:
- add KAS_PREMIRRORS: Specifies alternatives for repo URLs.
- use official openembedded project url in kas config files

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -13,12 +13,12 @@ repos:
   meta-qcom:
 
   oe-core:
-    url: https://github.com/openembedded/openembedded-core
+    url: https://git.openembedded.org/openembedded-core
     layers:
       meta:
 
   bitbake:
-    url: https://github.com/openembedded/bitbake
+    url: https://git.openembedded.org/bitbake
     layers:
       .: disabled
 
--- a/ci/qcom-distro.yml
+++ b/ci/qcom-distro.yml
@@ -11,7 +11,7 @@ repos:
     branch: main
 
   meta-openembedded:
-    url: https://github.com/openembedded/meta-openembedded
+    url: https://git.openembedded.org/meta-openembedded
     layers:
       meta-filesystems:
       meta-gnome:


.github/workflows/build-yocto.yml
```diff
@@ -20,6 +20,7 @@ concurrency:
 env:
   CACHE_DIR: /efs/qli/meta-qcom
   KAS_REPO_REF_DIR: /efs/qli/meta-qcom/kas-mirrors
+  KAS_PREMIRRORS: "https://git\\.openembedded\\.org/ https://github.com/openembedded/\n"
```
ndechesne, line 23: why are we doing this change? this is not obvious from the commit.

.github/workflows/build-yocto.yml
```diff
@@ -21,6 +21,7 @@ env:
   CACHE_DIR: /efs/qli/meta-qcom
   KAS_REPO_REF_DIR: /efs/qli/meta-qcom/kas-mirrors
   KAS_PREMIRRORS: "https://git\\.openembedded\\.org/ https://github.com/openembedded/\n"
+  KAS_CLONE_DEPTH: 1
```
ndechesne, line 22: should we get rid of KAS_REPO_REF_DIR ? both don't seem to make sense together, right?

.github/workflows/build-yocto.yml
quaresmajose: the next commit make it more clear.

.github/workflows/build-yocto.yml
quaresmajose: I've already been here to see if I should do this or not, and I still think we benefit from having it because we continue to have a small delta in the updates.

quaresmajose: I also added some timestamp to better understand the cost of using the KAS_REPO_REF_DIR.

quaresmajose: The time is considerably reduced from approximately 4 minutes to 40 seconds. I will eliminate KAS_REPO_REF_DIR

.github/workflows/build-yocto.yml
ricardosalveti: Yeah, please extend your commit message saying why we need KAS_PREMIRRORS, still not entirely clear to me.

What do we gain from using git.openembedded.org and later replacing with github.com/openembedded/?

ci/base.yml
```diff
@@ -13,12 +13,12 @@ repos:
   meta-qcom:
 
   oe-core:
-    url: https://github.com/openembedded/openembedded-core
+    url: https://git.openembedded.org/openembedded-core
```
ndechesne, line 16: you are reverting a change we made in 2d217e785c1819179070e6e63c61fbdd278c2c40, for a good reason. Why is it a good thing to revert?

.github/workflows/build-yocto.yml
```diff
@@ -36,13 +35,6 @@ jobs:
           wget -qO ${KAS_CONTAINER} https://raw.githubusercontent.com/siemens/kas/refs/tags/$LATEST/kas-container
           chmod +x ${KAS_CONTAINER}
 
-      - name: Update kas mirrors
```
ndechesne: this can/should be squashed with the previous commit.

ci/base.yml
quaresmajose: with the introduction of KAS_PREMIRRORS we can use the official git of the project and we can keep using the  github.com/openembedded on our ci build. the github mirror it's better because it's within the same infrastructure, so there shouldn't be any availability issues.

.github/workflows/build-yocto.yml
quaresmajose: I had already thought about that, and it becomes more obvious to explain the two if they are together. I'll put them together.

.github/workflows/build-yocto.yml
quaresmajose: I will squash it with the next one and it will be easy to explain.

ci/base.yml
ndechesne: I still do not understand why we are making this change. Sure, for our builders it won't change since you are setting KAS_MIRRORS in the github workflow, but it changes for users building locally.

we moved from git.oe.org to github because we agreed github.com was more reliable. we will end up with the same issues eventually if we merge this change.

ndechesne (commented):
i still have doubts about the first 2 patches, and we also agreed to squash the last 2, before we can merge.

ci/base.yml
koenkooi: These 2 commits move our config from "working for all users, all the time" to "only Github CI gets a reliable config", which is a significant regression.
I've commented on other PRs this week that we shouldn't hide unbreak-me options in KAS configs, this goes even further by moving such an option outside the reach of humans.

quaresmajose: > we moved from git.oe.org to github because we agreed github.com was more reliable. we will end up with the same issues eventually if we merge this change.

At the time I understood that this was for CI and not for everyone. The probability of having failures of this type in CI is substantially higher than locally given the number of builds.

quaresmajose: > These 2 commits move our config from "working for all users, all the time" to "only Github CI gets a reliable config", which is a significant regression.
> I've commented on other PRs this week that we shouldn't hide unbreak-me options in KAS configs, this goes even further by moving such an option outside the reach of humans.

I don't believe this will break our local users. With the death of poky, The git.openembedded.org is currently the official Yocto adress and that's the one that should be used. If there are too many fetching issues, they should be reported to the project so they can be fixed.

.github/workflows/build-yocto.yml
quaresmajose: I ended up not combining the two and added a description to explain them better. They explain themselves better together, but subject gets more complicated.

ci/base.yml
koenkooi: We did that last time and it took a few days for it to get fixed and worse, while it was being fixed, the URL structure changed. The blame is fully on the AI bots overloading the OE and YP git servers, but it didn't just affect the CI, it affected everyone using our KAS configs that needed to update. The change to use github mirrors was a response to an actual problem that happened.

So far, github as been more reliable (including this weeks outage!) and it doesn't case traffic on the OE and YP servers, so in my opinion there are no downsides to defaulting to github mirrors. If this situation changes, we should reevaluate.

lumag: I would also prefer to point our users to the GH mirros. Let's not increase the load on the free project without necessity. If you think it is confusing, let's add a brief comment instead.

ndechesne: I have said it before, I also prefer to use github mirrors

quaresmajose: The idea would simply be to promote the official OE-core website.

PR #1511, opened 2026-02-10
Ci rt config

DeewanS:
Updated following in meta-qcom/ci/linux-qcom-rt-6.18.yml:
1. kernel cmdline parameters in for Kodiak, Lemans, Monaco & Talos
2. cyclictest
3. stress-ng

orbit CR: https://orbit/CR/4434203

### DIFF
--- a/conf/machine/iq-8275-evk.conf
+++ b/conf/machine/iq-8275-evk.conf
@@ -28,3 +28,8 @@ QCOM_BOOT_FILES_SUBDIR = "qcs8300"
 QCOM_PARTITION_FILES_SUBDIR ?= "partitions/iq-8275-evk/ufs"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-qcs8300"
+
+KERNEL_CMDLINE_EXTRA .= "${@oe.utils.conditional('PREFERRED_PROVIDER_virtual/kernel',\
+   'linux-qcom-rt',\
+   ' rcupdate.rcu_expedited=1 rcu_nocbs=3 isolcpus=3 irqaffinity=0-2,3-7 cpuidle.off=1',\
+   '', d)}"
--- a/conf/machine/iq-9075-evk.conf
+++ b/conf/machine/iq-9075-evk.conf
@@ -28,3 +28,9 @@ QCOM_BOOT_FILES_SUBDIR = "qcs9100"
 QCOM_PARTITION_FILES_SUBDIR ?= "partitions/iq-9075-evk/ufs"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-qcs9100"
+
+KERNEL_CMDLINE_EXTRA .= "${@oe.utils.conditional('PREFERRED_PROVIDER_virtual/kernel',\
+   ' linux-qcom-rt',\
+   ' rcupdate.rcu_expedited=1 rcu_nocbs=7 isolcpus=7 irqaffinity=0-6 cpuidle.off=1',\
+   '', d)}"
+
--- a/conf/machine/qcm6490-idp.conf
+++ b/conf/machine/qcm6490-idp.conf
@@ -21,3 +21,8 @@ QCOM_BOOT_FILES_SUBDIR = "qcm6490"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-qcs6490"
 QCOM_PARTITION_FILES_SUBDIR ?= "partitions/qcm6490-idp/ufs"
+
+KERNEL_CMDLINE_EXTRA .= "${@oe.utils.conditional('PREFERRED_PROVIDER_virtual/kernel',\
+    'linux-qcom-rt',\
+    ' rcupdate.rcu_expedited=1 rcu_nocbs=7 isolcpus=7 irqaffinity=0-6 cpuidle.off=1',\
+    '', d)}"
--- a/conf/machine/qcs8300-ride-sx.conf
+++ b/conf/machine/qcs8300-ride-sx.conf
@@ -29,3 +29,8 @@ QCOM_BOOT_FILES_SUBDIR = "qcs8300"
 QCOM_PARTITION_FILES_SUBDIR ?= "partitions/qcs8300-ride-sx/ufs"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-qcs8300"
+
+KERNEL_CMDLINE_EXTRA .= "${@oe.utils.conditional('PREFERRED_PROVIDER_virtual/kernel',\
+     'linux-qcom-rt',\
+     ' rcupdate.rcu_expedited=1 rcu_nocbs=3 isolcpus=3 irqaffinity=0-2,4-7 cpuidle.off=1',\
+     '', d)}"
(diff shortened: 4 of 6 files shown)


vkraleti:
> Updated following in meta-qcom/ci/linux-qcom-rt-6.18.yml:
> 
> 1. kernel cmdline parameters in for Kodiak, Lemans, Monaco & Talos
> 2. cyclictest
> 3. stress-ng
> 
> orbit CR: https://orbit/CR/4434203

Don't provide links that are not accessible in public.

ci/linux-qcom-rt-6.18.yml
```diff
@@ -5,3 +5,14 @@ local_conf_header:
   kernelprovider: |
     PREFERRED_PROVIDER_virtual/kernel = "linux-qcom-rt"
     PREFERRED_VERSION_virtual/kernel = "6.18%"
+  cmdline: |
+    KERNEL_CMDLINE_EXTRA:append:iq-9075-evk = " rcupdate.rcu_expedited=1 rcu_nocbs=7 isolcpus=7 irqaffinity=0-6 cpuidle.off=1"
```
lumag: This should be set in machine configuration files

ci/linux-qcom-rt-6.18.yml
```diff
@@ -5,3 +5,14 @@ local_conf_header:
...
+    KERNEL_CMDLINE_EXTRA:append:qcs9100-ride-sx = " rcupdate.rcu_expedited=1 rcu_nocbs=7 isolcpus=7 irqaffinity=0-6 cpuidle.off=1"
+    KERNEL_CMDLINE_EXTRA:append:qcs615-ride = " rcupdate.rcu_expedited=1 rcu_nocbs=7 isolcpus=7 irqaffinity=0-6 cpuidle.off=1"
+    KERNEL_CMDLINE_EXTRA:append:rb3gen2-core-kit = " rcupdate.rcu_expedited=1 rcu_nocbs=7 isolcpus=7 irqaffinity=0-6 cpuidle.off=1"
+    KERNEL_CMDLINE_EXTRA:append:qcm6490 = " rcupdate.rcu_expedited=1 rcu_nocbs=7 isolcpus=7 irqaffinity=0-6 cpuidle.off=1"
+    KERNEL_CMDLINE_EXTRA:append:qcm6490-idp = " rcupdate.rcu_expedited=1 rcu_nocbs=7 isolcpus=7 irqaffinity=0-6 cpuidle.off=1"
+    KERNEL_CMDLINE_EXTRA:append:qcs8300-ride-sx = " rcupdate.rcu_expedited=1 rcu_nocbs=3 isolcpus=3 irqaffinity=0-2,4-7 cpuidle.off=1"
+    KERNEL_CMDLINE_EXTRA:append:iq-8275-evk = " rcupdate.rcu_expedited=1 rcu_nocbs=3 isolcpus=3 irqaffinity=0-2,4-7 cpuidle.off=1"
+  extra: |
+    IMAGE_INSTALL:append = " rt-tests stress-ng"
```
lumag: And this should be pulled through the normal image / packagegroups

ci/iq-9075-evk-rt.yml
```diff
@@ -0,0 +1,15 @@
...
+  - ci/base.yml
+
+machine: iq-9075-evk
+
+local_conf_header:
+  cmdline: |
+    KERNEL_CMDLINE_EXTRA:append = " rcupdate.rcu_expedited=1 rcu_nocbs=7 isolcpus=7 irqaffinity=0-6 cpuidle.off=1"
+  extra: |
+    IMAGE_INSTALL:append = " rt-tests stress-ng"
+
```
vkraleti: What is the necessity of this config when [iq-9075-evk.yml](https://github.com/qualcomm-linux/meta-qcom/blob/master/ci/iq-9075-evk.yml) and [linux-qcom-rt-6.18.yml](https://github.com/qualcomm-linux/meta-qcom/blob/master/ci/linux-qcom-rt-6.18.yml) are already present? 

Can't we use them together?

ci/linux-qcom-rt-6.18.yml
akdwived-dev: Sure will check conditional append in machine conf and would align, as we need to customize kernel command line only for RT recipe.

ci/linux-qcom-rt-6.18.yml
akdwived-dev: Ok

ci/linux-qcom-rt-6.18.yml
ricardosalveti: Also, please explain, in details, why each of these kernel command line arguments are required, as I would prefer reducing those if possible.

cpuidle.off=1 is a big hammer.

DeewanS: > Sure will check conditional append in machine conf and would align, as we need to customize kernel command line only for RT recipe.

Yes, I will align with this. The RT‑specific kernel command‑line parameters will be handled through machine configuration and the RT recipe, not CI. Thanks for the suggestion.

DeewanS: > This should be set in machine configuration files

Thanks, acknowledged. I will move these KERNEL_CMDLINE_EXTRA entries from the CI YAML into the appropriate machine .conf files so they are applied normally through the layer instead of CI.

DeewanS: > Also, please explain, in details, why each of these kernel command line arguments are required, as I would prefer reducing those if possible.
> 
> cpuidle.off=1 is a big hammer.

Acknowledged. I will provide detailed justification for each of the RT‑related kernel command‑line parameters in the commit message and also evaluate whether some of them (especially cpuidle.off=1) can be reduced or avoided.

ci/linux-qcom-rt-6.18.yml
DeewanS: > And this should be pulled through the normal image / packagegroups

Understood. I will remove rt-tests and stress-ng from the CI YAML and instead include them through the standard image/packagegroup mechanism

DeewanS:
> Missing commit messages, random changes which are reverted by later commits, this is not acceptable.

There were indeed unnecessary intermediate changes and reversions.
I’ll rewrite the commits to ensure:
1. clean, linear history
2. clear and complete commit messages
3. no reverted or noisy commits

This should resolve the concerns about missing explanations and random changes.

ci/iq-9075-evk-rt.yml
DeewanS: Thanks for pointing this out. The extra CI config was unintended duplication. The separate ci/iq-9075-evk-rt.yml (and the cmdline changes under ci/linux-qcom-rt-6.18.yml) were later reverted in subsequent commits, which caused confusion in the review.

Plan going forward:
1. Will not rely on CI YAML to carry RT kernel command line or test package changes.
2. The machine selection (e.g., iq-9075-evk.yml) + the RT kernel recipe (linux-qcom-rt) will be used together, but the kernel cmdline will come from the machine .conf / RT bbappend in the layer, not from CI.
3. Commit history will be cleaned to remove the add/revert noise and keep a single, coherent change set.

PR #1517, opened 2026-02-10
Replace individual SELinux and SOTA builds in CI with catchall build

vkraleti:
Update the GitHub workflow to build the qcom-distro-catchall variant
instead of running separate SELinux and SOTA builds. The catch-all
distro enables all relevant DISTRO_FEATURES in a single configuration,
reducing number of CI builds.


vkraleti:
Depends-on: https://github.com/qualcomm-linux/meta-qcom-distro/pull/156

Issue #1520, opened 2026-02-10
bitbake stuck at "Initialising tasks:" 44%


quaresmajose:
When bitbake starts from scracth it takes a lot of time to populate the Hash Equivalence database build/cache/hashserv.db.

The issue can be observed locally if we start from an empty build directory:

```
bitbake core-image-base -n
```
and bitbake gets stuck in `"Initialising tasks:" 44%` for quite a while.

In our builds, I noticed that in some cases the time reaches 20 minutes.
```
Tue, 10 Feb 2026 10:59:34 GMT 2026-02-10 10:59:34 - INFO     - 
Tue, 10 Feb 2026 11:19:02 GMT 2026-02-10 11:19:02 - INFO     - Initialising tasks...done.
Tue, 10 Feb 2026 11:19:28 GMT 2026-02-10 11:19:28 - INFO     - Checking sstate mirror object availability...done.
Tue, 10 Feb 2026 11:19:28 GMT 2026-02-10 11:19:28 - INFO     - Sstate summary: Wanted 5623 Local 4643 Mirrors 3 Missed 977 Current 0 (82% match, 0% complete)
```
https://github.com/qualcomm-linux/meta-qcom/actions/runs/21861466505/job/63093820565?pr=1509

quaresmajose:
The proposed PR https://github.com/qualcomm-linux/meta-qcom/pull/1523 To use the hash equivalence globally and share the database location, doesn't work because our persistent storage is on NFS, and this file system is not supported for this use case.

PR #1523, opened 2026-02-10
Use the hash equivalence globally and share the database location

quaresmajose:
- share the hash server database location
- use the hash equivalence globally

fixes https://github.com/qualcomm-linux/meta-qcom/issues/1520

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -34,13 +34,6 @@ jobs:
           wget -qO ${KAS_CONTAINER} https://raw.githubusercontent.com/siemens/kas/refs/tags/$LATEST/kas-container
           chmod +x ${KAS_CONTAINER}
 
-      - name: Update kas mirrors
-        run: |
-          for r in $(find ${KAS_REPO_REF_DIR}/* -maxdepth 0 -type d); do
-            echo "pre-fetch: $r"
-            git -C $r fetch --prune origin '+refs/*:refs/*'
-          done
-
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


ci/mirror.yml
```diff
@@ -7,3 +7,4 @@ local_conf_header:
   mirror: |
     BB_HASHSERVE_UPSTREAM = "wss://hashserv.yoctoproject.org/ws"
     SSTATE_MIRRORS = "file://.* http://sstate.yoctoproject.org/all/PATH;downloadfilename=PATH"
+    BB_SIGNATURE_HANDLER = "OEEquivHash"
```
lumag, line 12: Please apply this change in the qcom-distro configuration instead.

quaresmajose: No, the ideia is to have it applied globally in CI. This ensures that it's independent of the distro, which is the problem we need to fix.

lumag: This needs to be captured in the commit message. And could you please adjust qcom-distro config anyway.

koenkooi: And we should absolutely *not* have unbreak-me options in KAS configs, they need to be in the OE portion itself, be it `layer.conf`, `MACHINE.conf` or` DISTRO.conf`.

quaresmajose: I don't think we should force our users to use this or that sstate signature method. However, I'm going to close this one and open another because the one I had in mind isn't working.

lumag:
It's it possible to switch from NFS to any other solution, which will have correct fcntl semantics?

PR #1526, opened 2026-02-10
Add kas config for debug build

vkraleti:
Add `debug.yml` kas configuration to enable builds with DEBUG_BUILD set to "1".

### DIFF
--- a/ci/debug.yml
+++ b/ci/debug.yml
@@ -1,0 +1,6 @@
+header:
+  version: 14
+
+local_conf_header:
+  debug-build: |
+    DEBUG_BUILD = "1"


ricardosalveti:
How are we expecting to use the debug variants in our testing? These would probably be good targets for nightly instead.

vkraleti:
> > These images are normally used for stress and stability testing.
> > For now want to keep them as separate builds as there are concerns expressed to make debug fragment by default in PR/nightly builds. See [#1400 (comment)](https://github.com/qualcomm-linux/meta-qcom/pull/1400#discussion_r2714685185)
> 
> Right, having a debug yml fragment is fine, my concern is more towards what @lumag said, the concern that this will be the main image used during our testing and validation, and not the actual production one.

Regular images are not getting disturbed in any manner. They continue to serve PR pre-merge tests, nightly tests. Sanity will also continue on them. These are additional images for stress testing. If you really feel that they may be misused, I can move the debug build generation to an independent GitHub action so that the build can be generated on demand basis without tagging to any of existing CI jobs.

ricardosalveti:
> Ok, I dropped changes in build-yocto.yml for building these images by default. Just retained the kas config file so that users when needed can trigger these builds.

I think it is fine to build (via nightly or weekly), but for that we need to make sure that the teams are only using them for the correct purposes.

sbanerjee-quic (approved):

p-badge:
We need to generate debug builds for test team consumption at least for 2.0 (6.18.y). This is required for reaching our stability goals on commercialization branch. Without debug builds it is impossible to debug certain stability issues. 
We cannot generate debug builds on need basis as stability runs and stabilization process is continuous and incremental throughout the lifecycle of commercial branch.

knaveen-qc:
For QLI commercialization, it is necessary to validate the software baseline across stability, performance, and feature requirements. This validation can only be achieved by utilizing regular, debug, and performance build variants, as each variant enables targeted testing and ensures comprehensive coverage for commercial readiness.

To support the commercialization process and ensure thorough validation, it is essential to have regular, debug, and performance build variants available. These builds allow test teams to execute specialized test suites targeting stability, performance, and feature requirements, thereby providing comprehensive coverage and addressing different quality dimensions necessary for commercial readiness.

To effectively track and resolve issues identified during testing by various teams, it is essential to have official, trackable builds—including regular, debug, and performance variants—generated from the nightly CI pipeline. These builds ensure comprehensive validation for commercialization and align with standard practices across all programs.

Therefore, generating regular, debug, and performance build variants through the CI pipeline is essential to thoroughly validate the software for commercialization. The frequency of these builds can be further discussed with program PEs to ensure efficient resource utilization.

vkraleti:
> @knaveen-qc make regular = performance, there is no need to have two separate builds.

Lava tests are currently expecting console logs. Once @mwasilew and @smuppand confirm Lava coan work without them, make regular same as performance and drop separate performance build.

lumag:
There is no `console=quiet`. It's just `quiet`.

mwasilew:
> @mwasilew I'm fine with landing a qcom-specific patch which makes Linux version being always printed. Would that work for you?

yes. But it doesn't have to be there. If the bootloader logs are displayed they can be used as well. For example:
```
Exit EBS        [ 3834] UEFI End
```
This is good enough for LAVA as a "linux started booting" string - no need to carry Qualcomm specific patches.

PR #1534, opened 2026-02-11
qairt-sdk: align libcdsprpc dependency with provider SONAME

tjamula:
Some shared libraries listed an unversioned dependency on libcdsprpc.so, while the fastrpc package exports a versioned SONAME.

Update the DT_NEEDED entries in the binaries to depend on the provider SONAME instead, ensuring correct dependency resolution at both packaging and runtime.

### DIFF
--- a/recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
+++ b/recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
@@ -16,6 +16,10 @@ SRC_URI[sha256sum] = "c6306f7d70457e4d8736416bb406d8670f5186fb648dd6808db24e9b21
 
 S = "${UNPACKDIR}/qairt/${PV}"
 
+# We need host tools during the build:
+#  - patchelf-native: to patch DT_NEEDED entries in prebuilt .so files
+#  - binutils-native: for readelf to inspect NEEDED
+DEPENDS = "patchelf-native binutils-native"
 
 # The SDK ships multiple toolchain-specific lib directories with names
 # like "aarch64-oe-linux-gcc8.2", "aarch64-oe-linux-gcc9.3", etc.
@@ -68,8 +72,21 @@ do_install() {
     cp -r ${S}/bin/${PLATFORM_DIR}/* ${D}${bindir}
 }
 
-# Skip file-rdeps due to prebuilt libhta_hexagon_runtime_snpe with mismatched SONAME.
-INSANE_SKIP:${PN} += "file-rdeps"
+# Some shared libraries depend on the unversioned 'libcdsprpc.so',
+# while the provider (fastrpc) exports the versioned SONAME 'libcdsprpc.so.1'.
+# Rewrite DT_NEEDED from libcdsprpc.so to libcdsprpc.so.1 for affected libraries
+# to avoid packaging/runtime dependency mismatches.
+#
+# Can be dropped once fixed upstream (planned in QAIRT SDK v2.45)
+do_patch_qairt_needed_soname() {
+    for so in ${D}${libdir}/lib*.so; do
+        if readelf -d "$so" 2>/dev/null | grep -q "NEEDED.*libcdsprpc\.so"; then
+            patchelf --replace-needed libcdsprpc.so "libcdsprpc.so.1" "$so"
+        fi
+    done
+}
+
+addtask patch_qairt_needed_soname after do_install before do_package
 
 # SDK ships already-stripped proprietary binaries
 # which need not be re-striped or split into debug symbols.


recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -68,8 +68,42 @@ do_install() {
     cp -r ${S}/bin/${PLATFORM_DIR}/* ${D}${bindir}
 }
 
-# Skip file-rdeps due to prebuilt libhta_hexagon_runtime_snpe with mismatched SONAME.
-INSANE_SKIP:${PN} += "file-rdeps"
+# We need host tools during the build:
+#  - patchelf-native: to patch DT_NEEDED entries in prebuilt .so files
+#  - binutils-native: for readelf to inspect SONAME / NEEDED
+#  - fastrpc: so the provider library is present in the sysroot for SONAME lookup
+DEPENDS += "patchelf-native binutils-native fastrpc"
```
koenkooi: I'm not a big fan of having DEPENDS in multiple places or in the middle of a recipe, I can see why it's here, but I have a slight preference for moving this whole block to where DEPENDS normally lives (after SUMMARY/DESCRIPTION/LIC*). That should also get rid of the need for +=.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -68,8 +68,42 @@ do_install() {
...
+
+do_install:append() {
```
koenkooi: This is more thinking out loud than a review comment: would making this its own task fit the OE way of doing things better? `do_install` makes sense from a workflow point of view, but fix ups like this are generally given their own tasks.
Like I said, this is not a real review comment, this just made me realize that OE (and meta-qcom*) lack objective rules for this :)

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -68,8 +68,42 @@ do_install() {
...
+    # Locate the provider library in the recipe sysroot.
+    # We read its SONAME so we can patch QAIRT binaries to depend on the SONAME.
+    provider="${RECIPE_SYSROOT}${libdir}/libcdsprpc.so"
```
lumag: Why do we need to locate it? We know that we need to patch `libcdsprpc.so` -> `libcdsprpc.so.1`. With the next release of QAIRT this will go away at all (hopefully). So, this looks really overcomplicated for that task.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
ricardosalveti: Yeah, a task after compile or even as part of compile would work better here. Changing the content as part of do_install is not ideal.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
ricardosalveti: Yeah, just patch it directly to libcdsprpc.so.1, as this will be removed once there is a newer qairt release.

ricardosalveti:
Also please say that this will fixed with a newer qarit release (saying it will be released in weeks/months), and once released this will be reverted.

Add that to your commit message.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: The intent here is to obtain the SONAME from the library itself rather than hardcoding it, so that the solution remains robust even if the SONAME of libcdsprpc.so changes in future.
With that in mind, please let me know if you would still prefer hardcoding libcdsprpc.so.1, assuming the SONAME is expected to remain same.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: This is patching the binaries that are being packed to avoid packaging errors, so doing this patching to the installed libs makes more sense to me, if a separate task is needed, I would prefer having it like the following:
```bitbake
addtask patch_qairt_needed_soname after do_install before do_package
```
Let me know if that works?

tjamula:
> Also please say that this will fixed with a newer qarit release (saying it will be released in weeks/months), and once released this will be reverted.
> 
> Add that to your commit message.

Updated the commit message

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
ndechesne: I think this is really nitpicking here, this is just one shell command , and if it works let's keep it. the goal being to remove it very soon anyway. 

please convert to the task as discussed, and then we should be able to merge

tjamula: Converted to a task now.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -68,8 +73,33 @@ do_install() {
     cp -r ${S}/bin/${PLATFORM_DIR}/* ${D}${bindir}
 }
 
-# Skip file-rdeps due to prebuilt libhta_hexagon_runtime_snpe with mismatched SONAME.
-INSANE_SKIP:${PN} += "file-rdeps"
+do_patch_qairt_needed_soname() {
+    provider="${RECIPE_SYSROOT}${libdir}/libcdsprpc.so"
+
+    # Extract SONAME from provider so we can patch QAIRT binaries to depend on the same.
```
lumag: We _know_ it is libcdsprpcd.1.

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
```diff
@@ -68,8 +73,33 @@ do_install() {
...
+    if [ -z "$soname" ]; then
+        # Fallback keeps build deterministic if SONAME lookup fails for any reason.
+        soname="libcdsprpc.so.1"
+    fi
+
+    # Patch only the shared libraries that actually list 'NEEDED: libcdsprpc.so'.
+    # This aligns the dependency with what the provider exports (SONAME),
+    # fixing packaging/runtime dependency mismatch.
+    for so in ${D}${libdir}/lib*.so; do
+        [ -e "$so" ] || continue
```
lumag: Drop the check

lumag (changes requested):

recipes-ml/qairt/qairt-sdk_2.41.0.251128.bb
tjamula: Modified as suggested.

tjamula:
@lumag, @ricardosalveti, @ndechesne, @koenkooi

The comments have been addressed. Please review when you get a chance.

PR #1553, opened 2026-02-13
several fixes for DTB overlays support

lumag:
Fix indentation of variables, use DTBOs where applicable and drop non-standard entries which can not be selected anyways.


conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -97,4 +97,4 @@ FIT_DTB_COMPATIBLE[sm8450-hdk] = "qcom,sm8450-hdk"
 FIT_DTB_COMPATIBLE[sm8750-mtp] = "qcom,sm8750-mtp"
 FIT_DTB_COMPATIBLE[talos-evk] = "qcom,qcs615v1.1-iot"
 FIT_DTB_COMPATIBLE[talos-evk-camx] = "qcom,qcs615v1.1-iot-camx"
-FIT_DTB_COMPATIBLE[talos-evk-lvds-auo_g133han01] = "qcom,talos-evk-lvds-auo,g133han01"
+FIT_DTB_COMPATIBLE[talos-evk+talos-evk-lvds-auo_g133han01] = "qcom,talos-evk-lvds-auo,g133han01"
```
vkraleti, line 98: Just realized that underscore to comma conversion in [dtb_only_fitimage.py](https://github.com/qualcomm-linux/meta-qcom/blob/master/lib/qcom/dtb_only_fitimage.py) only happening to base DTB but not for overlays. Without fixing it above change will result in a broken config section.

`lookup_key = "+".join([os.path.splitext(dtb_id)[0].replace(',', '_')] + [os.path.splitext(ovl)[0] for ovl in ovl_list])`

lumag: Argh.

PR #1556, opened 2026-02-13
ci/base: fix incorrect regex for CodeLinaro yocto-mirrors 

vkraleti:
Fix an incorrect regex used for matching CodeLinaro yocto-mirrors URLs.
The previous pattern assumed a single path component, which caused
incorrect mirror resolution for repositories hosted under nested paths.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -46,7 +46,7 @@ local_conf_header:
   clo-mirrors: |
     MIRRORS:append = " \
     git://github.com git://git.codelinaro.org/clo/yocto-mirrors/github/ \
-    git://.*/.*/ git://git.codelinaro.org/clo/yocto-mirrors/ \
+    git://.*/ git://git.codelinaro.org/clo/yocto-mirrors/ \
     https://.*/.*/ https://artifacts.codelinaro.org/codelinaro-le/ \
     "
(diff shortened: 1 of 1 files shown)


ci/base.yml
```diff
@@ -46,6 +46,7 @@ local_conf_header:
   clo-mirrors: |
     MIRRORS:append = " \
     git://github.com git://git.codelinaro.org/clo/yocto-mirrors/github/ \
+    git://gitlab.*/ git://git.codelinaro.org/clo/yocto-mirrors/gitlab/ \
```
lumag: Is it really an issue? Do we see GitLab flakes?

quaresmajose: We only have 5 repos on this folder yocto-mirrors/gitlab/, why don't we move them (creating shortcuts) to yocto-mirrors/github/ ?

vkraleti: All 5 GitLab projects are now mirrored under git://git.codelinaro.org/clo/yocto-mirrors. I am dropping the commit that added yocto-mirrors/gitlab to MIRRORS. However, the regex fix is still valid, so retaining it.

lumag: You didn't drop it

PR #1557, opened 2026-02-13
Extend compatibles strings for camera variants

Kavinaya99:
Extend the compatible string to be used with base + camera variants.

### DIFF
--- a/conf/machine/include/fit-dtb-compatible.inc
+++ b/conf/machine/include/fit-dtb-compatible.inc
@@ -52,7 +52,10 @@ FIT_DTB_COMPATIBLE[qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine] = " \
     qcom,qcs5430-iot-subtype2 \
     qcom,qcs6490-iot-subtype2 \
 "
-FIT_DTB_COMPATIBLE[qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine-camx] = "qcom,qcs6490-iot-camx"
+FIT_DTB_COMPATIBLE[qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine-camx] = " \
+    qcom,qcs6490-iot-camx \
+    qcom,qcs6490-iot-subtype2-camx \
+"
 FIT_DTB_COMPATIBLE[qcs8300-ride] = "qcom,qcs8300-adp"
 FIT_DTB_COMPATIBLE[qcs8300-ride+qcs8300-ride-camx] = "qcom,qcs8300-adp-camx"
 FIT_DTB_COMPATIBLE[qcs9100-ride] = " \


Kavinaya99:
> None of the commit messages explains why the change is being performed. Please always start your commit messages with the statement that something doesn't work as expected, something is broken, etc.

Sure, will update the commit messages in the next patch.

conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -52,7 +52,10 @@ FIT_DTB_COMPATIBLE[qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine] = " \
     qcom,qcs5430-iot-subtype2 \
     qcom,qcs6490-iot-subtype2 \
 "
-FIT_DTB_COMPATIBLE[qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine-camx] = "qcom,qcs6490-iot-camx"
+FIT_DTB_COMPATIBLE[qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine-camx] = " \
+    qcom,qcs6490-iot-camx \
+    qcom,qcs6490-iot-subtype2-camx \
```
lumag, line 57: If we already have `qcom,qcs6490-iot-camx`, why do we need a separate `qcom,qcs6490-iot-subtype2-camx`?

Kavinaya99: If a user boots with camera config1 (i.e. qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine-camx) on vision mezzanine boards and then wants to switch to enable the downstream camera i.e. config2 (i.e. qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine-camx), this change would be necessary for that combination.

lumag: It's not the answer to the question that i asked. We already have `qcom,qcs6490-iot-camx` compatible for `qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine-camx`. Why do we need to add the one with the subtype? Isn't the more generic one also matched?

Kavinaya99: Vision kit has a different CDT. Hence the mapping needed to use camera on vision kit would be qcom,qcs6490-iot-subtype2-camx.

lumag: The non-subtype entry should match any subtypes. Am I wrong?

Kavinaya99: We have tried this experiment as well (i.e. using qcom,qcs6490-iot-camx) for vision kit to see if camx dtbos are getting picked up, but its not getting picked up.
Adding qcom,qcs6490-iot-subtype-camx will only ensure that camx dtbos are getting picked up for vision kits.

Kavinaya99: For the core kit, the configuration is:
qcom, qcs6490, iot, subtype5, emmc, 4GB+, softsku0, camx

The available compatible strings are:
qcom, qcs6490-iot
qcom, qcs6490-iot-subtype9
qcom, qcs6490-iot-subtype2

Since subtype5 does not exist in this list, the matching can only happen up to the common part: qcom, qcs6490, iot.
So the compatible string selected is: qcom, qcs6490-iot
Because the match is not subtype-specific, camx can be added independently, and the compatible string for camx can be: qcom, qcs6490-iot-camx

Vision Kit
For the vision kit, the configuration is:
qcom, qcs6490, iot, subtype2, emmc, 4GB+, softsku0

Using the same set of compatible strings, subtype2 is explicitly present: qcom, qcs6490-iot-subtype2
So the best match happens for:qcom, qcs6490, iot, subtype2
Because the match is now subtype-specific, camx must also be tied to this subtype. Hence, camx must always be passed as: qcom, qcs6490-iot-subtype2-camx

This ensures correct and consistent matching behavior for both core kit and vision kit.

lumag: ok, thanks for the explanation.

lumag (approved):

conf/machine/include/fit-dtb-compatible.inc
sbanerjeeqc: @lumag my understanding is that the non-subtype should match the subtype5

lumag: @sbanerjee-quic why / how? It doesn't match the identified rules for the matching the fallback (@Kavinaya99 's epxlanation does).

PR #1560, opened 2026-02-13
linux-qcom: specify dummy revision for devupstream

anujm1:
While correct, specifying AUTOREV in a recipe makes bitbake query the repository while parsing. That in turn leads to build failures in case of offline builds.

Specify the same revision as current fixed revision for devupstream as well to avoid this. It should be overridden by the user in local.conf.

### DIFF
--- a/recipes-kernel/linux/linux-qcom-next_git.bb
+++ b/recipes-kernel/linux/linux-qcom-next_git.bb
@@ -29,7 +29,9 @@ SRC_URI += " \
 # virtual/kernel provider to 'linux-qcom-next-upstream'
 BBCLASSEXTEND = "devupstream:target"
 PN:class-devupstream = "linux-qcom-next-upstream"
-SRCREV:class-devupstream ?= "${AUTOREV}"
+
+# Pin this to the qcom-next-6.19-rc6-20260207, override this in local.conf
+SRCREV:class-devupstream ?= "ccc1345fb4742e377d7d89658efe085c62c25164"
 
 S = "${UNPACKDIR}/${BP}"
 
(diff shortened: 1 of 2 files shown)


recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -29,7 +29,9 @@ SRC_URI += " \
 # virtual/kernel provider to 'linux-qcom-next-upstream'
 BBCLASSEXTEND = "devupstream:target"
 PN:class-devupstream = "linux-qcom-next-upstream"
-SRCREV:class-devupstream ?= "${AUTOREV}"
+
+# Pin this to the qcom-next-6.19-rc6-20260207, override this in local.conf
+SRCREV:class-devupstream ?= "ccc1345fb4742e377d7d89658efe085c62c25164"
```
lumag, line 34: This should be using something like linux-yocto-dev:
```
SRCREV ?= '${@oe.utils.conditional("PREFERRED_PROVIDER_virtual/kernel", "linux-qcom-next-upstream", "${AUTOREV}", "29594404d7fe73cd80eaa4ee8c43dcc53970c60e", d)}'
```
(yes, 3.7 was a good kernel)

anujm1: For an offline build with `linux-qcom-next-upstream`, it would still require a user to add an entry in local.conf to override `SRCREV`. So, does this approach offer any advantage?

linux-yocto-dev does not inherit `devupstream` while all others inheriting it expect users to override `SRCREV` in local.conf so it offers a consistent behaviour.

lumag: I find it wasier to switch PREFERRED_PROVIDER in local.conf than switching provider and setting the AUTOREV.

PR #1563, opened 2026-02-13
Use the sstate-cache basic hash signature globally

quaresmajose:
We are currently using different signatures depending on the distro variant used:

 qcom-distro: BB_SIGNATURE_HANDLER=OEBasicHash
 poky-altcfg: BB_SIGNATURE_HANDLER=OEEquivHash

The ci/base.yml is using hash equivalence signature because it uses the poky-altcfg
distro and it inherits what is defined in poky upstream.
The ci/qcom-distro.yml is using the basic signature because it uses the qcom-distro
distro and it inherits the bitbake nodistro.
Doesn't make sense to use two different signatures depending on the distribution in use.

Using the hash equivalence globally allows we to continue taking advantage of the
official Yocto sstate mirror. Unfortunately, this isn't possible for now because
our storage is on an NFS server and we can't share it between workers in any other way.

So we're left with the option of using the basic hash signatures. Since there is no database
here, we also fixed the problem of populating the database that occurs with hash equivalence.

Fixes https://github.com/qualcomm-linux/meta-qcom/issues/1520

### DIFF
--- a/ci/mirror.yml
+++ b/ci/mirror.yml
@@ -5,5 +5,4 @@ header:
 
 local_conf_header:
   mirror: |
-    BB_HASHSERVE_UPSTREAM = "wss://hashserv.yoctoproject.org/ws"
     SSTATE_MIRRORS = "file://.* http://sstate.yoctoproject.org/all/PATH;downloadfilename=PATH"


lumag:
I agree with the PR, but I will raise the same question here: can we switch our storage to something else than NFS?

ndechesne:
> I agree with the PR, but I will raise the same question here: can we switch our storage to something else than NFS?

we use AWS EFS. Do we have the same issues

         This variable cannot be set to a NFS mount and bitbake will error out then.
         The reason is that NFS implementations can have file locking issues, which
         can cause data loss and corruption when there are multiple writers operating
         on a file at the same time as explained in https://sqlite.org/faq.html#q5

ndechesne:
> > It is an NFS-like server, but Amazon EWS provides locks support (https://docs.aws.amazon.com/efs/latest/ug/limits.html#limits-client-specific)
> > > Each unique mount on the client can acquire up to a total of 65,536 locks per connection.
> 
> I can patch bitbake to avoid failure, we would carry this patch over to our end for some time to validation.

How worth is it to get the OEEquivHash instead of OEBasicHash?

quaresmajose:
There is a warning bug in the oe-core/meta/classes-global/sanity.bbclass. 
```
2026-02-13 15:33:01 - INFO     - WARNING: You are using a local hash equivalence server but have configured an sstate mirror. This will likely mean no sstate will match from the mirror. You may wish to disable the hash equivalence use (BB_HASHSERVE), or use a hash equivalence server alongside the sstate mirror.
```

The warning is only valid when BB_SIGNATURE_HANDLER=OEEquivHash. I'm going to propose a fix for this on the mailing list.

PR #1565, opened 2026-02-14
Add blacklist package for iris-video-dlkm

vkraleti:
The iris-vpu downstream video driver conflicts with upstream qcom_iris and Venus drivers. Install modprobe configuration files to blacklist these upstream kernel modules when using the iris-vpu driver.

Also, provide update-alternatives for the blacklist configuration to allow users to switch between upstream and downstream video drivers at runtime.

### DIFF
--- a/recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
+++ b/recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
@@ -5,10 +5,14 @@ LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=12f884d2ae1ff87c09e5b7ccc2c4ca7e"
 
 PV = "0.0+git"
 
-SRC_URI = "git://github.com/qualcomm-linux/video-driver.git;protocol=https;branch=video.qclinux.main"
+SRC_URI = " \
+    git://github.com/qualcomm-linux/video-driver.git;protocol=https;branch=video.qclinux.main \
+    file://blacklist-video.conf.venus \
+    file://blacklist-video.conf.vidc \
+"
 SRCREV  = "0e0fe75fb1910e011358485b078ea207a5c5f3e7"
 
-inherit module
+inherit module update-alternatives
 
 MAKE_TARGETS = "modules"
 
@@ -16,3 +20,30 @@ MAKE_TARGETS = "modules"
 # Therefore, builds for other architectures are not necessary and are explicitly excluded.
 COMPATIBLE_MACHINE = "^$"
 COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
+do_install:append() {
+    install -d ${D}${sysconfdir}/modprobe.d
+    install -Dm 0644 ${UNPACKDIR}/blacklist-video.conf.venus \
+            ${D}${sysconfdir}/modprobe.d/blacklist-video.conf.venus
+    install -Dm 0644 ${UNPACKDIR}/blacklist-video.conf.vidc \
+            ${D}${sysconfdir}/modprobe.d/blacklist-video.conf.vidc
+}
+
+PACKAGES += "${PN}-bl-venus ${PN}-bl-vidc"
+RDEPENDS:${PN} += "${PN}-bl-venus ${PN}-bl-vidc"
+
+FILES:${PN}-bl-venus += "${sysconfdir}/modprobe.d/blacklist-video.conf.venus"
+FILES:${PN}-bl-vidc += "${sysconfdir}/modprobe.d/blacklist-video.conf.vidc"
+
+ALTERNATIVE:${PN}-bl-vidc  = "blacklist-video"
+ALTERNATIVE_TARGET_${PN}-bl-vidc = "${sysconfdir}/modprobe.d/blacklist-video.conf.vidc"
+ALTERNATIVE_PRIORITY_${PN}-bl-vidc = "50"
+
+# On QCS615, prioritize blacklisting unsupported Vidc.
+ALTERNATIVE_PRIORITY_${PN}-bl-vidc:qcs615 = "150"
+
+ALTERNATIVE:${PN}-bl-venus = "blacklist-video"
+ALTERNATIVE_TARGET_${PN}-bl-venus = "${sysconfdir}/modprobe.d/blacklist-video.conf.venus"
+ALTERNATIVE_PRIORITY_${PN}-bl-venus = "100"
+
+ALTERNATIVE_LINK_NAME[blacklist-video] = "${sysconfdir}/modprobe.d/blacklist-video.conf"
--- a/recipes-kernel/iris-video-module/iris-video-dlkm/blacklist-video.conf.venus
+++ b/recipes-kernel/iris-video-module/iris-video-dlkm/blacklist-video.conf.venus
@@ -1,0 +1,7 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+blacklist qcom_iris
+blacklist venus_dec
+blacklist venus_enc
+blacklist venus_core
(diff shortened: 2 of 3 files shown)


recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
```diff
@@ -16,3 +19,16 @@ MAKE_TARGETS = "modules"
...
+
+do_install:append() {
+    install -Dm 0644 ${UNPACKDIR}/blacklist-upstream-video.conf \
+            ${D}${sysconfdir}/modprobe.d/blacklist-upstream-video.conf
+}
+
+# For QCS615 continue using upstream Venus Video Driver as Iris isn't supported
+do_install:append:qcs915() {
+    sed -i '/^blacklist/d' ${D}${sysconfdir}/modprobe.d/blacklist-upstream-video.conf
+}
```
sbanerjeeqc: Typo 915

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
```diff
@@ -16,3 +19,16 @@ MAKE_TARGETS = "modules"
...
 COMPATIBLE_MACHINE = "^$"
 COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
+do_install:append() {
+    install -Dm 0644 ${UNPACKDIR}/blacklist-upstream-video.conf \
+            ${D}${sysconfdir}/modprobe.d/blacklist-upstream-video.conf
+}
+
+# For QCS615 continue using upstream Venus Video Driver as Iris isn't supported
+do_install:append:qcs915() {
```
sbanerjeeqc: Instead of sed skipping the file from being installed for 615

vkraleti: Ok Sure.

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
vkraleti: corrected.

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
vkraleti: done

sbanerjee-quic:
@vkraleti is qcs615 unsupported by iris-video-dlkm? 
Or the case is iris-video-dlkm does support but as a product ask is to use upstream driver as the default for Talos?

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
```diff
@@ -16,3 +19,17 @@ MAKE_TARGETS = "modules"
...
 COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
+do_install:append() {
+    install -Dm 0644 ${UNPACKDIR}/blacklist-upstream-video.conf \
+            ${D}${sysconfdir}/modprobe.d/blacklist-upstream-video.conf
+}
+
+# For QCS615 continue using upstream Venus Video Driver as Iris isn't supported
+do_install:append:qcs615() {
+    rm -f ${D}${sysconfdir}/modprobe.d/blacklist-upstream-video.conf
```
lumag: I don't completely like this approach. Can we have two blacklists, one for iris/venus, one for vidc and switch between them using alternatives? 

This way we can set priorities as required (prefer vidc everywhere except for qcs615), but then our users can switch priorities without reflashing the device.

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
```diff
@@ -16,3 +19,17 @@ MAKE_TARGETS = "modules"
...
+}
+
+PACKAGE_BEFORE_PN += "${PN}-blacklist"
```
lumag: No need for before_pn

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
```diff
@@ -16,3 +19,17 @@ MAKE_TARGETS = "modules"
...
+FILES:${PN}-blacklist = "${sysconfdir}/modprobe.d"
+ALLOW_EMPTY:${PN}-blacklist = "1"
```
lumag: On the other hand, why do we need it as a separate package?

lumag (changes requested):

vkraleti:
> @vkraleti is qcs615 unsupported by iris-video-dlkm? Or the case is iris-video-dlkm does support but as a product ask is to use upstream driver as the default for Talos?

Iris won't support QCS615. There is no downstream Audio driver for this SoC.

vgarodia:
> > @vkraleti, in reference to qcom-multimedia-proprietary image we do not have a HW platform specific definition. The definition is the use audio, camera, graphics, video components maintained by Qualcomm. I.e., Audioreach, camx, adreno, iris-video-dlkm for all HW platforms.
> 
> This requirement may be derived assuming only Kodiak, Lemans, Monaco. iris-video-dlkm won't support Talos. So there is no possibility for a common set of video modules be used among Kodiak, Lemans, Monaco and Talos. @vgarodia please confirm my understanding.

To reconfirm, video-dlkm does not support Talos now, as well as in future, given the video IP (AR50) firmware does not support newer generation of HFIs, while iris-video-dlkm does not support older generation of HFIs.

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
vkraleti: Updated the implantation to use ALTERNATIVES to provide default blacklist conf.

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
```diff
@@ -5,14 +5,42 @@ LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=12f884d2ae1ff87c09e5b7ccc2c4ca7e"
...
+
+PACKAGES += "${PN}-bl-venus ${PN}-bl-vidc"
+RDEPENDS:${PN} += "${PN}-bl-venus ${PN}-bl-vidc"
+
+FILES:${PN}-bl-venus += "${sysconfdir}/modprobe.d/blacklist-video.conf.venus"
+FILES:${PN}-bl-vidc += "${sysconfdir}/modprobe.d/blacklist-video.conf.vidc"
+
+ALTERNATIVE:${PN}-bl-vidc  = "blacklist-video"
+ALTERNATIVE_TARGET_${PN}-bl-vidc = "${sysconfdir}/modprobe.d/blacklist-video.conf.vidc"
+ALTERNATIVE_PRIORITY_${PN}-bl-vidc = "50"
```
lumag, line 40: The priorities are incorrect on QCS615.

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
```diff
@@ -5,14 +5,42 @@ LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=12f884d2ae1ff87c09e5b7ccc2c4ca7e"
...
 MAKE_TARGETS = "modules"
 
 # This package is designed to run exclusively on ARMv8 (aarch64) machines.
 # Therefore, builds for other architectures are not necessary and are explicitly excluded.
 COMPATIBLE_MACHINE = "^$"
 COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
+do_install:append() {
+    install -d ${D}${sysconfdir}/modprobe.d
+    install -Dm 0644 ${UNPACKDIR}/blacklist-venus.conf \
```
lumag: Why can't we name the files using the expected names, then those can be installed as is.

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
vkraleti: The priorities are kept same for all SoCs assuming the switch would happen at run time. I'll update it.

recipes-kernel/iris-video-module/iris-video-dlkm_git.bb
vkraleti: Done.

PR #1566, opened 2026-02-14
workflows: exclude AudioRecord test on RB8

qcom-anilyada:
AudioRecord test when executed on RB8 causes device to freeze. This causes the test job to timeout. Since the timeout is rather long, and the test never completes it should be disabled until it's fixed.

This patch corresponds to:
qualcomm-linux/lava-test-plans#16

### DIFF
--- a/.github/actions/lava-test-plans/action.yml
+++ b/.github/actions/lava-test-plans/action.yml
@@ -27,7 +27,7 @@ runs:
         with:
           repository: qualcomm-linux/lava-test-plans
           path: lava-test-plans
-          ref: 1ab5e2f1d6cc3559ca4685941cc9fd17ab132c2d
+          ref: 6f79a2492a982d990238545efe3c66ccb87dbe22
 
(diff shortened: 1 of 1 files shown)


vkraleti (approved):

lumag:
What is the reason for the freeze?

lumag (approved):

koenkooi (approved):

bhargav0610:
> The issue needs to be root-caused

Currently, the SSR feature is not supported in audio because the existing ASoC core does not fully support all the cases for card rebinding. We are encountering a code path in the ASoC core that appears to be buggy.
 
To properly fix and enable card rebinding without any crashes, we estimate that it will take approximately 1-2 weeks of effort.
 
Given this situation, may we please remove the SSR test case in Audio until the feature is fully supported?

this was root caused by audio dev and requesting to get that fixed in 2 weeks

bhargav0610:
Issue has been fixed no longer needed to track the current PR

PR #1567, opened 2026-02-14
Allow listing overlay-only DTBs in KERNEL_DEVICETREE

vkraleti:
On some platforms, a base DTB is not usable on its own and must be combined with one or more DT overlays. In such cases, a standalone FIT_DTB_COMPATIBLE entry for the base DTB may not exist, even though valid base+overlay combinations are defined.

Allow DTBs listed in KERNEL_DEVICETREE without a standalone FIT_DTB_COMPATIBLE entry, as long as they are used as part of a valid base+overlay combination.

### DIFF
--- a/conf/machine/include/fit-dtb-compatible.inc
+++ b/conf/machine/include/fit-dtb-compatible.inc
@@ -102,6 +102,6 @@ FIT_DTB_COMPATIBLE[sa8775p-ride-r3+sa8775p-ride-camx] = " \
 FIT_DTB_COMPATIBLE[sdm845-db845c] = "qcom,sdm845"
 FIT_DTB_COMPATIBLE[sm8450-hdk] = "qcom,sm8450-hdk"
 FIT_DTB_COMPATIBLE[sm8750-mtp] = "qcom,sm8750-mtp"
-FIT_DTB_COMPATIBLE[talos-evk] = "qcom,qcs615v1.1-iot"
+FIT_DTB_COMPATIBLE[talos-evk+talos-evk-camera-imx577] = "qcom,qcs615v1.1-iot"
 FIT_DTB_COMPATIBLE[talos-evk-camx] = "qcom,qcs615v1.1-iot-camx"
 FIT_DTB_COMPATIBLE[talos-evk+talos-evk-lvds-auo_g133han01] = "qcom,talos-evk-lvds-auo,g133han01"


classes-recipe/dtb-fit-image.bbclass
```diff
@@ -66,22 +65,19 @@ python do_generate_qcom_fitimage() {
         if not parts:
             continue
 
-        base_stem = parts[0]
-        ovl_stems = parts[1:]
-
```
sbanerjeeqc, line 71: IMO these two variables make the code easier to read.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -46,11 +46,10 @@ python do_generate_qcom_fitimage() {
     shutil.copy(qcom_meta_src, qcom_meta_dst)
     root_node.fitimage_emit_section_dtb("qcom-metadata.dtb", qcom_meta_dst, compatible_str=None, dtb_type="qcom_metadata")
 
-    # KERNEL_DEVICETREE contains both .dtb and .dtbo
```
sbanerjeeqc, line 49: retain this line

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -46,11 +46,10 @@ python do_generate_qcom_fitimage() {
...
-    files_set = {os.path.basename(x) for x in (d.getVar('KERNEL_DEVICETREE') or "").split()}
+    kernel_devicetree_set = {os.path.basename(x) for x in (d.getVar('KERNEL_DEVICETREE') or "").split()}
```
lumag: Why do you need to rename it?

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -91,8 +87,8 @@ python do_generate_qcom_fitimage() {
         if fname.endswith(".dtb"):
             dtb_key = os.path.splitext(dtb_id)[0]
             compatible = d.getVarFlag("FIT_DTB_COMPATIBLE", dtb_key) or ""
-            if not compatible:
-                bb.fatal(f"FIT_DTB_COMPATIBLE[{dtb_key}] is not set for base DTB '{fname}'.")
+            if not compatible and fname not in overlay_groups:
```
lumag: This condition is not correct: fname can have commas, while overlay_groups don't.

lumag (changes requested):
Which "some platforms"? Why is it "not usable"? Be precise in your commit messages.

classes-recipe/dtb-fit-image.bbclass
vkraleti: No they aren't.

classes-recipe/dtb-fit-image.bbclass
vkraleti: It states very obvious thing. Such comments won't give any additional value to users reading the code.

classes-recipe/dtb-fit-image.bbclass
vkraleti: There are two datasets used here. One from Kernel device tree and another from the fit compatible strings, just mentioning "files_set" is confusing. It doesn't infer what type of data this variable holds.

lumag: There are no files (anymore) in the FIT compatible strings array,

vkraleti: So can we safely assume 'files' in this script mean kernel devicetree?

classes-recipe/dtb-fit-image.bbclass
vkraleti: updated to use `dtb_id`.

conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -102,6 +102,6 @@ FIT_DTB_COMPATIBLE[sa8775p-ride-r3+sa8775p-ride-camx] = " \
 FIT_DTB_COMPATIBLE[sdm845-db845c] = "qcom,sdm845"
 FIT_DTB_COMPATIBLE[sm8450-hdk] = "qcom,sm8450-hdk"
 FIT_DTB_COMPATIBLE[sm8750-mtp] = "qcom,sm8750-mtp"
-FIT_DTB_COMPATIBLE[talos-evk] = "qcom,qcs615v1.1-iot"
+FIT_DTB_COMPATIBLE[talos-evk+talos-evk-camera-imx577] = "qcom,qcs615v1.1-iot"
```
lumag, line 105: You are not adding anything, you are replacing. Also, please step away from the "This allows" pattern. Inside the commit message describe the problem before describing anything else..

lumag:
@vkraleti I've restarted the jobs.

lumag (approved):

koenkooi (approved):

PR #1568, opened 2026-02-14
Enable Talos CAMX DTB compilation for linux-qcom kernel 

cjha01:
Include the downstream camera DTB for linux-qcom kernels so it appears in
the final image when using linux-qcom-next, linux-qcom-rt, and linux-qcom-next-rt
kernels on the IQ-615-EVK platform.


lumag (changes requested):
Which recipe are you updating? (None)
Please write the commit message as a text, not as a bullet list. Start with the description of the problem that you are solving.

lumag (approved):

lumag (approved):

koenkooi (approved):

PR #1570, opened 2026-02-16
fit-dtb-compatible: update compatible string for kodiak lite core kit

nkumarsi:
Update QCS5430 Core Kit compatible string to use subtype5.

| Platform / Board       | CDT Subtype |
|------------------------|--------------- |
| Vision Mezzanine      | 2                    |
| Core Kit                     | 5                    |
| Industrial Mezzanine | 9                    |

### DIFF
--- a/conf/machine/include/fit-dtb-compatible.inc
+++ b/conf/machine/include/fit-dtb-compatible.inc
@@ -41,8 +41,8 @@ FIT_DTB_COMPATIBLE[qcs615-ride] = " \
     qcom,qcs615v1.1-adp \
 "
 FIT_DTB_COMPATIBLE[qcs6490-rb3gen2] = " \
-    qcom,qcs5430-iot \
-    qcom,qcs6490-iot \
+    qcom,qcs5430-iot-subtype5 \
+    qcom,qcs6490-iot-subtype5 \
 "
 FIT_DTB_COMPATIBLE[qcs6490-rb3gen2+qcs6490-rb3gen2-industrial-mezzanine] = " \
     qcom,qcs5430-iot-subtype9 \


conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -41,7 +41,7 @@ FIT_DTB_COMPATIBLE[qcs615-ride] = " \
     qcom,qcs615v1.1-adp \
 "
 FIT_DTB_COMPATIBLE[qcs6490-rb3gen2] = " \
-    qcom,qcs5430-iot \
+    qcom,qcs5430-iot-subtype5 \
```
sbanerjeeqc, line 44: subtype5 for 6490?

conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -41,7 +41,7 @@ FIT_DTB_COMPATIBLE[qcs615-ride] = " \
...
     qcom,qcs6490-iot \
 "
```
sbanerjeeqc, line 46: Rather than adding here first consider if DTB+DTBO is required if yes best approach is to create a new FIT_DTB_COMPATIBLE[key] that covers both 6490 and 5430 subtype5

conf/machine/include/fit-dtb-compatible.inc
nkumarsi: acknowledged

conf/machine/include/fit-dtb-compatible.inc
sbanerjeeqc: Can you explain why ``subtype5`` should use qcs6490-rb3gen2 ?

sbanerjeeqc: Why ``subtype5`` cannot use ``qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine``

Isn't ``subtype5`` required to support ``qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine-camx``

nkumarsi: > Can you explain why `subtype5` should use qcs6490-rb3gen2 ?

By default qcs6490-rb3gen2 core kit is configured to use subtype 5.

KODIAK	IOT	5	RB3Gen2 Core Kit (Moselle attach)	1.0	Default

akdwived-dev: > Why `subtype5` cannot use `qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine`
> 
> Isn't `subtype5` required to support `qcs6490-rb3gen2+qcs6490-rb3gen2-vision-mezzanine-camx`

IS the question valid ? we have specific CDT for vision mezz 

vision Mezz --> CDT 2
IA Mezz ------> CDT 9 
Core kit -------> CDT 5 

core kit +vision mezz + camex overlay shall still boot with CDT9 not CDT 5

lumag: Why do we need the subtype 5? Why does it not work if we don't specify any subtype?

nkumarsi: It works even without specifying subtype 5, but i added subtype 5 to compatible string to ensure an exact CDT match.

akdwived-dev: > Why do we need the subtype 5? Why does it not work if we don't specify any subtype?

Actually we dont need to update subtype for core kit, it should boot without subtype 5 chcek, but questions were raised on why there is no compatible DT for subtype 5 so the PR was raised. BUt question is why core kit is not booting ? in test setup do we really have taken latest metadata?

ricardosalveti: It is booting with latest metadata.

vkraleti:
Closing this PR as all agreed not to have a subtype for Core-Kit.

PR #1571, opened 2026-02-16
Use the latest protobuf version

gkhose-qipl:
Use the latest protobuf version for CamX libs and sensors recipes.

PR #1580, opened 2026-02-16
Camera, sensors: switch to main protobuf version

lumag:
Upgrade sensors and camera binaries to the version which is linked against protobuf version from meta-oe.

This is an alternative to #1571, keeping the protobuf-camx name for the sake of simplifying handling the of binary vs meta-oe handling in future. When the situation repeats there would be no need to switch between protobuf and protobuf-camx anymore.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.11.bb
+++ b/dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.11.bb
@@ -8,15 +8,15 @@ LICENSE = "LICENSE.qcom-2"
 LIC_FILES_CHKSUM = "file://usr/share/doc/${BPN}/LICENSE.QCOM-2.txt;md5=165287851294f2fb8ac8cbc5e24b02b0 \
                     file://usr/share/doc/${BPN}/NOTICE;md5=04facc2e07e3d41171a931477be0c690"
 
-PBT_BUILD_DATE = "260209.1"
+PBT_BUILD_DATE = "260211"
 SRC_URI = " \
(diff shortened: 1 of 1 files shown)


lumag:
> The problem that led to the exchange between protobuf-camx and protobuf was the library soname versions. I don't see how the proposed solution solves the problem when it comes again without copy the downgraded recipe from meta-oe. Can you explain it please?

If you take a look at #1491, we had to vendor in the version from meta-oe _and_ change all camera / sensors recipes to use `protobuf-camx`. With the proposed approach in place, the only change would be to change the protobuf-camx recipe from the empty stub to the full-featured protobuf implementation and back.

quaresmajose:
> > The problem that led to the exchange between protobuf-camx and protobuf was the library soname versions. I don't see how the proposed solution solves the problem when it comes again without copy the downgraded recipe from meta-oe. Can you explain it please?
> 
> If you take a look at #1491, we had to vendor in the version from meta-oe _and_ change all camera / sensors recipes to use `protobuf-camx`. With the proposed approach in place, the only change would be to change the protobuf-camx recipe from the empty stub to the full-featured protobuf implementation and back.

I think this will not work because your current protobuf-camx is empty. Therefore, it will not provide the dependencies that are necessary for its users recipes.

It seems to me that what you want to do would be easier using a virtual name with PROVIDES/RPROVIDES and flip it by the the backport from meta-oe when it breaks next time:
```
cat <'_EOF_' >> dynamic-layers/openembedded-layer/recipes-support/protobuf/protobuf_%.bbappend
PNCAMX:qcom = "${PN}-camx"
PROVIDES:qcom += "${PNCAMX}-compiler ${PNCAMX}-lite"
RPROVIDES:${PN}:qcom += "${PNCAMX}-compiler ${PNCAMX}-lite"
_EOF_
```

quaresmajose:
Lets sopose another component/package will need the protobuf-compiler on the target and he needs to add it to his RDEPENDS. How should he do this? Should he use `RDEPENDS${PN} = "protobuf-compiler"` or `RDEPENDS${PN} = "protobuf-camx-compiler"`? or none of them will work?

koenkooi (approved):

PR #1594, opened 2026-02-18
weston-init: add automatic multi-DPU detection and configuration

yashgupta2126:
weston-init: add automatic multi-DPU detection and configuration

To provide multi-DPU support in Weston across hardware with varying configurations, add automatic detection and configuration of display devices: separate GPU + DPUs, GPU+DPU combos, DPU-only systems, and non-deterministic probe order.

On some platforms DRM card numbering is not stable: either DPU can become card1 or card2 depending on probe order. Weston does not pick the primary KMS device by cardN; it enumerates devices via sysfs/device-tree order and selects the first valid KMS device it finds. If a wrapper relies on /sys/class/drm/card* numeric ordering, it can accidentally pass the same card Weston already selected as primary in --additional-devices, which can lead to startup failures (duplicate open of the primary).

Example observed on LEMANS dual-DPU:

Scenario A: 22000000.display-subsystem (DPU) probed as card2
- Boot mapping:
  - 22000000.display-subsystem/.../drm/card2
  - ae00000.display-subsystem/.../drm/card1
- Weston log:
   - Checking device: .../22000000.../drm/card2
   - Selected .../drm/card2 (first valid KMS device found)
- Without platform/by-path ordering the wrapper can end up passing card2 as an additional device as well, resulting in:
      - ERROR: DRM device 'card2' is not a KMS device.

Scenario B: 22000000.display-subsystem (DPU) probed as card1
- Boot mapping:
  - 22000000.display-subsystem/.../drm/card1
  - ae00000.display-subsystem/.../drm/card2
- Weston log:
  - Checking device: .../22000000.../drm/card1
  - Selected .../drm/card1 (first valid KMS device found)
- This is the same failure mode if a wrapper assumes fixed card numbering: the primary (card1 in this case) may be incorrectly passed again via --additional-devices, causing a duplicate-open startup failure.

Add a wrapper script that detects DRM cards with KMS capability by checking for connector directories in sysfs and uses /dev/dri/by-path symlinks sorted by platform address (stable) to ensure deterministic ordering regardless of probe sequence. The first KMS-capable card is treated as primary; remaining KMS cards are passed via --additional-devices.
    
A systemd drop-in overrides ExecStart to call the wrapper, maintaining backward compatibility while enabling automatic multi-DPU support.

### DIFF
--- a/recipes-graphics/wayland/weston-init.bbappend
+++ b/recipes-graphics/wayland/weston-init.bbappend
@@ -1,1 +1,26 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
 DEFAULTBACKEND:qcom ?= "drm"
+
+SRC_URI:append:qcom = " \
+    file://additional-devices.conf \
+    file://weston-start.sh \
+"
+
+do_compile:append:qcom() {
+    sed -i -e 's:@bindir@:${bindir}:g' ${WORKDIR}/sources/additional-devices.conf
+    sed -i -e 's:@bindir@:${bindir}:g' ${WORKDIR}/sources/weston-start.sh
+}
+
+do_install:append:qcom() {
+    install -d ${D}${systemd_system_unitdir}/weston.service.d
+    install -m 0644 ${WORKDIR}/sources/additional-devices.conf \
+        ${D}${systemd_system_unitdir}/weston.service.d/additional-devices.conf
+
+    install -d ${D}${bindir}
+    install -m 0755 ${WORKDIR}/sources/weston-start.sh \
+        ${D}${bindir}/weston-start.sh
+}
+
+FILES:${PN} += "${systemd_system_unitdir}/weston.service.d/additional-devices.conf"
+FILES:${PN} += "${bindir}/weston-start.sh"
--- a/recipes-graphics/wayland/weston-init/additional-devices.conf
+++ b/recipes-graphics/wayland/weston-init/additional-devices.conf
@@ -1,0 +1,5 @@
+[Service]
+# Clear the original ExecStart from weston.service
+ExecStart=
+# Use weston script to dynamically detect and configure additional DRM devices
+ExecStart=@bindir@/weston-start.sh
--- a/recipes-graphics/wayland/weston-init/weston-start.sh
+++ b/recipes-graphics/wayland/weston-init/weston-start.sh
@@ -1,0 +1,73 @@
+#!/bin/sh
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+# SPDX-License-Identifier: MIT
+#
+# Weston startup script
+# Dynamically detects DRM cards and configures additional-devices parameter
+
+# Initialize options
+opts="--modules=systemd-notify.so"
+additional_cards=""
+first_kms_card=""
+kms_cards=""
+
+# Build a sorted list of KMS-capable cards by their platform path
+# This ensures consistent ordering regardless of probe order
+for path_link in /dev/dri/by-path/platform-*-card; do
+    # Skip if not a symlink
+    [ -L "$path_link" ] || continue
+
+    # Resolve symlink to actual device
+    card=$(readlink -f "$path_link")
+    [ -n "$card" ] || continue
+
+    card_name=$(basename "$card")
+    [ -n "$card_name" ] || continue
+
+    # Check if this card has KMS capability by looking for connectors
+    has_kms=0
+    for connector in /sys/class/drm/${card_name}-*; do
+        if [ -d "$connector" ]; then
+            has_kms=1
+            break
+        fi
+    done
+
+    # Only process KMS-capable cards
+    if [ "$has_kms" -eq 1 ]; then
+        # Extract platform address for sorting
+        platform_addr=$(echo "$path_link" | sed 's/.*platform-\([0-9a-f]*\).*/\1/')
+
+        # Verify we extracted a valid address
+        if [ -n "$platform_addr" ]; then
+            kms_cards="$kms_cards $platform_addr:$card_name"
+        fi
+    fi
+done
+
+# Sort by platform address to get consistent ordering
+sorted_cards=$(echo "$kms_cards" | tr ' ' '\n' | grep -v '^$' | sort | cut -d: -f2)
+
+# First card in sorted list is primary (Weston will auto-select it)
+# Remaining cards are added to additional-devices
+for card_name in $sorted_cards; do
+    if [ -z "$first_kms_card" ]; then
+        # Mark first card as primary (we skip adding it)
+        first_kms_card="$card_name"
+    else
+        # Add subsequent cards to additional-devices list
+        if [ -z "$additional_cards" ]; then
+            additional_cards="$card_name"
+        else
+            additional_cards="$additional_cards,$card_name"
+        fi
+    fi
+done
+
+# Add additional-devices parameter if we found any
+if [ -n "$additional_cards" ]; then
+    opts="$opts --additional-devices=$additional_cards"
+fi
+
+# Execute weston with the constructed options
+exec @bindir@/weston $opts


recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -1 +1,12 @@
 DEFAULTBACKEND:qcom ?= "drm"
+
+# update weston.service for DUAL DPU in lemans target
+update_weston_service_exec_start() {
+    if [ -e ${D}${systemd_system_unitdir}/weston.service ]; then
+        sed -i "s#^ExecStart=.*#ExecStart=/bin/sh -c 'opts=\"--modules=systemd-notify.so\"; i=0; while [ \$i -lt 10 ]; do if [ -e /dev/dri/card2 ]; then opts=\"\$opts --additional-devices=card2\"; break; fi; sleep 1; i=\$((i+1)); done; exec ${bindir}/weston \$opts'#" ${D}${systemd_system_unitdir}/weston.service
+    fi
+}
+
+do_install:append:qcs9100-ride-sx() {
```
koenkooi: That's a very specific override that doesn't match the commit message. I'd rather have a patch in SRC_URI that applies to all configurations than a single-machine override.
For the actual content of this change I'll defer to @lumag

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -1 +1,12 @@
 DEFAULTBACKEND:qcom ?= "drm"
+
+# update weston.service for DUAL DPU in lemans target
+update_weston_service_exec_start() {
+    if [ -e ${D}${systemd_system_unitdir}/weston.service ]; then
+        sed -i "s#^ExecStart=.*#ExecStart=/bin/sh -c 'opts=\"--modules=systemd-notify.so\"; i=0; while [ \$i -lt 10 ]; do if [ -e /dev/dri/card2 ]; then opts=\"\$opts --additional-devices=card2\"; break; fi; sleep 1; i=\$((i+1)); done; exec ${bindir}/weston \$opts'#" ${D}${systemd_system_unitdir}/weston.service
```
lumag: This is unreadable. Please install it as a systemd service snippet. Also, the DRI card numbers are not fixed, they can change basing on the order of probing.

yashgupta2126: Hi @lumag Can you review the latest patch for weston-init service

recipes-graphics/wayland/weston-init.bbappend
yashgupta2126: Applied weston service change for all machine

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -1 +1,27 @@
 DEFAULTBACKEND:qcom ?= "drm"
+
+FILESEXTRAPATHS:prepend := "${THISDIR}:"
+
+SRC_URI += "file://weston.service.d/additional-devices.conf \
+            file://weston-init/weston-start.sh \
+           "
+
+do_install:append() {
```
quaresmajose: You need to make this qcom specific with: `do_install:append:qcom`. Otherwise, this will break other places where the layer is being used.

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -1 +1,27 @@
 DEFAULTBACKEND:qcom ?= "drm"
+
+FILESEXTRAPATHS:prepend := "${THISDIR}:"
```
quaresmajose: This usually goes on the first line and please use

```
FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
```

and create a files folder and move all the files porvided in the patch over there

recipes-graphics/wayland/weston-init/weston-start.sh
```diff
@@ -0,0 +1,59 @@
...
+# Initialize options
+opts="--modules=systemd-notify.so"
+additional_cards=""
+first_kms_card=""
+
+# Build a sorted list of KMS-capable cards by their platform path
+# This ensures consistent ordering regardless of probe order
+kms_cards=""
+
+for path_link in /dev/dri/by-path/platform-*-card; do
```
lumag, line 16: Could you please simplify this by iterating over `/sys/class/drm/card*`  without extra intermediate name transformations.

recipes-graphics/wayland/weston-init.bbappend
yashgupta2126: uploaded next patch with correction

recipes-graphics/wayland/weston-init.bbappend
yashgupta2126: uploaded next patch with correction

recipes-graphics/wayland/weston-init/weston-start.sh
yashgupta2126: Hi @lumag, can you review based on my findings with dual dpu testing for lemans ?

Problem
- DRM card numbers are not stable. Either DPU can become card1 or card2 depending on probe order.

What Weston uses (“sorting”)
- Weston does NOT select the primary KMS device by cardN order.
- Weston effectively follows sysfs/device-tree enumeration order (platform paths under /sys/devices/..., 
  typically lexicographic order of those path strings) and picks the “first valid KMS device found”.

Example we hit  error
- Boot A mapping:
  - 22000000.display-subsystem/.../drm/card2 (DPU at 2200...)
  - ae00000.display-subsystem/.../drm/card1  (DPU at ae0...)

- Weston log:
  - Checking device: .../22000000.../drm/card2
  - Selected .../drm/card2 (first valid KMS device found)

- “Simple” approach (no platform sorting) ended up passing skipped card1 as per script:
  - --additional-devices=card2
- Result:
  - using /dev/dri/card2 (primary)
  - adding secondary device /dev/dri/card2
  - ERROR: DRM device 'card2' is not a KMS device.

Why platform/by-path sorting fixed it
- /dev/dri/by-path/platform-22001000...-card -> card2
- /dev/dri/by-path/platform-ae01000...-card   -> card1
- Sorting by the platform address string gives a stable order independent of cardN.
- The script can reliably skip the first platform device (primary) and only pass the other as --additional-devices, 
  avoiding duplicate-open failures.

lumag: This needs to be explained in the commit message

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -1 +1,27 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
```
lumag: Use weston-init subdir, then you won't need this .

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -1 +1,27 @@
...
+SRC_URI += "file://additional-devices.conf \
+            file://weston-start.sh \
+           "
+
+do_install:append:qcom() {
+    if ${@bb.utils.contains('DISTRO_FEATURES','systemd','true','false',d)}; then
+        # Install systemd drop-in file
+        install -d ${D}${systemd_system_unitdir}/weston.service.d
+        install -m 0644 ${UNPACKDIR}/additional-devices.conf ${D}${systemd_system_unitdir}/weston.service.d/additional-devices.conf
+        sed -i -e s:@sysconfdir@:${sysconfdir}:g \
```
lumag: sed should be a part of do_compile.

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -1 +1,27 @@
...
+
 DEFAULTBACKEND:qcom ?= "drm"
+
+SRC_URI += "file://additional-devices.conf \
+            file://weston-start.sh \
+           "
+
+do_install:append:qcom() {
+    if ${@bb.utils.contains('DISTRO_FEATURES','systemd','true','false',d)}; then
+        # Install systemd drop-in file
```
lumag: Obvious comment, please drop.

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -1 +1,27 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
 DEFAULTBACKEND:qcom ?= "drm"
+
+SRC_URI += "file://additional-devices.conf \
```
lumag: Let each value have us own separate line. Also this will fail yocto-check-layers.

lumag (changes requested):

recipes-graphics/wayland/files/additional-devices.conf
```diff
@@ -0,0 +1,5 @@
+[Service]
+# Clear the original ExecStart from weston.service
+ExecStart=
+# Use weston script to dynamically detect and configure additional DRM devices
+ExecStart=@sysconfdir@/xdg/weston/weston-start.sh
```
ricardosalveti: Why `@sysconfdir@/xdg/weston`? It is just a normal script in the end, and this folder is used for configuration files.

recipes-graphics/wayland/weston-init/weston-start.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/sh
+# Weston startup script
+# Dynamically detects DRM cards and configures additional-devices parameter
```
ricardosalveti, line 6: Missing license + copyright.

recipes-graphics/wayland/weston-init/weston-start.sh
yashgupta2126: upload in next patchset with updated commit message

recipes-graphics/wayland/weston-init.bbappend
yashgupta2126: added weston-init as a subfolder but FILESEXTRAPATHS keeps this layer files fetch to workdir, otherwise failing in compilation as no such file found in path

recipes-graphics/wayland/weston-init.bbappend
yashgupta2126: Renoved in next patchset

recipes-graphics/wayland/weston-init.bbappend
yashgupta2126: made sed is part of do_compile in next patchset

recipes-graphics/wayland/weston-init.bbappend
yashgupta2126: Updated resolve in new patchset

recipes-graphics/wayland/files/additional-devices.conf
yashgupta2126: Removed from conf folder in new patchset

recipes-graphics/wayland/weston-init/weston-start.sh
yashgupta2126: added licesence in new patchset

lumag (approved):

recipes-graphics/wayland/weston-init/weston-start.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/sh
+# Copyright (c) 2025 Qualcomm Innovation Center, Inc. All rights reserved.
```
vkraleti: Please use QTI year-less copyright here.

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -1 +1,26 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
 DEFAULTBACKEND:qcom ?= "drm"
+
+SRC_URI += "\
```
ricardosalveti: Should be SRC_URI:append:qcom

recipes-graphics/wayland/weston-init/additional-devices.conf
```diff
@@ -0,0 +1,5 @@
+[Service]
+# Clear the original ExecStart from weston.service
+ExecStart=
+# Use weston script to dynamically detect and configure additional DRM devices
+ExecStart=@sysconfdir@/default/weston-start.sh
```
ricardosalveti: Why sysconfdir here? Can't we just install at /usr/bin or similar?

recipes-graphics/wayland/weston-init/weston-start.sh
yashgupta2126: Updated the QTI year-less copyright in latest patchset

recipes-graphics/wayland/weston-init.bbappend
yashgupta2126: updated the SRC_URI with append:qcom in latest patchset

recipes-graphics/wayland/weston-init/additional-devices.conf
yashgupta2126: Updated /usr/bin to place this weston script in latest patchset

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -1 +1,27 @@
...
+
 DEFAULTBACKEND:qcom ?= "drm"
+
+SRC_URI:append:qcom = " \
+    file://additional-devices.conf \
+    file://weston-start.sh \
+"
+
+do_compile:append:qcom() {
+    sed -i -e 's:@sysconfdir@:${sysconfdir}:g' ${WORKDIR}/sources/additional-devices.conf
```
anujm1: This shouldn't be needed now?

yashgupta2126: Updated with latest patchset

lumag (approved):

ricardosalveti (approved):

PR #1601, opened 2026-02-19
ci: actions/compile: disable SDK input control

quaresmajose:
Making the validation of the SDK build dependent on external conditions
is prone to fail and allows for the easy introduction of regressions.
Thus, this will be validated in all builds as all the others.

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -12,9 +12,6 @@ inputs:
     required: true
   cache_dir:
     required: true
-  sdk:
-    required: false
-    default: "0"
 outputs:
   url:
     description: Location of the published binaries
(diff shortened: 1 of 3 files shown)


ricardosalveti:
Converting to draft as it is bringing changes from https://github.com/qualcomm-linux/meta-qcom/pull/1603.

lumag (commented):
Should we skip populating SDK for non-nightly builds?

quaresmajose:
Yes, we talked about this somewhere. But I'd prefer to do this without adding new imput entries to the action. If it were possible to know that we were running the nightly in another way, for example using/reading an environment variable. I would prefer to investigate how to do this and I propose in a new PR.

lumag (approved):

vkraleti (approved):

ricardosalveti (approved):

PR #1603, opened 2026-02-19
Add monaco-evk-camera-imx577 overlay for linux-yocto

quaresmajose:
The monaco-evk-camera-imx577 dtbo is being reviewed upstream and has not yet been integrated.
Having Camera support on Monaco EVK is mandatory product ask.
So the solution carry the patch for linux-yocto/linux-yocto-dev till it lands in upstream

### DIFF
--- a/conf/machine/qcom-armv8a.conf
+++ b/conf/machine/qcom-armv8a.conf
@@ -32,6 +32,7 @@ KERNEL_DEVICETREE ?= " \
     qcom/lemans-evk.dtb \
     qcom/lemans-evk-camera-csi1-imx577.dtbo \
     qcom/monaco-evk.dtb \
+    qcom/monaco-evk-camera-imx577.dtbo \
     qcom/qcm6490-idp.dtb \
     qcom/qcs404-evb-4000.dtb \
     qcom/qcs615-ride.dtb \


recipes-kernel/linux/linux-yocto-6.18/workarounds/arm64_dts_qcom_monaco-evk-camera_Add_DT_overlay.patch
```diff
@@ -0,0 +1,229 @@
...
+X-Proofpoint-GUID: iKOI9c9VVzYYgGYcAuwz3pyGEcehlOXH
+X-Proofpoint-ORIG-GUID: iKOI9c9VVzYYgGYcAuwz3pyGEcehlOXH
+X-Proofpoint-Virus-Version: vendor=baseguard
+ engine=ICAP:2.0.293,Aquarius:18.0.1121,Hydra:6.1.51,FMLib:17.12.100.49
+ definitions=2026-02-13_02,2026-02-12_03,2025-10-01_01
+X-Proofpoint-Spam-Details: rule=outbound_notspam policy=outbound score=0
+ bulkscore=0 impostorscore=0 spamscore=0 suspectscore=0 priorityscore=1501
+ lowpriorityscore=0 adultscore=0 phishscore=0 malwarescore=0 clxscore=1015
+ classifier=typeunknown authscore=0 authtc= authcc= route=outbound adjust=0
+ reason=mlx scancount=1 engine=8.22.0-2601150000 definitions=main-2602130103
```
ricardosalveti: Please generate a patch without all these headers.

recipes-kernel/linux/linux-yocto-dev.bbappend
```diff
@@ -8,6 +8,7 @@ SRC_URI:append:qcom = " \
     file://qcm6490-board-dts/0001-PENDING-enable-xHCI.patch \
     file://workarounds/0001-QCLINUX-arm64-dts-qcom-qcm6490-disable-sdhc1-for-ufs.patch \
     file://workarounds/0001-PENDING-arm64-dts-qcom-Remove-voltage-vote-support-f.patch \
+    file://workarounds/arm64_dts_qcom_monaco-evk-camera_Add_DT_overlay.patch \
```
ricardosalveti: We should have this under monaco-evk-dts instead, as this is not really a workaround.

ricardosalveti:
The commit should also cover the reason why backporting this patch is required, which is basically because of our fit-dtb-compatible.inc implementation.

recipes-kernel/linux/linux-yocto-dev.bbappend
quaresmajose: moved to monaco-evk-dts

recipes-kernel/linux/linux-yocto-6.18/workarounds/arm64_dts_qcom_monaco-evk-camera_Add_DT_overlay.patch
quaresmajose: fixed

ricardosalveti:
The reason for the backport needs to be improved, as it is not really due a product ask but due our default configuration in conf/machine/include/fit-dtb-compatible.inc.

Why are forcing a dtbo combination by default at https://github.com/qualcomm-linux/meta-qcom/pull/1557/commits/2501211a0046af8d33bb3e1fa4867c6ef5b5bfdf? If this should indeed be the default, why is it not part of the main dtb instead?

IMHO the default configuration should be aligned with the default kernel dtb available, and overlays should be enabled after identifying at runtime (by the boot firmware directly or via efi var) or if explicitly asked by the user.

PR #1606, opened 2026-02-20
linux-qcom: enable SELinux kernel configs on LTS kernel

sasikuma-qti:


dynamic-layers/selinux/recipes-kernel/linux/linux-qcom_%.bbappend
```diff
@@ -0,0 +1 @@
+require recipes-kernel/linux/linux-yocto_selinux.inc
```
quaresmajose, line 1: Have the [changes](https://git.yoctoproject.org/meta-selinux/tree/recipes-kernel/linux/files/selinux.cfg) been applied? I believe not.

Please note that this chnages are for linux-yocto and the oe-core kernel recipe understand the *.cfg file but the linux-qcom not. On the linux-qcom this should be done explicity with [KBUILD_CONFIG_EXTRA](https://github.com/qualcomm-linux/meta-qcom/blob/daf8c98ae0042b53140ea7f335cf8516eecf1fce/recipes-kernel/linux/linux-qcom_6.18.bb#L42).

ricardosalveti: It actually is applied, because of `${@" ".join(find_cfgs(d))}`.

quaresmajose: Rigth, I didn't notice that part. Thanks

lumag (approved):

PR #1610, opened 2026-02-20
ci: earlycon enablement through command line for early bootup logs

sampnimm2629:
Enable earlycon to get the early console logs to debug early bootup crashes in CI testing.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -51,7 +51,7 @@ local_conf_header:
     https://.*/.*/ https://artifacts.codelinaro.org/codelinaro-le/ \
     "
   cmdline: |
-    KERNEL_CMDLINE_EXTRA:append = " qcom_scm.download_mode=1"
+    KERNEL_CMDLINE_EXTRA:append = " earlycon qcom_geni_serial.con_enabled=1  qcom_scm.download_mode=1"
   qcomflash: |
     IMAGE_CLASSES += "image_types_qcom"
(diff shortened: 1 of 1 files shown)


ci/base.yml
```diff
@@ -51,7 +51,7 @@ local_conf_header:
     https://.*/.*/ https://artifacts.codelinaro.org/codelinaro-le/ \
     "
   cmdline: |
-    KERNEL_CMDLINE_EXTRA:append = " qcom_scm.download_mode=1"
+    KERNEL_CMDLINE_EXTRA:append = " earlycon qcom_geni_serial.con_enabled=1  qcom_scm.download_mode=1"
```
ndechesne, line 54: this is likely not the right place to enable it. this would apply to all builds, all boards, every user. 
If that's the case, then it should be explained why in the commit message. 
what is the need here?

lumag:
Please point out failed jobs which require this to be enabled.

ci/base.yml
sampnimm2629: > this is likely not the right place to enable it. this would apply to all builds, all boards, every user. If that's the case, then it should be explained why in the commit message. what is the need here?

We need this to enable for all targets there in Lava CI, which will help us to get early console logs to debug early bootup crashes.

lumag: Please point out the jobs that failed early enough to be debuggable only with the help of earlycon.

PR #1612, opened 2026-02-20
Workflow to monitor GitHub_Token usage

steve345:
Workflow to monitor GitHub_Token usage. This will provide details at each run for the current API Limit usage. Runs every 30 minutes.


ricardosalveti:
Squash your changes, fix your commit message (see pattern in git log) and correct your user name, email and sign-off-by line (it seems you changed using the github interface, which is not recommended here).

PR #1616, opened 2026-02-23
iq-9075-evk: build mezzanine DTs as an overlays

nkumarsi:
Switch the mezzanine DT from a full DTB to a DTBO so it is built and packed as an overlay in multiDTB instead of a standalone base DTB. This aligns the lemans evk mezzanine hardware configuration with the intended hardware layering and avoids generating full DTB for mezzanine board.

### DIFF
--- a/conf/machine/include/fit-dtb-compatible.inc
+++ b/conf/machine/include/fit-dtb-compatible.inc
@@ -27,6 +27,10 @@ FIT_DTB_COMPATIBLE[lemans-evk+lemans-evk-camx+lemans-el2] = " \
     qcom,qcs9075-iot-camx-el2kvm \
     qcom,qcs9075v2-iot-camx-el2kvm \
 "
+FIT_DTB_COMPATIBLE[lemans-evk+lemans-evk-mezzanine] = " \
+    qcom,qcs9075-iot-subtype1 \
+    qcom,qcs9075v2-iot-subtype1 \
+"
 FIT_DTB_COMPATIBLE[monaco-evk+monaco-evk-camera-imx577] = "qcom,qcs8275-iot"
 FIT_DTB_COMPATIBLE[monaco-evk+monaco-evk-camx] = "qcom,qcs8275-iot-camx"
 FIT_DTB_COMPATIBLE[qcm6490-idp] = "qcom,qcm6490-idp"
--- a/conf/machine/iq-9075-evk.conf
+++ b/conf/machine/iq-9075-evk.conf
@@ -9,6 +9,7 @@ MACHINE_FEATURES += "efi pci tpm2"
 KERNEL_DEVICETREE ?= " \
                       qcom/lemans-evk.dtb \
                       qcom/lemans-evk-camera-csi1-imx577.dtbo \
+                      qcom/lemans-evk-mezzanine.dtbo \
                       "
 
 # These DTs are not upstreamed and currently exist only in linux-qcom kernels


conf/machine/iq-9075-evk.conf
```diff
@@ -9,6 +9,7 @@ MACHINE_FEATURES += "efi pci tpm2"
 KERNEL_DEVICETREE ?= " \
                       qcom/lemans-evk.dtb \
                       qcom/lemans-evk-camera-csi1-imx577.dtbo \
+                      qcom/lemans-evk-mezzanine.dtbo \
```
lumag, line 12: You are not switching anything, you are adding mezzanine. BTW, which mezzanine is it?

nkumarsi: yes, it is interface mezzanine which was recently posted upstream
lore link - https://lore.kernel.org/lkml/20260203193848.123307-1-umang.chheda@oss.qualcomm.com/

lumag: And it got the feedback that it can't be called simply mezzanine. It should be pointed out that that's an interface one.

conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -27,6 +27,10 @@ FIT_DTB_COMPATIBLE[lemans-evk+lemans-evk-camx+lemans-el2] = " \
     qcom,qcs9075-iot-camx-el2kvm \
     qcom,qcs9075v2-iot-camx-el2kvm \
 "
+FIT_DTB_COMPATIBLE[lemans-evk+lemans-evk-mezzanine] = " \
+    qcom,qcs9075-iot-subtype1 \
+    qcom,qcs9075v2-iot-subtype1 \
+"
```
ricardosalveti, line 33: Break this into a separated commit.

conf/machine/iq-9075-evk.conf
ricardosalveti: Also, only extend KERNEL_DEVICETREE if it is indeed available in all kernel recipes we use (linux-yocto, linux-qcom and linux-qcom-next).

umang-chheda: > And it got the feedback that it can't be called simply mezzanine. It should be pointed out that that's an interface one.

Hello @lumag ,
Currently in qcom-next & qcom 6.18 kernel -- v2 of the patch is merged as it had received R-B tag - the feedback for changing the name was received later - only after the patch was merged in qcom-next branch.
for now, can we go ahead merging with this name - and change the name later once the patch is accepted and comes into linux-next ?

Does it work ?

lumag: And then you'd need to change this again, once the patch in those trees is changed. I'd rather change the name now (also please adjust commits and PR description to mention which mezzanine you are working with).

umang-chheda: > And then you'd need to change this again, once the patch in those trees is changed. I'd rather change the name now

Hmm Yes I agree with this redundant efforts - But since currently for the upcoming RC1 release - the code freeze is done for the qcom-next and qcom-6.18y trees - we had to go through this route - as the patch getting accepted in upstream might take some more time.

If it's okay for this instance - can we merge with this name for the release and later change the names once the patches are merged in those trees ?

nkumarsi: > Also, only extend KERNEL_DEVICETREE if it is indeed available in all kernel recipes we use (linux-yocto, linux-qcom and linux-qcom-next).

Understood. KERNEL_DEVICETREE is available in all three kernel recipes.

conf/machine/include/fit-dtb-compatible.inc
nkumarsi: > Break this into a separated commit.

Acknowledged. I will move this into a separate commit.

conf/machine/iq-9075-evk.conf
lumag: I already wrote, no, it's not okay.

nkumarsi:
I am closing this PR. Once the board name change is approved on upstream, I will raise it again.

PR #1620, opened 2026-02-24
Add `quiet` option to kernel cmdline

vkraleti:
Add the `quiet` option to the kernel command line to reduce verbose kernel output during boot. This suppresses driver/module initialization messages, filesystem checks, and other non-critical logs, helping improve overall boot time.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -51,7 +51,7 @@ local_conf_header:
     https://.*/.*/ https://artifacts.codelinaro.org/codelinaro-le/ \
     "
   cmdline: |
-    KERNEL_CMDLINE_EXTRA:append = " qcom_scm.download_mode=1"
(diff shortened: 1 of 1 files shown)


ndechesne:
> Hmm, I thought that `quiet` will be added to a "performance" image, not to all images. In this form it breaks all testing.

agreed. it's in ci/performance.yml.

vkraleti:
> > Hmm, I thought that `quiet` will be added to a "performance" image, not to all images. In this form it breaks all testing.
> 
> agreed. it's in ci/performance.yml.

`quiet` is already available in ci/performance.yml  https://github.com/qualcomm-linux/meta-qcom/blob/master/ci/performance.yml

Current ask from @lumag is to make this option as default for all builds.

PR #1624, opened 2026-02-24
ci: actions/sync: added support for merging multiple prs during repo checkout

shivshah3400:
This github action will be used to sync multiple prs while repo checkout. This can be extensively used to build a set of PRs together saving resources and time.

### DIFF
--- a/.github/actions/sync/action.yml
+++ b/.github/actions/sync/action.yml
@@ -1,0 +1,66 @@
+name: 'Sync Repo [optionally with PRs]'
+description: 'Fetches and merges a list of open PRs into the current workspace.'
+
+inputs:
+  prs:
+    description: 'A space-separated list of open PR numbers to sync.'
+    required: false
+    default: ''
+  path:
+    description: 'The path on the runner where the repository is checked out.'
+    required: true
+  repo:
+    description: 'The full name of the repository (e.g., owner/repo).'
+    required: true
+  ref:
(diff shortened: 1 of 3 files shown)


shivshah3400:
@ndechesne This requirement was discussed with ricardo and dimitry last month. This feature is required for qli2.0 branch where we need to manually test PRs in batches. This process will also help optimize the CI/CD process for 2.0 by reducing resource consumption and improving efficiency. 
The change is breaking the builds due to introduction and usage of a new sync action which is yet to be merged in the repo. 
I will update the commit message with the exact requirement from 2.0 perspective. The idea was that this action will be generic and can be used for mainline as well.
 
 
@lumag Based on our discussion last month, it was decided that our team would be raising the PR and we would have follow up discussions on when to merge the PR. Sharing Minutes of Meeting from our last meeting:
 
> PR Batching Support
- meta-qcom maintainers raised concerns about enabling batching support citing that the feature is not necessary at the moment and should be enabled only when it is required. 
- LINT team will raise the feature proposal for batching of PRs in meta-qcom
- Follow-up discussion will be scheduled to finalize if and when the PR should be merged.
 
Also the image teams are against using test branches for batching PRs. This would require us to maintain a large number of branches and also manual effort in merging the PRs onto test branch and building. They want a simple approach which can take care of this seamlessly. We can schedule a call for this if required.

PR #1633, opened 2026-02-25
Upgrade boot firmware to r2.0_00001.1 release

vkraleti:
Update QCS615, QCS6490, QCS8300, QCS9100 boot firmware recipes to consume 
boot binaries from the r2.0_00001.1 release. Starting with this release, the firmware
artifacts are hosted using the new SPF 2.0 layout. Update the paths accordingly.

### DIFF
--- a/recipes-bsp/firmware-boot/firmware-qcom-boot-qcs6490_r2.0-00001.1-00004.bb
+++ b/recipes-bsp/firmware-boot/firmware-qcom-boot-qcs6490_r2.0-00001.1-00004.bb
@@ -1,0 +1,3 @@
+require firmware-qcom-boot-qcs6490.inc
+
+SRC_URI[bootbinaries.sha256sum] = "48d50c22d1b1e09ea2af1ada2adddee82bcf16a494fce656e8e3f5a694cef830"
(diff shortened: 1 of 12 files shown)


vkraleti:
Kept the PR in draft state as I am awaiting the release notes.

recipes-bsp/firmware-boot/firmware-qcom-boot-qcs6490_r2.0-00001.1-00004.bb
quaresmajose, line 1: It wold be better to remove the `r` in the PV and keep only numbers. Having letters easily breaks the order and can cause the version to backwards. consequently adjust the SRC_URI.

lumag (changes requested):
The PR doesn't describe, why are we switching versions. Is there a reason why we can't continue using current versioning scheme?

vkraleti:
> The PR doesn't describe, why are we switching versions. Is there a reason why we can't continue using current versioning scheme?

The current 1.0 SPF release line is going to be in maintenance mode post GA 1.8 release scheduled E/March. All new feature development will happen only in 2.0 SPF. I'll mention the same in commit messages.

recipes-bsp/firmware-boot/firmware-qcom-boot-qcs6490_r2.0-00001.1-00004.bb
ricardosalveti: Yeah, I'm having a hard time understanding this versioning scheme, can't we have like 2.0.0001.1 or similar?

ricardosalveti:
> Now, it all started at 0.something. Just continue that line, ending up with the point releases as a fixes for the stable branches with new releases incrementing minor revision as usual.

The main problem is that the SPF is part of the version, and as major instead of a minor. Ideally the component should have its own version, and we could have a stable release that is then consumed by a certain SPF.

The current scheme causes a mess to manage, because mainline can be newer but the 1.x / 2.x baselines will have a higher version.

PR #1636, opened 2026-02-25
Disable PAuth to allow all CPUs to boot with KVM on QCS8300

vkraleti:
On QCS8300 platforms, ARM64 Pointer Authentication (PAuth) is supported only on the boot CPUs (0–3) and not on the secondary CPUs (4–7). The ARM64 CPU feature discovery logic expects secondary CPUs to support all features enabled on the boot CPU. When this mismatch is detected, the secondary CPUs are prevented from booting, which is undesirable for production systems.

To ensure that all CPUs can boot correctly, disable PAuth by appending `arm64.nopauth` to the kernel cmdline.

This issue is observed only when Linux is booted in EL2 (i.e. KVM in use). When Linux runs under Gunyah (Type‑1 hypervisor), the PAuth feature is hidden from EL1 via HCR_EL2.TID3.

### DIFF
--- a/conf/machine/include/qcom-qcs8300.inc
+++ b/conf/machine/include/qcom-qcs8300.inc
@@ -7,6 +7,10 @@ require conf/machine/include/qcom-common.inc
 DEFAULTTUNE = "armv8-2a-crypto"
 require conf/machine/include/arm/arch-armv8-2a.inc
 
+# Disable PAuth via cmdline to allow all cores to boot with KVM.
+# More details at https://lore.kernel.org/all/3fcf6614-ee83-4a06-9024-83573b2e642e@quicinc.com/
+KERNEL_CMDLINE_EXTRA:append = " arm64.nopauth"
+
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-qcom-boot-essential \
     packagegroup-machine-essential-qcom-qcs8300-soc \


conf/machine/include/qcom-qcs8300.inc
```diff
@@ -7,6 +7,10 @@ require conf/machine/include/qcom-common.inc
 DEFAULTTUNE = "armv8-2a-crypto"
 require conf/machine/include/arm/arch-armv8-2a.inc
 
+# Disable PAuth via cmdline to allow all cores to boot with KVM.
+# More details at https://lore.kernel.org/all/3fcf6614-ee83-4a06-9024-83573b2e642e@quicinc.com/
+KERNEL_CMDLINE_EXTRA:append = " arm64.nopauth"
```
quaresmajose, line 12: Should this be conditional on `kvm` being in DISTRO_FEATURES?

vkraleti: Ideally yes, but I am not very sure if adding a DISTRO_FEATURE check in machine conf is right thing. 

As passing this option while booting with Gunyah has no impact and goal is to switch b/w KVM & Gunyah by switching an efivar, felt it is fine to append this unconditionally.

lumag: No need to. Let's not add extra difference between builds

PR #1639, opened 2026-02-25
CI: test: download build URLs from the correct GitHub run

lumag:
Before the commit 06b45e2959cf ("CI: test: stop downloading artifacts one more time") all build URLs were attempted to be downloaded twice, once in test.yml and once in the list-jobs/action.yml.

The commit dropped artifact downloading from test.yml, deferring all downloads to the list-jobs action, however the workflow passes the calling build id to the action rather than the current one, ending up with the list-jobs not finding any build URLs (because it tries to download them from the build run rather than the test run).

The issue didn't manifest in the 'Build on Push' workflows because the the build run and the test run are the same.

Fixes: 06b45e2959cf ("CI: test: stop downloading artifacts one more time")

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -68,7 +68,7 @@ jobs:
         uses: ./.github/actions/list-jobs
         with:
           prefix: "boottest"
-          build_id: ${{ inputs.build_id }}
+          build_id: ${{ github.run_id }}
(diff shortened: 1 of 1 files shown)


ricardosalveti:
Will merge as an attempt to get this behavior fixed, as all our current prs are on hold due this issue.

PR #1641, opened 2026-02-25
ci/base: Update kernel cmdline to enable ice wrapped keys

kuld-sing:
ice driver support both wrapped and standard keys and choice is on user to select key mechanism.

Wrapped key selection is given as userspace option and enabled on need basis. Enable use_wrapped_keys flag will still suffice both wrapped and standard keys hence, enable the flag via cmdline for meta-qcom users.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -51,7 +51,7 @@ local_conf_header:
     https://.*/.*/ https://artifacts.codelinaro.org/codelinaro-le/ \
     "
   cmdline: |
-    KERNEL_CMDLINE_EXTRA:append = " qcom_scm.download_mode=1"
+    KERNEL_CMDLINE_EXTRA:append = " qcom_scm.download_mode=1 qcom_ice.use_wrapped_keys=1"
   qcomflash: |
     IMAGE_CLASSES += "image_types_qcom"
     IMAGE_FSTYPES += "qcomflash"


ricardosalveti:
NAK for custom kernel command line options unless strictly required, with a strong justification and being per board, not generically added like this.

Why is this required?

arun-n-qc:
The Storage Inline Crypto Engine (ICE) is used to support the User data encryption. ICE supports both standard keys and hardware‑wrapped keys. To enable and use the hardware‑wrapped key feature, the ICE driver relies on a module parameter: https://elixir.bootlin.com/linux/v6.19.3/source/drivers/soc/qcom/ice.c#L102. 

Based on my understanding, the only way to enable this feature is by adding a kernel command‑line parameter. Could you please confirm if there are any alternative methods to enable this feature?

ricardosalveti:
> Based on my understanding, the only way to enable this feature is by adding a kernel command‑line parameter. Could you please confirm if there are any alternative methods to enable this feature?

Can't the driver itself check when it is available and enable it by default?

lumag:
> > Based on my understanding, the only way to enable this feature is by adding a kernel command‑line parameter. Could you please confirm if there are any alternative methods to enable this feature?
> 
> Can't the driver itself check when it is available and enable it by default?

It is a distro-level policy. Enabling hwkm means that non-hardware-wrapped keys become unavailable.

ricardosalveti:
> It is a distro-level policy. Enabling hwkm means that non-hardware-wrapped keys become unavailable.

But how is this going to be managed by distros? It is easy for meta-qcom-distro, but not for generic ones.

lumag:
I don't like it being enforced on a distro level. Please document how to enable it for the user builds (or when flashing the binary build).

sonineerajkumar:
The reasoning behind adding a control for module parameter is that wrapped keys cannot be used unconditionally as legacy mode and standard mode in ICE are mutually exclusive. Same reasoning is captured in ICE driver as well:

* Also, since HWKM and legacy mode are mutually exclusive, and
* ICE-capable storage driver(s) need to know early on whether to
* advertise support for raw keys or wrapped keys, HWKM cannot be used
* unconditionally.  A module parameter is used to opt into using it.

lumag:
You are forcing the decision on all users. Instead let users decide if they want HWKM or normal kerys on their own.

PR #1643, opened 2026-02-26
workflows: update lava-test-plans to 5e2441f

qcom-anilyada:
Changes:
- pre-merge: add list of expected tests
- Allow target_deploy_timeout to be overridden and increase it for db410c
- Reduce timeout values for faster test feedback
- Add fastrpc_test to pre-merge-basic suite
- Add display-gfx test plan
- Refactor BT suite to use loop-based test definitions

### DIFF
--- a/.github/actions/lava-test-plans/action.yml
+++ b/.github/actions/lava-test-plans/action.yml
@@ -27,7 +27,7 @@ runs:
         with:
           repository: qualcomm-linux/lava-test-plans
           path: lava-test-plans
-          ref: 1ab5e2f1d6cc3559ca4685941cc9fd17ab132c2d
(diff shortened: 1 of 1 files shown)


bhargav0610:
> There is no dependency, but I'd like to avoid changing 2 things at the same time.

as the PR is merged can we also merged this PR now

qcom-anilyada:
> We also are struggling with device availability, and I would appreciate if you helped in this area.

the device availability is better now.

> The jobs using this commit are expecting non-existing device tags and will fail.

the only tags being used in the changes are pvt and display, pvt has been updated to has-bt, shall I move the ref to the tip of lava-test-plans? Can we plan to merge this as this PR has important changes regarding optimization and adding testcases.

PR #1648, opened 2026-02-26
WIP: workflows: limit the testjobs count list-files action

mwasilew:
There was a typo in list-files action: ${{ input.prefix }}. Should be ${{ inputs.prefix }}. This causes unrelated files to be counted towards the list of test jobs to be executed. This patch fixes the typo and uses the prefix to filter the files that will be send to LAVA for execution.

### DIFF
--- a/.github/actions/list-jobs/action.yml
+++ b/.github/actions/list-jobs/action.yml
@@ -34,7 +34,7 @@ runs:
       id: listjobs
       shell: bash
       run: |
-        JOBFILES=$(find . -name "${{ input.prefix }}*.yaml")
+        JOBFILES=$(find . -name "*.yaml" | grep ${{ inputs.prefix }})
         RESULT_JSON=$(jq -n '{target: []}')
         for J in $JOBFILES
           do


mwasilew:
Pushed to `next`: https://github.com/qualcomm-linux/meta-qcom/actions/runs/22440383854

.github/actions/list-jobs/action.yml
```diff
@@ -34,7 +34,7 @@ runs:
       id: listjobs
       shell: bash
       run: |
-        JOBFILES=$(find . -name "${{ input.prefix }}*.yaml")
+        JOBFILES=$(find . -name "*.yaml" | grep ${{ inputs.prefix }})
```
lumag, line 37: Why do we need to grep instead of using the find as is?

lumag (changes requested):

.github/actions/list-jobs/action.yml
mwasilew: the path to the file is long. Prefix is prepended to the directory name. Is there a way to tell `find` to search for `{prefix}<wildcards>.yaml`? I didn't find such option. 

There are other yaml files in the repository and they also get added to the generated matrix. That's why the prefix is needed.

lumag: But isn't it exactly `find . -name "PREFIX*.yaml"`?

mwasilew: I tried - didn't work. With the typo we had it reduces to `find . -name "*.yaml"`

mwasilew: The full path looks sth like this
```
boottest-rb3gen2-core-kit-qcom-distro/rb3gen2-core-kit-qcom-distro-boot/projects/meta-qcom/devices/rb3gen2-core-kit/boot.yaml
```

lumag: @mwasilew  `find -path "${{inputs.prefix}}*/*.yaml"`?

mwasilew: @lumag 
I'm silly. What works is a simple as:
```
find boottest* -type f -name "*.yaml"
```
or using the workflow syntax
```
find ${{inputs.prefix}}* -type f -name "*.yaml"
```

lumag: I like the second option.

mwasilew:
This PR is obsolete. Closing.

PR #1660, opened 2026-02-27
ci: update nightly build runs time

ricardosalveti:
Minor update (minute) so we can merge with quic-yocto-ci, required to share the build notification via list.


ricardosalveti:
Please don't merge (but will need approvals), I need to merge this one with the quic-yocto-ci user.

PR #1666, opened 2026-03-02
Only create fit-dtb when QCOM_DTB_DEFAULT is set to multi-dtb

ricardosalveti:
Skip creating fit-dtb image when when QCOM_DTB_DEFAULT is set to a fixed 
device-tree.

### DIFF
--- a/classes/linux-qcom-dtbbin.bbclass
+++ b/classes/linux-qcom-dtbbin.bbclass
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
 
-inherit dtb-fit-image
+inherit_defer ${@bb.utils.contains('QCOM_DTB_DEFAULT', 'multi-dtb', 'dtb-fit-image', '', d)}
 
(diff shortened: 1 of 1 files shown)


vkraleti (dismissed):

koenkooi (dismissed):

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: BSD-3-Clause-Clear
 #
 
-inherit dtb-fit-image
+inherit ${@bb.utils.contains('QCOM_DTB_DEFAULT', 'multi-dtb', 'dtb-fit-image', '', d)}
```
quaresmajose: The `inherit_defer` should be used when conditions or variables are involved.
Perhaps it would be better to store the result of the condition in a new variable and use it later below when necessary.

ricardosalveti: Updated to use inherit_defer, didn't create a variable as I think it is easier to read the code by having the conditional.

anujm1 (approved):

vkraleti (approved):

PR #1671, opened 2026-03-03
workflows: replace poky-altcfg with nodistro and fix lava-test-plans

quaresmajose:
We name some of the ci task with the distro name, so better to use the nodistro since we are no longer use the poky-altcfg.


ricardosalveti:
Please rebase.

mwasilew:
@quaresmajose could you use this commit from lava-test-plans `d5b19bcc4d038a40849fe3de0847d4277d5c1305`? I added proper tags there.

PR #1673, opened 2026-03-04
Bump qcom-dtb-metadata to v0.3 and update FIT compatible strings

vkraleti:
Bump the qcom-dtb-metadata recipe to v0.3 to pick up support for the
new compatible string format. In this version, explicit `soc version` and 
`board revision` suffixes are introduced for DTB compatible string matching.
With this change, compatible strings that previously encoded SoC version or
board revision implicitly must be updated to the new format.


vkraleti:
@lumag @ricardosalveti UEFI changes to support this format are available from release 00122.0 onwards. Can you please check now?

vkraleti:
All test failures are due to known Bluetooth and Display issues. Since the updated compatible string structure impacts boot‑up, and boot validation has passed on all boards, this change is safe to merge.

PR #1675, opened 2026-03-04
Add debug build support in CI

vkraleti:
Extend the GitHub workflows to build debug variants in CI. These builds are produced
with `DEBUG_BUILD=1` and are intended for long‑running stability testing, providing
additional logging and diagnostics to help analyze system behavior and failures.


ricardosalveti:
> Let this change go, I'll submit another PR to limit some of the runs to only nightly.

Will merge to unblock the debug builds, but please work on moving them to nightly asap.

PR #1679, opened 2026-03-04
msm-gbm-backend: rename GBM backend library from msm_gbm to msm-kms_gbm

yashgupta2126:
The DRM driver name was updated from 'msm' to 'msm-kms'. Mesa's GBM backend selection uses the DRM driver name to auto-discover the backend library by looking for '<drm_driver_name>_gbm.so' in GBM_BACKENDS_PATH or the default path (/usr/lib/gbm/).

With the old library name 'msm_gbm.so', Mesa cannot find the backend for driver 'msm-kms' and silently falls back to the generic 'dri' backend

Rename the library to 'msm-kms_gbm.so' so that Mesa's auto-detection picks the correct MSM GBM backend without requiring the GBM_BACKEND environment variable to be set explicitly.

KERNEL Commit ID for reference: - 217ed15bd399980981f90f4332bc7ad4b05baa7e

### DIFF
--- a/recipes-graphics/msm-gbm-backend/files/0004-QCOM-libgbm-rename-GBM-backend-library-from-msm_gbm-.patch
+++ b/recipes-graphics/msm-gbm-backend/files/0004-QCOM-libgbm-rename-GBM-backend-library-from-msm_gbm-.patch
@@ -1,0 +1,46 @@
+From 6dc20b5b7c4e37443935107d2b14a0444964a4ac Mon Sep 17 00:00:00 2001
+From: Yash Gupta <quic_ygupt@quicinc.com>
+Date: Wed, 4 Mar 2026 22:39:55 +0530
+Subject: [PATCH] QCOM:libgbm: add support for both 'msm' and 'msm-kms' GBM
+ backend names
+
+The MSM DRM subsystem now exposes two driver names: 'msm' (legacy
+combined GPU+KMS driver) and 'msm-kms' (new KMS-only driver used on
+KGSL-based platforms). Mesa's GBM backend auto-detection looks for
+'<drm_driver_name>_gbm.so', so both names must be resolvable.
+
+Build the library as 'msm-kms_gbm.so' and install a compatibility
+symlink 'msm_gbm.so -> msm-kms_gbm.so' so that both driver names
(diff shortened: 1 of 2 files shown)


recipes-graphics/msm-gbm-backend/files/0004-QCOM-libgbm-rename-GBM-backend-library-from-msm_gbm-.patch
```diff
@@ -0,0 +1,32 @@
...
+Date: Wed, 4 Mar 2026 22:39:55 +0530
+Subject: [PATCH] QCOM:libgbm: rename GBM backend library from msm_gbm to
+ msm-kms_gbm
+
+With driver name 'msm-kms', Mesa looks for 'msm-kms_gbm.so'. Rename
+the library accordingly so GBM backend auto-detection works without
+requiring GBM_BACKEND env var to be set explicitly.
+
+Signed-off-by: Yash Gupta <quic_ygupt@quicinc.com>
+Upstream-Status: Pending
```
ricardosalveti, line 18: Why are we patching a repository we maintain?

yashgupta2126: The repository is shared between ubuntu deliverable in 6.8 kernel and mainline deliverable in 6.18+ kernel. The driver's name has changed with 'commit id 217ed15bd399980981f90f4332bc7ad4b05baa7e' in latest kernel version. Thus, raising a patch for mainline

ricardosalveti: Hm, NAK from my side, mainline should contain the latest changes as part of the repo (as part of a mainline / master / trunk branch), if we need to use an older rev / branch for ubuntu, then manage that separately.

ricardosalveti: Remember this will also later be used with debian and other distros (@lool fyi).

PR #1681, opened 2026-03-05
qcom-armv8a: use PARTLABEL for rootfs instead of hardcoded node

Kavinaya99:
This change updates the machine configuration to use PARTLABEL for selecting the root filesystem instead of a hardcoded device node. The root filesystem was previously identified using a fixed device node (for example /dev/sdaX).


lumag (changes requested):
Please rewrite the commit message.
Start with the description of the problem, continue with the actions necessary to be performed to solve it. Use imperative language for that.

conf/machine/qcom-armv8a.conf
```diff
@@ -58,7 +58,7 @@ LINUX_QCOM_KERNEL_DEVICETREE ?= " \
 "
 
 QCOM_BOOTIMG_PAGE_SIZE[apq8016-sbc] ?= "2048"
-QCOM_BOOTIMG_ROOTFS ?= "/dev/sda1"
+QCOM_BOOTIMG_ROOTFS ?= "PARTLABEL=rootfs"
 QCOM_BOOTIMG_ROOTFS[apq8016-sbc] ?= "/dev/mmcblk0p14"
```
koenkooi, line 62: The entries below still have some hardcoded entries, is there a plan to fix those as well. @lumag is 'system' correct for rb3gen2?

Kavinaya99: Currently, we are addressing the removal of hardcoded entries only for qcom-armv8a, as this machine is currently being used to validate the linux-yocto-dev kernel We have not extended these changes to the other entries yet.

lumag: 'system' was correct for the old LA / LE builds, which had system and userdata.

lumag: @Kavinaya99 you reponse doesn't make sense. Those entries are a part of the `qcom-armv8a` machine configuration.

ricardosalveti: rootfs for newer builds, system for older LA/LE as @lumag said.

ricardosalveti (approved):

conf/machine/qcom-armv8a.conf
vkraleti: How to differentiate between older and newer builds?

lumag: All the builds that we support are "new".

lumag: If I remember correctly, original software from Thundercomm has a typical mixture of Android and Linux userpace and thus had system and userdata partitions.

PR #1688, opened 2026-03-05
Fix zizmor findings

anujm1:
zizmor recommends that "checkout" action should be updated and specifically asked not to persist credentials [1] and that actions should be pinned by hash instead of tag [2]. Change all third party actions to use hash instead of tag.

[1] https://docs.zizmor.sh/audits/#artipacked
[2] https://docs.zizmor.sh/audits/#unpinned-uses

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -101,7 +101,7 @@ jobs:
           save_result_as_artifact: true
           save_job_details: true
           result_file_name: "${{ matrix.target.result_file }}"
-      - uses: mwasilew/github-action-matrix-outputs-write@v2
+      - uses: mwasilew/github-action-matrix-outputs-write@f7202d2224ebed937f287a2e2813e47fddd12bc8 # v2
         if: always()
         id: out
         with:
@@ -118,7 +118,7 @@ jobs:
     outputs:
       boot_result: "${{ steps.print-boot-result.outputs.boot_result }}"
     steps:
-      - uses: cloudposse/github-action-matrix-outputs-read@v1
+      - uses: cloudposse/github-action-matrix-outputs-read@33cac12fa9282a7230a418d859b93fdbc4f27b5a # v1
         id: read
         with:
           matrix-step-name: "submit-boot-job"
--- a/.github/workflows/publish-results.yml
+++ b/.github/workflows/publish-results.yml
@@ -59,7 +59,7 @@ jobs:
           private-key: ${{ secrets.TEST_REPORTING_APP_TOKEN }}
 
       - name: Publish Test Results
-        uses: EnricoMi/publish-unit-test-result-action@v2
+        uses: EnricoMi/publish-unit-test-result-action@c950f6fb443cb5af20a377fd0dfaa78838901040 # v2.23.0
         if: |
           always()
           && contains(steps.listfiles.outcome, 'success')
--- a/.github/workflows/repolinter.yml
+++ b/.github/workflows/repolinter.yml
@@ -19,16 +19,16 @@ jobs:
 
       - name: Verify repolinter config file is present
         id: check_files
-        uses: andstor/file-existence-action@v3
+        uses: andstor/file-existence-action@076e0072799f4942c8bc574a82233e1e4d13e9d6 # v3
         with:
           files: "repolint.json"
       - name: Run Repolinter with local repolint.json
         if: steps.check_files.outputs.files_exists == 'true'
-        uses: todogroup/repolinter-action@v1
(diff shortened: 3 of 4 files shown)


.github/actions/lava-test-plans/action.yml
```diff
@@ -23,7 +23,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-      - uses: actions/checkout@v6.0.2
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
lumag: I think the recommendation is only for 3rd party actions. I'd prefer to use `@v6`-style for GH Actions.

ricardosalveti: Yeah, I prefer to keep using tags for official GH actions and for the ones we own.

For 3rd party actions this is ok.

anujm1: I've removed the ones hosted at github.com/actions and foundriesio. When adding zizmor workflow, we'd need to configure which ones we trust.

.github/workflows/test.yml
```diff
@@ -101,7 +101,7 @@ jobs:
           save_result_as_artifact: true
           save_job_details: true
           result_file_name: "${{ matrix.target.result_file }}"
-      - uses: mwasilew/github-action-matrix-outputs-write@v2
+      - uses: mwasilew/github-action-matrix-outputs-write@f7202d2224ebed937f287a2e2813e47fddd12bc8 # v2
```
ricardosalveti, line 104: @mwasilew can we move this one under qualcomm / qualcomm-linux / foundriesio?

mwasilew: We can. I think I started OSR for this. Let me check it and report back here. There is also a chance I will eliminate it when refactoring the test.yml workflow.

mwasilew:
I think we should push this to `next` before merging.

.github/workflows/test.yml
mwasilew: OSR is still under review :(

steparce: As addressed in the OSR, we need @mwasilew to provide the correct Qualcomm organization as @ricardosalveti is stating there are a few options. 
@mwasilew what organization will this move under (qualcomm / qualcomm-linux / foundriesio)? In addition, what will be the repo name?

mwasilew: @steparce I replied in OSR. I'd like to move it to qualcomm-linux/github-action-matrix-outputs-write

mynameistechno:
> @mynameistechno I tried pushing to `next` but it looks like whitelist isn't correct: https://github.com/qualcomm-linux/meta-qcom/actions/runs/22919384312
> 
> ```
> The actions mwasilew/github-action-matrix-outputs-write@f7202d2224ebed937f287a2e2813e47fddd12bc8 and enricomi/publish-unit-test-result-action@c950f6fb443cb5af20a377fd0dfaa78838901040 are not allowed in qualcomm-linux/meta-qcom because all actions must be from a repository owned by your enterprise, created by GitHub, verified in the GitHub Marketplace, or match one of the patterns: EnricoMi/publish-unit-test-result-action@v2, Swatinem/rust-cache@*, andstor/file-existence-action@076e0072799f4942c8bc574a82233e1e4d13e9d6, andstor/file-existence-action@v1, andstor/file-existence-action@v3, cloudposse/github-action-matrix-outputs-write@v1, dcarbone/install-jq-action@*, egor-tensin/setup-clang@v1, foundriesio/lava-action@*, homebrew/actions/setup-homebrew@master, jgehrcke/github-repo-stats@release, mshick/add-pr-comment@*, mwasilew/github-action-matrix-outputs-write@v2, mynameistechno/qcom-reusable-workflows@*, newrelic/repolinter-action@v1, peter-evans/repository-dispatch@v3, pnpm/action-...
> ```

Ok I added both just now. Please retry.

.github/workflows/test.yml
steparce: Thanks for the response @mwasilew. As stated in OSR-20316 you still need Trademark approval. Please follow the instructions provided in the ticket.

ndechesne: thanks. However I think we should not block this PR. We can proceed with merging, and we will update when the repo is moved.

anujm1:
> https://github.com/qualcomm-linux/meta-qcom/actions/runs/23239763137 @anujm1 could you rebase this PR?

Yes, done. But, the build will probably fail right now because of layer compat issues.

PR #1697, opened 2026-03-06
iris-video-dlkm: update iris driver to the latest

venvenk:
This tag brings a below updates for Qcom Iris
- Aligns compose selection values with crop for non‑scaling IRIS variants
- Fixes FRMIVAL stepwise denominators by converting Q16 frame‑rate values
- Align encoder input resolution to fix green lines at the bottom


koenkooi:
I see a lot of tags in the upstream repo, can we stop using `0.0` as a version in this recipe?

venvenk:
> I see a lot of tags in the upstream repo, can we stop using `0.0` as a version in this recipe?

This recipe tracks the overlay branch video.qclinux.main. I checked video-driver.git: while the repo has many tags (e.g. LE.PRODUCT.*), none are reachable from origin/video.qclinux.main (git tag --merged origin/video.qclinux.main is empty and git describe --tags returns only the SHA). So for the current SRCREV there isn’t a tag I can use for PV.
If you want this recipe to follow a specific tag/branch, please point me to it.

PR #1703, opened 2026-03-06
Adds BSP layer download mirrors

quaresmajose:
Adds the official mirror of the BSP, which contains the necessary artifacts to minimize reproducibility issues as much as possible. The mirror uses the MIRRORS and thus works in fallback mode; when the official source is not available, the mirror is used.

### DIFF
--- a/classes/qli-mirrors.bbclass
+++ b/classes/qli-mirrors.bbclass
@@ -1,0 +1,25 @@
+# Qualcomm Linux download mirrors to minimize reproducibility issues
+#
+# Copyright (c) 2026 Qualcomm Innovation Center, Inc.
+#
+# SPDX-License-Identifier: MIT
+#
+
+QLI_MIRRORS_URI ?= "https://artifacts.codelinaro.org/aritfactory/qli-ci/downloads/${QLI_BASELINE}"
+
+QLI_MIRRORS ?= " \
+svn://.*/.*     ${QLI_MIRRORS_URI}/ \
+git://.*/.*     ${QLI_MIRRORS_URI}/ \
+gitsm://.*/.*   ${QLI_MIRRORS_URI}/ \
+hg://.*/.*      ${QLI_MIRRORS_URI}/ \
+p4://.*/.*      ${QLI_MIRRORS_URI}/ \
+https?://.*/.*  ${QLI_MIRRORS_URI}/ \
+ftp://.*/.*     ${QLI_MIRRORS_URI}/ \
+npm://.*/?.*    ${QLI_MIRRORS_URI}/ \
+s3://.*/.*      ${QLI_MIRRORS_URI}/ \
+crate://.*/.*   ${QLI_MIRRORS_URI}/ \
+gs://.*/.*      ${QLI_MIRRORS_URI}/ \
+"
+
+# Add Qualcomm Linux mirror so we can fallback to it
+MIRRORS += "${QLI_MIRRORS}"
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -28,6 +28,9 @@ BBFILES_DYNAMIC += " \
     selinux:${LAYERDIR}/dynamic-layers/selinux/*/*/*.bbappend \
 "
 
+# Qualcomm Linux baseline
+QLI_BASELINE = "main"
+
 # Set default provider for virtual-diag-router to avoid warnings
 # when both diag and diag-router are available
 PREFERRED_RPROVIDER_virtual-diag-router ?= "diag"
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -31,6 +31,8 @@ BBFILES_DYNAMIC += " \
 # Qualcomm Linux baseline
 QLI_BASELINE = "main"
 
+INHERIT += "qli-mirrors"
+
 # Set default provider for virtual-diag-router to avoid warnings
 # when both diag and diag-router are available
 PREFERRED_RPROVIDER_virtual-diag-router ?= "diag"
--- a/ci/mirror-download-disable.yml
+++ b/ci/mirror-download-disable.yml
@@ -1,0 +1,10 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+# This fragment should be used in a clean DL_DIR running all the bitbake fetch tasks
+# bitbake --runall fetch world
+
+header:
+  version: 14
+
+local_conf_header:
+  download_mirror_disable: |
+    INHERIT:remove = "qli-mirrors"
--- a/ci/mirror-download-test.yml
+++ b/ci/mirror-download-test.yml
@@ -1,0 +1,11 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+# This fragment should be used in a clean DL_DIR running all the bitbake fetch tasks
+# bitbake --runall fetch world
+
+header:
+  version: 14
+
+local_conf_header:
+  download_mirror_test: |
+    BB_FETCH_PREMIRRORONLY = "1"
+    PREMIRRORS += "${QLI_MIRRORS}"


classes/qcom-mirrors.bbclass
```diff
@@ -0,0 +1,25 @@
...
+#
+# SPDX-License-Identifier: MIT
+#
+
+# Add qcom mirror so we can fallback to it
+MIRRORS += "${MIRRORS_QCOM}"
+
+MIRRORS_QCOM_URI ?= "https://artifacts.codelinaro.org/aritfactory/qli-ci/downloads/${QCOM_RELEASE}"
+
+MIRRORS_QCOM ?= " \
```
ndechesne: you use qcom-mirror class, and the variable is MIRROR_QCOM, pick one :)

quaresmajose: changed everything to QCOM_ :) thanks

classes/qcom-mirrors.bbclass
```diff
@@ -0,0 +1,25 @@
+# Layer download mirror to minimize reproducibility issues 
+#
+# Copyright (c) 2026 Qualcomm Innovation Center, Inc.
+#
+# SPDX-License-Identifier: MIT
+#
+
+# Add qcom mirror so we can fallback to it
+MIRRORS += "${QCOM_MIRRORS}"
```
ricardosalveti: Can you move this assignment after QCOM_MIRRORS is defined? Easier to parse when reading this.

conf/layer.conf
```diff
@@ -28,6 +28,9 @@ BBFILES_DYNAMIC += " \
     selinux:${LAYERDIR}/dynamic-layers/selinux/*/*/*.bbappend \
 "
 
+# Release
+QCOM_RELEASE = "main"
```
ricardosalveti: QLI_RELEASE or QLI_BASELINE would work better from my perspective.

QCOM is too generic to be used here.

conf/layer.conf
```diff
@@ -31,6 +31,8 @@ BBFILES_DYNAMIC += " \
 # Release
 QCOM_RELEASE = "main"
 
+INHERIT += "qcom-mirrors"
```
ricardosalveti: Can we rename the file to clo-mirrors instead? It is hosted by codelinaro.

classes/qcom-mirrors.bbclass
```diff
@@ -0,0 +1,25 @@
...
+
+QCOM_MIRRORS_URI ?= "https://artifacts.codelinaro.org/aritfactory/qli-ci/downloads/${QCOM_RELEASE}"
```
lumag: Why do we need a release-specific dir?

lumag (changes requested):

conf/layer.conf
quaresmajose: I will change it to QLI_BASELINE.

classes/qcom-mirrors.bbclass
quaresmajose: I actually had some doubts about how to do that and that was the reason for introducing the variable with the version. This is necessary to separate the development version from the stable versions. It's safer to store them in different locations because files can be modified, and this probability is higher in development versions, which can unintentionally break the stable versions.

conf/layer.conf
craigez: Same comment for "QCOM_MIRRORS" above too

classes/qcom-mirrors.bbclass
lumag: Why do we want that? In the end, all the files are uniquelly identified, having two files with the same name is a mistake. Likewise,BitBake should handle not having the necessary revision in the mirror Git archive.

anujm1: It might be needed for git repositories if the tree was rebased. Having the unreachable commit will be helpful when we want to ensure older releases work. 

There seems to be a flag "[rebaseable](https://git.openembedded.org/bitbake/commit/?id=92701d4c5372db48847c70da4ebd0736d79fd54b)" that works around it by creating mirror tarball for each revision

classes/qcom-mirrors.bbclass
quaresmajose: done

conf/layer.conf
quaresmajose: change all QCOM_MIRRORS variables to QLI_MIRRORS

classes/qcom-mirrors.bbclass
quaresmajose: > Why do we want that? In the end, all the files are uniquelly identified, having two files with the same name is a mistake. Likewise,BitBake should handle not having the necessary revision in the mirror Git archive.

This issue is limited only to the git archive, please take a look at the yocto [mirror](https://downloads.yoctoproject.org/mirror/sources) on the archives git2_*tar.gz. Al of them have a generic name that is not dependent on the revisions included in the repo.

quaresmajose: > It might be needed for git repositories if the tree was rebased. Having the unreachable commit will be helpful when we want to ensure older releases work.
> 
> There seems to be a flag "[rebaseable](https://git.openembedded.org/bitbake/commit/?id=92701d4c5372db48847c70da4ebd0736d79fd54b)" that works around it by creating mirror tarball for each revision

I wasn't familiar with that flag, but I'd prefer not to use it for now, since it forces the SRC_URI of all recipes to be rewritten in the layer.

ricardosalveti: Yeah, without this flag the git snapshots are all named the same way, requiring a different directory.

ricardosalveti (dismissed):

classes/qcom-mirrors.bbclass
lumag: What is the current BCP from the Yocto side? I see that `oe-core/meta/classes-global/mirrors.bbclass` uses generic URI without the embedded branch.

anujm1 (dismissed):

classes/qcom-mirrors.bbclass
quaresmajose: > What is the current BCP from the Yocto side? I see that oe-core/meta/classes-global/mirrors.bbclass uses generic URI without the embedded branch.

It depends on how you look at it; in the first few lines, there are two locations for the same destination.
```
MIRRORS += "\
${DEBIAN_MIRROR}        http://snapshot.debian.org/archive/debian/20180310T215105Z/pool \
${DEBIAN_MIRROR}        http://snapshot.debian.org/archive/debian/20250101T023759Z/pool \
```

The above pattern is similar for the proposed tree in our mirror:

https://artifacts.codelinaro.org/aritfactory/qli-ci/downloads/main
https://artifacts.codelinaro.org/aritfactory/qli-ci/downloads/1.0
https://artifacts.codelinaro.org/aritfactory/qli-ci/downloads/2.0

We could list them all, but I don't see any advantages in that. Do you see any benefit?

quaresmajose: However, writing everything in the same location doesn't answer the question of how we can avoid file rewrites during development on the main branch in stable versions.

lumag: This brings us to the question, why do we want it. Your commit messages don't contain an answer to that question.

classes/qli-mirrors.bbclass
```diff
@@ -0,0 +1,25 @@
+# Qualcomm Linux download mirrors to minimize reproducibility issues
+#
+# Copyright (c) 2026 Qualcomm Innovation Center, Inc.
```
lumag, line 3: What issue are you trying to solve? Please start the commit message by describing the problem.

conf/layer.conf
```diff
@@ -31,6 +31,8 @@ BBFILES_DYNAMIC += " \
 # Qualcomm Linux baseline
 QLI_BASELINE = "main"
```
lumag, line 32: Can we use `${LAYERSERIES_CORENAMES}` instead?

lumag (changes requested):

conf/layer.conf
quaresmajose: Yes, I think we could use it, but we have our own versioning scheme which is a little different:

QLI 1.X: Yocto LTS kirkstone and scarthgap based
QLI 2.X: Yocto LTS wrynose 

Our 1.X extended to two upstream LTS versions.

classes/qli-mirrors.bbclass
quaresmajose: _minimize reproducibility issues as much as possible_, that is already on the commit

conf/layer.conf
lumag: So, what is the reason for having separate mirrors? Is it QLI version or is it a difference in the baselines (which means different commit IDs in the tarballs, etc)? In the former case, what is the expected name for QLI 2.x vs QLI 3.x vs mainline? In the latter case, QLI 1.x doesn't mean anything.

quaresmajose: The reason for having separate mirrors has already been discussed above. 

From the internal discussions that took place, I got the idea that mirror would be based on our QLI version. That's why I created this new variable. 

To my knowledge, the names for the versions were not discussed, and in my opinion, they aren't even necessary. We can use main, qli1, qli2, etc

classes/qli-mirrors.bbclass
lumag: Which issues? Do you see tarball changes? Other fetch issues? Please be more exact.
Also "This class adds ..." doesn't sound like a description of the issue.
If you observe a reproducibility issue, it can be desribed as "When there is a storm at Alpha Centauri, the stream of delta particles disturbs connectivity to YP.org, which causes build failures in the CI. To save on build attempts provide a stationary mirror for all tarballs and git repos."

quaresmajose: The issue is not because of tarball changes but git changes, the mirror will provide relaiable git repository archivestha can still be used regardless of whether these repositories disappear or are altered in their original sources.

What the mirror provides is easier to understand if you look at the variable.
https://docs.yoctoproject.org/bitbake/dev/bitbake-user-manual/bitbake-user-manual-ref-variables.html#term-BB_GENERATE_MIRROR_TARBALLS

> If you observe a reproducibility issue, it can be desribed as "When there is a storm at Alpha Centauri, the stream of delta particles disturbs connectivity to YP.org, which causes build failures in the CI. To save on build attempts provide a stationary mirror for all tarballs and git repos."

In this case, nothing is observed, and the intention of this change is to prevent anything from being observed.

lumag: Okay, so we are slowly getting to the point. "To prevent the build issues caused by the upstream retrospectively changing git branches / tags, use generated tarballs". At least this describes an exact issue (which I think everybody has stumbled upon at least once or twice in our work). Then it's (more or less) clear that:
- the tarballs should be per the LTS branch rather than per the QLI release: versions in kirkstone and scarthgap might already require incompatible repo layout
- there is little point in having tarballs for the main branch, it is a moving target which adapts to the repo changes

Also please document, how those tarballs are generated and updated.

quaresmajose: The generation of those tarballs is outside the context of integration for its use. Basically, it consists of using the variable BB_GENERATE_MIRROR_TARBALLS.

quaresmajose:
I would appreciate it if you could do a new review so it can be integrated in the qli-2.0.

The download mirror is also available and ready to use.

anujm1 (approved):

ricardosalveti (approved):

lumag (approved):

Issue #1454, opened 2026-01-29
Need alignment for QTEE TA installation path


harshaldev27:
Hello,

The minkipc repo is currently hosting Trusted Application (TA) binary blobs which need to be installed into the rootfs similar to how OP-TEE installs TA's in the /lib/optee-armtz path: https://github.com/OP-TEE/optee_client/blob/master/tee-supplicant/src/tee_supplicant.c#L68

See: https://github.com/qualcomm/minkipc/tree/main/libminkadaptor/tests/ta/tzecotestapp

We are looking for guidance on where to install these QTEE TAs in the rootfs. Earlier these were part of /lib/firmware, but since Trusted Applications aren't exactly firmware enabling a piece of hardware, we feel they are misplaced here.

Perhaps similar to OP-TEE, installing them in /lib/qtee-tas via minkipc recipe should be acceptable?

Do let us know your thoughts.

Regards,
Harshal

lumag:
Are they board-specific, platform-specific or generic? We have customer requests to support single rootfs running on different kinds of devices.

ndechesne:
we are still missing the per soc profiles to properly sign , i believe.

ricardosalveti:
Yeah, the path suggested would be fine if we could have a single set of TAs that were generic to all hardware platforms. With platform-specific TAs (due signature or qtee versions) we need to make sure the files are not going to conflict with each other once we generate a single rootfs that can support multiple hardware platforms.

harshaldev27:
> we are still missing the per soc profiles to properly sign , i believe.

Hi @ndechesne , thanks for pointing this out. I am trying to get a hold of the sectools person to figure out why this is missing. Let me update on this shortly.

harshaldev27:
Hi @ricardosalveti  and @lumag 

> Yeah, the path suggested would be fine if we could have a single set of TAs that were generic to all hardware platforms. With > platform-specific TAs (due signature or qtee versions) we need to make sure the files are not going to conflict with each other > once we generate a single rootfs that can support multiple hardware platforms.

Like I said, it cannot be a single-set of TA binaries which work for all platforms unfortunately.

> Please note, I wrote about the target system. It should have all the signed binaries for all the platforms it supports. It can not > be limited to supporting just one set of binaries.

[...]

PR #1707, opened 2026-03-07
minkipc: Update tag to v1.2.4 to add support for PKCS#11

harshaldev27:
Updated minkipc tag to v1.2.4 which brings support for PKCS#11 on Qualcomm platforms with QTEE support.

The support is provided through optee_client and optee_test repos v4.0.0 which are placed in the optee-client/ and optee-test/ sub directories. Only the PKCS#11 relevant library and test suites are built and supported, i.e.
libckteec_qtee and xtest_qtee (pkcs11-tests). The recipe license is updated to reflect the licenses of these repos.

The PKCS#11 test-signed Trusted Application binaries is also installed for supported targets (qcm6490, qcs9100, qcs8300 and qcs615) under `/lib/qtee-tas` as per conclusions in https://github.com/qualcomm-linux/meta-qcom/issues/1454.

On an RPMB provisioned device, the pkcs11-tests can be run by invoking the xtest_qtee binary.

The changes in this PR are as per conclusions based on discussions in the older PR: https://github.com/qualcomm-linux/meta-qcom/pull/1379

Depends on: PR https://github.com/qualcomm-linux/meta-qcom/pull/1733 and PR https://github.com/qualcomm-linux/meta-qcom/pull/1895


harshaldev27:
Hi @anujm1,

> Would it be better to include these explicitly in `SRC_URI` instead of using `gitsm://` so we know what is being fetched?

It's easy to check this by looking at the .gitmodules file in the minkipc repo. 😄

anujm1:
> Hi @anujm1,
> 
> > Would it be better to include these explicitly in `SRC_URI` instead of using `gitsm://` so we know what is being fetched?
> 
> It's easy to check this by looking at the .gitmodules file in the minkipc repo. 😄

I know how to check those. But I won't go to the repository every time I'm looking at the recipe. It also doesn't say which revision is being fetched. It also should be checked whether these are included in SBoM correctly.

b49020:
> > I had a chat with @b49020 and I think the better approach here would be to skip building minkipc packages for builds with OPTEE support instead of QTEE. Perhaps I should raise a change in `meta-qcom-distro` where we only include the `minkipc` package in `packgagegroup-qcom-security` if MACHINE_FEATURES =! "optee". Please let me know if you are fine with this approach.
> 
> This won't work, our target is to have one single rootfs that would work with both normal optee recipes and minkipc. And in your case they are not providing the same thing, just the same files, so we would need to rename them.

That sounds like nice end objective which is in line with standard distros like Debian too. @harshaldev27 let's just rename to following:
```
libckqteec.so*
xtest_qtee
```
Anyhow, in the longer run we need to start discussions in upstream OP-TEE project for proper integration of QTEE implementation.

harshaldev27:
> > > Agreed, but as per the original PR I did try to create a separate recipe for fetching the `optee-test` and `optee-client` repos for building these modules. However, the concern over there was that `meta-arm` layer is already providing these recipes, and so creating a new recipe in `meta-qcom` for these does not feel right. To address this, it was suggested to go the submodules way.
> > 
> > 
> > I didn't suggest separate recipes. I was asking about adding these to `SRC_URI` explicitly instead of using `gitsm://`.
> 
> Yeah, I would also prefer to have the submodules as part of SRC_URI, a lot easier to identify what the recipe is bringing on.

Ack.

> > > You are now installing components of the optee upstream project, make sure this is reflected in the LICENSE description as well.
> > 
> > 
> > So I realize that the sources we're building and shipping from optee-test and optee-client require us to include the BSD-2-Clause and GPL-2.0-only license. But I don't think there's a way to add checksums for these licenses by referring the paths for the submodules.
> 
> Add the submodules as part of SRC_URI and you will be able to include them.
> 

Ack.

> > Should I add these two licenses to the minkipc repo to allow checksum comparison? Legal did not comment on this but I think this should be done.
> 
> No need, we just need to point to the license files available as part of the other repos.
> 

Ack.

harshaldev27:
> > I'm trying to take a step back and really think through if this is the only way forward. My biggest concern is, once we have changed the name of the library (I don't care about `xtest` since it's a test binary anyway), and customers start linking to it, it will be very difficult to convince them in a subsequent release which re-names it again to `libckteec` (through meta-arm layer) to migrate back to `libckteec`. Not to mention that it will break their applications until they do so. Maybe we'll have to create `libckqteec` as a symlink to the final `libckteec` through the meta-arm layer?
> 
> If they are providing the same symbols, sure, a link would be enough. Otherwise we will just have to document the breaking changes saying they will have to rebuild against an updated ckteec implementation (from the upstream project).
> 
> We expect this to be temporary anyway.

Yeah, well I think that in the long run, being able to support pkcs11 for QTEE directly from tip of `meta-arm` and `optee-client` is going to be a big step that will require QTEE PKCS11 TA updates as well. We can continue to support `libckqteec` until that becomes possible. Once we have reached that milestone and customers have migrated to `libckteec` for latest PKCS11 version, we can deprecate `libckqteec` and corresponding old PKCS11 TA version support from subsequent `minkipc` releases.

So Ack.

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
```diff
@@ -8,13 +8,25 @@ qteesupplicant service is designed for invocation dispatch and handling callback
...
-LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2b1366ebba1ebd9ae25ad19626bbca93"
+LICENSE = "BSD-3-Clause & BSD-2-Clause & GPL-2.0-only"
 
 inherit cmake systemd pkgconfig
 
-SRC_URI = "git://github.com/qualcomm/minkipc.git;branch=main;protocol=https;tag=v${PV}"
-SRCREV = "307a2f368051d0436d450a9d1f5fa14ff0f94580"
+SRC_URI = "git://github.com/qualcomm/minkipc.git;branch=main;protocol=https;tag=v${PV};name=minkipc \
+           git://github.com/OP-TEE/optee_client.git;branch=master;protocol=https;name=opteec;destsuffix=${BPN}-${PV}/optee-client/optee_client \
+           git://github.com/OP-TEE/optee_test.git;branch=master;protocol=https;name=opteet;destsuffix=${BPN}-${PV}/optee-test/optee_test \
```
ricardosalveti: Do we have a particular tag to track here? I guess the 4.0.0 tag? If so, please add tag here as well for both.

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
```diff
@@ -8,13 +8,25 @@ qteesupplicant service is designed for invocation dispatch and handling callback
...
+           "
+
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2b1366ebba1ebd9ae25ad19626bbca93 \
+                    file://optee-client/optee_client/LICENSE;md5=69663ab153298557a59c67a60a743e5b \
+                    file://${COMMON_LICENSE_DIR}/BSD-2-Clause;md5=cb641bc04cda31daea161b1bc15da69f \
+                    file://${COMMON_LICENSE_DIR}/GPL-2.0-only;md5=801f80980d171dd6425610833a22dbe6"
```
ricardosalveti: What about the optee_test license?

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
```diff
@@ -36,6 +36,7 @@ EXTRA_OECMAKE = " \
     -DSYSTEMD_UNIT_DIR=${systemd_unitdir}/system \
     -DUDEV_DIR=${nonarch_libdir}/udev/rules.d \
     -DMINKIPC_LIBEXEC_DIR=${base_bindir} \
+    -DPERSIST_MOUNT_UNIT:STRING=var-lib-tee.mount \
```
ricardosalveti: I'm not so sure we should depend on this mount point, it is good to have but it can still work without it. I'm fine adding a before qtee-supplicant in var-lib-tee.mount, as then we will have the dependency set for both scenarios (with and without tee mount).

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
harshaldev27: Ack.

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
harshaldev27: Actually, at tag v4.0.0, there is a LICENSE.md file which simply provides URLS to generic BSD-2-Clause and GPL-2.0-only license: https://github.com/OP-TEE/optee_test/blob/4.0.0/LICENSE.md
So I've tried to reflect the same here.

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
harshaldev27: I was thinking from a distro agnostic perspective. Another systemd based distro (e.g. Ubuntu) might have a different name for this mount unit file. I felt that being able to configure that based on the mount unit name provided during the distro build is a better approach rather than asking the distro to make changes to their systemd unit file with a `Before=qteesupplicant` tag. But I am fine with either. Do let me know.

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
ricardosalveti: Then track this file as well here, LIC_FILES_CHKSUM is to track the license text from upstream and to detect changes.

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
ricardosalveti: Since you control the systemd job for qteesupplicant, that will be stable, then it is up for the distro to decide having the persist partition and its respective mount file, which is why I prefer changing the mount unit instead.

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
harshaldev27: Ack.

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
harshaldev27: `PERSIST_MOUNT_UNIT` is not passed to `qteesupplicant.service`. It is passed to `sfsconfig.service` which creates directories in the persist partition.
The dependency goes like this. `qteesupplicant.service` -> `sfsconfig.service` -> `var-lib-tee.mount`. So we need to add `Before=sfsconfig.service` in `var-lib-tee.mount`. I'll replace this commit with one that introduces this change instead.

ricardosalveti: > `PERSIST_MOUNT_UNIT` is not passed to `qteesupplicant.service`. It is passed to `sfsconfig.service` which creates directories in the persist partition.
> 
> The dependency goes like this. `qteesupplicant.service` -> `sfsconfig.service` -> `var-lib-tee.mount`. So we need to add `Before=sfsconfig.service` in `var-lib-tee.mount`. I'll replace this commit with one that introduces this change instead.


We should move away from sysconfig.service and use systemd tmpfiles instead, as I suggested in a previous pr, then this service won't be required.

harshaldev27: Sure @ricardosalveti, I have it in my worklog to move to tmpfiles from sfsconfig.

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
```diff
@@ -49,5 +49,5 @@ FILES:${PN}-qteesupplicant = "${bindir}/qtee_supplicant \
                               ${base_bindir}/sfs_config \
 "
 
-RDEPENDS:${PN}-qteesupplicant = "${PN}"
+RDEPENDS:${PN}-qteesupplicant = "${PN} mount-tee-partition"
```
ricardosalveti: This should be RRECOMMENDS instead, it works fine without it, but it is recommended to have it.

harshaldev27: Ack.

recipes-bsp/partition/mount-tee-partition/var-lib-tee.mount
```diff
@@ -4,7 +4,7 @@
 
 [Unit]
 Description=Mount persist partition at /var/lib/tee
-Before=local-fs.target
+Before=local-fs.target sfsconfig.service
```
lumag, line 7: It should be other way around. sfsconfig should be After the tee mount. We can't expect that the provider (mount unit) knows all consumers, but the consumers know what they need.

lumag (changes requested):

recipes-bsp/partition/mount-tee-partition/var-lib-tee.mount
ricardosalveti: I suggested this way because the service does not depend on var-lib-tee.mount, it is just a nice to have.

And since this mount point is just used by qtee-supplicant, this is fine.

lumag: I think in dowsntream we had several other users of the persist partition.

ricardosalveti: Well, we are mounting it under /var/lib/tee assuming it is for tee only, we asked and nobody complained.

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
```diff
@@ -8,13 +8,26 @@ qteesupplicant service is designed for invocation dispatch and handling callback
...
-SRCREV = "307a2f368051d0436d450a9d1f5fa14ff0f94580"
+SRC_URI = "git://github.com/qualcomm/minkipc.git;branch=main;protocol=https;tag=v${PV};name=minkipc \
+           git://github.com/OP-TEE/optee_client.git;branch=master;protocol=https;tag=4.0.0;name=opteec;destsuffix=${BPN}-${PV}/optee-client/optee_client \
+           git://github.com/OP-TEE/optee_test.git;branch=master;protocol=https;tag=4.0.0;name=opteet;destsuffix=${BPN}-${PV}/optee-test/optee_test \
+           "
+
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2b1366ebba1ebd9ae25ad19626bbca93 \
+                    file://optee-client/optee_client/LICENSE;md5=69663ab153298557a59c67a60a743e5b \
+                    file://optee-test/optee_test/LICENSE.md;md5=daa2bcccc666345ab8940aab1315a4fa \
+                    file://${COMMON_LICENSE_DIR}/BSD-2-Clause;md5=cb641bc04cda31daea161b1bc15da69f \
```
lumag: Well, no. DOn't checksum licneces in common dirs. They won't change.

lumag (changes requested):

dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.3.bb
harshaldev27: Ack.

ricardosalveti:
Please rebase to fix the CI issues.

ricardosalveti:
Sorry, please do another rebase as we fixed another CI issue.

koenkooi (dismissed):

lumag (dismissed):

harshaldev27:
> A minor nit: the license change isn't explicitly mentioned in the commit message.

Done.

lumag (dismissed):

vkraleti (dismissed):

koenkooi (dismissed):

ricardosalveti (changes requested):
```
root@rb3gen2-core-kit:~# xtest_qtee
Run test suite with level=0

TEE test application started over default TEE instance
######################################################
#
# regression+pkcs11
#
######################################################

* regression_1001 Core self tests
IGPAppClient_openSession failed: 15
mink_open_session() failed: 15
/usr/src/debug/minkipc/1.2.3/optee-test/optee_test/host/xtest/regression_1000.c:299: res has an unexpected value: 0xffff0000 = TEEC_ERROR_GENERIC, expec[  344.073683] systemd-coredump[1469]: Failed to get unit, ignoring: No such device or address
ted 0x0 = TEEC_SUCCESS
*** stack smashing detected ***: terminated
Aborted                    (core dumped) xtest_qtee
```

Since we are adding optee_test, I assume xtest_qtee needs to work, which is not the case here.

harshaldev27:
> > Sure, we can, I don't mind. But I thought there is a general aversion to maintaining patches in meta-qcom, so I didn't want to go that way and decided to keep it as it is for now. I can raise a patch to only run the applicable test cases.
> > > I would prefer having a clean xtest_qtee execution here, as then we can add to our test CI and avoid regressions as we move this forward.
> > 
> > 
> > I agree, let me raise a patch which stub the inapplicable regression test cases.
> 
> Thanks, I'm ok with a local patch selecting only the test cases that are known to work with qtee, as then we can instruct folks (and our CI) to just run xtest_qtee directly, without any additional argument.

Done. Here is the output of running xtest_qtee now:

```
root@rb3gen2-core-kit:/# xtest_qtee
Run test suite with level=0

TEE test application started over default TEE instance
######################################################
#
# pkcs11
#
######################################################

* pkcs11_1000 Initialize and close Cryptoki library
  pkcs11_1000 OK

* pkcs11_1001 PKCS11: List PKCS#11 slots and get information from
o pkcs11_1001.1 Test C_GetFunctionList()
  pkcs11_1001.1 OK
o pkcs11_1001.2 Test C_GetInfo()

[...]

pkcs11_1027 OK
pkcs11_1028 FAILED first error at /usr/src/debug/minkipc/1.2.3/optee-test/optee_test/host/xtest/pkcs11_1000.c:8854
+-----------------------------------------------------
3969 subtests of which 11 failed
29 test cases of which 7 failed
0 test cases were skipped
TEE test application done!
```

harshaldev27:
> > Here is the output of running xtest_qtee now, which can be enabled with CI once I release the PKCS#11 TA in next minkipc release immediately after this
> 
> Ah, it is not supported with this release, why can't you just release and bump this pull-request to the revision that includes the TA?
> 
> We can't merge a PR that adds support to something that is not functional.

Done. I have bumped up the release to v1.2.4 which installs the PKCS#11 TA for qcm6490, qcs9100, qcs8300 and qcs615 in `/lib/qtee-tas` as per conclusions on this issue: https://github.com/qualcomm-linux/meta-qcom/issues/1454

I have dropped the other two commits in this PR and moved them to a separate PR since they are un-related to this PR and only for mounting persist partition PR https://github.com/qualcomm-linux/meta-qcom/pull/1895. You should now be able to run the pkcs-11 test suite via `xtest_qtee` given that you take PR1895 in your workspace first. I hope we can merge this quickly since we already have approval on these changes from this current PR.

Please be aware that your device should be RPMB provisioned with test keys before it can run these test suites. The RPMB provisioning tool is planned for release publicly by the storage team soon. Meanwhile you can use the `smcinvoke_client` to provision the RPMB via following commands assuming you have access to our internal `smplap` TA. Let me know if you need it.
```
#define CLIENT_CMD14_RUN_RPMB_PROV 14
#define CLIENT_CMD15_RUN_RPMB_ERASE 15
#define CLIENT_CMD17_RUN_RPMB_RW 17

smcinvoke_client -i /<path to smplap>/ 15 1 # Erase RPMB
smcinvoke_client -i /<path to smplap>/ 14 1 # Provision RPMB with test keys
smcinvoke_client -i /<path to smplap>/ 17 1 # Test RPMB read/write

# Example
adb push smplap64.mbn /data
smcinvoke_client -i /data 15 1
```

I tried to ensure that only delta in minkipc recipe should be shown in the commit, but git keeps treating minkipc_1.2.4.bb as a brand new file despite my best efforts.

Thanks,
Harshal

harshaldev27:
> Your message still uses `adb push` and references `/data`, so it doesn't look integrated. Do we need to start a new repo with TA apps? Or should they go to linux-firmware?

I only referenced /data for smplap64.mbn because we don't yet have legal approval for releasing it's binary to minkipc and installing it in rootfs. For PKCS#11 tests you don't need to do anything. It's all integrated now.

ricardosalveti (approved):
```
Result of testsuite pkcs11:
pkcs11_1000 OK
pkcs11_1001 OK
pkcs11_1002 OK
pkcs11_1003 OK
pkcs11_1004 OK
pkcs11_1005 OK
pkcs11_1006 OK
pkcs11_1007 OK
pkcs11_1008 OK
pkcs11_1009 OK
pkcs11_1010 OK
pkcs11_1011 OK
pkcs11_1012 OK
pkcs11_1013 OK
pkcs11_1014 OK
pkcs11_1015 OK
pkcs11_1017 OK
pkcs11_1020 OK
pkcs11_1022 OK
pkcs11_1023 OK
pkcs11_1024 OK
+-----------------------------------------------------
3594 subtests of which 0 failed
21 test cases of which 0 failed
0 test cases were skipped
TEE test application done!
```

Finally and thanks for all the work on this PR.

ricardosalveti:
Merging based on previous approvals.

PR #1710, opened 2026-03-09
gst-plugins-imsdk: qtismartvencbin: add runtime dependency

Raja-Ganapathi-Busam:
Ensure that the Smart Video Encoder plugin can load the libVideoCtrl library at runtime by adding a runtime dependency on smart-venc-ctrl-algo. The plugin uses dlopen() to load this library, so the dependency guarantees its availability during execution.

### DIFF
--- a/recipes-multimedia/imsdk/gst-plugins-imsdk_0.1.0.bb
+++ b/recipes-multimedia/imsdk/gst-plugins-imsdk_0.1.0.bb
@@ -42,3 +42,6 @@ PACKAGECONFIG[videoproc]    = "-DENABLE_GST_VIDEOPROC_PLUGINS=1, -DENABLE_GST_VI
 
 # The TFLite plugin loads the TensorFlow Lite library at runtime using dlopen(). To ensure runtime availability, added runtime dependency on 'tensorflow-lite'.
 RDEPENDS:${PN}-qtimltflite += "tensorflow-lite"
+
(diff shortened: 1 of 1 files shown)


recipes-multimedia/imsdk/gst-plugins-imsdk_0.1.0.bb
```diff
@@ -42,3 +42,6 @@ PACKAGECONFIG[videoproc]    = "-DENABLE_GST_VIDEOPROC_PLUGINS=1, -DENABLE_GST_VI
 
 # The TFLite plugin loads the TensorFlow Lite library at runtime using dlopen(). To ensure runtime availability, added runtime dependency on 'tensorflow-lite'.
 RDEPENDS:${PN}-qtimltflite += "tensorflow-lite"
+
+# The Smart Video Encoder plugin loads the libVideoCtrl library at runtime using dlopen(). To ensure runtime availability, added runtime dependency on 'smart-venc-ctrl-algo'.
+RDEPENDS:${PN}-qtismartvencbin += "smart-venc-ctrl-algo"
```
lumag, line 47: RRECOMMENDS?

Raja-Ganapathi-Busam: `smart-venc-ctrl-algo` is not optional but required at runtime for `qtismartvencbin`.
Which one will be more suitable for this `RDEPENDS` or `RRECOMMENDS`...

PR #1715, opened 2026-03-09
meta-qcom: Update Dpdk to build for Qualcomm Linux

sarat576:
Upstream meta-dpdk does not determine the Arm Soc and leaves it to the vendor to define the SoC. The platform external property is  mandatory to be set by the SoC vendor. This change sets the property to the value 'generic'.

### DIFF
--- a/ci/qcom-distro.yml
+++ b/ci/qcom-distro.yml
@@ -44,6 +44,10 @@ repos:
       .:
       meta-tpm:
 
+  meta-dpdk:
+    branch: master
+    url: https://git.yoctoproject.org/meta-dpdk
+
 local_conf_header:
   virtualization:
     SKIP_META_VIRT_SANITY_CHECK = "1"
--- a/recipes-connectivity/dpdk/dpdk_%.bbappend
+++ b/recipes-connectivity/dpdk/dpdk_%.bbappend
@@ -1,0 +1,8 @@
+
+# This image is compatible only with aarch64 (ARMv8)
+COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+SRC_URI += "file://0001-config-arm-Allow-generic-platform.patch"
+
--- a/recipes-connectivity/dpdk/files/0001-config-arm-Allow-generic-platform.patch
+++ b/recipes-connectivity/dpdk/files/0001-config-arm-Allow-generic-platform.patch
@@ -1,0 +1,37 @@
+From 817d0657f35604a42bba1ca4e75cf7f7cb4d10a8 Mon Sep 17 00:00:00 2001
+From: Sarat Addepalli <sarata@qti.qualcomm.com>
+Date: Wed, 4 Mar 2026 10:18:07 +0530
+Subject: [PATCH] config/arm: Allow generic platform
+
+Upstream-Status: Inappropriate [ SoC Specific  ]
+
+Signed-off-by: Sarat Addepalli <sarata@qti.qualcomm.com>
+---
+ config/arm/meson.build | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/config/arm/meson.build b/config/arm/meson.build
+index 523b0fc0ed..d76c15befd 100644
+--- a/config/arm/meson.build
++++ b/config/arm/meson.build
+@@ -848,7 +848,7 @@ if dpdk_conf.get('RTE_ARCH_32')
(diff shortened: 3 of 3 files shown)


recipes-connectivity/dpdk/dpdk_%.bbappend
```diff
@@ -0,0 +1,8 @@
+
+# This image is compatible only with aarch64 (ARMv8)
+COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
+
+SRC_URI += "file://0001-config-arm-Allow-generic-platform.patch"
+
```
ricardosalveti, line 8: meta-dpdk is not even included here in this layer, we don't have the kas fragment, everything is missing.

recipes-connectivity/dpdk/files/0001-config-arm-Allow-generic-platform.patch
```diff
@@ -0,0 +1,37 @@
...
++        soc = meson.get_external_property('platform', 'generic')
+         if soc == ''
+             error('Arm SoC must be specified in the cross file.')
+         endif
+@@ -902,7 +902,7 @@ else
+         endif
+     else
+         # cross build
+-        soc = meson.get_external_property('platform', '')
++        soc = meson.get_external_property('platform', 'generic')
```
ricardosalveti, line 31: @anujm1 is this really the way forward here?

recipes-connectivity/dpdk/dpdk_%.bbappend
sarat576: Missed adding the change. Have enabled layer in latest commit.

ci/qcom-distro.yml
```diff
@@ -44,6 +44,10 @@ repos:
       .:
       meta-tpm:
 
+  meta-dpdk:
+    branch: master
+    url: https://git.yoctoproject.org/meta-dpdk
```
lumag, line 49: Separate commit.

recipes-connectivity/dpdk/files/0001-config-arm-Allow-generic-platform.patch
```diff
@@ -0,0 +1,37 @@
+From 817d0657f35604a42bba1ca4e75cf7f7cb4d10a8 Mon Sep 17 00:00:00 2001
+From: Sarat Addepalli <sarata@qti.qualcomm.com>
+Date: Wed, 4 Mar 2026 10:18:07 +0530
+Subject: [PATCH] config/arm: Allow generic platform
+
+Upstream-Status: Inappropriate [ SoC Specific  ]
```
lumag, line 6: It is not SoC-specfic, there is nothing about the "qcom" in it.

recipes-connectivity/dpdk/dpdk_%.bbappend
```diff
@@ -0,0 +1,8 @@
...
+# This image is compatible only with aarch64 (ARMv8)
```
lumag, line 2: Why?

recipes-connectivity/dpdk/dpdk_%.bbappend
```diff
@@ -0,0 +1,8 @@
...
+COMPATIBLE_MACHINE:aarch64 = "(.*)"
```
lumag, line 3: This will not pass yocto-check-layer.

recipes-connectivity/dpdk/files/0001-config-arm-Allow-generic-platform.patch
anujm1: No. For aarch64 builds, dpdk selects a pre-defined cross file from their repository by looking at a property defined in meson.cross. Since, we don't have that property defined, it fails. Correct solution would be to define that property in meta-dpdk. Something like this should work:

```
do_write_config:append() {
    cat >${WORKDIR}/dpdk.cross <<EOF
[properties]
platform = 'generic'
EOF
}

EXTRA_OEMESON:append:class-target = " --cross-file ${WORKDIR}/dpdk.cross"
```
@sarat576 I've not tested this. So, could you please test and make sure correct compiler tune/arch is used and propose patch for meta-dpdk instead.

anujm1: It should probably be user selectable so, 
```
RTE_PLATFORM ?= generic
```
and then set in cross file.

ci/qcom-distro.yml
sarat576: Will raise a new commit to include this layer

PR #1732, opened 2026-03-10
linux-firmware: fix qcom-kaanapali-audio topology 

quaresmajose:
Remove the empty package ${PN}-qcom-kaanapali-audio-tplg created in b88a23bbedb.
The package is empty and is not used in any image.


vivpuar:
@quaresmajose linux-firmware is merged in master of oe-core

PR #1733, opened 2026-03-10
Add systemd service to format persist partition.

vkraleti:
QTEE requires a valid filesystem on the persist partition mounted at `/var/lib/tee` to store encrypted data. If the partition is unformatted, mounting fails and dependent services such as qtee_supplicant cannot function correctly.

Add a systemd oneshot service that checks for an existing filesystem on the persist partition and formats it as ext4 when missing. The service runs once in early boot before `var-lib-tee.mount` unit, to ensure a valid filesystem is available at `/var/lib/tee` and disables itself after successful execution.

### DIFF
--- a/recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
+++ b/recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+#
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+set -e
+
+DEV=/dev/disk/by-partlabel/persist
+[ -b "$DEV" ] || exit 0
+
+export PATH=/sbin:/usr/sbin
+
+# Check if ext4 filesystem exists
+if [ "$(blkid -o value -s TYPE "$DEV" 2>/dev/null)" != "ext4" ]; then
+    echo "$DEV: no valid ext4 filesystem found, creating ext4"
+    mkfs.ext4 -F "$DEV"
+    exit 0
+fi
+
+# Run a read-only fsck check to detect corruptions
+e2fsck -n "$DEV" >/dev/null 2>&1
+err=$?
+if [ "$err" -gt 4 ]; then
+    echo "$DEV: filesystem corrupted (e2fsck error=$err), recreating"
+    mkfs.ext4 -F "$DEV"
+    exit 0
+else
+    echo "$DEV: filesystem is OK"
+fi
--- a/recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
+++ b/recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
@@ -1,0 +1,20 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+#
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+[Unit]
+Description=Create ext4 filesystem on persist partition if missing
+Requires=dev-disk-by\x2dpartlabel-persist.device
+After=dev-disk-by\x2dpartlabel-persist.device
+ConditionPathExists=/dev/disk/by-partlabel/persist
+Before=var-lib-tee.mount
+Wants=var-lib-tee.mount
+ConditionPathIsMountPoint=!/var/lib/tee
+
+[Service]
+Type=oneshot
+ExecStart=@sbindir@/check-tee-partition-fs.sh
+RemainAfterExit=yes
+
+[Install]
+WantedBy=local-fs-pre.target
--- a/recipes-bsp/partition/mount-tee-partition/var-lib-tee.mount
+++ b/recipes-bsp/partition/mount-tee-partition/var-lib-tee.mount
@@ -5,6 +5,8 @@
 [Unit]
 Description=Mount persist partition at /var/lib/tee
 Before=local-fs.target
+Requires=format-tee-partition.service
+After=format-tee-partition.service
 
 [Mount]
 DirectoryMode=0755
--- a/recipes-bsp/partition/mount-tee-partition_1.0.bb
+++ b/recipes-bsp/partition/mount-tee-partition_1.0.bb
@@ -4,9 +4,13 @@ encryped data and support security functions"
 LICENSE = "BSD-3-Clause-Clear"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/BSD-3-Clause-Clear;md5=7a434440b651f4a472ca93716d01033a"
 
-SRC_URI = "file://var-lib-tee.mount"
+SRC_URI = " \
+    file://var-lib-tee.mount \
+    file://check-tee-partition-fs.sh \
+    file://format-tee-partition.service \
+"
 
-inherit allarch features_check systemd
+inherit features_check systemd
 REQUIRED_DISTRO_FEATURES = "systemd"
 
 INHIBIT_DEFAULT_DEPS = "1"
@@ -17,9 +21,17 @@ do_compile[noexec] = "1"
 
 do_install() {
     install -Dm 0644 ${UNPACKDIR}/var-lib-tee.mount \
-            ${D}${systemd_unitdir}/system/var-lib-tee.mount
+            ${D}${systemd_system_unitdir}/var-lib-tee.mount
+    install -Dm 0755 ${UNPACKDIR}/check-tee-partition-fs.sh \
+            ${D}${sbindir}/check-tee-partition-fs.sh
+    install -Dm 0644 ${UNPACKDIR}/format-tee-partition.service \
+            ${D}${systemd_system_unitdir}/format-tee-partition.service
+    sed -i -e "s,@sbindir@,${sbindir},g" \
+            ${D}${systemd_system_unitdir}/format-tee-partition.service
 }
 
 PACKAGES = "${PN}"
 
-SYSTEMD_SERVICE:${PN} = "var-lib-tee.mount"
+SYSTEMD_SERVICE:${PN} = "var-lib-tee.mount format-tee-partition.service"
+
+RDEPENDS:${PN} += "e2fsprogs-e2fsck e2fsprogs-mke2fs util-linux-blkid"


recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
```diff
@@ -0,0 +1,28 @@
...
+Type=oneshot
+ExecStart=/bin/sh -c '\
+  DEV=/dev/disk/by-partlabel/persist; \
+  if [ -n "$(/sbin/blkid -o value -s TYPE "$DEV" 2>/dev/null || true)" ]; then \
+    echo "$DEV already has a filesystem, skipping mkfs"; \
+    exit 0; \
+  fi; \
+  /sbin/mkfs.ext4 "$DEV" \
+'
+ExecStartPost=/bin/systemctl disable format-tee-partition.service
```
lumag: Please don't disable it. If the filesystem gets corrupted, the unit should be able to recreate it (so maybe you should incorporate e2fsck call in addition or replacing the blkid).

ricardosalveti: Yeah, I would prefer a sequence of fsck + mkfs in case of issues.

vkraleti: Added a shell script that runs e2fsck and conditionally performs mkfs when required.

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
```diff
@@ -0,0 +1,26 @@
...
+    /sbin/mkfs.ext4 -F "$DEV"
+    exit 0
+fi
+
+# Run a read-only fsck check to detect corruptions
+/sbin/e2fsck -n "$DEV" >/dev/null 2>&1
+err=$?
+if [ "$err" -gt 4 ]; then
+    echo "$DEV: filesystem corrupted (e2fsck error=$err), recreating"
+    /sbin/mkfs.ext4 -F "$DEV"
```
anujm1: Wouldn't the recipe need to ensure this and `e2fsck` are always available?

vkraleti: thanks for pointing, added RDEPENDS.

anujm1: Thanks, we missed `blkid`

vkraleti: updated it as well.

recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
```diff
@@ -0,0 +1,20 @@
...
+#
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+
+[Unit]
+Description=Create ext4 filesystem on persist partition if missing
+Requires=dev-disk-by\x2dpartlabel-persist.device
+After=dev-disk-by\x2dpartlabel-persist.device
+ConditionPathExists=/dev/disk/by-partlabel/persist
+Before=var-lib-tee.mount
+Wants=var-lib-tee.mount
```
lumag, line 11: So, is it Before or Wants (= weak After)?

vkraleti: Both, intent is to format first, then mount.

`Before=var-lib-tee.mount` is to make sure the partition is formatted before mounting. 
`Wants=var-lib-tee.mount` ensures ordering even if the mount is pulled in by another unit.

lumag: @vkraleti `Wants` makes systemd  place this unit _after_ the mount unit. This is not what we expect here.

vkraleti: `Wants=` will only bring a requirement dependency but not ordering dependency. Adding _wants_ won't place the unit _after_ the mount unit.  `Before=` / `After=`  are the ones that decides the correct order of startup events.

A reference from Redhat documentation: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/using_systemd_unit_files_to_customize_and_optimize_your_system/assembly_working-with-systemd-unit-files_working-with-systemd#idm140554611903600 

If only `Before=`, is specified, `var-lib-tee.mount` might not start at all. Specifying both `Wants=` & `Before=` guarantees that both units are to be started in the same transaction.

lumag: Thanks. I misunderstood the Wants= clause. However your explanation points out a different issue. You are making `fsck/mkfs` depend on the `mount` unit. However the logic points out that the order of dependencies is different: it is the `mount` that depends on `mkfs`. I.e. if something pulls in the `var-lib-tee.mount` unit, then the `mkfs` unit should also be installed and enabled (because mounting the FS requires it to be in a good shape), however `mkfs` invocation has no other implications.

recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
```diff
@@ -0,0 +1,20 @@
...
+ConditionPathIsMountPoint=!/var/lib/tee
+
+[Service]
+Type=oneshot
+ExecStart=/sbin/check-tee-partition-fs.sh
```
ricardosalveti: We use usrmerge by default with systemd, so this is actually /usr/sbin.

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
```diff
@@ -0,0 +1,26 @@
...
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+set -e
+
+DEV=/dev/disk/by-partlabel/persist
+[ -b "$DEV" ] || exit 0
+
+# Check if a filesystem exists
+if [ -z "$(/sbin/blkid -o value -s TYPE "$DEV" 2>/dev/null || true)" ]; then
+    echo "$DEV: no filesystem found, creating ext4"
+    /sbin/mkfs.ext4 -F "$DEV"
```
ricardosalveti: Using /sbin works fine, but with systemd it is just a link to /usr/sbin.

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
```diff
@@ -0,0 +1,26 @@
...
+    exit 0
+fi
+
+# Run a read-only fsck check to detect corruptions
+/sbin/e2fsck -n "$DEV" >/dev/null 2>&1
+err=$?
+if [ "$err" -gt 4 ]; then
+    echo "$DEV: filesystem corrupted (e2fsck error=$err), recreating"
```
ricardosalveti, line 23: Any corruption will cause the fs to be erased, is that what we want here?

recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
koenkooi: The recipe should do something like `sed -i 's:=/sbin:=${sbindir}:g ${D}/foo.service` in `do_install()`, since we can't know where the DISTRO puts `${sbindir}`.

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
koenkooi: The recipe should do something like `sed -i 's:/sbin:=${sbindir}:g ${D}/foo.sh` in `do_install()`, since we can't know where the DISTRO puts `${sbindir}`.

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
koenkooi: Same ${sbindir} fixup comment here.

recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
vkraleti: Added `Requires=` and `After=` in `var-lib-tee.mount` unit so that whenever this unit get's pulled in, then `format-tee-partition.service` unit also gets installed.

recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
vkraleti: Thanks for pointing, added @sbindir@ placeholder in the unit and populated the same in do_install to mitigate this,

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
vkraleti: Added PATH in the script to search both `/sbin` and  `/usr/sbin` paths for these utilities.

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
vkraleti: Added PATH in the script to search both `/sbin` and  `/usr/sbin` paths for these utilities.

recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
ricardosalveti: @koenkooi systemd now requires usrmerge by default, so under the systemd context it is fine to assume sbin=/usr/sbin

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
ricardosalveti: @harshaldev27 is it ok to just recreate from scratch on errors or should we try fixing the fs first on errors?

recipes-bsp/partition/mount-tee-partition/var-lib-tee.mount
```diff
@@ -5,6 +5,8 @@
 [Unit]
 Description=Mount persist partition at /var/lib/tee
 Before=local-fs.target
+Requires=format-tee-partition.service
+After=format-tee-partition.service
```
ricardosalveti, line 9: This will cause the manual start of this mount unit to also run the fsck/mkfs logic, which is OK I would think but might create unexpected side effects to the user.

recipes-bsp/partition/mount-tee-partition_1.0.bb
```diff
@@ -17,9 +21,17 @@ do_compile[noexec] = "1"
...
 do_install() {
     install -Dm 0644 ${UNPACKDIR}/var-lib-tee.mount \
-            ${D}${systemd_unitdir}/system/var-lib-tee.mount
+            ${D}${systemd_system_unitdir}/var-lib-tee.mount
+    install -Dm 0755 ${UNPACKDIR}/check-tee-partition-fs.sh \
+            ${D}${sbindir}/check-tee-partition-fs.sh
+    install -Dm 0644 ${UNPACKDIR}/format-tee-partition.service \
+            ${D}${systemd_system_unitdir}/format-tee-partition.service
+    sed -i -e "s,@sbindir@,${sbindir},g" \
+            ${D}${systemd_system_unitdir}/format-tee-partition.service
```
ricardosalveti, line 30: As I said, we can assume sbindir with usrmerge (required by systemd), but won't hurt.

recipes-bsp/partition/mount-tee-partition/var-lib-tee.mount
vkraleti: Manual start also needs fs check, otherwise the partition may not be usable if there is any corruption. I couldn't think of any side effects except spending little extra time for integrity check.

recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
koenkooi: @rsalveti I'm more worried about DISTRO.confs doing something like `sbindir=/system-bins` or `sbindir=${TARGET_ARG}-sbin`, not whether it's in /usr or not.

lumag: It would be a very fancy distro, but I'd agree. We need to be using `${sbindir}`.

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
harshaldev27: As per the requirements for persist, it should be as resilient as possible, so it's a good-to-have to attempt to fix the FS on errors. However, I am fine if we have a subsequent patch adding this, anyways we are getting late waiting for this fix to be merged.

recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
vkraleti: Updated code to use `${sbindir}` provided by DISTRO. Please check the latest changes.

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
ricardosalveti: > The recipe should do something like `sed -i 's:/sbin:=${sbindir}:g ${D}/foo.sh` in `do_install()`, since we can't know where the DISTRO puts `${sbindir}`.

Which would probably break systemd :D

recipes-bsp/partition/mount-tee-partition/check-tee-partition-fs.sh
```diff
@@ -0,0 +1,28 @@
...
+# SPDX-License-Identifier: BSD-3-Clause-Clear
+set -e
+
+DEV=/dev/disk/by-partlabel/persist
+[ -b "$DEV" ] || exit 0
+
+export PATH=/sbin:/usr/sbin
+
+# Check if a filesystem exists
+if [ -z "$(blkid -o value -s TYPE "$DEV" 2>/dev/null || true)" ]; then
```
ricardosalveti: This will fail for cases where there is already a valid FS but it is not ext4.

e.g.

```
root@rb3gen2-core-kit:~# blkid -o value -s TYPE /dev/disk/by-partlabel/persist
vfat
```

Was testing the logic for handling bad FS and also using a different FS.

Check if the returned id is 'ext4' and erase if not.

ricardosalveti: Currently a vfat FS there will cause this unit to always fail, which is not good.

vkraleti: None of the devices I tested have a valid partition so thought it is always empty. But yes, there could be other FS types as well. I'll update script to handle those cases as well.

PR #1741, opened 2026-03-11
gst-plugins-imsdk: Add offline IFE and EIS mode flags to qmmfsrc

skomakul:
This patch updates the qmmfsrc component to enable two new build-time
configuration flags:

Feature_offline_IFE_support: Enables offline IFE feature for efficient usage of ISP in multistream usecases.

EIS_MODES_ENABLE: Adds support for enabling EIS (Electronic Image Stabilization) to have the feasibility to enable only for preview or preview and video in multistream usecases.


lumag (changes requested):
> Enables offline IFE for efficient usage of ISP in multicamera
and multistream scenarios.

What is offline IFE?

> Adds support to enable EIS for preview only or for both preview

What is EIS?

PR #1747, opened 2026-03-11
weston-init: Move Qdemo launcher entries to external weston.ini

muthumul:
weston-init: Move Qdemo launcher entries to external weston.ini and append during install

Move Qdemo launcher definitions into a separate weston.ini - qdemo-launcher-weston.ini file and append it during do_install for cleaner separation and easier maintenance. This will enable the Qdemo.png image to be displayed on the wayland display. Upon clicking this icon, gst-gui-launcher-app.py application will run.

### DIFF
--- a/recipes-graphics/wayland/weston-init.bbappend
+++ b/recipes-graphics/wayland/weston-init.bbappend
@@ -5,6 +5,7 @@ DEFAULTBACKEND:qcom ?= "drm"
 SRC_URI:append:qcom = " \
     file://additional-devices.conf \
     file://weston-start.sh \
+    file://weston-qdemo-launcher.ini \
 "
 
 do_compile:append:qcom() {
@@ -20,6 +21,10 @@ do_install:append:qcom() {
     install -d ${D}${bindir}
     install -m 0755 ${WORKDIR}/sources/weston-start.sh \
         ${D}${bindir}/weston-start.sh
+   
+    install -d ${D}${sysconfdir}/xdg/weston
+    echo "" >> ${D}${sysconfdir}/xdg/weston/weston.ini
+    cat ${WORKDIR}/sources/weston-qdemo-launcher.ini >> ${D}${sysconfdir}/xdg/weston/weston.ini
 }
 
 FILES:${PN} += "${systemd_system_unitdir}/weston.service.d/additional-devices.conf"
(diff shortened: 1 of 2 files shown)


recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -20,6 +21,10 @@ do_install:append:qcom() {
     install -d ${D}${bindir}
     install -m 0755 ${WORKDIR}/sources/weston-start.sh \
         ${D}${bindir}/weston-start.sh
+   
+    install -d ${D}${sysconfdir}/xdg/weston
```
ricardosalveti, line 25: No need, this is already created by the main recipe, here we're just appending.

recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -20,6 +21,10 @@ do_install:append:qcom() {
...
+    echo "" >> ${D}${sysconfdir}/xdg/weston/weston.ini
```
ricardosalveti, line 26: Why?

recipes-graphics/wayland/weston-init/weston-qdemo-launcher.ini
```diff
@@ -0,0 +1,8 @@
+[launcher]
+icon=/usr/share/weston/terminal.png
+path=/usr/bin/weston-terminal
+
+[launcher]
+icon=/usr/share/qdemo/Qdemo.png
+path=/usr/bin/Qdemo
+
```
ricardosalveti, line 8: Remove empty line.

recipes-graphics/wayland/weston-init.bbappend
muthumul: For clear separation from the previous lines that already exist in weston.ini. The cat command is appending the launcher code right below the existing code without leaving any space.

recipes-graphics/wayland/weston-init/weston-qdemo-launcher.ini
muthumul: Sure

muthumul:
> From your commit message, you are not moving anything here, you are adding the qdemo entries, so update to that.
> 
> The implementation looks OK, if all we need is to append to weston.ini, but it should be done at meta-qcom-distro level and not here (it has nothing BSP specific).

Hi @ricardosalveti. Given that the sample apps (non-python and python) and qdemo are part of meta-qcom, wouldn't it be better if we keep them in meta-qcom as Qdemo utilizes these sample-apps in the background?

PR #1748, opened 2026-03-11
Monthly cleanup

ndechesne:
The last monthly cleanup job timed out, and failed to delete all files from the cache. a few changes to improve how we manage the sstate cache.

### DIFF
--- a/.github/workflows/monthly.yml
+++ b/.github/workflows/monthly.yml
@@ -8,6 +8,7 @@ jobs:
   sstate-cache-cleanup:
     if: github.repository_owner == 'qualcomm-linux'
     runs-on: [self-hosted, gen-qlnx-prd-u2404-x64-sm-od-ephe]
+    timeout-minutes: 720
     steps:
       - name: Clean up the persistant sstate-cache dir
         run: |


.github/workflows/monthly.yml
```diff
@@ -2,7 +2,7 @@ name: Monthly job
 
 on:
   schedule:
-  - cron: "22 10 2 * *"   # montly job - at "random" time - top of hour can be busy in github
+  - cron: "22 10 5 * *"   # montly job - at "random" time - top of hour can be busy in github
```
ricardosalveti, line 5: Just a heads up that this will have to be merged by quic-yocto-ci, which I have access to, in order for the notifications to go to the user / list.

ricardosalveti: And the reason why nobody else was notified about the failure, was going just to you (https://github.com/qualcomm-linux/meta-qcom/actions/runs/22572160162).

.github/workflows/monthly.yml
```diff
@@ -2,7 +2,7 @@ name: Monthly job
 
 on:
   schedule:
-  - cron: "22 10 2 * *"   # montly job - at "random" time - top of hour can be busy in github
```
lool, line 5: FYI montly

.github/workflows/monthly.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   sstate-cache-cleanup:
     if: github.repository_owner == 'qualcomm-linux'
-    runs-on: [self-hosted, qcom-u2404, amd64-ssd]
+    runs-on: [self-hosted, gen-qlnx-prd-u2404-x64-sm-od-ephe]
```
lool, line 10: perhaps switch runner just for that job has more chances of breaking or getting surprising behavior

out of curiosity, I've checked here:
https://github.com/qualcomm-linux/meta-qcom/actions/metrics/usage?dateRangeType=DATE_RANGE_TYPE_PREVIOUS_MONTH
3070037 minutes in past month (all jobs/runners), so ~51167 hours.

and last monthly were between 2 and 6 hours (aborted because too long); even assuming 12h (new max), that's less than 0.025% of the total hours per month

ricardosalveti: The action will be the same (removing shared files), so it should behave the same way.

lool: Not sure which way you're advocating; I'm saying changing runners for this one action is premature optimization and not worth the risk

ricardosalveti: I'm saying that this new runner should behave the same, since the operation is not bound to the cpu / os  / etc.

lool: It requires 1) IT committing to provide/maintain this runner, 2) this runner staying with the same EFS setup

Anyway, probably not worth our time arguing over this :)

PR #1751, opened 2026-03-12
Add UEFI FMP capsule generation-related recipes

igoropaniuk:
Introduces `edk2-basetools-native` to build the required EDK2 host tools (`GenFf`s, `GenFv`) and f`irmware-qcom-capsule` which fetches capsule scripts from `cbsp-boot-utilities`, consumes boot binaries from virtual/bootbins, and produces a signed FMP capsule. The capsule is wired into the qcomflash image via a new `QCOM_CAPSULE_FIRMWARE` variable following the same pattern as `QCOM_BOOT_FIRMWARE`. 

Test PKI keys are provided vi for development and CI; production builds should override `CAPSULE_ROOT_CER`, `CAPSULE_CERT_PEM`, `CAPSULE_ROOT_PUB`, and `CAPSULE_SUB_PUB` with keys from a secure source.


recipes-devtools/edk2-basetools/edk2-basetools-native_1.0.bb
```diff
@@ -0,0 +1,81 @@
...
+# Only the three sub-directories needed by GenFfs/GenFv are built:
+#   Common, BrotliCompress, GenFfs, GenFv
+# VfrCompile (UEFI-forms compiler) and other BaseTools are intentionally
+# skipped - they are not needed for capsule generation and VfrCompile
+# requires the Pccts/antlr parser-generator which fails in a native build.
+
+DESCRIPTION = "EDK2 BaseTools host utilities (GenFfs, GenFv) for UEFI capsule generation"
+HOMEPAGE    = "https://github.com/tianocore/edk2"
+LICENSE     = "BSD-2-Clause-Patent"
+LIC_FILES_CHKSUM = "file://License.txt;md5=2b415520383f7964e96700ae12b4570a"
```
ricardosalveti: meta-arm provides ed2k-firmware and edk2-basetools-native, might be better to use those recipes and improve them there as needed.

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -0,0 +1,275 @@
...
+LICENSE     = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://uefi_capsule_generation/LICENSE;md5=8e1eb38e3de3966193d29f31f5d7e684"
+
+COMPATIBLE_MACHINE = "qcm6490|qcs9100|qcs8300|qcs615"
+
+# cbsp-boot-utilities - all Python capsule scripts
+SRCREV_cbsp = "58658d656953aeaa01f40b7ea75a65e7125d8488"
+
+SRC_URI = " \
+    git://github.com/quic/cbsp-boot-utilities.git;protocol=https;branch=main;name=cbsp \
```
ricardosalveti: Wonder if better to have a cbsp-boot-utilities recipe instead (with native) here, it might be useful to have it available out of the context of this recipe.

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -0,0 +1,275 @@
...
+# Set XBLCONFIG_DTB to the post-DDR DTB filename that lives inside the boot
+# binaries directory.  When set, the recipe will:
+#   1. dump XBLConfig sections
+#   2. patch QcCapsuleRootCert in the DTB with the converted root cert
+#   3. re-pack the updated DTB back into xbl_config.elf
+#
+# The default values below match the QCM6490 reference design; adjust for
+# other SoCs or board variants.
+XBLCONFIG_DTB         ?= ""
+XBLCONFIG_DTB:qcm6490 ?= "post-ddr-kodiak-1.0.dtb"
```
ricardosalveti: Should this be set at the board level?

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
ricardosalveti: Then we could move this logic into a class.

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -0,0 +1,275 @@
...
+    bbnote "Generate capsule file"
+    python3 GenerateCapsule.py \
+        -e \
+        -j config.json \
+        -o "${PN}.cap" \
+        --capflag PersistAcrossReset \
+        -v
+}
+
+do_install[noexec] = "1"
```
ricardosalveti: Do we want to package and make the capsule available in a package that could later be installed by an image?

ci/test-keys/QcFMPCert.pem
```diff
@@ -0,0 +1,57 @@
+Bag Attributes
```
ricardosalveti, line 1: I would prefer these files to be available under ci or similar, and only really enabled via a kas file that is only used by our CI.

We need to make sure that these keys are not used by the normal developer flow, other wise it will end up being used in production,

classes-recipe/image_types_qcom.bbclass
```diff
@@ -162,6 +164,12 @@ create_qcomflash_pkg() {
         install -m 0644 "${DEPLOY_DIR_IMAGE}/abl2esp-${ABL_SIGNATURE_VERSION}.elf" .
     fi
 
+    # capsule image
+    if [ -n "${QCOM_CAPSULE_FIRMWARE}" ] && \
+            [ -f "${DEPLOY_DIR_IMAGE}/${QCOM_CAPSULE_FIRMWARE}.cap" ]; then
+        install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_CAPSULE_FIRMWARE}.cap" .
```
ricardosalveti, line 179: It is useful to have in deploy, but should't it be also available as part of the rootfs?

igoropaniuk: That approach makes sense in LmP, where OSTree is used and the capsule is delivered through the OSTree sysroot deployment. In this case, however, I’m not sure it applies, as for example in case of LVFS usage it doesn’t make much sense to store the capsule in the root filesystem.

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
igoropaniuk: will drop that

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
igoropaniuk: Copy-paste from the original PR from meta-partner, will drop that

recipes-devtools/edk2-basetools/edk2-basetools-native_1.0.bb
igoropaniuk: yeah, make sense to use existing one

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
quaresmajose: yeah there is a lot of configuration variables here so fits better in new class.

quaresmajose (commented):
With the use of some recipes from meta-arm we should move the news ones who depend on them to dynamic layers bucause meta-arm layer is optional.

classes-recipe/image_types_qcom.bbclass
ricardosalveti: I think it always make sense to package the capsule as well, then we can decide to include or not in the image by simply adding the package to the image.

LVFS will also be available later as well, but we will always have demand for user-signed capsule which won't come via LVFS.

recipes-devtools/edk2-basetools/edk2-basetools-native_1.0.bb
igoropaniuk: added edk2-basetools-native bbappend

ci/test-keys/QcFMPCert.pem
igoropaniuk: done

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
igoropaniuk: moved to a separate recipe

dynamic-layers/meta-arm/recipes-bsp/edk2-basetools/edk2-basetools-native_%.bbappend
```diff
@@ -0,0 +1,65 @@
+# Extend meta-arm's edk2-basetools-native recipe to also build the C host
+# tools (GenFfs, GenFv) required for UEFI FV and capsule generation on
+# Qualcomm platforms.
```
ricardosalveti: Can you try sending this to meta-arm? Explain that we want to generate capsules on meta-qcom.

recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,35 @@
+# Build-host toolchain for UEFI FMP capsule generation on Qualcomm platforms.
```
ricardosalveti: Add DESCRIPTION with this line.

recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,35 @@
...
+# Installs the uefi_capsule_generation/ scripts from cbsp-boot-utilities into
+# the native sysroot (${STAGING_DATADIR_NATIVE}/cbsp-boot-utilities/) and pulls
+# in all other native tools required for capsule generation:
+#   - edk2-basetools-native  : GenFfs, GenFv, GenerateCapsule.py, Common/
+#   - dtc-native             : fdtdump (XBLConfig section dumping)
+#   - python3-pyelftools-native : ELF parsing for xblconfig_parser.py
+#   - python3-dtc-native     : pylibfdt bindings for set_dtb_property.py
```
ricardosalveti: I think we can drop this, or move to the git commit message.

recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,35 @@
...
+
+S = "${UNPACKDIR}/${BPN}-${PV}"
+
+inherit native
+
+DEPENDS = " \
+    edk2-basetools-native \
+    dtc-native \
+    python3-pyelftools-native \
+    python3-dtc-native \
```
ricardosalveti: Sort the list.

recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,35 @@
...
+"
+
+do_configure[noexec] = "1"
+do_compile[noexec]   = "1"
+
+do_install() {
+    install -d "${D}${datadir}/cbsp-boot-utilities"
+    cp -r "${S}/uefi_capsule_generation/." "${D}${datadir}/cbsp-boot-utilities/"
```
ricardosalveti: Can we be a bit more specific and copy only the python and xml files?

recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,35 @@
...
+HOMEPAGE = "https://github.com/quic/cbsp-boot-utilities"
+LICENSE  = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://uefi_capsule_generation/LICENSE;md5=8e1eb38e3de3966193d29f31f5d7e684"
+
+SRC_URI = "git://github.com/quic/cbsp-boot-utilities.git;protocol=https;branch=main"
+SRCREV  = "58658d656953aeaa01f40b7ea75a65e7125d8488"
+
+S = "${UNPACKDIR}/${BPN}-${PV}"
+
+inherit native
```
ricardosalveti: Missing python3native?

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -0,0 +1,294 @@
...
+    # Populate tools provided by edk2-basetools-native
+    # -----------------------------------------------------------------------
+    cp "${STAGING_BINDIR_NATIVE}/GenFfs" "${CAPSULE_DIR}/"
+    cp "${STAGING_BINDIR_NATIVE}/GenFv"  "${CAPSULE_DIR}/"
+    chmod +x "${CAPSULE_DIR}/GenFfs" "${CAPSULE_DIR}/GenFv"
+
+    cp "${STAGING_DATADIR_NATIVE}/edk2-basetools/GenerateCapsule.py" \
+       "${CAPSULE_DIR}/"
+    cp -r "${STAGING_DATADIR_NATIVE}/edk2-basetools/Common" \
+       "${CAPSULE_DIR}/"
```
ricardosalveti: Can't we use all these tools from staging? Why do we need to copy them?

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -0,0 +1,294 @@
...
+
+    # -----------------------------------------------------------------------
+    # FvUpdate.xml selection
+    # SRC_URI:append override replaces the default file from the upstream repo.
+    # -----------------------------------------------------------------------
+    if [ -f "${WORKDIR}/FvUpdate.xml" ]; then
+        cp "${WORKDIR}/FvUpdate.xml" "${CAPSULE_DIR}/FvUpdate.xml"
+        bbnote "Using custom FvUpdate.xml from SRC_URI"
+    else
+        bbnote "Using default FvUpdate.xml from cbsp-boot-utilities"
```
ricardosalveti: I prefer to not have such bbnote calls here, this is just a normal recipe in the end.

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -0,0 +1,294 @@
+# Yocto recipe for UEFI capsule generation on Qualcomm platforms.
+#
+# All Python capsule-generation scripts are provided by cbsp-boot-utilities-native
+# (installed to ${STAGING_DATADIR_NATIVE}/cbsp-boot-utilities/).  The EDK2 host
+# tools (GenFfs, GenFv, GenerateCapsule.py, Common/) are provided by
+# edk2-basetools-native.  Boot firmware binaries are consumed from the
+# firmware-qcom-bootbins deploy directory (virtual/bootbins provider).
```
ricardosalveti: I still thin that most of this should be moved to a class, as we expect customers to set these variables in order for this to be useful.

ricardosalveti: Have a class-recipe and then a simple recipe to produce the capsule. Then customers can also create their own recipes if needed.

dynamic-layers/meta-arm/recipes-bsp/edk2-basetools/edk2-basetools-native_%.bbappend
lumag: +1

dynamic-layers/meta-arm/recipes-bsp/edk2-basetools/edk2-basetools-native_%.bbappend
```diff
@@ -0,0 +1,65 @@
...
+#
+# meta-arm's recipe copies the Python BaseTools to the native sysroot but
+# intentionally skips building C tools.  This bbappend:
```
lumag: This is more of the commit message. Don't repeat yourself

dynamic-layers/meta-arm/recipes-bsp/edk2-basetools/edk2-basetools-native_%.bbappend
```diff
@@ -0,0 +1,65 @@
...
+# meta-arm bumps the edk2 SRCREV.
+#   edk2-stable202502 (b158dad1): brotli f4153a09
+#   edk2-stable202602 (b7a715f7): brotli e230f474
+
+# Re-declare SRC_URI with explicit destsuffix=${BPN}-${PV} on the edk2 entry
+# so that both the source directory and brotli destsuffix are version-agnostic
+# and consistent with BitBake's auto-S convention (${UNPACKDIR}/${BPN}-${PV}).
+# SRCREV_edk2 lazily expands to whatever SRCREV the base recipe sets.
+SRC_URI = " \
+    git://github.com/tianocore/edk2.git;branch=master;protocol=https;name=edk2;destsuffix=${BPN}-${PV} \
```
lumag: This really should not be necessary. We should not be adding destsuffix to the main project's Git URIs, unless really necessary.

dynamic-layers/meta-arm/recipes-bsp/edk2-basetools/edk2-basetools-native_%.bbappend
```diff
@@ -0,0 +1,65 @@
...
+
+do_compile() {
+    BASE_C="${S}/BaseTools/Source/C"
+
+    for d in Common BrotliCompress GenFfs GenFv; do
+        make -C "${BASE_C}/${d}" clean 2>/dev/null || true
+    done
+    rm -f "${BASE_C}/bin/GenFfs" "${BASE_C}/bin/GenFv"
+
+    make -C "${BASE_C}/Common"         -j"${BB_NUMBER_THREADS}"
```
lumag: `oe_runmake` ? This would remove the need for `-j`.

recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,35 @@
...
+
+SUMMARY  = "CBSP boot utilities - UEFI capsule generation scripts (native)"
+HOMEPAGE = "https://github.com/quic/cbsp-boot-utilities"
+LICENSE  = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://uefi_capsule_generation/LICENSE;md5=8e1eb38e3de3966193d29f31f5d7e684"
+
+SRC_URI = "git://github.com/quic/cbsp-boot-utilities.git;protocol=https;branch=main"
+SRCREV  = "58658d656953aeaa01f40b7ea75a65e7125d8488"
+
+S = "${UNPACKDIR}/${BPN}-${PV}"
```
lumag: No need to

recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
lumag: Can we work with the (Qualcomm)  upstream and provide distutils / setupitils integration instead?

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -0,0 +1,294 @@
...
+#
+# ---------------------------------------------------------------------------
+# Integrator checklist
+# ---------------------------------------------------------------------------
+# 1.  Pin SRCREV in cbsp-boot-utilities-native_1.0.bb to the commit you validated.
```
lumag: Huh?

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -0,0 +1,294 @@
...
+# 5.  Adjust CAPSULE_FW_VERSION / CAPSULE_FW_LSV for each OTA release.
+# ---------------------------------------------------------------------------
+
+DESCRIPTION = "UEFI Capsule generation for Qualcomm platforms"
+# This recipe contains no source of its own; all scripts are provided by
+# cbsp-boot-utilities-native (BSD-3-Clause-Clear, license tracked there).
+LICENSE     = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+COMPATIBLE_MACHINE = "qcm6490|qcs9100|qcs8300|qcs615"
```
lumag, line 5: What about qcom-armv8a? Don't limit it too much.

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -0,0 +1,294 @@
...
+#   SRC_URI:append = " file://my-board-FvUpdate.xml;subdir=fvupdate"
+# The recipe detects a custom FvUpdate.xml placed in ${WORKDIR} and uses
+# it in place of the upstream default.
+
+# ---------------------------------------------------------------------------
+# Capsule parameters (override in machine / distro / local.conf)
+# ---------------------------------------------------------------------------
+
+# Firmware version embedded in the capsule header
+CAPSULE_FW_VERSION ?= "0.0.1.2"
```
lumag: Where did 0.0.1.2 come from? Does it make sense to provide default here and for the next variable?

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -53,12 +53,12 @@ CAPSULE_FV_TYPE    ?= "SYS_FW"
...
+# OEM PKI material (must be supplied by the integrator)
 # ---------------------------------------------------------------------------
-# Paths to the PKI artefacts produced by the OpenSSL cert-generation step
-# in ci.yml.  In a product build these come from a secure location outside
-# the source tree (e.g., a secrets manager, a separate signing recipe, or
-# a confidential layer).
+# These variables must be set explicitly - there are no built-in defaults.
+# For CI builds, include ci/capsule-test-keys.yml which sets them to the
+# test keys stored under ci/test-keys/.  For production builds, point them
+# at keys from a secure location (secrets manager, signing recipe, etc.).
```
lumag: This should be squashed to the previous patch instead.

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -171,6 +171,14 @@ do_compile() {
         bbfatal "CAPSULE_CERT_PEM, CAPSULE_ROOT_PUB and CAPSULE_SUB_PUB must all be set for capsule signing"
     fi
 
+    case "${CAPSULE_ROOT_CER}" in
```
lumag: Turn this into the QA check and let it be skipped for the CI builds?

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -29,6 +29,14 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 
 COMPATIBLE_MACHINE = "qcm6490|qcs9100|qcs8300|qcs615"
 
+# Capsule assembly requires edk2-basetools-native (GenFfs, GenFv,
+# GenerateCapsule.py), which is only available when meta-arm is present.
+python() {
+    if 'meta-arm' not in (d.getVar('BBFILE_COLLECTIONS') or '').split():
+        raise bb.parse.SkipRecipe('firmware-qcom-capsule requires meta-arm ' \
+                                  '(provides edk2-basetools-native)')
+}
```
lumag: Just move it to `dynamic-layers/meta-arm`, there is no need to create a non-standard mechanism.

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -35,6 +35,9 @@ python() {
     if 'meta-arm' not in (d.getVar('BBFILE_COLLECTIONS') or '').split():
         raise bb.parse.SkipRecipe('firmware-qcom-capsule requires meta-arm ' \
                                   '(provides edk2-basetools-native)')
+    if not d.getVar('CAPSULE_ROOT_CER'):
```
lumag: This should be squashed into the corresponding patch.

lumag (changes requested):

dynamic-layers/meta-arm/recipes-bsp/edk2-basetools/edk2-basetools-native_%.bbappend
igoropaniuk: sent RFC patch https://lists.yoctoproject.org/g/meta-arm/message/6960

recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
igoropaniuk: `uefi_capsule_generation` doesn't have a proper package structure yet (for example, there is no`pyproject.toml`), it's in my todo list

So we can't use setuptools

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
igoropaniuk: The effective MACHINEOVERRIDES for `rb3gen2-core-kit` is:

`aarch64:armv8-2a:qcom:qcm6490:rb3gen2-core-kit`

We probably should address this first

recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
igoropaniuk: will do

dynamic-layers/meta-arm/recipes-bsp/edk2-basetools/edk2-basetools-native_%.bbappend
```diff
@@ -0,0 +1,32 @@
+SRC_URI += " \
```
ricardosalveti: Include in your commit that you also sent this to meta-arm, with the link to the patch, as it should be reverted once merged there.

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,31 @@
+SUMMARY  = "CBSP boot utilities - UEFI capsule generation scripts (native)"
```
ricardosalveti: Drop extra whitespace.

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,31 @@
...
+DESCRIPTION = "Build-host toolchain for UEFI FMP capsule generation on Qualcomm platforms."
+HOMEPAGE = "https://github.com/quic/cbsp-boot-utilities"
+LICENSE  = "BSD-3-Clause-Clear"
```
ricardosalveti: Same.

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,31 @@
...
+LIC_FILES_CHKSUM = "file://uefi_capsule_generation/LICENSE;md5=8e1eb38e3de3966193d29f31f5d7e684"
+
+SRC_URI = "git://github.com/quic/cbsp-boot-utilities.git;protocol=https;branch=main"
+SRCREV  = "58658d656953aeaa01f40b7ea75a65e7125d8488"
```
ricardosalveti: Same.

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,31 @@
...
+
+do_configure[noexec] = "1"
+do_compile[noexec]   = "1"
+
+do_install() {
+    install -d "${D}${datadir}/cbsp-boot-utilities"
+    cp "${S}/uefi_capsule_generation/"*.py  "${D}${datadir}/cbsp-boot-utilities/"
+    cp "${S}/uefi_capsule_generation/"*.xml "${D}${datadir}/cbsp-boot-utilities/"
+    # FVCreation.py locates GenFfs/GenFv relative to its own __file__, so they
+    # must be installed alongside the Python scripts.
```
ricardosalveti: Can't we fix this upstream? It should try to call both from the system standard path instead (which is exposed as part of sysroot-native).

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -0,0 +1,31 @@
...
+    python3-dtc-native \
+    python3-pyelftools-native \
+"
+
+do_configure[noexec] = "1"
+do_compile[noexec]   = "1"
+
+do_install() {
+    install -d "${D}${datadir}/cbsp-boot-utilities"
+    cp "${S}/uefi_capsule_generation/"*.py  "${D}${datadir}/cbsp-boot-utilities/"
```
ricardosalveti: datadir?

classes-recipe/qcom-capsule.bbclass
```diff
@@ -0,0 +1,222 @@
...
+XBLCONFIG_DTB         ?= ""
+XBLCONFIG_DTB_SECTION ?= ""
+
+# ---------------------------------------------------------------------------
+# Boot binaries location
+# ---------------------------------------------------------------------------
+# FVCreation.py resolves firmware paths using the <InputPath> field in
+# FvUpdate.xml relative to BOOTBINS_DIR.
+# QCOM_BOOT_IMG_SUBDIR / SOC_FAMILY is set per-SoC in the machine include files.
+BOOTBINS_DIR ?= "${DEPLOY_DIR_IMAGE}/${SOC_FAMILY}"
```
ricardosalveti: Should't it just use QCOM_BOOT_IMG_SUBDIR instead of SOC_FAMILY?

ricardosalveti: QCOM_BOOT_FILES_SUBDIR actually.

classes-recipe/qcom-capsule.bbclass
```diff
@@ -0,0 +1,222 @@
...
+# ---------------------------------------------------------------------------
+# To provide a board/project-specific capsule layout, append your file to
+# SRC_URI and name it FvUpdate.xml, e.g. in a .bbappend or local.conf:
+#   SRC_URI:append = " file://my-board-FvUpdate.xml;subdir=fvupdate"
+# The class detects a custom FvUpdate.xml placed in ${WORKDIR} and uses
+# it in place of the upstream default.
+
+inherit python3native deploy
+
+DEPENDS = " \
```
ricardosalveti: += or make it a task dependency of do_compile.

classes-recipe/qcom-capsule.bbclass
```diff
@@ -0,0 +1,222 @@
...
+    skip += (d.getVar('INSANE_SKIP:' + pn) or '').split()
+    if 'test-pki-keys' in skip:
+        return
+    root_cer = d.getVar('CAPSULE_ROOT_CER') or ''
+    if 'test-keys' in root_cer:
+        bb.warn('%s: built with test PKI keys; replace CAPSULE_ROOT_CER, '
+                'CAPSULE_CERT_PEM, CAPSULE_ROOT_PUB and CAPSULE_SUB_PUB '
+                'with production keys before shipping' % pn)
+}
+addtask do_qa_test_pki_keys after do_compile before do_install
```
ricardosalveti: This could probably just be an anonymous python and not a task, as you know this at parse time anyway.

classes-recipe/qcom-capsule.bbclass
```diff
@@ -0,0 +1,222 @@
...
+
+# Ensure boot binaries are deployed before we try to consume them
+do_compile[depends] += "${@'${QCOM_BOOT_FIRMWARE}:do_deploy' if d.getVar('QCOM_BOOT_FIRMWARE') else ''}"
+
+do_compile() {
+    CAPSULE_DIR="${WORKDIR}/capsule_gen"
+    CBSP_SCRIPTS="${STAGING_DATADIR_NATIVE}/cbsp-boot-utilities"
+    EDK2_SCRIPTS="${STAGING_DATADIR_NATIVE}/edk2-basetools"
+    rm -rf "${CAPSULE_DIR}"
+    mkdir -p "${CAPSULE_DIR}"
```
ricardosalveti: We can have:

```
do_compile[dirs] = "${CAPSULE_DIR}"
do_compile[cleandirs] = "${CAPSULE_DIR}"
```

classes-recipe/qcom-capsule.bbclass
```diff
@@ -0,0 +1,222 @@
...
+do_deploy() {
+    install -d "${DEPLOYDIR}"
+    install -m 0644 "${WORKDIR}/capsule_gen/${PN}.cap" "${DEPLOYDIR}/"
+
+    # When XBLConfig was patched with the OEM root cert, deploy the updated
+    # binary under a distinct name to avoid a deploy-manifest conflict with
+    # firmware-qcom-bootbins (which already owns xbl_config.elf).
+    if [ -f "${WORKDIR}/capsule_gen/.xbl_patched" ]; then
+        install -m 0644 "${WORKDIR}/capsule_gen/bootbins/xbl_config.elf" \
+            "${DEPLOYDIR}/xbl_config-patched.elf"
```
ricardosalveti: We should name this differently as patched doesn't really mean much. Maybe -with-oem-cert or similar.

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -0,0 +1,13 @@
+DESCRIPTION = "UEFI FMP capsule for Qualcomm platforms"
+LICENSE     = "MIT"
```
ricardosalveti: Drop extra whitespace.

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
ricardosalveti: At this point it really depends on the availability of our edk2-downstream firmware release, and that might also be machine and not soc specific, but I guess having it soc specific is ok for now.

ci/capsule-test-keys.yml
```diff
@@ -0,0 +1,17 @@
...
+
+header:
+  version: 14
+
+local_conf_header:
+  capsule-test-keys: |
+    CAPSULE_ROOT_CER = "${META_QCOM_DIR}/ci/test-keys/QcFMPRoot.cer"
+    CAPSULE_CERT_PEM = "${META_QCOM_DIR}/ci/test-keys/QcFMPCert.pem"
+    CAPSULE_ROOT_PUB = "${META_QCOM_DIR}/ci/test-keys/QcFMPRoot.pub.pem"
+    CAPSULE_SUB_PUB  = "${META_QCOM_DIR}/ci/test-keys/QcFMPSub.pub.pem"
```
ricardosalveti: Please describe how these keys and certs where generated as part of your commit message.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -6,6 +6,11 @@ inherit image_types
 IMAGE_TYPES += "qcomflash"
 
 QCOM_BOOT_FIRMWARE ?= ""
+QCOM_CAPSULE_FIRMWARE ?= ""
```
ricardosalveti: If we expect customers to create their own capsule recipe, we should probably depend on a virtual provider.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -6,6 +6,11 @@ inherit image_types
...
+
+# When a capsule recipe is set, include its package in the rootfs so the
+# capsule is available on-device (e.g. for fwupd / LVFS delivery).
+IMAGE_INSTALL:append = "${@' ${QCOM_CAPSULE_FIRMWARE}' if d.getVar('QCOM_CAPSULE_FIRMWARE') else ''}"
```
ricardosalveti: This should not be done as part of this class, as it is a decision the user should take. Maybe via MACHINE_EXTRA_RRECOMMENDS, but then it will have to be defined at a common soc/qcom class.

classes-recipe/image_types_qcom.bbclass
```diff
@@ -162,6 +168,12 @@ create_qcomflash_pkg() {
         install -m 0644 "${DEPLOY_DIR_IMAGE}/abl2esp-${ABL_SIGNATURE_VERSION}.elf" .
     fi
 
+    # capsule image
+    if [ -n "${QCOM_CAPSULE_FIRMWARE}" ] && \
```
ricardosalveti, line 177: And what about xbl patched? Should qcomflash pick that one when available as default?

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
igoropaniuk: Created PR to address these issues (and similar one for qcom-ptool)

https://github.com/quic/cbsp-boot-utilities/pull/32

igoropaniuk:
The patch for edk2 recipe landed into meta-arm, so dropped bbappend

https://github.com/jonmason/meta-arm/commit/990f2fa7e854fc9480d5c1e6296b7923c9bbfdb2

ricardosalveti:
> The patch for edk2 recipe landed into meta-arm, so dropped bbappend
> 
> [jonmason/meta-arm@990f2fa](https://github.com/jonmason/meta-arm/commit/990f2fa7e854fc9480d5c1e6296b7923c9bbfdb2)

Did you update this pr? Also, there are still several comments which are not yet solved.

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/files/0001-FVCreation-fall-back-to-PATH-when-locating-GenFfs-Ge.patch
```diff
@@ -0,0 +1,75 @@
...
+Subject: [PATCH] FVCreation: fall back to PATH when locating GenFfs/GenFv
+
+When --edk2-path is not supplied, resolve GenFfs/GenFv via shutil.which
+before falling back to the script directory.  This lets callers rely on
+the tools being present in PATH (e.g. staged in a sysroot's bindir) and
+removes the need to copy the binaries next to FVCreation.py.
+
+Precedence: --edk2-path > PATH > script directory.
+
+Upstream-Status: Submitted [https://github.com/quic/cbsp-boot-utilities]
```
igoropaniuk: I'll post this patch soon and will update the link here

igoropaniuk:
@ricardosalveti just force pushed the updated version
There are still a few unresolved comments, since addressing them depends on landing some fixes in the upstream repo first (https://github.com/quic/cbsp-boot-utilities)

lumag (dismissed):

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/files/0001-FVCreation-fall-back-to-PATH-when-locating-GenFfs-Ge.patch
ricardosalveti: Sure, current status is Pending then.

classes-recipe/qcom-capsule.bbclass
```diff
@@ -0,0 +1,234 @@
+#
+# Copyright (c) 2024 Qualcomm Innovation Center, Inc. All rights reserved.
```
ricardosalveti: Either drop the year (which is fine from legal side) or update to 2026 :-)

classes-recipe/qcom-capsule.bbclass
```diff
@@ -0,0 +1,234 @@
...
+# Firmware volume type label passed to FVCreation.py / UpdateJsonParameters.py
+CAPSULE_FV_TYPE    ?= "SYS_FW"
+# FMP ESRT GUID that identifies this firmware on the target
+CAPSULE_GUID       ?= "6F25BFD2-A165-468B-980F-AC51A0A45C52"
+
+# ---------------------------------------------------------------------------
+# OEM PKI material (must be supplied by the integrator)
+# ---------------------------------------------------------------------------
+# These variables must be set explicitly - there are no built-in defaults.
+# For CI builds, include ci/capsule-test-keys.yml which sets them to the
```
ricardosalveti, line 21: Change commit order, since ci/capsule-test-keys.yml is only added after this class.

classes-recipe/qcom-capsule.bbclass
ricardosalveti: You didn't comment on this issue, opening it again.

classes-recipe/qcom-capsule.bbclass
```diff
@@ -0,0 +1,234 @@
...
+        FVUPDATE_XML="${CBSP_SCRIPTS}/FvUpdate.xml"
+    fi
+
+    # Validate mandatory PKI variables before doing any real work
+    if [ -z "${CAPSULE_ROOT_CER}" ]; then
+        bbfatal "CAPSULE_ROOT_CER must point to the DER-encoded root CA certificate (QcFMPRoot.cer)"
+    fi
+    if [ -z "${CAPSULE_CERT_PEM}" ] || [ -z "${CAPSULE_ROOT_PUB}" ] || [ -z "${CAPSULE_SUB_PUB}" ]; then
+        bbfatal "CAPSULE_CERT_PEM, CAPSULE_ROOT_PUB and CAPSULE_SUB_PUB must all be set for capsule signing"
+    fi
```
ricardosalveti: Can't we check the 4 variables in one go? Otherwise the user will set CAPSULE_ROOT_CER and it will probably fail at the next line.

ricardosalveti (changes requested):

classes-recipe/qcom-capsule.bbclass
```diff
@@ -0,0 +1,234 @@
...
+        -e \
+        -j config.json \
+        -o "${PN}.cap" \
+        --capflag PersistAcrossReset \
+        -v
+}
+
+do_install() {
+    install -d "${D}${datadir}/firmware"
+    install -m 0644 "${CAPSULE_DIR}/${PN}.cap" "${D}${datadir}/firmware/"
```
ricardosalveti: Should we really install the capsule under datadir/firmware? Why not /usr/lib/firmware/efi or similar? Where is this commonly stored with other vendors and implementations?

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
ricardosalveti: Extra whitespace still here.

ci/capsule.yml
```diff
@@ -0,0 +1,16 @@
...
+# Usage: kas build ci/base.yml:ci/<machine>.yml:ci/capsule.yml
+#        kas build ci/base.yml:ci/<machine>.yml:ci/capsule.yml:ci/capsule-test-keys.yml
+
+header:
+  version: 14
+
+local_conf_header:
+  capsule: |
+    PREFERRED_PROVIDER_virtual/qcom-capsule-firmware = "firmware-qcom-capsule"
+    VIRTUAL-RUNTIME_qcom-capsule-firmware = "firmware-qcom-capsule"
```
ricardosalveti, line 16: Split this into a different commit.

conf/machine/include/qcom-common.inc
```diff
@@ -18,6 +18,12 @@ XSERVER ?= " \
 
 PREFERRED_PROVIDER_virtual/kernel ??= "linux-yocto"
 
+# When a UEFI FMP capsule provider is configured (via
+# VIRTUAL-RUNTIME_qcom-capsule-firmware, e.g. from ci/capsule.yml),
+# recommend it into the rootfs so the .cap file is available on-device
+# (e.g. for fwupd / LVFS delivery).
+MACHINE_EXTRA_RRECOMMENDS += "${@d.getVar('VIRTUAL-RUNTIME_qcom-capsule-firmware') or ''}"
```
ricardosalveti: This should also be in a separated commit.

classes-recipe/qcom-capsule.bbclass
igoropaniuk: using `${QCOM_BOOT_FILES_SUBDIR}` now

```
BOOTBINS_DIR ?= "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}"
```

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
igoropaniuk: Will address this in a next PR, when https://github.com/quic/cbsp-boot-utilities is more mature, so we can package it properly and there won't be any need to have all these hacks

ci/capsule.yml
```diff
@@ -0,0 +1,15 @@
...
+#
+# Usage: kas build ci/base.yml:ci/<machine>.yml:ci/capsule.yml
+#        kas build ci/base.yml:ci/<machine>.yml:ci/capsule.yml:ci/capsule-test-keys.yml
+
+header:
+  version: 14
+
+local_conf_header:
+  capsule: |
+    PREFERRED_PROVIDER_virtual/qcom-capsule-firmware = "firmware-qcom-capsule"
```
ricardosalveti, line 15: Please split the inclusion of ci/capsule.yml in a separated commit.

ci/capsule.yml
ricardosalveti: You could squash this into a new commit that adds ci/capsule.yml.

Try to avoid mixing layer changes with ci related changes.

ci/capsule.yml
igoropaniuk: done, thanks!

ci/capsule.yml
ricardosalveti: This wasn't really resolved.

igoropaniuk: got what you meant, fixed

ricardosalveti:
Can you add one extra entry at https://github.com/qualcomm-linux/meta-qcom/blob/master/.github/workflows/build-yocto.yml#L278 so we can build-test this in our CI as well? Fine to build for just one machine for now (and nodistro is probably enough).

conf/layer.conf
```diff
@@ -1,6 +1,10 @@
 # We have a conf and classes directory, add to BBPATH
 BBPATH .= ":${LAYERDIR}"
 
+# Expose the layer root so downstream recipes can reference layer-local files
+# (e.g. ci/test-keys/) without depending on the build-directory layout.
+META_QCOM_DIR := "${LAYERDIR}"
```
quaresmajose: Maybe LAYERDIR_qcom is moore alligned with the variables that already exist
```diff
LAYERDEPENDS_qcom = "core"
+# Expose the layer root so downstream recipes can reference layer-local files
+# (e.g. ci/test-keys/) without depending on the build-directory layout.
+LAYERDIR_qcom =  := "${LAYERDIR}" 
LAYERRECOMMENDS_qcom = "openembedded-layer meta-arm"
LAYERSERIES_COMPAT_qcom = "wrynose"
```

igoropaniuk:
@ricardosalveti done!

classes-recipe/image_types_qcom.bbclass
```diff
@@ -6,6 +6,7 @@ inherit image_types
 IMAGE_TYPES += "qcomflash"
 
 QCOM_BOOT_FIRMWARE ?= ""
+QCOM_CAPSULE_FIRMWARE ?= "${@'virtual/qcom-capsule-firmware' if d.getVar('PREFERRED_PROVIDER_virtual/qcom-capsule-firmware') else ''}"
```
quaresmajose: I belive this can be replaced by:
```
PREFERRED_PROVIDER_virtual/qcom-capsule-firmware ?= ""
QCOM_CAPSULE_FIRMWARE ?= "${PREFERRED_PROVIDER_virtual/qcom-capsule-firmware}"
```

conf/machine/include/qcom-common.inc
quaresmajose: Same here and maybe there is a better palce for VIRTUAL-RUNTIME_qcom-capsule-firmware
```
VIRTUAL-RUNTIME_qcom-capsule-firmware ?= ""
MACHINE_EXTRA_RRECOMMENDS += "${VIRTUAL-RUNTIME_qcom-capsule-firmware}"
```

ricardosalveti (dismissed):

quaresmajose (dismissed):
I don't know if my suggestions will work, but I think they simplified things and removed some inline Python that's always heavy for Bitbake to parse.

.github/workflows/build-yocto.yml
```diff
@@ -283,6 +283,14 @@ jobs:
...
                 dirname: ""
                 yamlfile: ""
+          - machine: rb3gen2-core-kit
+            distro:
+                name: nodistro
+                yamlfile: ':ci/meta-arm.yml:ci/capsule.yml:ci/capsule-test-keys.yml'
+            kernel:
+                type: default
+                dirname: ""
+                yamlfile: ""
```
ricardosalveti, line 293: Looking at the matrix it ended up creating 2 build-pr / rb3gen2-core-kit/nodistro runs, we can probably "abuse" dirname here to set something which makes this unique.

igoropaniuk: fixed

conf/machine/include/qcom-common.inc
igoropaniuk: fixed, thanks!

conf/layer.conf
igoropaniuk: fixed, thanks!

classes-recipe/image_types_qcom.bbclass
igoropaniuk: done

quaresmajose (approved):

lumag (approved):

ricardosalveti (approved):

PR #1763, opened 2026-03-16
ci: create a reusable action for building the build summary

ndechesne:
The builds summary table is generated in more places now, such as meta-qcom-distro. Create a reusable action to avoid duplication.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -263,7 +263,7 @@ jobs:
   publish_summary:
     needs: compile
     if: always()
-    runs-on: [self-hosted, qcom-u2404, amd64]
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
         with:


lumag (dismissed):

ricardosalveti (dismissed):

.github/workflows/build-yocto.yml
```diff
@@ -264,59 +264,8 @@ jobs:
     needs: compile
     runs-on: [self-hosted, qcom-u2404, amd64]
     steps:
-      - name: 'Download build URLs'
-        uses: actions/download-artifact@v7
+      - uses: ./.github/actions/print-build-output
```
mwasilew, line 272: This might not work witout checkout before using the action. See line 119.

mwasilew (changes requested):

.github/workflows/build-yocto.yml
ndechesne: right. fixed now.

PR #1777, opened 2026-03-18
iris-video-dlkm: update iris driver to v1.0.2

venvenk:
This tag v1.0.2 brings below update for Qcom Iris

This release fixes an mmap issue when a client calls reqbuf -> mmap without export_buf. Separate helpers are introduced for direct mmap and export -> mmap paths to correctly handle both use cases.


anujm1:
@venvenk Can you please rebase?

PR #1779, opened 2026-03-18
Add Wrynose compatibility until it lands upstream

quaresmajose:
They will all be reverted as they are integrated into their respective layers.

### DIFF
--- a/0001-meta-arm-layers-update-for-wrynose-release-series.patch
+++ b/0001-meta-arm-layers-update-for-wrynose-release-series.patch
@@ -1,0 +1,53 @@
+From deac07d5f27674d39e9e629ca9fbdc2af2e25fd4 Mon Sep 17 00:00:00 2001
+From: Jose Quaresma <jose.quaresma@oss.qualcomm.com>
+Date: Wed, 18 Mar 2026 12:23:48 +0000
+Subject: [PATCH] layers: update for wrynose release series
+
+Upstream-Status: Submitted [https://lists.yoctoproject.org/g/meta-arm/message/6961]
+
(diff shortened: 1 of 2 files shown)


0001-meta-openembedded-layers-update-for-wrynose-release-series.patch
```diff
@@ -0,0 +1,157 @@
...
+From: Khem Raj <raj.khem@gmail.com>
+Date: Tue, 17 Mar 2026 16:59:34 -0700
+Subject: [PATCH] layers: update for wrynose release series
+
+Drop walnascar from supported release series
+
+Upstream-Status: Pending
+
+Signed-off-by: Khem Raj <raj.khem@gmail.com>
+Signed-off-by: Jose Quaresma <jose.quaresma@oss.qualcomm.com>
```
ricardosalveti, line 11: Where did you get this patch from?

quaresmajose: from main-next branch https://git.openembedded.org/meta-openembedded/commit/?h=master-next&id=d1c00fc7e8e64ebb67ef5cb7d5b0416b0203ebf9. please note this is rebased which will disappear. I looked through the mailinglist but didn't see anything there.

ricardosalveti: Hm, so not really pending, guess more like submitted (khem just pushed directly).

quaresmajose: Yeah it was pushed but I don't have the rigth url link so I chose Pending. I think that one will be the first merged upstream. Therefore, it will be the first to revert.

PR #1792, opened 2026-03-20
android-tools-conf-configfs: add SoC serial to USB product string

vkraleti:
Axiom requires a stable and unique device identifier that is consistent across EDL, fastboot, and ADB modes. The ADB serial number cannot be used for this purpose, as it is not available in EDL or fastboot modes. However, the SoC serial number is accessible in all boot modes.

Including the SoC serial number in the USB gadget product string (strings/0x409/product), allows Axiom to reliably read and use it as a stable identifier for uniquely tracking devices.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf-configfs/qcom/android-gadget-setup.machine
+++ b/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf-configfs/qcom/android-gadget-setup.machine
@@ -1,5 +1,7 @@
 manufacturer=Qualcomm
 model=`hostname`
+[ -e /sys/devices/soc0/soc_id ] && model="${model}($(< /sys/devices/soc0/soc_id))"
+[ -e /sys/devices/soc0/serial_number ] && model="${model}_SN:$(printf '%x' "$(tr -d ' \n' < /sys/devices/soc0/serial_number)")"
 androidserial="$(sed -n -e '/androidboot.serialno/  s/.*androidboot.serialno=\([^ ]*\).*/\1/gp ' /proc/cmdline)"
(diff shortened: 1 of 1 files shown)


lumag (changes requested):
I'm sorry, I don't understand. The serial number matches between fastboot and the ADB. Do you observe a different behaviour?

With this in place the behaviour starts to depend on qcom_socinfo being probed by the time adbd starts or not (which might make a difference for the root-less builds).

Not to mention that this suddenly exports the serial number of the SoC over the USB, possibly exporting the trackable number to the outside world.

dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf-configfs/qcom/android-gadget-setup.machine
```diff
@@ -1,5 +1,7 @@
 manufacturer=Qualcomm
 model=`hostname`
+[ -e /sys/devices/soc0/soc_id ] && model="${model}($(< /sys/devices/soc0/soc_id))"
+[ -e /sys/devices/soc0/serial_number ] && model="${model}_SN:$(printf '%x' "$(tr -d ' \n' < /sys/devices/soc0/serial_number)")"
```
lumag, line 4: S/N is defiintely not a part of the model. Don't hack it.

vkraleti: 'model' is the variable name used for populating `adb/strings/0x409/product`, which is for capturing product information. Link: https://git.openembedded.org/meta-openembedded/tree/meta-oe/recipes-devtools/android-tools/android-tools-conf-configfs/android-gadget-setup#n33 

IMO, SoC Number can be part of product descriptor. If not this, any better variable to capture this info?

lumag (approved):

ricardosalveti (approved):

Issue #1796, opened 2026-03-23
Publish summary executes even if builds are failing


quic-viskuma:
https://github.com/qualcomm-linux/meta-qcom/actions/runs/23424536313/job/68137258105

We rely on successful completion of the publish_summary checkrun as a pre-requisite for triggering subsequent jobs via various automations. This includes tagging workflows on meta-qcom-buildconf and triggering test jobs in Axiom.

Can we revert this - https://github.com/qualcomm-linux/meta-qcom/commit/e9542ed9c8be673d3df3c118bdd4a1eebbc7622d

PR #1800, opened 2026-03-23
ci: add post-compile success job

quic-viskuma:
The publish-summary checkrun was used by various automations to trigger
subsequent CI jobs on build success. This step was updated to always run
regardless of build compilation failures making it an unreliable trigger.
(https://github.com/qualcomm-linux/meta-qcom/commit/e9542ed9c8be673d3df3c118bdd4a1eebbc7622d)

This change adds a success-only job after all compile matrix builds pass.
The resulting check can be reused by downstream workflows.

Fixes: #1796

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -286,3 +286,9 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           pattern: build-url*
 
+  build_successful:
+    needs: compile
+    runs-on: ubuntu-latest
+    steps:
+      - name: Build completed successfully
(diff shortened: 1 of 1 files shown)


.github/workflows/build-yocto.yml
```diff
@@ -268,6 +268,13 @@ jobs:
           cache_dir: ${CACHE_DIR}
           sdk: ${{inputs.sdk}}
 
+  build_successful:
```
lumag: Let it be executed after publish-summary, please.

quic-viskuma: ok

PR #1802, opened 2026-03-23
protobuf-camx: provide 6.33.5

lumag:
meta-oe has upgraded protobuf to 6.33.6, breaking the CamX and sensors binaries. Provide the backport of the older version as temporary measure until CamX and sensors are upgraded to use new protobuf.

### DIFF
--- a/recipes-multimedia/imsdk/gst-plugins-imsdk/0001-gst-plugins-drop-using-imported-include-dirs.patch
+++ b/recipes-multimedia/imsdk/gst-plugins-imsdk/0001-gst-plugins-drop-using-imported-include-dirs.patch
@@ -1,0 +1,911 @@
+From 3f8196eeca8b436c941053e0f7eebba68966d2a4 Mon Sep 17 00:00:00 2001
+From: Dmitry Baryshkov <dmitry.baryshkov@oss.qualcomm.com>
+Date: Wed, 25 Mar 2026 06:55:09 +0200
+Subject: [PATCH] gst-plugins: drop using imported include dirs
+
+Using include dirs imported from other non-installed IMSDK plugins
+result in include path being poisoned with /usr/include/gstreamer-1.0.
+Drop those from the include path.
+
+Upstream-Status: Submitted [https://github.com/qualcomm/gst-plugins-imsdk/pull/8]
+Signed-off-by: Dmitry Baryshkov <dmitry.baryshkov@oss.qualcomm.com>
+---
+ gst-plugin-batch/CMakeLists.txt                    |  1 -
+ gst-plugin-camimgreproc/CMakeLists.txt             |  3 ---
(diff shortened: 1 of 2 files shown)


lumag:
cc @gkhose-qipl @bmasta018 @vsegu-qcom

ci/base.yml
```diff
@@ -14,6 +14,7 @@ repos:
 
   oe-core:
     url: https://github.com/openembedded/openembedded-core
+    commit: 2f41668002846dd1dbcff37dc59a8c8312eb9c57
```
quaresmajose: Thus, we are losing everything that is being integrated. In my opinion, this should be done by reverting the specific pkgconf changes until the affected packages are fixed in our side. It should also be in another PR because this change is quite intrusive.

ricardosalveti: Yeah, this should be done separately, and we should also have an open issue with the build failures caused by the pkgconf updates.

lumag: Updated with the workaround by Ross.

dynamic-layers/qcom-distro/recipes-ml/tflite/tensorflow-lite_%.bbappend
```diff
@@ -0,0 +1,2 @@
+DEPENDS:remove:qcom = "protobuf protobuf-native"
+DEPENDS:append:qcom = "protobuf-camx protobuf-camx-native"
```
ricardosalveti: Missing whitespace with append.

lumag:
> I looked at it briefly. It happens only for imsdk because it relies on uninstalled `.pc` files built by itself. When `PKG_CONFIG_SYSROOT_DIR` is defined, these files won't give sysroot prepended to paths and start pointing to system path. Probably because of [pkgconf/pkgconf#226](https://github.com/pkgconf/pkgconf/pull/226)

Indeed. Thanks for the pointer! I'll implement the fix then.

ci/patches/meta-openembedded/0001-protobuf-disable-ptests-for-now.patch
```diff
@@ -0,0 +1,35 @@
...
+
+oe-core just moved from pkgconfig to pkgconf, which has broken the
+ptest buikd due to how fragile the compilation was.
+
+This will be revisited to build the tests properly, but for now simply
+disable the ptests.
+
+Signed-off-by: Ross Burton <ross.burton@arm.com>
+Signed-off-by: Khem Raj <khem.raj@oss.qualcomm.com>
+Upstream-Status: Submitted [https://lore.kernel.org/r/20260324105318.2150394-1-ross.burton@arm.com/]
```
anujm1: This has been merged now, so can be dropped as well.

lumag: Indeed. It wasn't, when I checked.

PR #1810, opened 2026-03-25
ci: fix template injection warnings

anujm1:
The template in delimiters ${{ ... }} is expanded before workflow execution and can, in case of some events, lead to code injection by the user that triggered the event [1][2].

Use an intermediate environment variable that is set to these values to remediate for all instances flagged by zizmor.

[1] https://securitylab.github.com/resources/github-actions-untrusted-input/
[2] https://docs.zizmor.sh/audits/#template-injection

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -51,11 +51,13 @@ jobs:
         id: pr_comment_prep
         run: |
           echo "## Test run [workflow](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})" > pr-comment.txt
-          echo "## Test jobs for commit ${{ github.event.workflow_run.head_sha }}" >> pr-comment.txt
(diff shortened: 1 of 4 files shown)


.github/workflows/test-pr.yml
```diff
@@ -51,11 +51,13 @@ jobs:
         id: pr_comment_prep
         run: |
           echo "## Test run [workflow](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})" > pr-comment.txt
-          echo "## Test jobs for commit ${{ github.event.workflow_run.head_sha }}" >> pr-comment.txt
```
ndechesne, line 54: can head_sha be considered unsecure? It's not mentioned in https://securitylab.github.com/resources/github-actions-untrusted-input/

anujm1: May be not with the way this is used here. But, I think there's no disadvantage to replacing it, if this works. We would otherwise need to evaluate and ignore the warning etc.

PR #1811, opened 2026-03-25
Enable Hamoa upstream camera DTBO compilation for qcom-next kernel

wenmliu:
Update recipe to pack upstream Camera dtbo into final dtb image of Hamoa evk platform.

### DIFF
--- a/conf/machine/include/fit-dtb-compatible.inc
+++ b/conf/machine/include/fit-dtb-compatible.inc
@@ -7,7 +7,7 @@
 FIT_DTB_COMPATIBLE[apq8016-sbc] = "qcom,apq8016-sbc"
 FIT_DTB_COMPATIBLE[apq8096-db820c] = "qcom,apq8096-sbc"
 FIT_DTB_COMPATIBLE[glymur-crd] = "qcom,glymur-crd"
-FIT_DTB_COMPATIBLE[hamoa-iot-evk] = " \
+FIT_DTB_COMPATIBLE[hamoa-iot-evk+hamoa-iot-evk-camera-imx577] = " \
     qcom,hamoa-evk \
     qcom,hamoav2.1-evk \
 "


conf/machine/iq-x7181-evk.conf
```diff
@@ -10,6 +10,10 @@ KERNEL_DEVICETREE ?= " \
                       qcom/hamoa-iot-evk.dtb \
                       "
 
+KERNEL_DEVICETREE:pn-linux-qcom-next += " \
+                      qcom/hamoa-iot-evk-camera-imx577.dtbo \
+                     "
+
```
ricardosalveti, line 13: Why only linux-qcom-next?

ricardosalveti: We have LINUX_QCOM_KERNEL_DEVICETREE which covers both linux-qcom, linux-qcom-next (and rts).

ricardosalveti: And if available upstream, just use KERNEL_DEVICETREE.

ricardosalveti:
> code has merged in linux-qcom 6.18, but meta-qcom tag not update, after tag update, will remove pn-linux-qcom-next.

Can't we wait for this tag update then? I would prefer having it in both and just changing it once, if possible.

conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -7,6 +7,10 @@
 FIT_DTB_COMPATIBLE[apq8016-sbc] = "qcom,apq8016-sbc"
 FIT_DTB_COMPATIBLE[apq8096-db820c] = "qcom,apq8096-sbc"
 FIT_DTB_COMPATIBLE[glymur-crd] = "qcom,glymur-crd"
+FIT_DTB_COMPATIBLE:pn-linux-yocto[hamoa-iot-evk] = " \
```
lumag: Will UEFI loader cope with two entries in the fit image having the same set of capabilities?

wenmliu: This attempt was made due to an error caused by compiling linux-yocto, and I will request review after I find an appropriate way to solve this problem.

ricardosalveti:
Please order your commits correctly, every commit should be functional, first add the missing dtbo, then change the machine, then the fit class, etc.

PR #1815, opened 2026-03-26
ci: reduce load on pr runs

ricardosalveti:
Reduce the load on our PR runs by reducing the amount of warm_up runs (for better cache usage) and skipping debug and performance builds.

Debug builds are only consumed from nightly runs and performance has no major difference from the normal build, so an extra run to catch possible issues there is not that useful in the end.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -108,22 +108,10 @@ jobs:
             yamlfile: ""
           - name: qcom-distro
             yamlfile: ':ci/qcom-distro-prop-image.yml'
-          - name: qcom-distro-catchall
-            yamlfile: ':ci/qcom-distro-catchall.yml'
-          - name: debug
-            yamlfile: ':ci/qcom-distro-prop-image.yml:ci/debug.yml'
-          - name: performance
-            yamlfile: ':ci/qcom-distro-prop-image.yml:ci/performance.yml'
         kernel:
           - type: default
             dirname: ""
             yamlfile: ""
-          - type: 6.18
-            dirname: "_linux-qcom-6.18"
-            yamlfile: ":ci/linux-qcom-6.18.yml"
-          - type: rt-6.18-distro-kvm
-            dirname: "_linux-qcom-rt-6.18_qcom-distro-kvm"
-            yamlfile: ":ci/linux-qcom-rt-6.18.yml:ci/qcom-distro-kvm.yml"
     name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
     steps:
       - uses: actions/checkout@v6
@@ -158,10 +146,12 @@ jobs:
           - iq-x7181-evk
           - kaanapali-mtp
           - qcm6490-idp
+          - qcom-armv8a
           - qcs615-ride
           - qcs8300-ride-sx
           - qcs9100-ride-sx
           - rb1-core-kit
+          - rb3gen2-core-kit
           - sm8750-mtp
         distro:
           - name: nodistro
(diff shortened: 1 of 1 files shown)


ricardosalveti:
@vkraleti while we have https://github.com/qualcomm-linux/meta-qcom/pull/1418 for introducing weekly, I wanted to first optimize our prs runs by moving a few to nightly, as the current pr load is way too much.

.github/workflows/build-yocto.yml
```diff
@@ -180,6 +174,40 @@ jobs:
...
+            distro:
+                name: qcom-distro
+                yamlfile: ":ci/qcom-distro-prop-image.yml"
+            kernel:
+                type: default
+                dirname: ""
+                yamlfile: ""
         include:
           # Additional builds for specific machines
           - machine: qcom-armv8a
```
shivshah3400, line 201: I see this combination is added as part of "excludes" and is already build in compile_warm_up. Anything listed in exclude will still run if it also appears in include. 
Remove this from includes list.

.github/workflows/build-yocto.yml
```diff
@@ -180,6 +170,40 @@ jobs:
           - type: rt-6.18-distro-kvm
             dirname: "_linux-qcom-rt-6.18_qcom-distro-kvm"
             yamlfile: ":ci/linux-qcom-rt-6.18.yml:ci/qcom-distro-kvm.yml"
+        exclude:
+          # Exclude jobs from compile_warm_up
+          - machine: qcom-armv8a
+            distro:
+                name: nodistro
+                yamlfile: ""
```
ndechesne: you don't need to include the addition property when excluding a specific combination. as long as you give enough attribute to uniquely identify the matrix combination to exclude. so this would be equivalent and much simpler to read/maintain

```
         - machine: qcom-armv8a
            distro:
                name: nodistro
            kernel:
                type: default
          - machine: qcom-armv8a
            distro:
                name: qcom-distro
            kernel:
                type: default
          - machine: rb3gen2-core-kit
            distro:
                name: nodistro
            kernel:
                type: default
          - machine: rb3gen2-core-kit
            distro:
                name: qcom-distro
            kernel:
                type: default          
```

.github/workflows/build-yocto.yml
ricardosalveti: It is actually not included by default there, this extra job is to validate qcom-armv8a with linux-yocto-dev.

.github/workflows/build-yocto.yml
ricardosalveti: Yup, makes sense, updated.

.github/workflows/build-yocto.yml
```diff
@@ -256,11 +260,14 @@ jobs:
                 yamlfile: ""
     name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
     steps:
-      - uses: actions/checkout@v6
+      - name: checkout
+        if: inputs.profile != 'pr' || (matrix.distro.name != 'debug' && matrix.distro.name != 'performance')
```
ndechesne, line 264: can you not use github.event_name  here , instead of making an additional inputs?

ricardosalveti: For pr, yes, but I was thinking that having an extra argument makes it more explicit and also not bound to a potential event rename. We also expect to have a different permutation between nighly / weekly / monthly as well, so that could use the same input variable.

.github/workflows/build-yocto.yml
```diff
@@ -256,11 +260,14 @@ jobs:
                 yamlfile: ""
     name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
```
quaresmajose, line 273: It wouldn't be possible to use an `if` statement here? that way we would only have one.

ricardosalveti: I don't think so, it needs to be in the steps.

quaresmajose: I also didn't find any examples outside of the steps.

ricardosalveti: We could change to have a dynamic matrix in place, but that is a more intrusive change.

PR #1818, opened 2026-03-26
workflows/build-yocto: pass value of cache_dir correctly

anujm1:
Make sure that we actually pass the value of CACHE_DIR from env context to compile action.

Suggested-by: Naveen Kumar <kumarn@qti.qualcomm.com>

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -134,7 +134,7 @@ jobs:
           distro_name: ${{matrix.distro.name}}
           kernel_yaml: ${{matrix.kernel.yamlfile}}
           kernel_dirname: ${{matrix.kernel.dirname}}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-yocto.yml
```diff
@@ -134,7 +134,7 @@ jobs:
           distro_name: ${{matrix.distro.name}}
           kernel_yaml: ${{matrix.kernel.yamlfile}}
           kernel_dirname: ${{matrix.kernel.dirname}}
-          cache_dir: ${CACHE_DIR}
+          cache_dir: ${env.CACHE_DIR}
```
ndechesne: ${{ .. }}

anujm1: :( I looked at the documentation before making the change and still overlooked the brackets. Thanks, fixed now.

PR #1825, opened 2026-03-27
workflows: revert list-jobs action variable fix

mwasilew:
When "input" is changed to "inputs" fixing a typo, the action breaks. Full refactoring is on the way. Leaving the action with a typo as it works.

### DIFF
--- a/.github/actions/list-jobs/action.yml
+++ b/.github/actions/list-jobs/action.yml
@@ -34,7 +34,9 @@ runs:
       id: listjobs
       shell: bash
       run: |
-        JOBFILES=$(find . -name "${{ inputs.prefix }}*.yaml")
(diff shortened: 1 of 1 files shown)


.github/actions/list-jobs/action.yml
```diff
@@ -34,7 +34,10 @@ runs:
       id: listjobs
       shell: bash
       run: |
-        JOBFILES=$(find . -name "${{ inputs.prefix }}*.yaml")
+        # This is intentionally left not using inputs.prefix.
+        # Fixing the typo breaks the action
```
ricardosalveti: There is no typo anymore, this line can be removed.

mwasilew: Fixed

PR #1832, opened 2026-03-27
ci: nightly: temporarily disable building SDK

ricardosalveti:
As discussed at
https://github.com/qualcomm-linux/meta-qcom-distro/pull/237, IMSDK requires proper package (or recipe) split between OSS and proprietary dependencies as the main -dev package (used by the SDK) depends on both, causing qcom-multimedia-image to bring proprietary components.

This revert is temporarily needed to unlock nighly CI runs, until the SDK can be generated for both qcom-multimedia-image and qcom-multimedia-proprietary-image.

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -17,7 +17,7 @@ jobs:
   build-nightly:
     uses: ./.github/workflows/build-yocto.yml
     with:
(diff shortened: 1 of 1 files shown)


vkraleti:
This must be reverted ASAP as QSC-CLI, QLauncher workflows, sample apps development are based on SDK.

PR #1835, opened 2026-03-28
actions/compile: buildcharts and environment changes

quaresmajose:
- actions/compile: process the buildcharts for all builds

Create the buildcharts that are missing in some builds.

- compile/action: prepare all the environment in the same step

Do everything in one step and remove code duplication.
Create the KAS_YAMLS env with the common yaml files for kas.

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -41,62 +41,58 @@ runs:
         echo "KAS_CONTAINER=$KAS_CONTAINER" >> $GITHUB_ENV
         chmod +x $KAS_CONTAINER
 
-    - name: Setup build variables and sstate-cache
+    - name: Setup build variables
       shell: bash
       run: |
         # use a monthly sstate cache folder
         echo "DL_DIR=${INPUTS_CACHE_DIR}/downloads" >> $GITHUB_ENV
         echo "SSTATE_DIR=${INPUTS_CACHE_DIR}/sstate-cache-$(date '+%Y-%m')" >> $GITHUB_ENV
-        echo "KAS_WORK_DIR=$PWD/../kas" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/actions/compile/action.yml
```diff
@@ -41,65 +41,64 @@ runs:
...
         echo "DL_DIR=${INPUTS_CACHE_DIR}/downloads" >> $GITHUB_ENV
         echo "SSTATE_DIR=${INPUTS_CACHE_DIR}/sstate-cache-$(date '+%Y-%m')" >> $GITHUB_ENV
-        echo "KAS_WORK_DIR=$PWD/../kas" >> $GITHUB_ENV
+        echo "KAS_WORK_DIR=${PWD}/../kas" >> $GITHUB_ENV
+        echo "KAS_YAMLS=ci/mirror.yml:ci/${INPUTS_MACHINE}.yml${INPUTS_DISTRO_YAML}${INPUTS_KERNEL_YAML}" >> $GITHUB_ENV
+        echo "INPUTS_CACHE_DIR=${INPUTS_CACHE_DIR}" >> $GITHUB_ENV
+        echo "INPUTS_MACHINE=${INPUTS_MACHINE}" >> $GITHUB_ENV
+        echo "INPUTS_DISTRO_YAML=${INPUTS_DISTRO_YAML}" >> $GITHUB_ENV
+        echo "INPUTS_KERNEL_YAML=${INPUTS_KERNEL_YAML}" >> $GITHUB_ENV
+        echo "INPUTS_SDK=${INPUTS_SDK}" >> $GITHUB_ENV
```
anujm1, line 56: Could you please also run zizmor? Use of `GITHUB_ENV` should be avoided, but it is probably okay if these are not user controllable values. 

If this action is reused outside of this repository, we might not want to trust the inputs ..

PR #1842, opened 2026-03-30
weston-init: add repaint-window=10 to [core] config

yashgupta2126:
Weston's default repaint-window is 7ms. At 60 fps the vsync interval is ~16ms. Under load (multi-surface composition or 4K fullscreen weston-simple-egl), the compositor cannot complete repaints within this window, causing up to 50% FPS drops.

Changing repaint-window to 10ms gives the compositor sufficient time to complete repaints before vblank while retaining a 6ms buffer acquisition window, improving frame delivery consistency without introducing additional frame latency.


quaresmajose:
@yashgupta2126 please submit your proposed change to the oe-core email list.

PR #1849, opened 2026-03-31
Add recipe for QWES service binaries.

maniSankarJ:
Add a recipe to package QWES service components, including the qwesd daemon
 and the qwes_cli native utility. The daemon enables loading the QWES TA and
 initializing the license store when enabled, while the CLI provides command-line 
access to check few use-cases supported by QWES.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
+++ b/dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
@@ -1,0 +1,37 @@
+SUMMARY = "Prebuilt Qualcomm Wireless Edge Services binaries, setup scripts and utility application"
+DESCRIPTION = "Qualcomm Wireless Edge Services provide a suite of features Platform feature management, \
+device attestation and secure provisioning. This recipe includes the daemon and scripts which setup \
+the store and optionally load QcWES TA to provide these features."
+
+LICENSE = "LICENSE.qcom-2"
+LIC_FILES_CHKSUM = "file://usr/share/doc/${BPN}/LICENSE.qcom-2;md5=165287851294f2fb8ac8cbc5e24b02b0"
+
+PBT_BUILD_DATE = "260409"
+
+SRC_URI = "https://softwarecenter.qualcomm.com/nexus/generic/software/chip/component/sec-userspace.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qwes_${PV}_armv8a.tar.gz"
+SRC_URI[sha256sum] = "2fc38e033aa32e59a1749017526d29e9ad71f38f168ff652383eb06a05be3985"
+
+S = "${UNPACKDIR}"
+
+inherit systemd
+
+DEPENDS += "curl minkipc qmi-framework glibc"
+
+# This package is currently only used and tested on ARMv8 (aarch64) machines.
+# Therefore, builds for other architectures are not necessary and are explicitly excluded.
+COMPATIBLE_MACHINE = "^$"
+COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
+SYSTEMD_SERVICE:${PN} = "qwesd.service"
+
+do_install() {
+    install -d ${D}${bindir}
+    install -d ${D}${systemd_system_unitdir}
+    install -d ${D}${docdir}/${BPN}
+
+    # Install binaries
+    install -m 0755 ${S}/usr/bin/* ${D}${bindir}/
+    install -m 0644 ${S}/usr/lib/systemd/system/qwesd.service ${D}${systemd_system_unitdir}/qwesd.service
+    install -m 0644 ${S}/usr/share/doc/${BPN}/NOTICE.txt ${D}${docdir}/${BPN}
+    install -m 0644 ${S}/usr/share/doc/${BPN}/LICENSE.qcom-2 ${D}${docdir}/${BPN}
+}


dynamic-layers/openembedded-layer/recipes-security/qwes/files/LICENSE.qcom-2
quaresmajose, line 1: This license is already on the layer in licenses/LICENSE.qcom-2

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
```diff
@@ -0,0 +1,41 @@
...
+device attestation and secure provisioning. These recipe includes the daemon and scripts which setup \
+the store and optionally load QcWES TA to provide these features."
+
+LICENSE = "LICENSE.qcom-2"
+LIC_FILES_CHKSUM = "file://LICENSE.qcom-2;md5=165287851294f2fb8ac8cbc5e24b02b0"
+
+PBT_BUILD_DATE = "260325"
+
+SRC_URI = "https://softwarecenter.qualcomm.com/nexus/generic/software/chip/component/sec-userspace.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qwes_${PV}_armv8a.tar.gz;name=prebuiltTarball"
+SRC_URI += "file://LICENSE.qcom-2"
```
quaresmajose: This is not needed please remove it and update the LIC_FILES_CHKSUM to use the licenses/LICENSE.qcom-2

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
```diff
@@ -0,0 +1,41 @@
...
+inherit systemd
+
+DEPENDS += "curl minkipc qmi-framework qcbor glib-2.0 libxml2 libtinyxml2 glibc"
+
+# This package is currently only used and tested on ARMv8 (aarch64) machines.
+# Therefore, builds for other architectures are not necessary and are explicitly excluded.
+COMPATIBLE_MACHINE = "^$"
+COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
+FILES:${PN} = "${bindir}/qwesd \
```
quaresmajose: Move FILES to the end of recipe after do_install and also try removing it completely because I don't think it's necessary.

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
```diff
@@ -0,0 +1,41 @@
...
+do_install() {
+    install -d ${D}${bindir}
+    install -d ${D}${systemd_system_unitdir}
+
+    # Install binaries
+    install -m 0755 ${S}/usr/bin/* ${D}${bindir}/
+    install -m 0644 ${S}${systemd_system_unitdir}/qwesd.service ${D}${systemd_system_unitdir}/qwesd.service
+}
+
+SYSTEMD_SERVICE:${PN} = "qwesd.service"
```
quaresmajose: Move this up before do_install

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
```diff
@@ -0,0 +1,41 @@
...
+               ${bindir}/qwes_license_store \
+               ${systemd_system_unitdir}/qwesd.service \
+"
+do_install() {
+    install -d ${D}${bindir}
+    install -d ${D}${systemd_system_unitdir}
+
+    # Install binaries
+    install -m 0755 ${S}/usr/bin/* ${D}${bindir}/
+    install -m 0644 ${S}${systemd_system_unitdir}/qwesd.service ${D}${systemd_system_unitdir}/qwesd.service
```
quaresmajose: The first systemd_system_unitdir in the src should include the service full path. Because this varies, and in other distributions that we don't control, it may have a different value.

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
```diff
@@ -0,0 +1,41 @@
...
+DESCRIPTION = "Qualcomm Wiseless Edge Servcies provide a suit of features Platform feature management, \
+device attestation and secure provisioning. These recipe includes the daemon and scripts which setup \
+the store and optionally load QcWES TA to provide these features."
+
+LICENSE = "LICENSE.qcom-2"
+LIC_FILES_CHKSUM = "file://LICENSE.qcom-2;md5=165287851294f2fb8ac8cbc5e24b02b0"
+
+PBT_BUILD_DATE = "260325"
+
+SRC_URI = "https://softwarecenter.qualcomm.com/nexus/generic/software/chip/component/sec-userspace.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qwes_${PV}_armv8a.tar.gz;name=prebuiltTarball"
```
lumag: Drop name

lumag (changes requested):

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
```diff
@@ -0,0 +1,41 @@
+SUMMARY = "Prebuilt Qualcomm Wiseless Edge Servcies binaries, setup scirpts and utility application"
```
ricardosalveti: Fix your typos.

conf/layer.conf
```diff
@@ -15,6 +15,8 @@ LAYERSERIES_COMPAT_qcom = "wrynose"
 
 LICENSE_PATH += "${LAYERDIR}/licenses"
 
+QCOM_LICENSE_DIR = "${LAYERDIR}/licenses"
```
lumag: No, drop this please.

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
```diff
@@ -0,0 +1,34 @@
+SUMMARY = "Prebuilt Qualcomm Wireless Edge Services binaries, setup scripts and utility application"
+DESCRIPTION = "Qualcomm Wireless Edge Services provide a suite of features Platform feature management, \
+device attestation and secure provisioning. This recipe includes the daemon and scripts which setup \
+the store and optionally load QcWES TA to provide these features."
+
+LICENSE = "LICENSE.qcom-2"
+LIC_FILES_CHKSUM = "file://${QCOM_LICENSE_DIR}/LICENSE.qcom-2;md5=165287851294f2fb8ac8cbc5e24b02b0"
```
lumag: Okay. THis is a definite no-go. Don't checksum some random files. The checksum is verifying that there is no unnoticed changes of the licence. As such, it should be checksumming licence-related files in the source tree. Not something outside of it.

lumag (changes requested):

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
quaresmajose: I have asked to change this because this licence is on the layer and in a previous version it was added. Can LICENSE_PATH not be used here?

However, I also agree that the ideal solution would be to include it along with the code tree.

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
```diff
@@ -0,0 +1,34 @@
...
+PBT_BUILD_DATE = "260325"
+
+SRC_URI = "https://softwarecenter.qualcomm.com/nexus/generic/software/chip/component/sec-userspace.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qwes_${PV}_armv8a.tar.gz"
+SRC_URI[sha256sum] = "174994699c993bc647b8eb0d949f50f69dc532e2b005de4b497a60ecf106494d"
+
+S = "${UNPACKDIR}"
+
+inherit systemd
+
+DEPENDS += "curl minkipc qmi-framework qcbor glib-2.0 libxml2 libtinyxml2 glibc"
```
anujm1: Are these dependencies correct? The binaries need both libxml2 and libtinyxml2? The binaries in package don't seem to be linking against either one of them. The `.pc` in package also seems to be talking about a library but that is not present or installed?

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
lumag: It's not just the ideal solution. It's the only thing that makes sense. If there is no license in the tarball, it must be added.

quaresmajose: But why add it if it's already present in the layer?

lumag: What is the licence for the tarball contents? It is a separate artifact, it can get used by Debian, Fedora or any other distro, it can be downloaded directly. There should be no dependency on meta-qcom.

quaresmajose: _However, I also agree that the ideal solution would be to include it along with the code tree._

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
maniSankarJ: We have a library libqwes. But I am not enabling it now as there are still some dependencies not available on upstream. I missed these in cleanup. I removed them now.

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
maniSankarJ: I have included the license into tarball. Updated recipe to refer the license from it.

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
```diff
@@ -0,0 +1,37 @@
...
+# Therefore, builds for other architectures are not necessary and are explicitly excluded.
+COMPATIBLE_MACHINE = "^$"
+COMPATIBLE_MACHINE:aarch64 = "(.*)"
+
+SYSTEMD_SERVICE:${PN} = "qwesd.service"
+
+do_install() {
+    install -d ${D}${bindir}
+    install -d ${D}${systemd_system_unitdir}
+    install -d ${D}${datadir}/doc/${BPN}
```
anujm1: `${datadir}/doc` -> `${docdir}` here and below.

maniSankarJ: Changed as suggested.

anujm1 (approved):

conf/layer.conf
maniSankarJ: Dropped it.

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
maniSankarJ: Dropped name

quaresmajose (approved):

lumag (changes requested):
> Add license file to the tarball.

What does this mean in the commit message? Are you doing something as a part of the recipe? No.

maniSankarJ:
> > Add license file to the tarball.
> 
> What does this mean in the commit message? Are you doing something as a part of the recipe? No.

The QWES tarball has been updated to include the license file. But, as you pointed the license is not being added by this recipe. So, I removed it now.

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
maniSankarJ: Fixed them

lumag (approved):

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
maniSankarJ: Moved the files.

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.0.bb
maniSankarJ: Added license file in source tarball and referring it from recipe.

PR #1874, opened 2026-04-04
gst-plugins-imsdk: add mosquitto runtime dependency for MQTT adaptor

Raja-Ganapathi-Busam:
The MQTT adaptor uses libmosquitto and loads it at runtime via dlopen(). In addition, the mosquitto broker service is required to execute the supported MQTT use cases.

Since the mosquitto package provides the broker service and has an RDEPENDS on libmosquitto, add a runtime dependency on mosquitto to ensure both the library and service are available at runtime.

### DIFF
--- a/recipes-multimedia/imsdk/gst-plugins-imsdk_0.1.0.bb
+++ b/recipes-multimedia/imsdk/gst-plugins-imsdk_0.1.0.bb
@@ -49,3 +49,7 @@ RDEPENDS:${PN}-qtimltflite += "tensorflow-lite"
 
 # The Smart Video Encoder plugin loads the libVideoCtrl library at runtime using dlopen(). To ensure runtime availability, added runtime dependency on 'smart-venc-ctrl-algo'.
 RDEPENDS:${PN}-qtismartvencbin += "smart-venc-ctrl-algo"
+
+# The MQTT adaptor loads the mosquitto library at runtime using dlopen(). MQTT server also needed to exercise the use-case.
+# To ensure runtime availability, added runtime dependency on 'mosquitto'.
+RDEPENDS:libgstqtimqttadaptor += "mosquitto"


lumag (approved):

recipes-multimedia/imsdk/gst-plugins-imsdk_0.1.0.bb
```diff
@@ -49,3 +49,7 @@ RDEPENDS:${PN}-qtimltflite += "tensorflow-lite"
 
 # The Smart Video Encoder plugin loads the libVideoCtrl library at runtime using dlopen(). To ensure runtime availability, added runtime dependency on 'smart-venc-ctrl-algo'.
 RDEPENDS:${PN}-qtismartvencbin += "smart-venc-ctrl-algo"
+
+# The MQTT adaptor loads the mosquitto library at runtime using dlopen(). MQTT server also needed to exercise the use-case.
+# To ensure runtime availability, added runtime dependency on 'mosquitto'.
+RDEPENDS:libgstqtimqttadaptor += "mosquitto"
```
anujm1, line 55: mosquitto is in meta-networking and meta-qcom doesn't require it. Considering the number of recipes gst-plugins-imsdk requires from meta-oe etc. now, I think this recipe itself should be moved to `dynamic-layers/networking-layer`.

lumag: Well... It's RDEPENDS, maybe exactly for that reason. I would prefer to have a direct linkage instead. 
Instead of  moving to `dynamic-layers/networking-layer`, I would rather declare dependency in the main package and the enable packageconfig separately.

Raja-Ganapathi-Busam: Yes, it's RDEPENDS. It was designed as runtime loading due to some requirements. 

Currently,  we have separate packageconfig opinion for the entire messaging plugin and it's protocol adapters. Separate packageconfig option for each protocol adapter will be covered with the ongoing refactoring efforts. Anyways, RDEPENDS is declared for the MQTT adapter only right now.

CherukuriNaveenQC: @ricardosalveti can you please review this change ?

ricardosalveti (approved):

recipes-multimedia/imsdk/gst-plugins-imsdk_0.1.0.bb
ricardosalveti: I'm fine having the dependency here and the conditional under dynamic-layers.

PR #1899, opened 2026-04-08
image_types_qcom: include soccp.bin and qsahara_device_programmer.xml

vivpuar:
Kaanapali uses qsahara_device_programmer.xml as device programmer while flashing, and soccp.bin file is flashed in soccp_a/b partiiton.
Including these files in the qcomflash directory.

### DIFF
--- a/recipes-bsp/firmware-boot/firmware-qcom-boot-common.inc
+++ b/recipes-bsp/firmware-boot/firmware-qcom-boot-common.inc
@@ -21,6 +21,7 @@ do_deploy() {
     find "${S}" -maxdepth 1 -name '*.lzma' -exec install -m 0644 {} ${DEPLOYDIR}/${QCOM_BOOT_IMG_SUBDIR} \;
     find "${S}" -maxdepth 1 -name '*.mbn' -exec install -m 0644 {} ${DEPLOYDIR}/${QCOM_BOOT_IMG_SUBDIR} \;
     find "${S}" -maxdepth 1 -name '*.melf' -exec install -m 0644 {} ${DEPLOYDIR}/${QCOM_BOOT_IMG_SUBDIR} \;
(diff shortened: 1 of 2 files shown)


lumag (changes requested):
Please don't shout. It's `Kaanapali-MTP`. `Kaanapali` is not an abbreviation to be upper cased.

lumag:
MTP _is_ an abbreviation, so it should be uppercased.

recipes-bsp/firmware-boot/firmware-qcom-boot-common.inc
```diff
@@ -21,6 +21,7 @@ do_deploy() {
     find "${S}" -maxdepth 1 -name '*.lzma' -exec install -m 0644 {} ${DEPLOYDIR}/${QCOM_BOOT_IMG_SUBDIR} \;
     find "${S}" -maxdepth 1 -name '*.mbn' -exec install -m 0644 {} ${DEPLOYDIR}/${QCOM_BOOT_IMG_SUBDIR} \;
     find "${S}" -maxdepth 1 -name '*.melf' -exec install -m 0644 {} ${DEPLOYDIR}/${QCOM_BOOT_IMG_SUBDIR} \;
+    find "${S}" -maxdepth 1 -name '*.xml' -exec install -m 0644 {} ${DEPLOYDIR}/${QCOM_BOOT_IMG_SUBDIR} \;
```
ricardosalveti: Please don't copy all xml here as this will copy all the partitions related files as well, which we don't use or consume here.

vivpuar: Updated the rule so that it installs only qsahara_*.xml file, please review once

lumag (approved):

koenkooi (approved):

ricardosalveti (approved):

PR #1903, opened 2026-04-08
ci/ci: use the same package manager for all builds

quaresmajose:
Currently we are using rpm in qcom-distro and ipk in nodistro.
This causes non-determinism between distributions and has implications in
build performance. Basically, we can't use packages in sstate-cache from
distro1 in distro2, and everything has to be recompiled.

### DIFF
--- a/ci/ci.yml
+++ b/ci/ci.yml
@@ -10,3 +10,4 @@ local_conf_header:
     BB_SIGNATURE_HANDLER = "OEBasicHash"
     FIRMWARE_COMPRESSION:qcom-armv8a = "zst"
     INHERIT:remove = "rm_work"
+    PACKAGE_CLASSES = "package_rpm"


ci/ci.yml
```diff
@@ -8,5 +8,5 @@ header:
 local_conf_header:
   ci: |
```
lumag, line 9: Missing commit message.

ci/ci.yml
```diff
@@ -8,5 +8,7 @@ header:
...
     BB_SIGNATURE_HANDLER = "OEBasicHash"
+    EXTRA_IMAGE_FEATURES += "ptest-pkgs"
```
lumag: The commit message tells, what happens. But it's not written, why do we want it to happen. Please start by describing the issue or the problem you are trying to solve.

ci/ci.yml
```diff
@@ -12,3 +12,4 @@ local_conf_header:
     EXTRA_IMAGE_FEATURES:remove:sota = "ptest-pkgs"
     FIRMWARE_COMPRESSION:qcom-armv8a = "zst"
     INHERIT:remove = "rm_work"
+    PACKAGE_CLASSES = "package_rpm"
```
lumag, line 13: I'd rather not change the nodistro too much, unless it's really required.

lumag (changes requested):
P

koenkooi:
As I said before: we shouldn't be changing DISTRO settings, since that configuration would cease to be that DISTRO. Adding a `nodistro-qcom.conf` variant would be fine, changing `nodistro` without renaming is not.

quaresmajose:
I was wrong in my analysis, and using different packages doesn't impact build performance. The sstate chain stores first the packge output withou any formaing and after that it stores another one already packed with the format used on the distro

do_package_setscene -> do_package_write_[pkg]_setscene

The first one is used on the native build that does not use a package manager and the second one is used for the target and is properly packaged.

PR #1916, opened 2026-04-09
Add devicetree overlays to support QPS615 on Rb3Gen2, IQ-8275 IFP Mezz and IQ-9075 IFP Mezz

ayaan-anwar:
This series adds Device Tree overlay (DTBO) support for the QPS615 PCIe Ethernet switch across three Qualcomm platforms. The QPS615 is a PCIe switch with dual Ethernet ports, present on the IFP Mezzanine board (IQ-9075 EVK and IQ-8275 EVK) and integrated on the RB3Gen2 Core Kit. Its Ethernet functionality is enabled by an out-of-tree kernel driver (qps615-dlkm) that requires platform-specific devicetree properties to access hardware resources.

These are "staging" DTBOs due to their downstream nature, they live in the LINUX_QCOM_KERNEL_DEVICETREE list and are referenced via FIT_DTB_COMPATIBLE overlay combinations in the FIT image.

The changes are 3 pairs - one pair each for Rb3Gen2, IQ-9075 EVK IFP Mezz and IQ-8275 EVK Mezz. For each pair, one commit updates `fit-dtb-compatible.inc` to register the staging overlay combination, and the second adds the DTBO to the machine configuration:

**Exception details:** Until the QPS615 driver is upstreamed to the kernel (third party ETA: end of 2026), an exception has been approved to include the driver + DT overlay (QLIJIRA-99).

### DIFF
--- a/conf/machine/include/fit-dtb-compatible.inc
+++ b/conf/machine/include/fit-dtb-compatible.inc
@@ -39,6 +39,7 @@ FIT_DTB_COMPATIBLE[lemans-evk+lemans-evk-ifp-mezzanine+lemans-evk-staging] = " \
 FIT_DTB_COMPATIBLE[monaco-evk+monaco-evk-camera-imx577] = "qcom,qcs8275-iot"
 FIT_DTB_COMPATIBLE[monaco-evk+monaco-evk-camx] = "qcom,qcs8275-iot-camx"
 FIT_DTB_COMPATIBLE[monaco-evk+monaco-evk-ifp-mezzanine] = "qcom,qcs8275-iot-subtype3"
+FIT_DTB_COMPATIBLE[monaco-evk+monaco-evk-ifp-mezzanine+monaco-evk-staging] = "qcom,qcs8275-iot-subtype3-staging"
 FIT_DTB_COMPATIBLE[purwa-iot-evk] = "qcom,purwa-evk"
 FIT_DTB_COMPATIBLE[qcm6490-idp] = "qcom,qcm6490-idp"
 FIT_DTB_COMPATIBLE[qcom-apq8064-asus-nexus7-flo] = "qcom,apq8064"
--- a/conf/machine/include/fit-dtb-compatible.inc
+++ b/conf/machine/include/fit-dtb-compatible.inc
@@ -56,6 +56,10 @@ FIT_DTB_COMPATIBLE[qcs6490-rb3gen2] = " \
     qcom,qcs5430-iot \
     qcom,qcs6490-iot \
 "
+FIT_DTB_COMPATIBLE[qcs6490-rb3gen2+qcs6490-rb3gen2-staging] = " \
+    qcom,qcs5430-iot-staging \
+    qcom,qcs6490-iot-staging \
+"
 FIT_DTB_COMPATIBLE[qcs6490-rb3gen2+qcs6490-rb3gen2-industrial-mezzanine] = " \
     qcom,qcs5430-iot-subtype9 \
     qcom,qcs6490-iot-subtype9 \
--- a/conf/machine/include/fit-dtb-compatible.inc
+++ b/conf/machine/include/fit-dtb-compatible.inc
@@ -32,6 +32,10 @@ FIT_DTB_COMPATIBLE[lemans-evk+lemans-evk-ifp-mezzanine] = " \
     qcom,qcs9075-iot-subtype1 \
     qcom,qcs9075v2-iot-subtype1 \
 "
+FIT_DTB_COMPATIBLE[lemans-evk+lemans-evk-ifp-mezzanine+lemans-evk-staging] = " \
+    qcom,qcs9075-iot-subtype1-staging \
+    qcom,qcs9075v2-iot-subtype1-staging \
+"
 FIT_DTB_COMPATIBLE[monaco-evk+monaco-evk-camera-imx577] = "qcom,qcs8275-iot"
 FIT_DTB_COMPATIBLE[monaco-evk+monaco-evk-camx] = "qcom,qcs8275-iot-camx"
 FIT_DTB_COMPATIBLE[monaco-evk+monaco-evk-ifp-mezzanine] = "qcom,qcs8275-iot-subtype3"
--- a/conf/machine/iq-8275-evk.conf
+++ b/conf/machine/iq-8275-evk.conf
@@ -15,6 +15,7 @@ KERNEL_DEVICETREE ?= " \
 LINUX_QCOM_KERNEL_DEVICETREE ?= " \
                       qcom/monaco-evk-camx.dtbo \
                       qcom/monaco-evk-ifp-mezzanine.dtbo \
+                      qcom/monaco-evk-staging.dtbo \
                       "
 
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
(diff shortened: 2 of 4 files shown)


conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -46,7 +46,7 @@ FIT_DTB_COMPATIBLE[qcs615-ride] = " \
     qcom,qcs615-adp \
     qcom,qcs615-socv1.1-adp \
 "
-FIT_DTB_COMPATIBLE[qcs6490-rb3gen2] = " \
+FIT_DTB_COMPATIBLE[qcs6490-rb3gen2+qcs6490-rb3gen2-staging] = " \
```
ricardosalveti: Isn't this going to change the default behavior / option?

ayaan-anwar: Yes, this would change the default behavior. The Rb3Gen2 core kit is different from IQ9/IQ8 in that it has the QPS615 as part of the default hardware configuration, instead of a separate mezzanine board.

ricardosalveti: Right, and we should not have a staging dtbo applied by default, this should be a user choice via a custom efi variable that will load such overlay, same way as we do with camx.

How is this going to work with other kernel recipes, such as linux-yocto?

ayaan-anwar: I see your point. We might need to add a separate entry for the staging overlay.
Should I remove rb3gen2 from this PR and work on it separately, i.e. do you think the handling for IQ8/IQ9 mezz is fine?

shashim-quic: > Right, and we should not have a staging dtbo applied by default, this should be a user choice via a custom efi variable that will load such overlay, same way as we do with camx.
> 
> How is this going to work with other kernel recipes, such as linux-yocto?

Here the overlay is just to segregate the (still) downstream change. However the change is corresponding to the fixed board definition. Hence I dont think we should give it as an option to user here.

Once change upstreams, it continues to be part of same fixed board  defn.

ricardosalveti: Staging should be handled the same way for all targets.

And we can't really use it as the default dtb as it won't be compliant, any user trying to run a generic distro (or ACS) after flashing an image which contains the dtb with staging (here it would be the default image) will also be affected by this.

shashim-quic: > Staging should be handled the same way for all targets.
> 
> And we can't really use it as the default dtb as it won't be compliant, any user trying to run a generic distro (or ACS) after flashing an image which contains the dtb with staging (here it would be the default image) will also be affected by this.

I believe most common user expectation will be to get a dtb that corresponds to full board definition (that includes ethernet working). Agree that ACS and other folks who want compliant to upstream will choose without overlay. However comparatively former is more common expectation, the later can always select what they need.
Isn't it ?

Forcing all common usage to boot and then set efi variable to pick the overlay and reboot is not out of the box experience that product is expecting.

ricardosalveti: Something we should discuss, I don't like the idea of using a DT that is not compliant with upstream by default, it will create issues to a class of users and won't send the right message from my perspective.

It would be better if we could add runtime logic that is only available as part of our images to enable the overlay by default (via qcom-distro) and not have it as a BSP decision. Remember that we will have users bringing their own kernel here, and these external kernel trees won't have the overlay by default, which will also cause a build failure at the BSP level as well (since it is enabled by default here).

@ndechesne @lumag @lool ^

lumag: My preference is not a surprise: get corresponding bits at least agreed upon with the upstream, then get them merged here. Otherwise it's not better than any other random downstream overlay.

shashim-quic: > My preference is not a surprise: get corresponding bits at least agreed upon with the upstream, then get them merged here. Otherwise it's not better than any other random downstream overlay.

So your preference is to maintain like a camx ? (same as Ricardo opinion)

lumag: If we can't get those upstream, then yes.

shashim-quic: > If we can't get those upstream, then yes.

These additions are to be upstreamed. The DT changes are held up just because their drivers + dt-bindings are under submission.

lumag: Which means that they can't be upstreamed now.

ayaan-anwar:
I have updated the commits to support enabling the overlay via an efi variable.

A quick test on Rb3Gen2 by writing the string "staging" into the VendorDtbOverlays efi variable shows that the overlay is successfully happening.

ricardosalveti:
`2026-04-15 16:36:18 - INFO     - | make[3]: *** No rule to make target 'arch/arm64/boot/dts/qcom/qcs6490-rb3gen2-staging.dtbo'.  Stop.
`

ayaan-anwar:
> `2026-04-15 16:36:18 - INFO - | make[3]: *** No rule to make target 'arch/arm64/boot/dts/qcom/qcs6490-rb3gen2-staging.dtbo'. Stop. `


This error should go away now that https://github.com/qualcomm-linux/meta-qcom/pull/1956 is merged.

ayaan-anwar:
+@sgaud-quic / @shashim-quic 
I think we need qcom-next to be updated as well.

> 2026-04-16 04:25:11 - INFO     - | make[3]: *** No rule to make target 'arch/arm64/boot/dts/qcom/qcs6490-rb3gen2-staging.dtbo'.  Stop.
> 2026-04-16 04:25:11 - INFO     - | make[2]: *** [/work/build/tmp/work-shared/rb3gen2-core-kit/kernel-source/scripts/Makefile.build:548: arch/arm64/boot/dts/qcom] Error 2
> 2026-04-16 04:25:11 - INFO     - | make[1]: *** [/work/build/tmp/work-shared/rb3gen2-core-kit/kernel-source/Makefile:1557: qcom/qcs6490-rb3gen2-staging.dtbo] Error 2
> 2026-04-16 04:25:11 - INFO     - | make: *** [/work/build/tmp/work-shared/rb3gen2-core-kit/kernel-source/Makefile:248: __sub-make] Error 2
> 2026-04-16 04:25:11 - INFO     - | ERROR: oe_runmake failed
> 2026-04-16 04:25:11 - INFO     - | WARNING: exit code 1 from a shell command.
> 2026-04-16 04:25:11 - INFO     - NOTE: recipe linux-qcom-next-6.19+7.0-rc6+git-r0: task do_compile: Failed

ricardosalveti:
Please rebase.

conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -32,6 +32,10 @@ FIT_DTB_COMPATIBLE[lemans-evk+lemans-evk-ifp-mezzanine] = " \
     qcom,qcs9075-iot-subtype1 \
     qcom,qcs9075v2-iot-subtype1 \
 "
+FIT_DTB_COMPATIBLE[lemans-evk+lemans-evk-ifp-mezzanine+lemans-evk-staging] = " \
+    qcom,qcs9075-iot-subtype1-staging \
+    qcom,qcs9075v2-iot-subtype1-staging \
+"
```
ricardosalveti, line 38: Please change the order of the commits, first add the dtbo into the machine conf, then add the fit-dtb-compatible. Same for other targets.

ayaan-anwar: Done

lumag (approved):

ricardosalveti (approved):

PR #1917, opened 2026-04-09
[DRAFT] meta-qcom: update OSS recipes for sample app packaging

shanmaju:
## Summary
- Remove camera service dependency from the sample app package config
- Add the sample app in the OSS `.bbappend` so it can be built with the OSS image

## Details
This change updates the recipe/package configuration so that the sample app can be built as part of the OSS image without depending on camera service. Camera service support remains available through the proprietary recipe.

## Notes
- Opened as a draft PR for early review
- Validation completed in oss and prop recipies


shanmaju:
issues in PR update

Issue #1709, opened 2026-03-09
Talos ADP IOT Soc stuck in systemd boot


knaveen-qc:
On Talos ADP IOT board, we are seeing bootup stuck in systemd boot.

Lava Jobs:
https://lava.infra.foundries.io/scheduler/job/157083
https://lava.infra.foundries.io/scheduler/job/156510
https://lava.infra.foundries.io/scheduler/job/156727

[[0;32m  OK  [0m] Found device [0;1;39m/dev/tee0[0m.
**[[0;1;31mFAILED[0m] Failed to mount [0;1;39mefi.mount[0m.
See 'systemctl status efi.mount' for details.**

[   11.943123] msm_dsi_phy ae94400.phy: supply vdda not found, using dummy regulator
[   11.951152] msm_dsi ae94000.dsi: supply refgen not found, using dummy regulator
         Starting [0;1;39mLoad/Save RF Kill Switch Status[0m...
[[0;32m  OK  [0m] Finished [0;1;39mVirtual Console Setup[0m.
[[0;32m  OK  [0m] Started [0;1;39mLoad/Save RF Kill Switch Status[0m.
[   12.052185] Bluetooth: hci0: Frame reassembly failed (-84)
[   12.080770] ath10k_pci 0000:01:00.0: Failed to find firmware-N.bin (N between 2 and 6) from ath10k/QCA6174/hw3.0: -2
[   12.091610] ath10k_pci 0000:01:00.0: could not fetch firmware files (-2)
[   12.100962] ath10k_pci 0000:01:00.0: could not probe fw (-2)
[[0;32m  OK  [0m] Finished [0;1;39mRebuild Dynamic Linker Cache[0m.
[   14.049260] Bluetooth: hci0: command 0xfc00 tx timeout
[   14.054680] Bluetooth: hci0: Reading QCA version information failed (-110)
[   14.061824] Bluetooth: hci0: Retry BT power ON:0
[   14.144508] Bluetooth: hci0: Frame reassembly failed (-84)
[[0m[0;31m*     [0m] (1 of 3) A start job is running for…Loader Random Seed (2s / no limit)
M
[K[[0;1;31m*[0m[0;31m*    [0m] (1 of 3) A start job is running for…Loader Random Seed (3s / no limit)
M
[K[[0;31m*[0;1;31m*[0m[0;31m*   [0m] (1 of 3) A start job is running for…Loader Random Seed (3s / no limit)
[...]

PR #1919, opened 2026-04-10
mount-tee-partition: fix early-boot format/mount ordering cycle

### LINKED ISSUE #1938: bluetoothd service is not running causing BT on/off failure
We are seeing BT failure on latest PR builds can you pls help checking the issue

Lava job: [LAVA | Scheduler | Jobs | 175829](https://lava.infra.foundries.io/scheduler/job/175829)

[INFO] 1970-01-01 00:04:14 - ------------------------------------------------------------[INFO] 1970-01-01 00:04:14 - Starting BT_ON_OFF Testcase[INFO] 1970-01-01 00:04:14 - Checking dependency: bluetoothctl[INFO] 1970-01-01 00:04:14 - Checking if bluetoothd is running...[WARN] 1970-01-01 00:04:14 - bluetoothd not running, retrying in 5s...[WARN] 1970-01-01 00:04:19 - bluetoothd not running, retrying in 5s...[WARN] 1970-01-01 00:04:24 - bluetoothd not running, retrying in 5s...[FAIL] 1970-01-01 00:04:29 - Bluetooth daemon not detected after 3 attempts.+ /home/lava-175829/0/tests/9_BT_ON_OFF/Runner/utils/send-to-lava.sh BT_ON_OFF.res<<<LAVA_SIGNAL_TESTCASE TEST_CASE_ID=BT_ON_OFF RESULT=FAIL>>>

Thanks,
Bhargav

ricardosalveti:
Remove mutual Wants/Requires dependencies between
format-tee-partition.service and var-lib-tee.mount and rely on ordering-only logic to enforce format before mount.

Also disable DefaultDependencies for format-tee-partition, as units started from local-fs-pre.target should not implicitly depend on basic.target. The default dependency added an ordering loop involving basic.target, local-fs.target, and var-lib-tee.mount, causing early-boot services to be skipped.

With this change, format is guaranteed to run before the persist partition is mounted, without creating systemd ordering cycles during boot.

### DIFF
--- a/recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
+++ b/recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
@@ -4,11 +4,11 @@
 
 [Unit]
 Description=Create ext4 filesystem on persist partition if missing
+DefaultDependencies=no
 Requires=dev-disk-by\x2dpartlabel-persist.device
 After=dev-disk-by\x2dpartlabel-persist.device
 ConditionPathExists=/dev/disk/by-partlabel/persist
 Before=var-lib-tee.mount
-Wants=var-lib-tee.mount
 ConditionPathIsMountPoint=!/var/lib/tee
 
 [Service]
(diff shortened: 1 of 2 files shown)


lumag (approved):

koenkooi (approved):

vkraleti:
I tested multiple back to back reboots on IQ-8075-EVK with this change and didn't observe any boot issues. Hence approving. The problem reported in https://github.com/qualcomm-linux/meta-qcom/issues/1709  due to /efi not mounting is still seen ocationally.

vkraleti (approved):

PR #1920, opened 2026-04-10
ci: make PR origin available for forked pull request workflows

rishi-external:
Problem: Downstream automation needs to determine which pull request
triggered a CI workflow run so that validation results can be associated
with the correct submission and used to drive subsequent CR state
transitions. This requires access to the source PR, including repository
and branch information. Without this data, workflow runs cannot be
reliably linked to the pull request that caused them, preventing correct
CR state updates.

For pull requests originating from forks, workflow completion events do
not consistently expose enough information to establish this association.

Approaches considered: Several fields provided by the GitHub workflow run
API appear suitable but cannot be relied on as a stable solution.

The .pull_requests[] field is not consistently populated for workflows
triggered from forked repositories.

The referenced_workflows[].ref field is optional and depends on workflow
topology. It only resolves to PR-style references when reusable workflows
are present and reside in the same repository. If reusable workflows are
moved to a shared repository, or if multiple reusable workflows are involved,
the reference resolves to a branch or becomes ambiguous.

As a result, none of the available metadata fields provide a reliable way to
determine the originating pull request for forked workflows across different
CI configurations.

Solution: To ensure reliable PR identification, the required PR context is
made available explicitly at workflow runtime. This provides a stable and
predictable source of PR origin information, independent of repository
layout, reusable workflow structure, or future CI refactoring.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -12,6 +12,7 @@ permissions:
   pull-requests: write
   contents: read
   packages: read
+  actions: read
 
 jobs:
   event-file:
@@ -23,6 +24,15 @@ jobs:
       with:
         name: Event File
         path: ${{ github.event_path }}
+        
+  pr-to-workflow-dependency:
+    name: "PR To Workflow Dependency"
+    uses: qualcomm/qcom-reusable-workflows/.github/workflows/qswat-reusable-pr-to-workflow-dependency.yml@qswat-release-v1
+    with:
+      workflow_id: ${{ github.run_id }}
+      run_attempt: ${{ github.run_attempt }}
+      pr_number: ${{ github.event.pull_request.number }}
+      
   build-pr:
     uses: ./.github/workflows/build-yocto.yml
     with:


lumag (changes requested):
The commit message is impossible to follow or understand. The commit subject doesn't follow the established practice. 
Also, why would the team change any CRs based on the PR workflows?

ricardosalveti:
> Also, why would the team change any CRs based on the PR workflows?

They are tracking component specific changes and verifying from the submitter pull-request if these components are being updated, and to which revision, as with that information they can automate the process of moving the CR state (e.g. change proposed, in-flight).

But commit message here needs improvement, focus on explaining that by enabling this workflow we can track the CR state internally as PRs are proposed to meta-qcom.

lumag:
After checking the action runs, I can see a reference to the PR. See https://api.github.com/repos/qualcomm-linux/meta-qcom/actions/runs/24253979371, `referenced_workflow.ref`. Is it not enough for identifying the source? Or am I missing something?

ricardosalveti:
> After checking the action runs, I can see a reference to the PR. See https://api.github.com/repos/qualcomm-linux/meta-qcom/actions/runs/24253979371, `referenced_workflow.ref`. Is it not enough for identifying the source? Or am I missing something?

IIRC the details of the origin of the pull request is not fully there, and they want to receive the repo and branch name so they can evaluate the content and see if there are relevant changes, but @rishi-external would know better.

rishi-external:
> > After checking the action runs, I can see a reference to the PR. See https://api.github.com/repos/qualcomm-linux/meta-qcom/actions/runs/24253979371, `referenced_workflow.ref`. Is it not enough for identifying the source? Or am I missing something?
> 
> IIRC the details of the origin of the pull request is not fully there, and they want to receive the repo and branch name so they can evaluate the content and see if there are relevant changes, but @rishi-external would know better.

Hay @ricardosalveti, thanks for the follow-up comments

Hay @lumag,  
**See https://api.github.com/repos/qualcomm-linux/meta-qcom/actions/runs/24253979371, referenced_workflow.ref. Is it not enough for identifying the source? Or am I missing something?**
For this we can get the details only if the PR is created from a head branch from the same repo

but, in case if the PR is created to this repo from any forked PR, there is a limitation from GitHub where we cannot extract PR details based on the workflow number, and there is no other way from the GitHub were we can get the exact PR details on which a particular workflow ran in case of forked repos..!
image

So here we are injected a dedicated workflow from the reusable workflows, and at the end we will extract the details in the form of artifacts..
example:  https://github.com/qualcomm-linux/meta-qcom/actions/runs/24228240239
image

lumag:
@rishi-external but isn't the following information enough? Then you can use GH API to extract information about the PR. 

```
referenced_workflows": [
    {
      "path": "qualcomm-linux/meta-qcom/.github/workflows/build-yocto.yml@446e01f643965b49fd04425e72d548684f13120f",
      "sha": "446e01f643965b49fd04425e72d548684f13120f",
      "ref": "refs/pull/1928/merge"
    }
 ],
```

rishi-external:
> referenced_workflows": [
>     {
>       "path": "qualcomm-linux/meta-qcom/.github/workflows/build-yocto.yml@446e01f643965b49fd04425e72d548684f13120f",
>       "sha": "446e01f643965b49fd04425e72d548684f13120f",
>       "ref": "refs/pull/1928/merge"
>     }
>  ],

@lumag , you are correct but, this parameter in the workflow details is optional parameter and will behave differently in different cases.

case 1: this we will only have in the workflow details, if there is any reusable workflows in the main workflow, right now we have this **"uses: ./.github/workflows/build-yocto.yml"**, due to which the details are there, and if in future if the flow gets change, it will be a problem again for the CR transition flow

case 2: and even though there is a re-usable workflow(build-yocto.yml), that workflow(build-yocto.yml, here this is in the same repo) should be inside the same repo, then only we can get this PR details, if this reusable worklfow(build-yocto.yml) is moved to any common workflows repo and referring to that workflow then we will not get those details.
instead of **"ref": "refs/pull/143/merge",** we will get **"ref": "refs/heads/reusabkeworkflows"**

case 3: and in few cases we will have multiple entries in the referenced_workflows, it's all depends on the number of reusable workflows we are using (here this is not a big problem, almost all the case we will have the same PR noumber)

So, by considering all the above possible scenarios, we opted this robust approach where in any case by including this reusable workflow we can get all the details and if there is a need of more data, we can always enhance this reusable workflow.

lumag:
Please work on improving your subject and the commit message. For the established style please see the git log.
Please start your commit message by describing the problem, then continue by explaining why other approaches are not working (`.pull_requests[]`, `.referenced_workflows[].ref`). Don't describe patch contents and especially not in the "we are doing" style.

rishi-external:
> Please work on improving your subject and the commit message. For the established style please see the git log. Please start your commit message by describing the problem, then continue by explaining why other approaches are not working (`.pull_requests[]`, `.referenced_workflows[].ref`). Don't describe patch contents and especially not in the "we are doing" style.

hay @lumag , I have added the detailed commit message, could you please check now

lumag:
@rishi-external missing SoB.

lumag (commented):
You start by describing GitHub limitations:

> CI workflows triggered on pull request events do not consistently expose
enough information to uniquely identify the originating pull request when
the PR is created from a fork.

Please instead start by describing the problem that you are solving. Why do you need the data about the PR.

rishi-external:
> Please instead start by describing the problem that you are solving. Why do you need the data about the PR.

@lumag , I did the changes, hope everything is fine..! Could you please review once

lumag (approved):

koenkooi:
Can you fix the title of this PR as well? It will end up in the merge commit.

rishi-external:
> Can you fix the title of this PR as well? It will end up in the merge commit.

Done..! Please check now

ricardosalveti (approved):

PR #1926, opened 2026-04-10
Pin upstream layer commits for RC3 preparation

vkraleti:
In preparation for the Qualcomm Linux 2.0 RC3 release, add ci/base.lock.yml
file to stop tracking the moving tips of upstream meta layer repositories.

The lock file pins all external layers to fixed commit revisions
corresponding to the Yocto 6.0 (M3) release. This ensures reproducible
CI builds and mitigates unexpected breakages caused by upstream changes
during the release phase.

### DIFF
--- a/ci/base.lock.yml
+++ b/ci/base.lock.yml
@@ -1,0 +1,24 @@
+header:
+    version: 14
+overrides:
+    repos:
+        oe-core:
+            commit: f235ffc43c7783a84e1ac0dce8489fcba8354e12
+        bitbake:
+            commit: 5d722b5d65e4eef7befe6376983385421e993f86
+        meta-arm:
+            commit: d987a5945802dcbbc06be91a0d34f00431ea840e
+        meta-qcom-distro:
+            commit: dc4051f119c08e081dfa932567ee6d88f32d1c54
+        meta-openembedded:
+            commit: 96a803a50d55a455b0584d8a81168351d0f8c697
+        meta-virtualization:
+            commit: 4e6c583591c1da7e898254dd33eca5cc04c739a9
+        meta-audioreach:
+            commit: 75402c8d1ea1203dcf3a63792161f7d5d93946bb
+        meta-selinux:
+            commit: 4904b87b12c51a68efc5b9cd45c7b519fdc48af1
+        meta-updater:
+            commit: bcea37a06cb8b17ddcd7f3209b4a3c5643fee4d5
+        meta-security:
+            commit: 8028c573db6923525c2918724f2bd36d4a420e0b
--- a/ci/meta-arm.yml
+++ b/ci/meta-arm.yml
@@ -4,6 +4,7 @@ header:
 repos:
   meta-arm:
     url: https://git.yoctoproject.org/meta-arm
+    branch: master
     layers:
       meta-arm:
       meta-arm-toolchain:
--- a/ci/qcom-distro.yml
+++ b/ci/qcom-distro.yml
@@ -12,6 +12,7 @@ repos:
 
   meta-openembedded:
     url: https://github.com/openembedded/meta-openembedded
+    branch: master
     layers:
       meta-filesystems:
       meta-gnome:


lumag (commented):
Please expand the commit message describing the way to update those locked commit ids (if there is a good way to do so).

lumag:
Also, I think this PR should be targeting the wrynose branch rather than the master branch.

ricardosalveti:
Can't we have a separated base.yml.lock file instead of changing the main file? Not sure if this would work without also changing meta-qcom-distro, but something to check.

ricardosalveti:
> Also, I think this PR should be targeting the wrynose branch rather than the master branch.

I would prefer to stay on master until oe-core also branches out, as we would have to start merging in master and then backporting to wrynose, which will impact the workflow. Can we start working on wrynose after rc3?

quaresmajose:
I also think it would be preferable to have base.yml.lock and update it as needed. That way we also don't pollute the base.yml history and when we no longer need it, we delete it.

ricardosalveti:
Tested the following just in meta-qcom and validated that it is also consumed by meta-qcom-distro and meta-qcom-3rdparty, so we can have it all in one single place, which is nice:

```
commit e360455fb2727eb5f94ff8b5ed23d99fbd539031 (HEAD -> gst-dev)
Author: Ricardo Salveti <ricardo.salveti@oss.qualcomm.com>
Date:   Fri Apr 10 21:00:44 2026 -0300

    ci: add base.lock.yml for RC3

    As part of the Qualcomm Linux 2.0 RC3 preparation, add base.lock.yml in
    order to lock all the repositories used by this layer until the
    milestone [1] gets completed.

    Generated with:
    $ kas lock --update ci/base.yml:ci/meta-arm.yml:ci/qcom-distro.yml

    [2]
    https://github.com/qualcomm-linux/meta-qcom/milestone/3

    Signed-off-by: Ricardo Salveti <ricardo.salveti@oss.qualcomm.com>

diff --git a/ci/base.lock.yml b/ci/base.lock.yml
new file mode 100644
index 00000000..e5309d5f
--- /dev/null
+++ b/ci/base.lock.yml
@@ -0,0 +1,24 @@
+header:
+    version: 14
+overrides:
+    repos:
+        oe-core:
+            commit: 5a60dca0fb81695e26149e45aad9a9eee7c955ec
+        bitbake:
+            commit: 5d722b5d65e4eef7befe6376983385421e993f86
+        meta-arm:
+            commit: d987a5945802dcbbc06be91a0d34f00431ea840e
+        meta-qcom-distro:
+            commit: fde6b0a7113941bc78a45b7fa618e837ce9f13a4
+        meta-openembedded:
+            commit: d793c367e067c49956d38caf6eb84cb112c9c9b7
+        meta-virtualization:
+            commit: 4dddf149832ef19233c13af31987bcd6b1faf4cd
+        meta-audioreach:
+            commit: 75402c8d1ea1203dcf3a63792161f7d5d93946bb
+        meta-selinux:
+            commit: 4904b87b12c51a68efc5b9cd45c7b519fdc48af1
+        meta-updater:
+            commit: bcea37a06cb8b17ddcd7f3209b4a3c5643fee4d5
+        meta-security:
+            commit: 8028c573db6923525c2918724f2bd36d4a420e0b
```

Then we can just remove this file after RC3 gets released.

lumag:
@rsalveti but then it's also too easy to miss the layer without a lock. E.g. if we merge meta-dpdk addition, it won't have a lock (which easily can get unnoticed).

quaresmajose:
It's a detail to keep in mind that whenever a new layer is added while the lock is active, the lock will have to be updated to include the layer. Given that the lock's presence is short, I don't see this as anything disturbing.

lumag:
> It's a detail to keep in mind that whenever a new layer is added while the lock is active, the lock will have to be updated to include the layer. Given that the lock's presence is short, I don't see this as anything disturbing.

Lock presense will be forever for the LTS branch.

ricardosalveti:
Yeah, adding a layer is a major change, while merges can happen without reflecting on the lock file, I would also say this is a minor issue for the RC3 cycle, as we expect to drop this lock by the end of the week.

For LTS we don't really expect to add new layers, if done, it would require a major justification for doing so.

At least with a separated lock the git history of the fragments will be clean.

lumag:
> as we expect to drop this lock by the end of the week.

Why? I'd definitely keep it and train updating the lock files via pull requests.

lumag:
> Error: Unable to download artifact(s): Artifact not found for name: kas-lockfile

So, we need to fix the workflow too.

ci/base.yml
```diff
@@ -2,6 +2,8 @@
 
 header:
   version: 14
+  includes:
+  - ci/base.lock.yml
```
kraj: lock files corresponding to manifests are automatically read by kas. So this include might be redundant. Please validate.

ricardosalveti: Yeah, this include is not required.

vkraleti: base.lock.yml only applies to projects listed in base.yml. Additional projects introduced via meta-arm.yml and qcom-distro.yml will not be covered by base.lock.yml. For those, we need separate lock files or use a common lock file (may be name it something like upstream.lock.yml) and include in base.yml.

vkraleti: Please ignore above, single base.lock.yml worked for all repos. Updated the commit.

quaresmajose (approved):

koenkooi (approved):

PR #1929, opened 2026-04-11
image_types_qcom.bbclass: include xbl_config_devprg.elf for Kaanapali-MTP

vivpuar:
[PR#1191](https://github.com/qualcomm-linux/meta-qcom/pull/1191) adds KVM support which excludes **xbl_config\*.elf** files and installs xbl_config.elf or xbl_config_kvm.elf based on the kvm DISTRO_FEATURE.

xbl_config_devprg.elf is required by Kaanapali-MTP while flashing, so including it in the qcomflash directory.

### DIFF
--- a/classes-recipe/image_types_qcom.bbclass
+++ b/classes-recipe/image_types_qcom.bbclass
@@ -102,7 +102,8 @@ create_qcomflash_pkg() {
                 -name 'logfs_*.bin' -o \
                 -name 'qsahara_*.xml' -o \
                 -name 'sec.dat' -o \
-                -name 'soccp*.bin'` ; do
+                -name 'soccp*.bin' -o \
+                -name 'xbl_config_devprg.elf'` ; do
             install -m 0644 ${bfw} .
         done
 


lumag (changes requested):
Assuming it's another fancy name for the firehose flasher, can we settle on one name for those binaries? Or at least one pattern?

vivpuar:
> Assuming it's another fancy name for the firehose flasher, can we settle on one name for those binaries? Or at least one pattern?

qsahara_device_programmer.xml is the firehose used, and it is having dependency on xbl_config_devprg.elf. Tried removing xbl_config_devprg.elf but it's giving below error.

```
STATUS	Build loading completed on the device - Qualcomm USB Composite Device 90DB (0015)
STATUS	-----------------------------------------------------------------------------------
Status: FAILURE
Error: Sahara image Id: 38 could not be found in the build path
Suggestion: Please verify the sahara image from build path. If the issue persists, contact below meta build or target for assistance
POC: Target Team
```

@apaigava Can you please check on this and comment?

apaigava:
@vivpuar qsahara_device_programmer.xml lists multiple files. Sahara tries to load all files, including xbl_config_devprg.elf. Since this file is listed but not present, the Sahara image load fails and throws the error

Each target has its own dependencies so the naming and binaries can vary. To avoid any issues it would be best to keep everything as it is for now

lumag (approved):

koenkooi (approved):

lumag:
Please rebase on the tip to solve CI issues.

ricardosalveti (approved):

PR #1939, opened 2026-04-13
ci: base.lock.yml: remove meta-qcom-distro from the list

ricardosalveti:
meta-qcom-distro is owned by us and we do expect changes there which are RC3 related, so remove from the lock to simplify the release process.

### DIFF
--- a/ci/base.lock.yml
+++ b/ci/base.lock.yml
@@ -8,8 +8,6 @@ overrides:
             commit: 5d722b5d65e4eef7befe6376983385421e993f86
         meta-arm:
             commit: d987a5945802dcbbc06be91a0d34f00431ea840e
-        meta-qcom-distro:
-            commit: dc4051f119c08e081dfa932567ee6d88f32d1c54
         meta-openembedded:
             commit: 96a803a50d55a455b0584d8a81168351d0f8c697
         meta-virtualization:


ricardosalveti:
We can keep meta-qcom-distro in the lock file as well, but since we own and maintain the layer, removing will make the integration easier.

Created as draft for discussion, fine to have the lock in place for it as well if we think we want to lock it the same way.

vkraleti:
May be meta-audioreach can also be excluded?

ricardosalveti:
> May be meta-audioreach can also be excluded?

Is the development there also bound to the QLI 2.0 release machinery?

lumag:
> May be meta-audioreach can also be excluded?

Definitely no.

lumag:
> meta-qcom-distro is owned by us and we do expect changes there which are RC3 related, so remove from the lock to simplify the release process.

Why? This would mean that there is no way to rebuild RC3 "as is" from its source. I'd rather not do this.

ricardosalveti:
> Why? This would mean that there is no way to rebuild RC3 "as is" from its source. I'd rather not do this.

We could either update before RC3 or reuse what is already going to be published in meta-qcom-releases.

If we keep distro locked down, we will have to bump meta-qcom in every distro change we merge, which will force 2 sets of build / pr / merge rounds, for almost no gain.

lumag:
> > Why? This would mean that there is no way to rebuild RC3 "as is" from its source. I'd rather not do this.
> 
> We could either update before RC3 or reuse what is already going to be published in meta-qcom-releases.
> 
> If we keep distro locked down, we will have to bump meta-qcom in every distro change we merge, which will force 2 sets of build / pr / merge rounds, for almost no gain.

Ack, let's lock it down just before the release.

lumag (approved):

anujm1 (approved):

PR #1940, opened 2026-04-13
Revert "weston: handle shader compiler differences across GPU impleme…

quaresmajose:
…ntations"

This reverts commit 4502bf5c6c230bbb3e0fe7df6d9b958c6d953bf9.

The patch is merged upstream https://git.openembedded.org/openembedded-core/commit/?id=8c008f36d7af1e63cb3f4a2ba763efd1f1e5fe4d


ricardosalveti:
This will also require a lock bump, which brings to the question if we want to keep RC3 aligned with M3, because if so then we should only merge this PR after RC3.

The bad side effect of holding this change is that meta-qcom will cause the autobuilder builds to fail upstream, as we now have a patch conflict.

lumag:
> This will also require a lock bump, which brings to the question if we want to keep RC3 aligned with M3, because if so then we should only merge this PR after RC3.

I don't see M3 tag in OE-Core.

> The bad side effect of holding this change is that meta-qcom will cause the autobuilder builds to fail upstream, as we now have a patch conflict.

I think it's bad from the community and everything else point of view, but I think our release goals are more important, especially since we are going to fix up the fallouts in a week.

ricardosalveti:
> > This will also require a lock bump, which brings to the question if we want to keep RC3 aligned with M3, because if so then we should only merge this PR after RC3.
> 
> I don't see M3 tag in OE-Core.

Still waiting for QA: https://lists.yoctoproject.org/g/yocto/message/66399

> > The bad side effect of holding this change is that meta-qcom will cause the autobuilder builds to fail upstream, as we now have a patch conflict.
> 
> I think it's bad from the community and everything else point of view, but I think our release goals are more important, especially since we are going to fix up the fallouts in a week.

+1

quaresmajose:
> > > This will also require a lock bump, which brings to the question if we want to keep RC3 aligned with M3, because if so then we should only merge this PR after RC3.
> > 
> > 
> > I don't see M3 tag in OE-Core.
> 
> Still waiting for QA: https://lists.yoctoproject.org/g/yocto/message/66399
> 
> > > The bad side effect of holding this change is that meta-qcom will cause the autobuilder builds to fail upstream, as we now have a patch conflict.
> > 
> > 
> > I think it's bad from the community and everything else point of view, but I think our release goals are more important, especially since we are going to fix up the fallouts in a week.
> 
> +1

Well, at the moment we don't have many options. And after this one, another one might come during our freeze.

quaresmajose:
Perhaps it would be a good idea to consider removing our layer on the autobuilder during this phase so as not to disturb the yocto. @anujm1 what is your opinion about this?

anujm1:
> Perhaps it would be a good idea to consider removing our layer on the autobuilder during this phase so as not to disturb the yocto. @anujm1 what is your opinion about this?

I think we should consider merging this change and bumping. There might be bugs in rc3 build which is in QA so there's no guarantee that we'll be using M3 commit for rc3.

I don't think we should disable those builds just for this brief period.

quaresmajose:
I added a smaller oe-core bump to be able to integrate the revert.
```
    openembedded-core changes:
    - 8c008f36d7 weston: backport gl-shaders fix to avoid shader compiler crashes
    - 83b100bf2e devtool: ide-sdk: use TOOLCHAIN not TCOVERRIDE
    - ae24057e6a oe-selftest/cpp-example: fix conf file ownership with static UIDs/GIDs
    - a42c2beedb oe-selftest: devtool: use assertRegex to match test output for meson
    - f8ec611e86 oe-selftest: devtool: GDB breakpoint after std::vector is constructed
    - c4d86d5226 package: update the comment block explaining 'emit_pkgdata'
```
The Yocto autobuilder is currently blocked by meta-qcom layer. The bump is ahead of m3 in order to unblock autobuilder.
https://lists.openembedded.org/g/openembedded-core/message/235485

quaresmajose:
None of the changes have an impact on the BSP side.

ricardosalveti (approved):
I think this is a good trade-off, changes shouldn't impact us (comment, devtool and self-tests).

lumag (approved):

PR #1941, opened 2026-04-13
workflows: build-yocto: skip lock update when lock is already available

ricardosalveti:
Avoid running `kas lock --update` when `ci/base.lock.yml` is already available as part of the repository, to avoid unexpected lock updates during the build.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -43,6 +43,7 @@ jobs:
           persist-credentials: false
 
       - name: Run kas lock
+        if: ${{ hashFiles('ci/base.lock.yml') == '' }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build-yocto.yml
```diff
@@ -43,6 +43,7 @@ jobs:
           persist-credentials: false
 
       - name: Run kas lock
+        if: ${{ hashFiles('ci/base.lock.yml') == '' }}
```
kraj, line 46: LGTM. is there a separate GH action for updating the locks?

ricardosalveti: Not yet, that is planned post RC3.

lumag (approved):

ricardosalveti:
Merging to get our CI back to what we expect it to be after merging the lock file.

PR #1942, opened 2026-04-14
mount-tee-partition: make persist handling event driven

ricardosalveti:
On platforms without a persist partition, the current systemd setup can
cause the boot process to block waiting for /dev/disk/by-partlabel/persist,
resulting in long delays or timeouts during early boot.

As the persist partition is optional, solutions based on static systemd
service dependencies, generators, or ordering against generic targets
(such as multi-user.target) are not reliable in this case, as
/dev/disk/by-partlabel/persist is only created after udev has processed
all block device events, and systemd explicitly discourages depending on
udev-settle for correctness.

Convert persist partition formatting and mounting to an event-driven
model without hard systemd dependencies. Mounting is triggered
directly by a udev event when the persist partition appears, ensuring
that systems without the partition do not block during boot, while
preserving strict ordering so the filesystem check and format service
always runs before the mount.

### DIFF
--- a/recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
+++ b/recipes-bsp/partition/mount-tee-partition/format-tee-partition.service
@@ -5,8 +5,6 @@
 [Unit]
 Description=Create ext4 filesystem on persist partition if missing
 DefaultDependencies=no
-Requires=dev-disk-by\x2dpartlabel-persist.device
-After=dev-disk-by\x2dpartlabel-persist.device
 ConditionPathExists=/dev/disk/by-partlabel/persist
 Before=var-lib-tee.mount
 ConditionPathIsMountPoint=!/var/lib/tee
@@ -15,6 +13,3 @@ ConditionPathIsMountPoint=!/var/lib/tee
 Type=oneshot
 ExecStart=@sbindir@/check-tee-partition-fs.sh
 RemainAfterExit=yes
-
-[Install]
-WantedBy=local-fs-pre.target
--- a/recipes-bsp/partition/mount-tee-partition/persist.rules
+++ b/recipes-bsp/partition/mount-tee-partition/persist.rules
@@ -1,0 +1,1 @@
+SUBSYSTEM=="block", ENV{DEVTYPE}=="partition", ENV{PARTNAME}=="persist", ACTION=="add", TAG+="systemd", ENV{SYSTEMD_WANTS}+="var-lib-tee.mount"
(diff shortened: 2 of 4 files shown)


ricardosalveti:
This is to avoid the service timeout we are seeing on devices without the persist partition.

lumag (changes requested):
The commit message doesn't capture the reason for the changes. Please start it by describing the reason for the commit.

Next, I don't think that using event mechanism is a correct way to handle partitions which are knowingly present on the storage device.

koenkooi:
Do need a custom generator?

ricardosalveti:
> Next, I don't think that using event mechanism is a correct way to handle partitions which are knowingly present on the storage device.

The problem is that a direct bind will force systemd to expect the partition to be available in every device, not only on the ones with the persist partition, just see the boot log for db820 for example.

The only way is via udev or via a generator, as suggested by @koenkooi.

ricardosalveti:
> Do need a custom generator?

Yeah, that might work, will have a look today.

lumag:
> > Next, I don't think that using event mechanism is a correct way to handle partitions which are knowingly present on the storage device.
> 
> The problem is that a direct bind will force systemd to expect the partition to be available in every device, not only on the ones with the persist partition, just see the boot log for db820 for example.
> 
> The only way is via udev or via a generator, as suggested by @koenkooi.

Would it be better if we had a Condition or starting qtee service and let qtee request the mounting of the partition? Wouldn't mounting be skipped if there is no Qtee device for example?

ricardosalveti:
> Would it be better if we had a Condition or starting qtee service and let qtee request the mounting of the partition? Wouldn't mounting be skipped if there is no Qtee device for example?

I believe only if mounting / formatting the partition is done at a later stage in the boot process (after udev and local-fs is finished), as then we can use ConditionPathExists and let systemd decide when to run these services. If done at an earlier stage (as done now), the only reliably way would be to move it to be event or generator driven.

ricardosalveti:
Will play with this a bit more today and see what are the working options with systemd.

ricardosalveti:
As the images are generic and can be used on targets without the persist partition, the only safe way to have the mount unit not block the systemd boot process, when it is not available, is by changing to be mounted via udev events.

While generators could be used to link the mount unit on a specific target, there is no guarantee that the event creating the `/dev/disk/by-partlabel/persist` link (since we need to use partition name / label) will be executed before the mount unit. If we bind the unit with the device (as done today), then systemd will wait for the device to show up and delay the boot on devices without such partition in place.

Moved back from draft and updated the commit message.

ricardosalveti:
> starting qtee service and let qtee request the mounting of the partition?

Qtee can work with and without the persist partition, so we can't force that dependency.

lumag (approved):

koenkooi (approved):

PR #1944, opened 2026-04-14
README: Add reference to meta-qcom milestone release repository

shivshah3400:
A meta-qcom build pulls metadata from 10+ layers (oe-core, bitbake, meta-oe, ..). Hence, we need a lock file to
store all the meta layers shas for a milestone release which can then be used by customers for reproducing a stable build.
Introduced a Releases section in the README that points users to the meta‑qcom‑releases
repository which holds the lock file for meta-qcom milestone releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,15 @@ Make sure that ModemManager is not running, disable it if necessary.
    READ64 image: 13 offset: 0x0 length: 0x40
    ```
 
+## Releases
+
+Milestone releases for meta-qcom are managed in a separate repository:
+[meta-qcom-releases](https://github.com/qualcomm-linux/meta-qcom-releases)
+
+The meta-qcom-releases repository contains kas lock files corresponding to each official milestone release.
+These lock files capture the exact revisions of meta-qcom and all dependent meta-layers used during a
+given release, ensuring reproducible and stable builds.
+
 ## Contributing
 
 Please submit any patches against the `meta-qcom` layer (branch **master**)


README.md
```diff
@@ -170,6 +170,15 @@ Make sure that ModemManager is not running, disable it if necessary.
...
    ```
 
+## Releases
+
+Milestone releases for meta-qcom are managed in a separate repository:
+[meta-qcom-releases](https://github.com/qualcomm-linux/meta-qcom-releases)
+
+The meta-qcom-releases repository contains kas lock files corresponding to each official milestone release.
+These lock files capture the exact revisions of meta-qcom and all dependent meta-layers used during a
+given release, ensuring reproducible and stable builds.
```
lumag, line 180: How is that supposed to work? And more importantly, why?

lumag (changes requested):

README.md
shivshah3400: Please refer build instructions on meta-qcom-releases for more details. 
We have been using this repo for all our RC releases and this is also documented as part of Qualcomm's offical build guide [here](https://docs.qualcomm.com/doc/80-80021-254/topic/build_from_source_github_intro.html?product=895724676033554725&facet=Build%20Guide&version=2.0-rc2#build-from-source-github-intro)

lumag: Your comment does not answer on the 'why' question. Why do we store the lock file in the external repo? It needs to be captured at least in the commit message.

Also, please expand the README so that it actually documents the process. The parts that you added don't provide any actual documentation (nor would it be a good idea to link the limited access docs).

shivshah3400: > Your comment does not answer on the 'why' question. Why do we store the lock file in the external repo? It needs to be captured at least in the commit message.

Updated the commit message to capture why a lock file is needed.


> Also, please expand the README so that it actually documents the process. The parts that you added don't provide any actual documentation (nor would it be a good idea to link the limited access docs).

The build instructions are already documented as part of the README in meta-qcom-releases. If you feel anything is missing or needs to be updated there, feel free to create an issue. As for meta-qcom-releases repo, its a public repository so everyone will have access to it.

lumag: > > Your comment does not answer on the 'why' question. Why do we store the lock file in the external repo? It needs to be captured at least in the commit message.
> 
> Updated the commit message to capture why a lock file is needed.

You still didn't answer, why are we pointing users to a separate repo rather than having a lock file here.

> > Also, please expand the README so that it actually documents the process. The parts that you added don't provide any actual documentation (nor would it be a good idea to link the limited access docs).
> 
> The build instructions are already documented as part of the README in meta-qcom-releases. If you feel anything is missing or needs to be updated there, feel free to create an issue. As for meta-qcom-releases repo, its a public repository so everyone will have access to it.

Definitely. The docs should be here rather than some other repo.

Last, but not least, please work on fixing your environment. Your commit message is wrongly wrapped. Your Git is not setup properly, etc.

shivshah3400: > You still didn't answer, why are we pointing users to a separate repo rather than having a lock file here.

Are you open to keeping the tags here in meta-qcom? We can then maintain a lock file for wrynose branch and tag the releases here in meta-qcom itself.
If not, then maintaining a seperate repo makes sense and ideally the build steps should be a part of README of that repo.

lumag: I'm definitely open to maintaining tags in meta-qcom-distro and meta-qcom.

ricardosalveti: We didn't have the lock file before, and we think we can keep maintaining the lock file for the wrynose branch as part of meta-qcom, then we could indeed just tag and be done.

PR #1945, opened 2026-04-14
CAMX: revision update for Kodiak, Lemans, Talos

gkhose-qipl:
The versions used here now include the same set of features and changes as those in QLI 1.8.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.16.bb
+++ b/dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.16.bb
@@ -8,15 +8,15 @@ LICENSE = "LICENSE.qcom-2"
 LIC_FILES_CHKSUM = "file://usr/share/doc/${BPN}/LICENSE.QCOM-2.txt;md5=165287851294f2fb8ac8cbc5e24b02b0 \
                     file://usr/share/doc/${BPN}/NOTICE;md5=04facc2e07e3d41171a931477be0c690"
 
-PBT_BUILD_DATE = "260404"
+PBT_BUILD_DATE = "260411"
 SRC_URI = " \
    https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/camx.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/${BPN}_${PV}_armv8-2a.tar.gz;name=camxlib \
    https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/camx.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/camx-kodiak_${PV}_armv8-2a.tar.gz;name=camx \
    https://qartifactory-edge.qualcomm.com/artifactory/qsc_releases/software/chip/component/camx.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/chicdk-kodiak_${PV}_armv8-2a.tar.gz;name=chicdk \
    "
-SRC_URI[camxlib.sha256sum] = "bd9652a8337af92449ca844db5299c4110baae99ea5d326f67abad1638cdf301"
-SRC_URI[camx.sha256sum] = "d8e3b778be981d6c390713294f11ab150f1faff0c62d5e1ce37e5cd2ba3b9c52"
-SRC_URI[chicdk.sha256sum] = "bbcda983a28383853ecea0a64dde001b1dc960b6ef92616d5a53a2441d7299ca"
+SRC_URI[camxlib.sha256sum] = "e816f8b5fd0996bc98a149e16e02873d94567a3ee711c5872c7ecb5c56c65e8d"
+SRC_URI[camx.sha256sum] = "e9f7ac3c99b08f38af75f7e7e5f45bab96762362ae8e1f188dae11257746082a"
+SRC_URI[chicdk.sha256sum] = "b1280a7c398352564ba9c29cf8ddcb8f22a812ee7c62984379bfb877d3a5f6c2"
 
 S = "${UNPACKDIR}"
 
--- a/dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.17.bb
+++ b/dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.17.bb
@@ -1,12 +1,12 @@
 PLATFORM = "lemans"
-PBT_BUILD_DATE = "260403"
+PBT_BUILD_DATE = "260411"
 
 require common.inc
 
-SRC_URI[camxlib.sha256sum]     = "c3a2899a19078c0ced7ae6ec4605b84834eae33893820f05a34e864f198bdfc9"
-SRC_URI[camx.sha256sum]        = "91bfc42dad35adef895656dfef48f4eacb4059e49b813dec9cbece9a4b30a309"
-SRC_URI[chicdk.sha256sum]      = "4c83df2a4690759979087d112633d43420599ab4906f233109591528107d4769"
-SRC_URI[camxcommon.sha256sum]  = "89ee02d0b4bd09dcca091be2f2339ff50b6a34fc50ca51c612b0e633768a956d"
+SRC_URI[camxlib.sha256sum]     = "3ddbe79a9753b3238f3fb6072022a9a572be946576e1c9c43c383795a1f9cb4e"
+SRC_URI[camx.sha256sum]        = "5733fecdfb12339c9a4063e0058badf1c69a79327afb0be754a33832c577b149"
+SRC_URI[chicdk.sha256sum]      = "d26a55a95c11dcd8d350f6161465ee81845b5f0cdb9231997092c85fc95a60e5"
+SRC_URI[camxcommon.sha256sum]  = "e035bb3a34231dd12f3d2c73f87c10d116c1ec8bc6b289efd20924d9a6fbd00e"
 
 DEPENDS += "${@bb.utils.contains('DISTRO_FEATURES', 'opencl', 'qcom-adreno virtual/libopencl1', '', d)}"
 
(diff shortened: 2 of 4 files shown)


dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.16.bb
```diff
@@ -105,6 +105,7 @@ FILES:${PN} = "\
     ${libdir}/camx/kodiak/*${SOLIBS} \
     ${libdir}/camx/kodiak/hw/*${SOLIBS} \
     ${libdir}/camx/kodiak/camera/components/*${SOLIBS} \
+    ${libdir}/camx/kodiak/bm2n08.bin \
```
lumag: What kind of file is this? Why is it a binary under `${libdir}`?

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.17.bb
```diff
@@ -27,4 +27,7 @@ FILES:camx-nhx = "\
     ${sysconfdir}/camera/test/NHX/ \
 "
 
-FILES:${PN} += "${@bb.utils.contains('DISTRO_FEATURES', 'opencl', '${libdir}/camx/${PLATFORM}/*.cl ${libdir}/camx/${PLATFORM}/libmctf_cl_program.bin', '', d)}"
+FILES:${PN} += "\
+    ${libdir}/camx/${PLATFORM}/bm4a*.bin \
```
lumag: Same question. What kind of file is this? Why is it a binary under ${libdir}?

lumag (changes requested):

ricardosalveti:
> Aligns 0.0 with QLI1.8 parity.

This doesn't really mean much here, what are the versions of these components in QLI 1.8?

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.17.bb
gkhose-qipl: @lumag ,
bins packed in lib is helper binary and not executed directly so kept in libdir.

gkhose-qipl:
> > Aligns 0.0 with QLI1.8 parity.
> 
> This doesn't really mean much here, what are the versions of these components in QLI 1.8?

QLI 1.8 is latest release (version was not maintained) in QLI 1.0 component.

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.17.bb
lumag: What kind of helper binary? How is it used?

gkhose-qipl: > What kind of helper binary? How is it used?

it's DL based binary it contents network graph architecture and weight. we open file and read contents into buffer and pass buffer into nsp.

ricardosalveti:
> QLI 1.8 is latest release (version was not maintained) in QLI 1.0 component. 

Sorry, not maintained how? Don't we have a version of these components in qli 1.0?

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.17.bb
lumag: What is 'DL'? If it is passed to NSP, it should not be in `${libdir}`.

gkhose-qipl:
> > QLI 1.8 is latest release (version was not maintained) in QLI 1.0 component.
> 
> Sorry, not maintained how? Don't we have a version of these components in qli 1.0?

for every release in QLI 1.0 we have 1.0.0 version for all camera packages.

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.17.bb
gkhose-qipl: > What is 'DL'? If it is passed to NSP, it should not be in `${libdir}`.

deep learning.

ricardosalveti:
> > > QLI 1.8 is latest release (version was not maintained) in QLI 1.0 component.
> > 
> > 
> > Sorry, not maintained how? Don't we have a version of these components in qli 1.0?
> 
> for every release in QLI 1.0 we have 1.0.0 version for all camera packages.

Then please say that the respective versions used here are now containing the same set of features / changes as the one used by QLI 1.8 instead of just saying 0.0 (which means nothing to our external users).

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.17.bb
lumag: The state is as is. And yes, still not `${libdir}`.

gkhose-qipl:
> respective versions used here are now containing the same set of features / changes as the one used by QLI 1.8

sure will update commit text as 
respective versions used here are now containing the same set of features / changes as the one used by QLI 1.8

gkhose-qipl:
@lumag / @ricardosalveti ,
can you please review, due to rb3 priority removed DLFD changes and only kept fast-forward changes.

lumag (approved):

koenkooi (approved):

dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-lemans_1.0.17.bb
kkalyanr: The binary files are deep learning models that contain the neural network’s architecture and its trained weights, but they are not executable files. To run the network on the NSP, the binary file is first loaded into a memory buffer, which serves as intermediate storage. The NSP reads this buffer to extract the model graph and weights required for execution. Once the NSP has successfully extracted all the necessary information, the buffer is freed to release memory.

The data directory path is meant to store device‑specific files, but the network binary files are common across Lemans, its variants, and Monaco. If we place these common network binaries inside device‑specific data directories, the same files would be duplicated for each device. This would cause unnecessary redundancy and increase memory usage. Since these binaries are tightly coupled with the CAMX code, it makes more sense to store them in /usr/lib/camx/lemans. From this shared location, the binaries can be accessed by all Lemans variants and Monaco, avoiding duplication and keeping the memory footprint minimal.

lumag: @kkalyanr no. /usr/lib is only for the host library files. Please refer to the Linux FHS. Please move those to /usr/share/

kkalyanr: @lumag 
The BITML binary files (bm4a*.bin) located under ${libdir}/camx/${PLATFORM} are common across Lemans, its variants, and Monaco platforms. Since these binaries are shared by all targets, installing them into a target‑specific directory results in unnecessary duplication.
To avoid this redundancy, we plan to install these common binaries in a centralized location under the data directory. A suitable and consistent location would be /usr/share/qcom/camx, as it is accessible to all targets. This approach ensures better maintainability, reduces storage duplication, and clearly separates common resources from target‑specific components.

lumag: Just `/usr/share/camx`, please. Also, if they are common, that needs to be specified in the commit message. My understanding was that they are per-target.

kkalyanr: @lumag 
small correction from my side regarding binary files .
For Kodiak, we use bm2n08.bin. The remaining BIN files that start with bm4a*.bin are common across all LeMans variants and Monaco. Since these BIN files are target‑specific, the paths should be used accordingly: /usr/share/camx/lemans for LeMans and /usr/share/camx/kodiak for Kodiak.

lumag: If there is no conflict in the names, please install all bins under `/usr/share/camx`.

kkalyanr: Hi @lumag,
In the camxlib-lemans and camxlib-kodiak packages, all .so files are currently present in libdir. However, our BITML binaries and .so files are intended to be installed in datadir (/usr/share).

The issue is that Yocto's QA check is flagging our BITML .so file because it expects all .so files to reside in libdir:
ERROR: camxlib-lemans-1.0.18-r0 do_package_qa: QA Issue: camxlib-lemans: 
found library in wrong location: 
/usr/share/qcom/sa8775p/Qualcomm/SA8775P-RIDE/dsp/cdsp/libbitml_nsp_73nb_skel.so [libdir]

The root cause is that the BITML DSP skel library is a 32-bit library, while Yocto expects it to be 64-bit in libdir.

Skip QA checks for all .so files in the camxlib packages.
INSANE_SKIP:${PN} += "arch libdir already-stripped"

can we add the above rule to resolve BITML lib installation issue in camx-lib package

lumag: Split SKEL libraries to a separate packages (one per Hexagon arch). Provide INSANE_SKIP for those packages only.

kkalyanr: Thanks @lumag 
What could be the name  for hexagon arch specific packages?

lumag: While we are waiting for FastRPC to support loading the system libraries from standard locations, there best option would be `camx-qcs8300-ride-dsp`

PR #1968, opened 2026-04-16
gstreamer1.0-plugins-bad: Update NV12_Q08C patch to fix format check

Raja-Ganapathi-Busam:
Use self->render_info instead of self->video_info when checking for NV12_Q08C format. The video_info field is assigned from render_info later in the function, so checking video_info at this point would use stale data from the previous caps instead of the newly parsed format information.

### DIFF
--- a/recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad/0001-wayland-Add-support-for-NV12_Q08C-compressed-8-bit-f.patch
+++ b/recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad/0001-wayland-Add-support-for-NV12_Q08C-compressed-8-bit-f.patch
@@ -69,7 +69,7 @@ index 19a3e8c..030fe1b 100644
              &self->render_info, DRM_FORMAT_MOD_LINEAR))
        gst_video_info_dma_drm_init (&self->drm_info);
 +
-+    if (GST_VIDEO_INFO_FORMAT (&self->video_info) == GST_VIDEO_FORMAT_NV12_Q08C) {
++    if (GST_VIDEO_INFO_FORMAT (&self->render_info) == GST_VIDEO_FORMAT_NV12_Q08C) {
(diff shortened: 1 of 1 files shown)


koenkooi (changes requested):
Multiple things:

1. If you change the patch, the patch header needs to be updated as well for its Upstream-Status. If that isn't changed, note that in the commit message.
2. This is a change to the gstreamer recipe, not the wayland recipe. Please fix the commit message
3. This is a change to a patch, that needs to be clear in the commit message as well

Raja-Ganapathi-Busam:
> Multiple things:
> 
> 1. If you change the patch, the patch header needs to be updated as well for its Upstream-Status. If that isn't changed, note that in the commit message.
> 2. This is a change to the gstreamer recipe, not the wayland recipe. Please fix the commit message
> 3. This is a change to a patch, that needs to be clear in the commit message as well

Updated in the commit message.

koenkooi:
> > Multiple things:
> > 
> > 1. If you change the patch, the patch header needs to be updated as well for its Upstream-Status. If that isn't changed, note that in the commit message.
> > 2. This is a change to the gstreamer recipe, not the wayland recipe. Please fix the commit message
> > 3. This is a change to a patch, that needs to be clear in the commit message as well
> 
> Updated in the commit message.

Thanks!

ricardosalveti:
> Note: No change to the patch header as this is a fix to the existing patch.

Can you also add that no change to the patch header was done as this patch was denied upstream and it is maintained here? Just for clarity.

Raja-Ganapathi-Busam:
> > Note: No change to the patch header as this is a fix to the existing patch.
> 
> Can you also add that no change to the patch header was done as this patch was denied upstream and it is maintained here? Just for clarity.

Done

anujm1 (approved):

ricardosalveti (approved):

lumag (approved):

koenkooi (approved):

PR #1976, opened 2026-04-16
packagegroup-rb3gen2: include firmware for LT9611UXC

lumag:
RB3 Gen2 uses LT9611UXC DSI-to-HDMI bridge. Include firmware for this chip into the device-specific packagegroup to let users reflash it.


koenkooi (approved):
Thanks! Reflashing my board is on my TODO, this makes it a lot easier!

ricardosalveti (approved):

PR #1979, opened 2026-04-16
Fix lemans-el2 DTBO name typo in fit-dtb-compatible.inc

vkraleti:
Fix a typo in the DT overlay name used in FIT_DTB_COMPATIBLE for the
qcs9100-ride-r3 platform.

Replace the incorrect `fdt-lemans-el2` reference with `lemans-el2` so the
base+overlay combination is matched correctly during FIT image generation.


ricardosalveti (approved):

lumag (approved):

lumag:
And it fails to boot on QCS9100:
https://lava.infra.foundries.io/scheduler/job/180145

ricardosalveti:
> And it fails to boot on QCS9100: https://lava.infra.foundries.io/scheduler/job/180145

Wonder if this is also related to https://github.com/qualcomm-linux/meta-qcom/issues/1709, it seems there is some sort of systemd issue during the first boot which is happening with all devices, but not on every first boot, probably a race condition.

Issue #1973, opened 2026-04-16
IMSDK proprietary plugins missing in nightly image


bkosuri:
\\swayam\QLI_Builds\Yocto\24365713647_Nightly_Build_488\performance_linux-qcom-6.18\qcom-multimedia-proprietary-image-rb3gen2-core-kit

IMSDK proprietary plugins are not available in above nightly build.
Missing libs:
/usr/lib/gstreamer-1.0/libgstqtimlqnn.so
/usr/lib/gstreamer-1.0/libgstqtimlsnpe.so
/usr/lib/gstreamer-1.0/libgstqtiqmmfsrc.so
/usr/lib/gstreamer-1.0/libgstqtismartvencbin.so
Sample apps

anujm1:
It is not installing the weak dependency gst-plugins-imsdk-prop-meta because it's unable to:

```
--> Starting dependency resolution
--> Finished dependency resolution
Error:
Problem: package gst-plugins-imsdk-prop-meta-0.1.1-r0.armv8_2a from oe-repo requires gst-plugins-imsdk-prop-apps, but none of the providers can be installed
  - conflicting requests
  - nothing provides gst-plugins-imsdk-prop-media needed by gst-plugins-imsdk-prop-apps-0.1.1-r0.armv8_2a from oe-repo
(try to add '--skip-broken' to skip uninstallable packages)
```

The media package should move to oss recipe if the contents are enabled through a PACKAGECONFIG enabled there.

anujm1:
I got this when trying to explicitly install gst-plugins-imsdk-prop-meta

Raja-Ganapathi-Busam:
Thanks Anuj.
c7551b2a4aa2f514b21440739de8b598314c59ec commit is introducing the ${PN}-media package, which is present in common packaging file used by both imsdk-oss and imsdk-prop packages.

anujm1:
Fixed by https://github.com/qualcomm-linux/meta-qcom/pull/1988

PR #1988, opened 2026-04-17
gst-plugins-imsdk-oss: move media packaging from packaging.inc to oss

muthumul:
Move the media packaging from packaging.inc to oss as it is preventing the proprietary plugins from getting packed into the image.

### DIFF
--- a/recipes-multimedia/imsdk/gst-plugins-imsdk-oss_0.1.1.bb
+++ b/recipes-multimedia/imsdk/gst-plugins-imsdk-oss_0.1.1.bb
@@ -7,3 +7,11 @@ DESCRIPTION = "Open-source Qualcomm IMSDK GStreamer multimedia plugins"
 DEPENDS += "gst-plugins-imsdk-base"
 
 PACKAGECONFIG ??= "sw tools videoproc"
+
+PACKAGES += "${PN}-media"
+
+FILES:${PN}-media = " \
+    ${datadir}/qdemo/* \
+"
+
+RDEPENDS:${PN}-apps += "${@bb.utils.contains('PACKAGECONFIG', 'python-apps', '${PN}-media', '', d)}"
--- a/recipes-multimedia/imsdk/gst-plugins-imsdk-packaging.inc
+++ b/recipes-multimedia/imsdk/gst-plugins-imsdk-packaging.inc
@@ -13,12 +13,6 @@ python split_imsdk_module_packages () {
     do_split_packages(d, imsdk_root, r'([^/]+)/modules/[^/]+\.so', d.expand('${PN}-%s-modules'), 'IMSDK runtime modules for %s plugin', recursive=True, extra_depends='', match_path=True)
 }
 
-PACKAGES += "${PN}-configs ${PN}-media"
+PACKAGES += "${PN}-configs"
 
 FILES:${PN}-configs = "${sysconfdir}/configs/"
-
-FILES:${PN}-media = " \
-    ${datadir}/qdemo/* \
-"
-
-RDEPENDS:${PN}-apps += "${PN}-media"


muthumul:
This change moves the {PN}-media package from shared packaging include to the gst-plugins-imsdk-oss
recipe to make sure that the proprietary plugins are getting packed into the image, addressing the issue posted here: https://github.com/qualcomm-linux/meta-qcom/issues/1973

quaresmajose (dismissed):

recipes-multimedia/imsdk/gst-plugins-imsdk-oss_0.1.1.bb
```diff
@@ -7,3 +7,11 @@ DESCRIPTION = "Open-source Qualcomm IMSDK GStreamer multimedia plugins"
 DEPENDS += "gst-plugins-imsdk-base"
 
 PACKAGECONFIG ??= "sw tools videoproc"
+
+PACKAGES += "${PN}-media"
+
+FILES:${PN}-media = " \
+    ${datadir}/qdemo/* \
```
anujm1, line 14: Is this `qdemo` only present when sample-apps or any other PACKAGECONFIG is enabled? If yes, then I think we should also add `ALLOW_EMPTY:${PN}-media = "1"` for corner case when it is disabled.

muthumul: > Is this `qdemo` only present when sample-apps or any other PACKAGECONFIG is enabled? If yes, then I think we should also add `ALLOW_EMPTY:${PN}-media = "1"` for corner case when it is disabled.

After splitting gst-plugins-imsdk into base, oss, and prop, we should be able to ensure that any package defined within a specific recipe only affects that recipe and not the others, if I am not wrong? Correct me please.

However, in the current setup, the ${PN}-media package is being pulled in by ${PN}-apps. python-apps is enabled via the oss.bbappend, which means the qdemo artifacts are getting pulled in by both multimedia and prop images unconditionally.

Given this, do you still want me to handle this corner case, Anuj?

quaresmajose: I think the point here is not the dependencie but lets supose the `python-apps` is disabled in the oss.bbappend, in this case the will it be something produced and installed in `${datadir}/qdemo`?
If no, you need to add `ALLOW_EMPTY:${PN}-media = "1"`. You can verify this disbaling `python-apps` and build the `gst-plugins-imsdk-oss`.

anujm1: Yes, thanks Jose. Otherwise, if that package is not generated, the ${PN}-apps package below and anything that depends on it will fail to install (which was the issue for proprietary recipe since this qdemo wasn't present in that recipe).

muthumul: > I think the point here is not the dependencie but lets supose the `python-apps` is disabled in the oss.bbappend, in this case the will it be something produced and installed in ${datadir}/qdemo ? If yes, you need to add `ALLOW_EMPTY:${PN}-media = "1"`. You can verify this disbaling `python-apps` and build the `gst-plugins-imsdk-oss`.

Sure, Jose. I will try it out and update the PR.

muthumul: > Yes, thanks Jose. Otherwise, if that package is not generated, the ${PN}-apps package below and anything that depends on it will fail to install (which was the issue for proprietary recipe since this qdemo wasn't present in that recipe).

Sure, Anuj. I will update the PR. Thanks for the suggestion.

recipes-multimedia/imsdk/gst-plugins-imsdk-oss_0.1.1.bb
```diff
@@ -7,3 +7,13 @@ DESCRIPTION = "Open-source Qualcomm IMSDK GStreamer multimedia plugins"
...
+"
+
+ALLOW_EMPTY:${PN}-media = "1"
```
ricardosalveti: Please explain why this is getting added as part of the commit message.

lumag (changes requested):
> Move the {PN}-media package from shared packaging include to the gst-plugins-imsdk-oss
recipe to make sure that the proprietary plugins are getting packed into the image.

There is no description of the issue, no analysys. Why are we performing these changes?

recipes-multimedia/imsdk/gst-plugins-imsdk-oss_0.1.1.bb
muthumul: > Please explain why this is getting added as part of the commit message.

It is done, Ricardo.

muthumul:
> > Move the {PN}-media package from shared packaging include to the gst-plugins-imsdk-oss
> > recipe to make sure that the proprietary plugins are getting packed into the image.
> 
> There is no description of the issue, no analysys. Why are we performing these changes?



> > Move the {PN}-media package from shared packaging include to the gst-plugins-imsdk-oss
> > recipe to make sure that the proprietary plugins are getting packed into the image.
> 
> There is no description of the issue, no analysys. Why are we performing these changes?

Please let me know if the updated commit message addresses the concerns highlighted by you. Thank you.

lumag (commented):
> IMSDK proprietary plugins are not available in the latest nightly builds.
Missing libs:
/usr/lib/gstreamer-1.0/libgstqtimlqnn.so
/usr/lib/gstreamer-1.0/libgstqtimlsnpe.so
/usr/lib/gstreamer-1.0/libgstqtiqmmfsrc.so
/usr/lib/gstreamer-1.0/libgstqtismartvencbin.so
Sample apps

Drop. Just sate which package is broken and how.

recipes-multimedia/imsdk/gst-plugins-imsdk-oss_0.1.1.bb
lumag: I'd rather see a PACKAGECONFIG-dependent dependecy rather than having an empty -qmedia package.

muthumul:
> > IMSDK proprietary plugins are not available in the latest nightly builds.
> > Missing libs:
> > /usr/lib/gstreamer-1.0/libgstqtimlqnn.so
> > /usr/lib/gstreamer-1.0/libgstqtimlsnpe.so
> > /usr/lib/gstreamer-1.0/libgstqtiqmmfsrc.so
> > /usr/lib/gstreamer-1.0/libgstqtismartvencbin.so
> > Sample apps
> 
> Drop. Just sate which package is broken and how.

Sure, will do.

recipes-multimedia/imsdk/gst-plugins-imsdk-oss_0.1.1.bb
muthumul: > I'd rather see a PACKAGECONFIG-dependent dependecy rather than having an empty -qmedia package.

As python-apps is what installs qdemo artifacts, the ${PN}-media dependency is now conditional on that PACKAGECONFIG.

ricardosalveti (approved):

lumag (approved):

PR #1969, opened 2026-04-16
iris-video-dlkm: update iris driver to v1.0.5

venvenk:
This tag v1.0.5 brings below updates for Qcom Iris

- Add PAS context for KVM with Linux-managed IOMMU support
- Enable secure firmware loading and authentication

PR #1993, opened 2026-04-17
linux-yocto: Add support for KVM across multiple SoCs

gouravk-qualcomm:
Add generic API support which provide the platform to enable KVM for multiple client like video. It enables for targets like kodiak, lemans, monaco etc.

Qualcomm SoCs running with QHEE (EL2) have traditionally relied on the Peripheral Authentication Service (PAS) from TrustZone (TZ) firmware to securely authenticate and reset remote processors using a single SMC call (auth_and_reset). This call is first trapped by QHEE, which then invokes TZ for authentication. Once authentication is complete, control returns to QHEE, where the IOMMU translation scheme for the remote processors is configured before bringing them out of reset. Due to the design of the Qualcomm EL2 hypervisor, the Linux host OS running at EL1 is not permitted to configure IOMMU translations for remote processors, and only a single-stage translation is supported.

To make the remote processor bring-up (PAS) sequence hypervisor-independent, the auth_and_reset SMC call is now handled entirely by TrustZone. However, IOMMU configuration remains a challenge in scenarios where the KVM host is running at EL2 and does not have visibility into the remote processors’ IOMMU requirements. This is addressed by dynamically overlaying the IOMMU properties when the SoC runs a Linux host at EL2. An SMC interface provided by the TrustZone firmware allows the KVM host to retrieve the resource table for a given subsystem.
In addition, several remote processors—such as video, camera, and graphics—do not use the remoteproc framework to manage their lifecycle and instead rely on the Qualcomm PAS service to authenticate firmware. These processors must also be brought out of reset when Linux operates at EL2. Their client drivers use the MDT loader to load and authenticate firmware and, similar to the Qualcomm remoteproc PAS driver, are required to retrieve the resource table, create a shared memory bridge (shmbridge), and map the necessary resources before releasing the processors from reset.

This patch series enables Qualcomm infrastructure drivers such as SCM, remoteproc, and mdtloader to support the Secure PIL bring-up sequence when Linux runs as a hypervisor. All related patches have already been accepted upstream, except for one patch that is still FROMLIST. This remaining patch ensures that remote processor memory is unmapped from the Linux host before being handed over to the remote processor. Without this step, the memory region would remain mapped in the HLOS, allowing speculative CPU access while the remote processor is active, which poses a security risk.

### DIFF
--- a/recipes-kernel/linux/linux-yocto-6.18/generic-drivers/mdtloader/0004-soc-qcom-mdtloader-Add-PAS-context-aware-qcom_mdt_pa.patch
+++ b/recipes-kernel/linux/linux-yocto-6.18/generic-drivers/mdtloader/0004-soc-qcom-mdtloader-Add-PAS-context-aware-qcom_mdt_pa.patch
@@ -1,0 +1,150 @@
+From 5df4e58dcc528bcdb294b74236b166899d8d49f5 Mon Sep 17 00:00:00 2001
+From: Mukesh Ojha <mukesh.ojha@oss.qualcomm.com>
+Date: Mon, 5 Jan 2026 18:52:54 +0530
+Subject: [PATCH 04/12] soc: qcom: mdtloader: Add PAS context aware
+ qcom_mdt_pas_load() function
+
+Introduce a new PAS context-aware function, qcom_mdt_pas_load(), for
+remote processor drivers. This function utilizes the PAS context
+pointer returned from qcom_scm_pas_ctx_init() to perform firmware
+metadata verification and memory setup via SMC calls.
+
+The qcom_mdt_pas_load() and qcom_mdt_load() functions are largely
+similar, but the former is designed for clients using the PAS
+context-based data structure. Over time, all users of qcom_mdt_load()
+can be migrated to use qcom_mdt_pas_load() for consistency and
+improved abstraction.
+
+As the remoteproc PAS driver (qcom_q6v5_pas) has already adopted the
+PAS context-based approach, update it to use qcom_mdt_pas_load().
+
+Reviewed-by: Konrad Dybcio <konrad.dybcio@oss.qualcomm.com>
+Signed-off-by: Mukesh Ojha <mukesh.ojha@oss.qualcomm.com>
+Upstream-Status: Backport [https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git 8a4fcffde6c8]
+---
+ drivers/remoteproc/qcom_q6v5_pas.c  | 24 +++++-----------------
+ drivers/soc/qcom/mdt_loader.c       | 31 +++++++++++++++++++++++++++++
+ include/linux/soc/qcom/mdt_loader.h | 10 ++++++++++
+ 3 files changed, 46 insertions(+), 19 deletions(-)
+
+diff --git a/drivers/remoteproc/qcom_q6v5_pas.c b/drivers/remoteproc/qcom_q6v5_pas.c
+index 5f90976e5b41..0c4ec72352e3 100644
+--- a/drivers/remoteproc/qcom_q6v5_pas.c
++++ b/drivers/remoteproc/qcom_q6v5_pas.c
+@@ -239,15 +239,9 @@ static int qcom_pas_load(struct rproc *rproc, const struct firmware *fw)
+ 			return ret;
+ 		}
+ 
+-		ret = qcom_mdt_pas_init(pas->dev, pas->dtb_firmware, pas->dtb_firmware_name,
+-					pas->dtb_pas_id, pas->dtb_mem_phys,
+-					pas->dtb_pas_ctx);
+-		if (ret)
+-			goto release_dtb_firmware;
+-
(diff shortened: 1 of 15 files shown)


lumag:
You can bring the API here, but would also bring corresponding patches to Debian and other targets? I'd rather suggest fixing the driver instead to check for the kernel version.

ricardosalveti:
What is the goal here? Do we want to support KVM with linux-yocto or are you just trying to get https://github.com/qualcomm-linux/meta-qcom/pull/1969 to build with it?

You could change the driver to check for the base kernel version used, but it might not work perfectly as linux-yocto is based on the same version as linux-qcom, and there we do have the required APIs.

gouravk-qualcomm:
> What is the goal here? Do we want to support KVM with linux-yocto or are you just trying to get #1969 to build with it?
> 
> You could change the driver to check for the base kernel version used, but it might not work perfectly as linux-yocto is based on the same version as linux-qcom, and there we do have the required APIs.

yes @ricardosalveti , we are building against linux-yocto-6.18 and linux-yocto-dev for  [#1969](https://github.com/qualcomm-linux/meta-qcom/pull/1969) .

Without these patches even build is stalled.

gouravk-qualcomm:
> You can bring the API here, but would also bring corresponding patches to Debian and other targets? I'd rather suggest fixing the driver instead to check for the kernel version.

@lumag , since we're not including the overlay #1969 changes for qcom-deb-image, we currently don't need to implement the API change for qcom-deb-image either.

ricardosalveti:
Patches are ok from my perspective, but please extend the commit message here explaining why it is necessary.

mukeshojha-linux:
Reasoning:

Qualcomm SoCs running with QHEE (EL2) have been utilizing the Peripheral Authentication Service (PAS) from TrustZone (TZ) firmware to securely authenticate and reset remote processors via a single SMC call, auth_and_reset. This call is first trapped by QHEE, which then invokes TZ for authentication. Once authentication is complete, the call returns to QHEE, which sets up the IOMMU translation scheme for the remote processors and subsequently brings them out of reset. The design of the Qualcomm EL2 hypervisor dictates that the Linux host OS running at EL1 is not permitted to configure IOMMU translation for remote processors, and only a single-stage translation is configured.
To make the remote processor bring-up (PAS) sequence hypervisor-independent, the auth_and_reset SMC call is now handled entirely by TZ. However, the issue of IOMMU configuration remains unresolved—for example, in scenarios where the KVM host running at EL2 has no knowledge of the remote processors’ IOMMU settings. This is addressed by overlaying the IOMMU properties when the SoC runs a Linux host at EL2. An SMC call is provided by the TrustZone firmware to retrieve the resource table for a given subsystem.
There are also remote processors, such as those for video, camera, and graphics, that do not use the remoteproc framework to manage their lifecycle. Instead, they rely on the Qualcomm PAS service to authenticate their firmware. These processors also need to be brought out of reset when Linux is running at EL2. The client drivers for these processors use the MDT loader to load and authenticate firmware. Similar to the Qualcomm remoteproc PAS driver, they also need to retrieve the resource table, create a shared memory bridge (shmbridge), and map the resources before bringing the processors out of reset.

This set of patches enables Qualcomm infrastructure drivers, such as SCM, remoteproc, and mdtload, to support the Secure PIL sequence when Linux runs as a hypervisor. All patches have been accepted upstream except one patch, which is still FROMLIST. This remaining patch unmaps the remote processor memory from Linux before handing over the memory to the remote processor. Without this unmapping, the region would remain mapped in the HLOS, allowing the CPU to speculatively access the region while the remote processor is running.

ricardosalveti:
Please extend the commit message instead.

lumag:
The purpose of this recipe is to serve as the upstream reference. I am really not convinced that taking a huge series in benefits that purpose.

gouravk-qualcomm:
> Please extend the commit message instead.

Thanks for the feedback @ricardosalveti. Extended the commit message to include the full context.

sbanerjee-quic:
> The purpose of this recipe is to serve as the upstream reference. I am really not convinced that taking a huge series in benefits that purpose.

@lumag yes there are quite a big number of patches. But I am thinking this is the way if we do not want to add preferred kernel checks. 
Once we have the Wrynose LTS this is how many backports will look like. 
Are you thinking otherwise? I.e., not to build overlays against the linux-yocto-6.18?

lumag:
Again, this is a check that modules work with the _upstream_, not with the patched upstream. If you need the extra API for the video driver, please convert that API to the be provided by the additional module. We most likely want the vidc driver to work with standard kernels from Debian/Ubuntu/Fedora. Those distros won't patch the kernel for us.

vgarodia:
Modules working with upstream kernel is understandable when the ask is for yocto-dev, but why the ask if for linux-yocto-6.18 ?

> If you need the extra API for the video driver, please convert that API to the be provided by the additional module

@lumag could you please elaborate on this approach of extending new APIs ontop of existing kernel drivers - remoteproc, scm, mdtloader, as modules and have vidc driver compile still on upstream ?

lumag (approved):

vkraleti (approved):

PR #1995, opened 2026-04-19
qps615-dlkm: Add patch to implement phy-mode DT overlay for SGMII

ayaan-anwar:
Add a patch to the QPS615 DLKM recipe that implements phy-mode DT overlay support for SGMII on the Rb3Gen2 Industrial Kit.

The first port on the Rb3Gen2 Industrial Kit is wired to an RTL8221 SGMII PHY, but the TC956X driver defaults to 10GBASE-R for port 0. Without this patch, the port fails to link up on that board variant.

The new patch implements tc956x_platform_port_interface_overlay() to read the "phy-mode" property from the device tree node at probe time. When SGMII is specified, the function configures the required parameters for RTL8221. For any other or absent "phy-mode" value the function returns 0 and the driver proceeds with its default configuration.

The upstream change has been added to the Qualcomm Platform support PR at https://github.com/TC956X/TC9564_Host_Driver/pull/6.

### DIFF
--- a/recipes-kernel/qps615-module/qps615-dlkm/0013-net-ethernet-tc956x-Implement-phy-mode-DT-overlay-fo.patch
+++ b/recipes-kernel/qps615-module/qps615-dlkm/0013-net-ethernet-tc956x-Implement-phy-mode-DT-overlay-fo.patch
@@ -1,0 +1,72 @@
+From bfb38c3349dcd4aeb2af8f1339a30710e70ad75d Mon Sep 17 00:00:00 2001
+From: Mohd Ayaan Anwar <mohd.anwar@oss.qualcomm.com>
+Date: Thu, 16 Apr 2026 10:49:06 +0530
+Subject: [PATCH 13/13] net: ethernet: tc956x: Implement phy-mode DT overlay
+ for SGMII
+
+The first port on the RB3 Gen2 Industrial Kit uses an SGMII PHY,
+whereas the driver defaults to 10GBASE-R. Implement
+tc956x_platform_port_interface_overlay() to read the phy-mode
+property from the DT node and configure the port interface when
+SGMII is specified.
+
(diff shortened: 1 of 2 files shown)


recipes-kernel/qps615-module/qps615-dlkm/0013-net-ethernet-tc956x-Implement-phy-mode-DT-overlay-fo.patch
```diff
@@ -0,0 +1,72 @@
...
++		return 0;
++	}
++
++	switch (phy_iface) {
++	case PHY_INTERFACE_MODE_SGMII:
++		res->port_interface = ENABLE_SGMII_INTERFACE;
++		res->c45_state = false;
++		res->mdc_clk = TC956XMAC_XGMAC_MDC_CSR_62;
++		break;
++	default:
```
lumag, line 62: This should be an error, unless I miss something.

lumag (changes requested):

recipes-kernel/qps615-module/qps615-dlkm/0013-net-ethernet-tc956x-Implement-phy-mode-DT-overlay-fo.patch
ayaan-anwar: That's true, I think it should print something like "unsupported mode, reverting to the default interface mode".

lumag: No. Just return -EINVAL.

ricardosalveti (approved):

lumag (approved):

PR #1971, opened 2026-04-16
linux: import fix for fastrpc driver

lumag:
Backport a patch from the upstream, fixing the double-free of the remote heap, which should hopefully fix CI failures on db820c.

PR #1999, opened 2026-04-20
linux-qcom-6.18: Add missing bug fixes in memory areas

quic-anane:
Add missing bug fixes in memory areas. This patch series fixes multiple memory handling issues in the FastRPC
driver, primarily around the Audio PD remote heap.

The Audio PD uses a reserved memory-region that is shared between HLOS
and the DSP. Allocating and freeing this memory from userspace is unsafe,
as the kernel cannot reliably determine when the DSP has finished using
the buffers.

To address this, the entire reserved memory-region for the Audio PD is
now fully assigned to the DSP during remoteproc boot-up, and its lifetime
is tied to the rpmsg channel.


quic-anane:
looks like the change is already raised 

https://github.com/qualcomm-linux/meta-qcom/pull/1971

closing this PR

lumag:
@quic-anane that PR is for linux-yocto-6.18, rather than linux-qcom-6.18. For linux-qcom-foo please get those changes merged into the corresponding repo.

PR #2016, opened 2026-04-22
uboot: Bump uboot version to 2026.07-rc1

aswinm94:
Update the uboot version from 2026.04-rc1 to 2026.07-rc1 release

U-Boot v2026.04 introduces a build failure in the mkeficapsule host tool
when building via Yocto/OpenEmbedded.

This is a known upstream issue reported in both U-Boot and OE-Core,
and the required fixes are not yet present in the OpenEmbedded layer,
which has not been updated to fully support v2026.04.

As a temporary workaround, CONFIG_TOOLS_MKEFICAPSULE is disabled to
allow the U-Boot build to succeed until the upstream fixes are
available and integrated.

### DIFF
--- a/recipes-bsp/u-boot/files/disable-eficapsule-tool.cfg
+++ b/recipes-bsp/u-boot/files/disable-eficapsule-tool.cfg
@@ -1,0 +1,1 @@
+# CONFIG_TOOLS_MKEFICAPSULE is not set
--- a/recipes-bsp/u-boot/u-boot-qcom_git.bb
+++ b/recipes-bsp/u-boot/u-boot-qcom_git.bb
@@ -5,9 +5,9 @@ DEPENDS += "bc-native dtc-native gnutls-native python3-pyelftools-native qtestsi
 
 COMPATIBLE_MACHINE:aarch64 = "(qcom)"
 
-PV = "2026.01+2026.04-rc1+git"
+PV = "2026.04+2026.07-rc1+git"
 
-SRCREV = "7b9fb537f19db0fa48b824ecc883c9c8512f0e21"
+SRCREV = "7a82a69bc2e56bdf0df04378b2b6902f995790a6"
 SRCBRANCH = "nobranch=1"
 
 SRC_URI = "git://github.com/qualcomm-linux/u-boot.git;${SRCBRANCH};protocol=https;name=uboot"
--- a/recipes-bsp/u-boot/u-boot-qcom_git.bb
+++ b/recipes-bsp/u-boot/u-boot-qcom_git.bb
@@ -11,6 +11,7 @@ SRCREV = "7a82a69bc2e56bdf0df04378b2b6902f995790a6"
 SRCBRANCH = "nobranch=1"
 
 SRC_URI = "git://github.com/qualcomm-linux/u-boot.git;${SRCBRANCH};protocol=https;name=uboot"
+SRC_URI += "file://disable-eficapsule-tool.cfg"
 
 python __anonymous() {
     ubootconfig = (d.getVar('UBOOT_CONFIG') or "").split()


recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -5,9 +5,9 @@ DEPENDS += "bc-native dtc-native gnutls-native python3-pyelftools-native qtestsi
 
 COMPATIBLE_MACHINE:aarch64 = "(qcom)"
 
-PV = "2026.01+2026.04-rc1+git"
+PV = "2026.04+git"
 
-SRCREV ?= '${@oe.utils.conditional("PREFERRED_PROVIDER_virtual/bootloader", "u-boot-qcom-upstream", "${AUTOREV}", "7b9fb537f19db0fa48b824ecc883c9c8512f0e21", d)}'
+SRCREV := '${@oe.utils.conditional("PREFERRED_PROVIDER_virtual/bootloader", "u-boot-qcom-upstream", "${AUTOREV}", "37ab40bbbf58e9f4d4d019776fdcc137392df7cd", d)}'
```
lumag: The `:=` doesn't seem right.

lumag (changes requested):

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: @lumag, SRCREV is already set in u-boot-common.inc, so it has to be overridden in u-boot-qcom using `:= `

lumag: No. You can overwrite it with the normal '=' after including the `u-boot-common.inc`

aswinm94: @lumag, if we use `=` getting the below error

`/local/mnt/workspace/aswinm/le_linux_lemans/qli_2_0/build/../meta-qcom/recipes-bsp/u-boot/u-boot-qcom_git.bb: AUTOREV/SRCPV set too late for the fetcher to work properly, please set the variables earlier in parsing. Erroring instead of later obtuse build failures.
ERROR: Parsing halted due to errors, see error messages above`

lumag: Ok, so it's not related to version bump. Please document and submit separately (at least as a separate commit).

The main issue with `:=` is that it is expanded at the exact time, so if `PREFERRED_PROVIDER_virtual/bootloader` is changed later, it will be ignored. Likewise, I'm not sure if BB will handle it correctly if one has `SRCREV` override in the conf file.

Would it help if we move class-devupstream and AUTOREV to `u-boot-common.inc`?

quaresmajose: > @lumag, SRCREV is already set in u-boot-common.inc, so it has to be overridden in u-boot-qcom using := 

This means it was already broken before; you must first fix this in a specific commit and then do the bump.

recipes-bsp/u-boot/u-boot-qcom_git.bb
```diff
@@ -19,6 +19,8 @@ SRC_URI = "git://github.com/qualcomm-linux/u-boot.git;${SRCBRANCH};protocol=http
 BBCLASSEXTEND = "devupstream:target"
 PN:class-devupstream = "u-boot-qcom-upstream"
 
+EXTRA_OEMAKE += "CONFIG_TOOLS_MKEFICAPSULE=n"
```
quaresmajose: It would be good to describe the reason for this change in the commit.

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: > The main issue with `:=` is that it is expanded at the exact time, so if `PREFERRED_PROVIDER_virtual/bootloader` is changed later, it will be ignored. Likewise, I'm not sure if BB will handle it correctly if one has `SRCREV` override in the conf file.
> 
> Would it help if we move class-devupstream and AUTOREV to `u-boot-common.inc`?

@lumag 
u-boot-common.c is part of oe-core layer, so we can't modify it.
if `:= ` is not advisable, whether can we remove **devupstream** class support as it may cause unnecessary issues?

lumag: > u-boot-common.c is part of oe-core layer, so we can't modify it.

Of course we can. You just need to send a patch to oe-core ML.

lumag: Anyway, the change is irrelevant for the version bump. Please open a separate PR for discussing it.

recipes-bsp/u-boot/u-boot-qcom_git.bb
ricardosalveti: Yeah, why is this set together with the version bump?

If you need to change the config, do so via a config fragment instead.

lumag:
Please rebase.

b49020:
NAK for this U-Boot bump since it's based on a downstream U-Boot branch: https://github.com/qualcomm-linux/u-boot/commits/u-boot-mainline-1.0/ and not on qcom-next.

@aswinm94 didn't we discussed about that the meta-qcom integration has to closely follow U-Boot upstream and not a dump of downstream patches.

recipes-bsp/u-boot/u-boot-qcom_git.bb.save
```diff
@@ -0,0 +1,52 @@
+require recipes-bsp/u-boot/u-boot-common.inc
```
ricardosalveti: Drop this file.

recipes-bsp/u-boot/files/disable-eficapsule-tool.cfg
```diff
@@ -0,0 +1 @@
+# CONFIG_TOOLS_MKEFICAPSULE is not set
```
ricardosalveti, line 1: Please point to the discussion reported in u-boot and oe-core as part of the commit.

aswinm94: Added the discussion thread as part of commit

aswinm94:
> NAK for this U-Boot bump since it's based on a downstream U-Boot branch: https://github.com/qualcomm-linux/u-boot/commits/u-boot-mainline-1.0/ and not on qcom-next.
> 
> @aswinm94 didn't we discussed about that the meta-qcom integration has to closely follow U-Boot upstream and not a dump of downstream patches.

the srcrev has been updated to point to qcom-next

recipes-bsp/u-boot/u-boot-qcom_git.bb
aswinm94: as suggested now its been handled through config fragment

lumag (dismissed):

aswinm94:
Since uboot qcom-nxt branch has been rebased to 2026.07-rc1 recently, updated the PR to align with it

b49020:
> Since uboot qcom-nxt branch has been rebased to 2026.07-rc1 recently, updated the PR to align with it

Bump to qcom-next based on 2026.07-rc1 sounds fine to me.

b49020 (approved):

ricardosalveti:
But why are you updating to a newer version after this PR was proposed and approved? Since this is a major change I would prefer updating in steps, first to the final 04 release and then to the 07-rc1 (since this is rc1 and it might have other issues).

aswinm94:
> But why are you updating to a newer version after this PR was proposed and approved? Since this is a major change I would prefer updating in steps, first to the final 04 release and then to the 07-rc1 (since this is rc1 and it might have other issues).

@ricardosalveti , when the PR was created the qcom-next branch was based on 2026.04, yesterday we had a 2026.07-rc1 release from uboot which has many of the changes from Qualcomm also the qcom-next branch was rebased to the newer version, hence i updated PR to point to 2026.07-rc1

ricardosalveti:
> > But why are you updating to a newer version after this PR was proposed and approved? Since this is a major change I would prefer updating in steps, first to the final 04 release and then to the 07-rc1 (since this is rc1 and it might have other issues).
> 
> @ricardosalveti , when the PR was created the qcom-next branch was based on 2026.04, yesterday we had a 2026.07-rc1 release from uboot which has many of the changes from Qualcomm also the qcom-next branch was rebased to the newer version, hence i updated PR to point to 2026.07-rc1

Sure, but please avoid changing PRs which were already approved for the previous release, it is perfectly fine to propose a new one later for the new bump, which also helps when we want to bisect possible regressions and issues.

aswinm94:
> > > But why are you updating to a newer version after this PR was proposed and approved? Since this is a major change I would prefer updating in steps, first to the final 04 release and then to the 07-rc1 (since this is rc1 and it might have other issues).
> > 
> > 
> > @ricardosalveti , when the PR was created the qcom-next branch was based on 2026.04, yesterday we had a 2026.07-rc1 release from uboot which has many of the changes from Qualcomm also the qcom-next branch was rebased to the newer version, hence i updated PR to point to 2026.07-rc1
> 
> Sure, but please avoid changing PRs which were already approved for the previous release, it is perfectly fine to propose a new one later for the new bump, which also helps when we want to bisect possible regressions and issues.

@ricardosalveti , qli2.0 required features & bug fixes are present in 2026.07-rc1, so it will be good to bump to that version directly.

ricardosalveti:
> > Sure, but please avoid changing PRs which were already approved for the previous release, it is perfectly fine to propose a new one later for the new bump, which also helps when we want to bisect possible regressions and issues.
> 
> @ricardosalveti , qli2.0 required features & bug fixes are present in 2026.07-rc1, so it will be good to bump to that version directly.

That's a good thing to discuss, as wrynose (which is the release QLI 2.0 will be based on) uses u-boot 2026.04 by default, so this will move us ahead of oe-core, which is fine for master but we might want to discuss for wrynose.

I'm OK using a newer version for wrynose as there is no LTS for u-boot, but I'm not sure releasing on 07-rc1 is ideal, I would prefer to only bump wrynose once we have the final 07 release.

@b49020 @lumag @ndechesne what do you guys thing?

Meanwhile can you move back to the 2026.04 based rev? We can merge it first and then decide how to bump to 07-rc1.

lumag:
> Meanwhile can you move back to the 2026.04 based rev? We can merge it first and then decide how to bump to 07-rc1.

Unless @b49020 recommends otherwise, I'd go with 2026.04.

b49020:
> That's a good thing to discuss, as wrynose (which is the release QLI 2.0 will be based on) uses u-boot 2026.04 by default, so this will move us ahead of oe-core, which is fine for master but we might want to discuss for wrynose.

Not sure if we tightly follow that practice from kernel point of view as well.

> I'm OK using a newer version for wrynose as there is no LTS for u-boot, but I'm not sure releasing on 07-rc1 is ideal, I would prefer to only bump wrynose once we have the final 07 release.

U-Boot LTS has been long discussed but it doesn't exist at all. The U-Boot team's effort are towards mainline enablement only and I think that's the feasible approach only given the limited number of folks working upstream. Surely there will be regular rcX bumps for U-Boot as we already do for the kernel rcX but U-Boot doesn't have the LTS model as the kernel does.

That means all the fixes even after U-Boot main releases land in master branch only. Regarding Yocto wrynose vs mainline, even if we decide 2026.04 then there aren't any plans to support any feature updates or fixes there. Not sure if that would be useful for the customers if regular updates don't come there.

However, if we stick with mainline then for sure it will go through the testing process and any fixes required will get merged in staging tree as well as in mainline.

ricardosalveti:
> Regarding Yocto wrynose vs mainline, even if we decide 2026.04 then there aren't any plans to support any feature updates or fixes there.

So the only "supported" path from our side would be to keep moving it to newer releases even in wrynose. Guess with that it is fine to move it to 07-rc1 then, as it will be the baseline we support.

b49020:
> > Regarding Yocto wrynose vs mainline, even if we decide 2026.04 then there aren't any plans to support any feature updates or fixes there.
> 
> So the only "supported" path from our side would be to keep moving it to newer releases even in wrynose. Guess with that it is fine to move it to 07-rc1 then, as it will be the baseline we support.

Given the resourcing constraints and U-Boot no LTS model, I think that's a reasonable approach to take. Also, we would be tracking the U-Boot patch-set delta in staging tree to be well under control as the upstream work continues.

ricardosalveti:
> As a temporary workaround, CONFIG_TOOLS_MKEFICAPSULE is disabled to allow the U-Boot build to succeed until the upstream fixes are available and integrated.

Just a note to link to an issue so we can revert the changes later.

ricardosalveti (approved):

PR #2023, opened 2026-04-22
mount-tee-partition: add usb_fw mount and xHCI bind rule for qcm6490

akakum-qualcomm:
Add qcm6490-specific installation of a systemd mount unit to mount the usb_fw partition at /var/usbfw during local-fs startup.

Also add a udev rule to bind xHCI PCI devices to xhci_hcd and create a firmware symlink from /lib/firmware/renesas_usb_fw.mem to the firmware image stored in /var/usbfw.

This ensures the Renesas USB firmware is available from the mounted partition and accessible through the standard firmware path.

### DIFF
--- a/recipes-bsp/partition/mount-tee-partition/var-usbfw.mount
+++ b/recipes-bsp/partition/mount-tee-partition/var-usbfw.mount
@@ -1,0 +1,15 @@
+#Copyright (c) 2024 Qualcomm Innovation Center, Inc. All rights reserved.
+#SPDX-License-Identifier: BSD-3-Clause-Clear
+
+[Unit]
+Description=mount usb_fw unit
+Before=local-fs.target
+
+[Mount]
+What=/dev/disk/by-partlabel/usb_fw
+Where=/var/usbfw
+DirectoryMode=0744
+Type=ext4
+
+[Install]
+WantedBy=local-fs.target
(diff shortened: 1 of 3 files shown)


lumag (commented):
This has nothing to do with mounting of TEE partition. Please don't insert unrelated changes into the existing code.

lumag (changes requested):
Not to mention the missing SoB.

recipes-bsp/partition/mount-tee-partition/var-usbfw.mount
```diff
@@ -0,0 +1,15 @@
+#Copyright (c) 2024 Qualcomm Innovation Center, Inc. All rights reserved.
+#SPDX-License-Identifier: BSD-3-Clause-Clear
+
+[Unit]
+Description=mount usb_fw unit
+Before=local-fs.target
+
+[Mount]
+What=/dev/disk/by-partlabel/usb_fw
```
ricardosalveti, line 9: There is no usb_fw partition: https://github.com/qualcomm-linux/qcom-ptool/blob/main/platforms/qcs6490-rb3gen2/ufs/partitions.conf

Also, this is just a really bad workaround for a firmware distribution issue.

How is this supposed to be handled by other OSs such as Debian?

akakum-qualcomm:
> This has nothing to do with mounting of TEE partition. Please don't insert unrelated changes into the existing code.

sure will add separately for usb_fw.

akakum-qualcomm:
> Not to mention the missing SoB.

will udpate in next push.

vkraleti:
I believe this is already discussed in https://github.com/qualcomm-linux/qcom-ptool/pull/57 and decided not to add this partition as it can't be scaled across distributions.

lumag:
I agree with @vkraleti 's comment. Closing.

PR #2024, opened 2026-04-22
u-boot-qcom: drop devupstream AUTOREV support and pin SRCREV

aswinm94:
The devupstream class previously allowed building U-Boot from the tip of the qcom-next branch using AUTOREV. This setup introduced additional complexity and resulted in fragile fetch-time behavior.

Since U-Boot must be built using explicit, validated SRCREV values, remove the devupstream class extension and pin the recipe to a fixed validated revision unconditionally.


lumag (approved):

quaresmajose (approved):

ricardosalveti (approved):

koenkooi (approved):
AUTOREV and its analogues are a bad idea in public facing repos, so I'm glad this is getting removed.

PR #2035, opened 2026-04-23
ci: nighty-build: move back to one nightly per day

ricardosalveti:
The expected stream of changes reduced considerably now that we are approaching RC3, so move back nightly to run only once per day at 0:22 UTC, which should be enough for it to be completed at early IST.

This will help reducing the build & CI pressure we currently have.

This needs to be merged by quic-yocto-ci.


lumag (approved):

koenkooi (approved):

ricardosalveti:
Merging as RC3 is under testing and we will be using new builds from the wrynose branch.

PR #2037, opened 2026-04-24
firmware-qcom-nhlos: make sure that meta firmware package is empty

lumag:
Follow the logic of linux-firmware recipe and set FILES list for PN to an empty value, ensuring that all firmware files are correctly handled and pushed to the individual packages.

### DIFF
--- a/recipes-bsp/firmware/firmware-qcom-nhlos.inc
+++ b/recipes-bsp/firmware/firmware-qcom-nhlos.inc
@@ -64,3 +64,7 @@ do_install:prepend() {
         done
     fi
 }
+
+# Make sure that there is nothing in the meta-package, all firmware must go to
+# a particular package.
+FILES:${PN} = ""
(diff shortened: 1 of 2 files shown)


ricardosalveti (approved):

anujm1 (approved):

recipes-bsp/firmware/firmware-qcom-nhlos.inc
```diff
@@ -50,15 +50,17 @@ do_compile:prepend() {
 }
 
 do_install:prepend() {
-    install -d ${D}${FW_QCOM_PATH}
+    if [ -n "${NHLOS_URI}${PROPRIETARY_URI}" ] ; then
```
quaresmajose, line 53: You can negate the logic and return immediately, and the rest of the function remains the same.

lumag: I'd rather not, because this would prevent some ways to extend the `do_install` in the recipe.

quaresmajose: Good point! Forget what I said.

quaresmajose (approved):

ricardosalveti:
This change is isolated enough that can justify adding to rc3.

PR #2044, opened 2026-04-24
Restrict the SDK for multimedia images

quaresmajose:
We don't need to validate the SDK in the qcom distro and we can restrict it to only multimedia images:
- qcom-multimedia-image
- qcom-multimedia-proprietary-image

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -85,7 +85,9 @@ runs:
         if [ "${INPUTS_MACHINE}" = "qcom-armv8a" ]; then
           # SDK only with the default kernel and qcom-distro
           if [ "${INPUTS_KERNEL_YAML}" = "" ] && [ "${INPUTS_DISTRO_YAML}" = ":ci/qcom-distro.yml" ] ; then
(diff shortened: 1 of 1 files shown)


ricardosalveti:
I believe we only want to generate with the qcom-distro (and maybe qcom-distro-catchall), don't think we need to generate them for the debug and performance builds.

ricardosalveti:
Wonder if better to have this under a specific kas fragment and also as a specific CI matrix job.

quaresmajose:
> Wonder if better to have this under a specific kas fragment and also as a specific CI matrix job.

This is not possible because we need to run with `--task populate_sdk`

quaresmajose:
> I believe we only want to generate with the qcom-distro (and maybe qcom-distro-catchall), don't think we need to generate them for the debug and performance builds.

Rigth, I will change it to just run on qcom-distro.

ricardosalveti (approved):

lumag (approved):

ricardosalveti:
Merging as this is needed to unblock our CI.

PR #2045, opened 2026-04-24
ci/base: override wget fetch command to set curl User-Agent

vkraleti:
Some servers like crates.io, rejecting default wget requests when
no User-Agent header is present, causing fetcher failures.

To overcome this, set FETCHCMD_wget in local.conf with `curl` as a
compatible User-Agent and add retry-connrefused option to retry
on connection refusal.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -45,6 +45,8 @@ local_conf_header:
     git://.*/ git://git.codelinaro.org/clo/yocto-mirrors/ \
     https://.*/.*/ https://artifacts.codelinaro.org/codelinaro-le/ \
     "
+  fetchcmd: |
+    FETCHCMD_wget = "/usr/bin/env wget --tries=2 --timeout=100 --user-agent='curl'"
   cmdline: |
     KERNEL_CMDLINE_EXTRA:append = " qcom_scm.download_mode=1"
   qcomflash: |


quaresmajose:
Can you please say on the commit that the default of FETCHCMD_wget is `/usr/bin/env wget --tries=2 --timeout=100`

quaresmajose (dismissed):

lumag:
And what is the default UA sent by wget?

ci/base.yml
```diff
@@ -45,6 +45,8 @@ local_conf_header:
     git://.*/ git://git.codelinaro.org/clo/yocto-mirrors/ \
     https://.*/.*/ https://artifacts.codelinaro.org/codelinaro-le/ \
     "
+  fetchcmd: |
+    FETCHCMD_wget = "/usr/bin/env wget --tries=2 --timeout=100 --retry-connrefused --user-agent='curl'"
```
ricardosalveti: Agent itself could potentially cause issues, but we can try, but do we really need --retry-connrefused? That would change the flow.

vkraleti:
> And what is the default UA sent by wget?

None. The default command is `/usr/bin/env wget --tries=2 --timeout=100`

ci/base.yml
vkraleti: Felt retry would help to make the command more robust. But to address present issue with crates.io it is not needed. I'll drop it.

ricardosalveti:
Please rebase to fix the CI issues.

lumag (changes requested):
> Some servers like crates.io, rejecting default wget requests when
no User-Agent header is present, causing fetcher failures.

Please specify, what is actually failing. If there is a path in OE that makes wget go to crates.io and fail, it must be fixed in OE-Core rather than being fixed here.

vkraleti:
With https://github.com/rust-lang/simpleinfra/pull/988 Rust maintainers rate limited number of downloads. Seems from Qualnet, this limit is exhausting quite quickly. Need a broader discussion with IT to overcome this situation. 

For now moving this PR to draft as I am not sure if this is right fix.

vkraleti:
A fix for the crates.io fetcher failures has been merged into the BitBake repository:
https://git.openembedded.org/bitbake/commit/?id=f3904634889e737861b47d6bf4068faac4d4661d

Closing this PR since this issue is now resolved in upstream.

PR #1890, opened 2026-04-07
workflows: refactor test workflow

mwasilew:
Test job template rendering was done for each machine/distro combination in a separate workflow. This caused delays due to worker availability and also increased the number of uploaded/downloaded files. With growing number of builds and machines the number of github API calls exceeded the rate limits.

This patch addresses the issue above by removing the list-jobs action and moving all test job generation to lava-test-plans action. For this reason the lava-test-plans action accepts the list of devices (machines) as an input instead of one name as it used to.

On top of that the test flow was split for each distro build making it easier to distinguish the failures based on the distro. For "nodistro" only boot tests are executed and it's clearly visible in the workflow page and the result summary. Current implementation covers: nodistro, qcom-distro and qcom_distro with 6.18 kernel. It's easy to add/remove distro/kernel combinations by adding/removin entries to test.yml.

It's worth noting that test jobs for each distro will be run in parallel. As there are multiple pre-merge test jobs for each board it will result in the number of test jobs greater than the number of available devices for any one build. In case there are more builds happening in parallel github job timeouts are likely to occur.

PR #2047, opened 2026-04-25
ci: support CI on wrynose

ricardosalveti:
Current CI settings are specific to master, separate workflows so they can be branch based and add support for specific wrynose-based workflows.

Needs to be merged by quic-yocto-ci due the cron related changes.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -7,6 +7,10 @@ on:
         required: false
         type: string
         default: "full" # values: pr | full
+      git_ref:
+        required: false
+        type: string
+        default: ""
     outputs:
       artifacts_url:
         description: "URL to retrieve build artifacts"
@@ -37,6 +41,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
+          ref: ${{ inputs.git_ref }}
 
       - name: Run kas lock
         if: ${{ hashFiles('ci/base.lock.yml') == '' }}
@@ -63,6 +68,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
+          ref: ${{ inputs.git_ref }}
 
       - name: Download kas lockfile
         uses: actions/download-artifact@v7
@@ -115,6 +121,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
+          ref: ${{ inputs.git_ref }}
 
       - name: Run kas build
         uses: ./.github/actions/compile
@@ -293,6 +300,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           persist-credentials: false
+          ref: ${{ inputs.git_ref }}
 
       - name: Run kas build
         if: inputs.profile != 'pr' || (matrix.distro.name != 'debug' && matrix.distro.name != 'performance')
@@ -314,6 +322,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
+          ref: ${{ inputs.git_ref }}
 
       - uses: ./.github/actions/print-build-output
         with:
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,4 +1,5 @@
 name: Build on push
+run-name: "${{ github.ref_name }}: ${{ github.event.head_commit.message }}"
 
 on:
   push:
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - wrynose
 
 permissions:
   checks: write
--- a/.github/workflows/repolinter.yml
+++ b/.github/workflows/repolinter.yml
@@ -2,9 +2,9 @@ name: QuIC Organization Repolinter
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, wrynose ]
   pull_request:
-    branches: [ master ]
+    branches: [ master, wrynose ]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/nightly-build-wrynose.yml
+++ b/.github/workflows/nightly-build-wrynose.yml
@@ -1,0 +1,54 @@
+name: Nightly Build (wrynose)
+
+on:
+  schedule:
+  # NOTE - changes to the cron spec should be pushed by https://github.com/quic-yocto-ci
+  # so that build notification emails will be sent out properly.
+  # At 23:22 UTC everyday - picked a random "minute" as top of hour can be busy in github
+  - cron: "22 23 * * *"
+
+permissions:
+  checks: write
+  pull-requests: write
+  contents: read
+  packages: read
+
+jobs:
+  # Resolve wrynose to a single SHA so all downstream jobs build the same commit,
+  # and so test results are attributed to the correct wrynose commit (not master's
+  # github.sha, which is what scheduled workflows expose by default).
+  resolve-ref:
+    runs-on: ubuntu-latest
+    outputs:
+      sha: ${{ steps.get-sha.outputs.sha }}
+    steps:
+      - name: Resolve wrynose SHA
+        id: get-sha
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          SHA=$(gh api repos/${{ github.repository }}/git/ref/heads/wrynose --jq '.object.sha')
+          echo "sha=$SHA" >> $GITHUB_OUTPUT
+
+  build-nightly:
+    needs: resolve-ref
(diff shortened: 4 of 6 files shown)


vkraleti:
Before merging, shouldn't we retarget the wrynose branch to the current master tip?

ricardosalveti:
> Before merging, shouldn't we retarget the wrynose branch to the current master tip?

Already did that, this PR is to reflect that it caused no run after the update :-)

.github/workflows/push.yml
```diff
@@ -3,7 +3,7 @@ name: Build on push
 on:
   push:
     branches:
-      - master
+      - wrynose
```
lumag, line 8: This should be executed for both branches.

.github/workflows/repolinter.yml
```diff
@@ -2,9 +2,9 @@ name: QuIC Organization Repolinter
 
 on:
   push:
-    branches: [ master ]
+    branches: [ wrynose ]
   pull_request:
-    branches: [ master ]
+    branches: [ wrynose ]
```
lumag: Again, both branches.

lumag (changes requested):
You are effectively disabling CI for the master branch. NAK.

ricardosalveti:
> You are effectively disabling CI for the master branch. NAK.

The target branch was supposed to be wrynose, updated that now.

.github/workflows/push.yml
lumag, line 1: Can we please also add a branch name to the workflows so that it's more obvious, for which branch were they executed? I assume we'd need to change the pr.yaml.

lumag (changes requested):

lumag:
> > You are effectively disabling CI for the master branch. NAK.
> 
> The target branch was supposed to be wrynose, updated that now.

Argh. It makes sense with the change of target branch.

.github/workflows/push.yml
ricardosalveti: Yeah, would be nice to name it differently, will have a look.

ricardosalveti: Separated all the runs and changed the target branch to master as we can land all the changes needed there.

ricardosalveti:
Changed back target branch to master and renamed the workflows so they can be separated per branch.

lumag (dismissed):
LGTM

.github/workflows/pr-wrynose.yml
```diff
@@ -0,0 +1,41 @@
+name: Build on PR (wrynose)
```
mwasilew: I don't think this will work. Github only allows to run one version of the workflow with `workflow_run` trigger - from `master` branch (or any other branch designated as "most important").

mwasilew (commented):
Running PRs testing will require some other approach. For now the tests are the same so it will work. Should the tests diverge between `master` and `wrynrose` this implementation will cause issues.

quaresmajose:
Duplicating some of the workflows doesn't seem right to me. This will involve code duplication and won't scale as branches appear. This, in turn, also makes backporting between branches more difficult. This will also, in turn, make backporting between branches more difficult.

We should try to make the workflows the same between branches and introduce the necessary parameters to be executed from different branches.

ricardosalveti:
> Duplicating some of the workflows doesn't seem right to me. This will involve code duplication and won't scale as branches appear

We wanted to have separated workflows as that allows us to easily identify runs based on the workflow / branch used. If we share the same workflow we won't really know from the actions / notifications the type of build, unless you open and see in details.

As we're focused on maintaining master and LTS branches at this point, we won't really have a lot of duplication in the end.

.github/workflows/pr-wrynose.yml
ricardosalveti: But in this case we're defining both workflows in the master branch, and since we're reusing the same test definitions from master, it should work, no?

ricardosalveti:
> Running PRs testing will require some other approach. For now the tests are the same so it will work. Should the tests diverge between `master` and `wrynrose` this implementation will cause issues.

I think we should push to reuse testing the most we can here, since later we will expect changes landing in master and then being backported to wrynose.

If we ever end up with different test sets per release, then we can split this up.

ricardosalveti:
Rebased on top of https://github.com/qualcomm-linux/meta-qcom/pull/1890.

.github/workflows/test-pr.yml
```diff
@@ -1,10 +1,10 @@
-name: Test PR build
+name: Test PR build (master)
```
lumag: If we are testing both master and wrynose, why do we have 'master' in the name?

ricardosalveti: Yes, this is not needed, leftover from the initial duplication, removed.

lumag:
LGTM. @mwasilew, What do you think?

.github/workflows/push-wrynose.yml
```diff
@@ -0,0 +1,31 @@
+name: Build on push (wrynose)
```
ndechesne: push.yaml and push-wrynose.yaml are the same, no? why don't you just add 'wrynose' in push.branches property in push.yaml?

ricardosalveti: Because then the workflow will be the same in github and you won't really know if what was merged was from master or from wrynose without opening up the run.

ndechesne: can you use a variable to get the target branch?

ndechesne: i was kind of hoping we would define the workflow in the branch itself, and leverage reusable workflow from master when possible. adding all the branch specific workflows in the master branch seems odd.

ricardosalveti: Not in name, but there is a run-name which can potentially be used, let me give it a shot.

lumag: > can you use a variable to get the target branch?

It is possible, using the [`run-name`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#run-name)

lumag: > i was kind of hoping we would define the workflow in the branch itself, and leverage reusable workflow from master when possible. adding all the branch specific workflows in the master branch seems odd.

I was hoping for that too. But at this moment I'd really prefer to get it working and unblock the master branch. We can move corresponding woflow bits to the branch later.

ricardosalveti: > i was kind of hoping we would define the workflow in the branch itself

Not all can be defined in there, we will need to reuse from master.

ricardosalveti: Added run-name.

ricardosalveti: I'm not sure I like using run-name, it removes the PR name as we had before and it is still based on the same github workflow: https://github.com/qualcomm-linux/meta-qcom/actions/workflows/pr.yml

This makes it harder to find and identify the specifics of the workflows. Even if we add again the PR name to it, it is not going to be that separated.

.github/workflows/repolinter.yml
```diff
@@ -2,9 +2,9 @@ name: QuIC Organization Repolinter
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, wrynose ]
```
quaresmajose, line 5: Having this in a list with elements per line is easier to understand in my opinion.
```
    branches:
      - master
      - wrynose
```

quaresmajose (dismissed):
LGTM

.github/workflows/build-yocto.yml
```diff
@@ -37,6 +41,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
+          ref: ${{ inputs.git_ref }}
```
ndechesne, line 44: when calling build-pr from master, input.git_ref is not set, so we use the default empty string. 
I am surprised to see that it works. did you do that by 'chance', or did you read somewhere that it should work?   I was expect ref to be set to "" here, but clearly, looking at jobs they checkout  with:

`/usr/bin/git checkout --progress --force refs/remotes/pull/2047/merge
`

ndechesne:
Looking at https://github.com/qualcomm-linux/meta-qcom/actions/runs/25067626837/usage, some builds did not happen (see builds such as build-pr / glymur-crd/debug which are a few seconds, there are a douzen of them). I see no reason why this should have happened..

mwasilew (dismissed):

mwasilew:
This looks OK to me

mwasilew (approved):

quaresmajose (approved):

ricardosalveti:
> Looking at https://github.com/qualcomm-linux/meta-qcom/actions/runs/25067626837/usage, some builds did not happen (see builds such as build-pr / glymur-crd/debug which are a few seconds, there are a douzen of them). I see no reason why this should have happened..

This is not related to this change, we are already not building for debug and performance on prs.

PR #2054, opened 2026-04-26
pg-machine-essential: add missing kernel modules for qcm2290-soc and qcs6490-soc

ricardosalveti:
Add missing kernel modules based on the EVK used by the soc family into the family specific package group.

This should only affect images not importing kernel-modules directly, as the case for the ones from qcom-distro.

### DIFF
--- a/recipes-bsp/packagegroups/packagegroup-machine-essential.bb
+++ b/recipes-bsp/packagegroups/packagegroup-machine-essential.bb
@@ -141,7 +141,15 @@ RRECOMMENDS:${PN}-qcom-qcm2290-soc += " \
     ${PN}-qcom-generic \
     kernel-module-ath10k-snoc \
     kernel-module-dispcc-qcm2290 \
+    kernel-module-dwc3-qcom \
     kernel-module-gpucc-qcm2290 \
+    kernel-module-q6adm \
+    kernel-module-q6afe \
+    kernel-module-q6afe-clocks \
+    kernel-module-q6asm \
+    kernel-module-q6core \
+    kernel-module-qcom-pm8008 \
+    kernel-module-qcom-pm8008-regulator \
 "
 
 RRECOMMENDS:${PN}-qcom-qcs615-soc += " \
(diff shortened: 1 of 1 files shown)


recipes-bsp/packagegroups/packagegroup-machine-essential.bb
```diff
@@ -155,11 +155,13 @@ RRECOMMENDS:${PN}-qcom-qcm2290-soc += " \
 RRECOMMENDS:${PN}-qcom-qcs615-soc += " \
     ${PN}-board-generic \
     ${PN}-qcom-generic \
+    kernel-module-ath11k-ahb \
```
lumag: No, this one is for kodiak. I don't think we have AHB-attached WCN6750 on QCS615 boards.

recipes-bsp/packagegroups/packagegroup-machine-essential.bb
```diff
@@ -155,11 +155,13 @@ RRECOMMENDS:${PN}-qcom-qcm2290-soc += " \
...
     kernel-module-ath11k-pci \
     kernel-module-camcc-qcs615 \
     kernel-module-dispcc-qcs615 \
     kernel-module-gpucc-qcs615 \
     kernel-module-pwrseq-qcom-wcn \
+    kernel-module-snd-soc-da7213 \
```
lumag: This module is not SoC-specific, it is used for the on-board device.

lumag (changes requested):

recipes-bsp/packagegroups/packagegroup-machine-essential.bb
ricardosalveti: Yeah, this is indeed only in kodiak, will drop.

recipes-bsp/packagegroups/packagegroup-machine-essential.bb
ricardosalveti: Yeah, ended up mixing the final board with the shared soc level ones, will drop form this patch at least.

lumag (approved):

koenkooi (approved):
Yay!

PR #2066, opened 2026-04-29
PENDING: ARM: dts: msm: add rpi panel support for IQ9 board

qti-ssmaurya:
Add panel and corresponding regulator nodes.


koenkooi (changes requested):
Please have a look at other commits to that folder to see how the commit messages are supposed to look.

lumag:
```
Upstream-Status: Inappropriate [RPI Debug]
```
From this message I assume this is an internal PR which was opened by mistake. Please don't use meta-qcom PRs in such a case.
If this is intended to be merged, I'd second @koenkooi . Please see how other patches and other commits look like. Please communicate your intentions (in the form of the commit messages and the PR descriptions. Otherwise I will close this PR soon.

ricardosalveti:
A few issues here:
- We don't really want to merge `Inappropriate` patches for recipes and repos we control.
- Reason shows `RPI Debug`, if that is really the idea then this should not land here (as debug related changes should be done locally only)
- This is a linux-qcom patch and should ideally be provided directly by the qcom kernel tree.
- Commit messages not really describing why these patches were proposed.

lumag:
Closing, please get the patches through the kernel repo.

PR #2069, opened 2026-04-29
linux-yocto-6.18: drop fastrpc driver fix

ekanshibu:
Remote heap fix was backported to yocto 6.18 tree. Drop the patch from bbappend to avoid compilation failures.

Source tree changelog: https://git.yoctoproject.org/linux-yocto/log/drivers/misc/fastrpc.c?h=v6.18/standard/base


lumag:
This should only be merged together with the lock file bump.

ricardosalveti:
Merged as part of https://github.com/qualcomm-linux/meta-qcom/pull/2075, thanks!

PR #2073, opened 2026-04-29
ci/qcom-distro: set defaults branch

quaresmajose:
We can simplify and define a default branch for all layers and specify only the branch for divergent cases.

### DIFF
--- a/ci/qcom-distro.yml
+++ b/ci/qcom-distro.yml
@@ -5,6 +5,10 @@ header:
 
 distro: qcom-distro
 
+defaults:
+  repos:
+    branch: master
+
 repos:
   meta-qcom-distro:
     url: https://github.com/qualcomm-linux/meta-qcom-distro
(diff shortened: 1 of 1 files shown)


quaresmajose:
We only need to validate the kas checkout and we don't need to validate a complete build for this, so I cancelled it.

koenkooi:
This is basically a revert of https://github.com/qualcomm-linux/meta-qcom/commit/6292cb2f3c9aa3d98a8feb7b884f90aff8052783, so which school of thought do we want to follow: default branch + exceptions or explicit branch names everywhere?

quaresmajose:
I think _default branch + exceptions_ is preferable because the exceptions are fewer. Only one is different on the master and in the wrynose branch it is expected that there will be no exceptions.

lumag (approved):

vkraleti (approved):

ricardosalveti (approved):

PR #2076, opened 2026-04-30
ci: add workflow for backports

ricardosalveti:
Add workflow to enable support to automatically backport pull-requests as they are merged, based on the tags used at the time of the merge.

Workflow uses https://github.com/korthout/backport-action, at a fixed revision based on the 4.5.0 release.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,21 @@
+name: Backport merged pull request
+on:
+  pull_request_target:
+    types: [closed]
+permissions:
+  contents: write # so it can comment
+  pull-requests: write # so it can create pull requests
+jobs:
+  backport:
+    name: Backport pull request
+    runs-on: ubuntu-latest
+    # Don't run on closed unmerged pull requests
+    if: >
+      github.event.pull_request.merged == true &&
+      github.event.pull_request.base.ref == 'master'
+    steps:
+      - uses: actions/checkout@v6
+      - name: Create backport pull requests
+        uses: korthout/backport-action@7c3f6cd5843cac11bc59a04a1b7699af93261670 # 4.5.0
+        with:
+          label_pattern: "^backport wrynose$"


.github/workflows/backport.yml
```diff
@@ -0,0 +1,17 @@
+name: Backport merged pull request
+on:
+  pull_request_target:
+    types: [closed]
+permissions:
+  contents: write # so it can comment
```
mynameistechno, line 6: I think this is so it can create branches

.github/workflows/backport.yml
```diff
@@ -0,0 +1,17 @@
...
+  pull-requests: write # so it can create pull requests
```
mynameistechno, line 7: This is so it can comment on PRs I imagine

.github/workflows/backport.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  backport:
+    name: Backport pull request
+    runs-on: ubuntu-latest
+    # Don't run on closed unmerged pull requests
+    if: github.event.pull_request.merged
+    steps:
+      - uses: actions/checkout@v6
+      - name: Create backport pull requests
+        uses: korthout/backport-action@7c3f6cd5843cac11bc59a04a1b7699af93261670 # 4.5.0
```
mynameistechno, line 19: Whitelisted this action

mynameistechno (dismissed):
Appears to be fairly low risk despite broad write permissions. pull_request_target is necessary for the write permissions from forked PRs. 

One option to consider is to run it on push to master - working out the PR from the merge commit `GET /repos/{owner}/{repo}/commits/{sha}/pulls` (there are fetch utilities that make this easier) and then passing the pr number to the backport action. Not sure if it would work

ricardosalveti:
> One option to consider is to run it on push to master - working out the PR from the merge commit `GET /repos/{owner}/{repo}/commits/{sha}/pulls` (there are fetch utilities that make this easier) and then passing the pr number to the backport action. Not sure if it would work

It could potentially work since we are not squashing nor rebasing commits on merge, and also every change we merge needs to be done via a pr, but it would need to change the action and make it work for our own use case only (won't work generically as a direct push would not contain PR information).

.github/workflows/backport.yml
ricardosalveti: Thanks!

anujm1:
Should we add a label_pattern as well to match only wrynose at this point?

ricardosalveti:
> Should we add a label_pattern as well to match only wrynose at this point?

Yeah, making it specific to wrynose won't cause any harm at this stage.

ricardosalveti:
Made specific to wrynose and also specific to the 'backport wrynose' label.

anujm1 (dismissed):

.github/workflows/backport.yml
```diff
@@ -0,0 +1,20 @@
...
+    types: [closed]
+permissions:
+  contents: write # so it can comment
+  pull-requests: write # so it can create pull requests
+jobs:
+  backport:
+    name: Backport pull request
+    runs-on: ubuntu-latest
+    # Don't run on closed unmerged pull requests
+    if: github.event.pull_request.merged
```
quaresmajose: Shouldn't we somehow say that these are merged PRs on the master branch? What happens if the PR is merged into a branch other than master?

quaresmajose:
Does this action add the SHA of the commit from which it came? like commits created with -x

anujm1:
> Does this action add the SHA of the commit from which it came? like commits created with -x

Looks like it: https://github.com/korthout/backport-action#how-it-works says

> The commits are cherry-picked with the [-x](https://git-scm.com/docs/git-cherry-pick#Documentation/git-cherry-pick.txt--x) flag.

ricardosalveti:
> Does this action add the SHA of the commit from which it came? like commits created with -x

Yup, that is taken care by the bot.

.github/workflows/backport.yml
ricardosalveti: The action could support backports from any branch (e.g. backporting a change that is not available in master from one LTS into another LTS), so we could restrict but I don't see this as an issue as atm it would show up as wrynose -> wrynose, which wouldn't work anyway.

ndechesne: but then when we merge a PR in master, backport will create a PR , and when we merge that PR, backport will also run, no?

ndechesne: we probably want to be very explicit that we want to merge PR merged in master?

quaresmajose: > but then when we merge a PR in master, backport will create a PR , and when we merge that PR, backport will also run, no?

yeah that was my concern too motivation for the question.

quaresmajose:
I gained more by reading the documentation before asking questions.

.github/workflows/backport.yml
ricardosalveti: > but then when we merge a PR in master, backport will create a PR , and when we merge that PR, backport will also run, no?

The new PR won't have the label and it won't trigger the action.

ndechesne: maybe i did not understand how it works.. but i understood that with target_branches set, then it will backport any merged PR automatically, even the ones without the label.

ricardosalveti: yeah, we don't actually need both, moved to label pattern only.

Also added a protection to only run the action when merged in master only.

anujm1 (approved):

quaresmajose (approved):

PR #2085, opened 2026-04-30
[Wrynose] Switch default kernel to linux-qcom-6.18

vkraleti:
On QLI 2.0, linux-qcom-6.18 is the default kernel. Since the wrynose
 branch is available now for QLI 2.0 development, set linux-qcom-6.18
 as the default kernel for machines that were previously building
 against linux-qcom-next.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -228,9 +228,9 @@ jobs:
           - type: default
             dirname: ""
             yamlfile: ""
-          - type: 6.18
-            dirname: "_linux-qcom-6.18"
-            yamlfile: ":ci/linux-qcom-6.18.yml"
+          - type: qcom-next
+            dirname: "_linux-qcom-next"
+            yamlfile: ":ci/linux-qcom-next.yml"
           - type: rt-6.18-distro-kvm
             dirname: "_linux-qcom-rt-6.18_qcom-distro-kvm"
             yamlfile: ":ci/linux-qcom-rt-6.18.yml:ci/qcom-distro-kvm.yml"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,9 +40,9 @@ jobs:
       distro_name: "nodistro"
       lava_test_plans_ref: "${{ needs.prepare-env.outputs.lava-test-plans-ref }}"
       testkit_ref: "${{ needs.prepare-env.outputs.testkit-ref }}"
-  test-qcom-distro:
+  test-qcom-distro_linux-qcom-next:
     needs: [prepare-env]
-    name: "Test qcom-distro"
+    name: "Test qcom-distro_linux-qcom-next"
     uses: ./.github/workflows/test-distro.yml
     secrets: inherit
     with:
@@ -51,11 +51,12 @@ jobs:
       devices_premerge: "glymur-crd,iq-8275-evk,iq-9075-evk,kaanapali-mtp,qcm6490-idp,qcs615-ride,rb3gen2-core-kit,qcs8300-ride-sx,qcs9100-ride-sx,rb1-core-kit"
       project_name: "meta-qcom"
       distro_name: "qcom-distro"
+      distro_suffix: "_linux-qcom-next"
       lava_test_plans_ref: "${{ needs.prepare-env.outputs.lava-test-plans-ref }}"
       testkit_ref: "${{ needs.prepare-env.outputs.testkit-ref }}"
-  test-qcom-distro_linux-qcom-6-18:
+  test-qcom-distro:
     needs: [prepare-env]
-    name: "Test qcom-distro_linux-qcom-6.18"
+    name: "Test qcom-distro"
     uses: ./.github/workflows/test-distro.yml
     secrets: inherit
     with:
@@ -64,13 +65,12 @@ jobs:
       devices_premerge: "iq-8275-evk,iq-9075-evk,qcm6490-idp,qcs615-ride,rb3gen2-core-kit,qcs8300-ride-sx,qcs9100-ride-sx,rb1-core-kit"
       project_name: "meta-qcom"
       distro_name: "qcom-distro"
-      distro_suffix: "_linux-qcom-6.18"
       lava_test_plans_ref: "${{ needs.prepare-env.outputs.lava-test-plans-ref }}"
       testkit_ref: "${{ needs.prepare-env.outputs.testkit-ref }}"
 
   collect-ids:
     name: "Collect Summary artifact IDs"
-    needs: [test-nodistro, test-qcom-distro, test-qcom-distro_linux-qcom-6-18]
+    needs: [test-nodistro, test-qcom-distro, test-qcom-distro_linux-qcom-next]
     if: |
       always()
       && contains(needs.*.result, 'success')
@@ -85,4 +85,4 @@ jobs:
       - name: Summary IDs
         id: generate-summary
         run: |
-          echo "artifact_ids=${{ needs.test-nodistro.outputs.summary_id }},${{ needs.test-qcom-distro.outputs.summary_id }},${{ needs.test-qcom-distro_linux-qcom-6-18.outputs.summary_id }}" >> $GITHUB_OUTPUT
+          echo "artifact_ids=${{ needs.test-nodistro.outputs.summary_id }},${{ needs.test-qcom-distro.outputs.summary_id }},${{ needs.test-qcom-distro_linux-qcom-next.outputs.summary_id }}" >> $GITHUB_OUTPUT
--- a/conf/machine/include/qcom-base.inc
+++ b/conf/machine/include/qcom-base.inc
@@ -1,7 +1,7 @@
 # Common configurations and variables for Qualcomm platforms.
 
 # Provider for linux kernel
-PREFERRED_PROVIDER_virtual/kernel ?= "linux-qcom-next"
+PREFERRED_PROVIDER_virtual/kernel ?= "linux-qcom"
 
 KERNEL_IMAGETYPE ?= "Image"
 KERNEL_ALT_IMAGETYPE ?= "vmlinux"


ricardosalveti:
This might bring the need to have different testing workflows (or at least per branch changes?), since the default run is now 6.18 based and we won't have 6.18 anymore as a separated flow.

Guess it would be smart enough to not generate the 6.18 runs and trigger the standard ones (which will be 6.18 based)?

@mwasilew ^^

vkraleti:
While we finalize test workflows, is it acceptable to build qcom-6.18 kernel in both default and 6.18 specific builds?

mwasilew:
> This might bring the need to have different testing workflows (or at least per branch changes?), since the default run is now 6.18 based and we won't have 6.18 anymore as a separated flow.
> 
> Guess it would be smart enough to not generate the 6.18 runs and trigger the standard ones (which will be 6.18 based)?
> 

It would be easy to fix if there were only push based test workflows. I don't know how to address it with PR workflows. For push we can just update the workflow file in the branch.

.github/workflows/build-yocto.yml
```diff
@@ -173,9 +173,6 @@ jobs:
           - type: default
             dirname: ""
             yamlfile: ""
-          - type: 6.18
-            dirname: "_linux-qcom-6.18"
-            yamlfile: ":ci/linux-qcom-6.18.yml"
```
sbanerjeeqc, line 233: On Wrynose we retain 6.18. Isn't so?

lumag:
> While we finalize test workflows, is it acceptable to build qcom-6.18 kernel in both default and 6.18 specific builds?

Wouldn't that mean duplicate builds? I'd rather drop one.

.github/workflows/build-yocto.yml
vkraleti: Yes. This PR switches default kernel to 6.18  and dropping duplicate 6.18 workflow.

lumag (changes requested):
> Also remove qcom-next-rt
entry for iq-9075-evk as it is no longer needed.

Why? Also usually all the "also foo" should be split to separate commits.

koenkooi (dismissed):

lumag:
Fails to build

.github/workflows/build-yocto.yml
ricardosalveti: Should these files also be removed from the branch? They won't be tested once removed from our CI, and won't also be necessarily useful for our users.

ricardosalveti: Same for the linux-qcom-next recipe.

vkraleti: Yes, they can be dropped.

lumag (changes requested):
As  discussed today, please switch the default to qcom-6.18, but keep qcom-next in the repo (and in the CI loop).

.github/workflows/build-yocto.yml
vkraleti: Latest guidance is to support two LTS kernels on QLI 2.0. To catch up with next LTS, linux-qcom-next recipes are needed.  Will consider dropping them post 2026 Kernel LTS release.

conf/machine/include/qcom-base.inc
```diff
@@ -1,7 +1,8 @@
 # Common configurations and variables for Qualcomm platforms.
 
 # Provider for linux kernel
-PREFERRED_PROVIDER_virtual/kernel ?= "linux-qcom-next"
+PREFERRED_PROVIDER_virtual/kernel ?= "linux-qcom"
+PREFERRED_VERSION_virtual/kernel ?= "6.18%"
```
quaresmajose: Specifying the version is redundant because it's the only one we have.

quaresmajose (commented):
The first commit needs to be written because it is no longer up to date.

conf/machine/include/qcom-base.inc
vkraleti: QLI 2.0 plans to support two linux-qcom LTS kernels. If a preferred version is not explicitly specified, newly added kernel recipe gets selected automatically by the build. Explicitly specifying the kernel version avoids this auto selection. However this can be changed at the time of introducing the new kernel recipe.

lumag: And it was discussed that SoC might switch from one to another on their own. So, no, this is not a correct way to do it,

lumag: Ah, this is the bb include, not the CI include. Then I'm sorry, my comment is not relevant.

lumag (changes requested):
> Support for linux-qcom-next will be dropped in subsequent commits.

This is no longer true.

conf/machine/include/qcom-base.inc
quaresmajose: I understand and agree that it will be necessary when we have versions to choose from. However, there's nothing wrong with the PREFERRED_VERSION being defined; it's just redundant.

vkraleti:
> > Support for linux-qcom-next will be dropped in subsequent commits.
> 
> This is no longer true.

Updated commit message.

quaresmajose (dismissed):

lumag (dismissed):

lumag:
@mwasilew ?

ricardosalveti (dismissed):

ricardosalveti:
We will also have to update the axiom related jobs for wrynose.

mwasilew:
> @mwasilew ?

Sorry, I was off yesterday. It seems that the change in `test.yml` is missing. 
https://github.com/qualcomm-linux/meta-qcom/blob/master/.github/workflows/test.yml#L56
This will fail as there is no 6.18 build any more.

mwasilew:
Here is the failed workflow: https://github.com/qualcomm-linux/meta-qcom/actions/runs/26052024580

vkraleti:
Thank you @mwasilew . Updated test workflows as well. Please check.

mwasilew (dismissed):
This should work.

mwasilew:
> Thank you @mwasilew . Updated test workflows as well. Please check.

It looks correct. Should we push it to `next` to try out or just merge and hope for the best?

.github/workflows/test.yml
```diff
@@ -47,8 +47,8 @@ jobs:
     secrets: inherit
     with:
       build_id: "${{ inputs.build_id }}"
-      devices: "glymur-crd,iq-8275-evk,iq-9075-evk,kaanapali-mtp,qcm6490-idp,qcs615-ride,rb3gen2-core-kit,qcs8300-ride-sx,qcs9100-ride-sx,rb1-core-kit"
-      devices_premerge: "glymur-crd,iq-8275-evk,iq-9075-evk,kaanapali-mtp,qcm6490-idp,qcs615-ride,rb3gen2-core-kit,qcs8300-ride-sx,qcs9100-ride-sx,rb1-core-kit"
+      devices: "iq-8275-evk,iq-9075-evk,qcm6490-idp,qcs615-ride,rb3gen2-core-kit,qcs8300-ride-sx,qcs9100-ride-sx,rb1-core-kit"
+      devices_premerge: "iq-8275-evk,iq-9075-evk,qcm6490-idp,qcs615-ride,rb3gen2-core-kit,qcs8300-ride-sx,qcs9100-ride-sx,rb1-core-kit"
```
lumag: No. There is a 6.18 build few lines below. Adapt this one and the 6.18 configurations to build corresponding kernels rather than dropping tests here.

.github/workflows/build-yocto.yml
```diff
@@ -228,9 +228,6 @@ jobs:
           - type: default
             dirname: ""
             yamlfile: ""
-          - type: 6.18
```
lumag, line 231: Why are you dropping it instead of changing to qcom-next?

.github/workflows/build-yocto.yml
```diff
@@ -231,6 +231,9 @@ jobs:
           - type: rt-6.18-distro-kvm
             dirname: "_linux-qcom-rt-6.18_qcom-distro-kvm"
             yamlfile: ":ci/linux-qcom-rt-6.18.yml:ci/qcom-distro-kvm.yml"
+          - type: qcom-next
+            dirname: "_linux-qcom-next"
+            yamlfile: ":ci/linux-qcom-next.yml"
```
lumag, line 233: I see. Squash into the previous commit.

.github/workflows/test.yml
```diff
@@ -53,9 +53,9 @@ jobs:
       distro_name: "qcom-distro"
       lava_test_plans_ref: "${{ needs.prepare-env.outputs.lava-test-plans-ref }}"
       testkit_ref: "${{ needs.prepare-env.outputs.testkit-ref }}"
-  test-qcom-distro_linux-qcom-6-18:
+  test-qcom-distro_linux-qcom-next:
```
lumag: And now we miss tests on glymur.

lumag (changes requested):

.github/workflows/build-yocto.yml
vkraleti: Done.

.github/workflows/test.yml
vkraleti: Swapped qcom-6.18 and qcom-next entries in workflow. Same set of targets should build now.

.github/workflows/build-yocto.yml
vkraleti: Squashed commits to indicate it is a swap.

mwasilew (commented):
I assume the list of devices the tests run on is correct for each build. Please double check if this is the case. Currently for `qcom-distro` (6.18) the following devices are tested:
```
iq-8275-evk,iq-9075-evk,qcm6490-idp,qcs615-ride,rb3gen2-core-kit,qcs8300-ride-sx,qcs9100-ride-sx,rb1-core-kit
```
For `qcom-next` the following devices are tested:
```
glymur-crd,iq-8275-evk,iq-9075-evk,kaanapali-mtp,qcm6490-idp,qcs615-ride,rb3gen2-core-kit,qcs8300-ride-sx,qcs9100-ride-sx,rb1-core-kit
```

The difference is `glymur-crd,kaanapali-mtp`. Are these not supported and not built in 6.18?

vkraleti:
Thanks for cross checking @mwasilew.  In current master branch tip, Glymur-crd and Kaanapali-MTP don't have support on the 6.18 kernel.  So omitting these is fine.

lumag (approved):

quaresmajose (approved):

ricardosalveti (approved):
LGTM, but let's just understand first the axiom integration changes that are required, otherwise that will be completely broken once this is merged.

vkraleti:
@smuppand @bhargav0610 Can you please update Axiom configurations to accommodate this switch?

quic-viskuma:
> @smuppand @bhargav0610 Can you please update Axiom configurations to accommodate this switch?

Coral/Axiom configuration updates are required after this PR is merged; otherwise, other PRs and nightly runs may be affected.

vkraleti:
Thanks for confirming. I am merging this PR. Please update Coral/Axiom configurations. @smuppand @bhargav0610 FYI.

bhargav0610:
> Thanks for confirming. I am merging this PR. Please update Coral/Axiom configurations. @smuppand @bhargav0610 FYI.

sure

PR #2088, opened 2026-04-30
ci: backport: add missing capture group

ricardosalveti:
Action uses a matching capture group to define the target branch to use, so make sure wrynose is inside a group in order to make it work for wrynose labels.


ricardosalveti:
Merging as an attempt to have a working action, will cancel the merged workflow.

PR #2098, opened 2026-05-03
ci: skip nightly builds when inputs are unchanged

### LINKED ISSUE #1662: Skip Duplicate Nightly Runs When No New Commits
Currently, the nightly GitHub Actions are being triggered twice after including this change:
https://github.com/qualcomm-linux/meta-qcom/pull/1652.

Below actions are being triggered with same commit: 
https://github.com/qualcomm-linux/meta-qcom/commit/1a7b6b0363fa75232d46ffb9fd706b9539872e10 

https://github.com/qualcomm-linux/meta-qcom/actions/runs/22517222402
https://github.com/qualcomm-linux/meta-qcom/actions/runs/22528490430
https://github.com/qualcomm-linux/meta-qcom/actions/runs/22539595186

It would be better to avoid triggering the nightly jobs when there are no new commits or changes since the previous run. we can prevent redundant executions and unnecessary resource usage/memory.

ricardosalveti:
Limit nightly CI runs to cases where effective build inputs change, avoiding duplicate runs with identical results.

Trigger the nightly build only when one or more of the following changes:
- the meta-qcom layer itself
- updated ci/base.lock.yml (based on kas-setup)

Use the content of base.lock.yml to detect changes in pinned dependencies.

Derive and store a combined hash from:
- updated base.lock.yml
- meta-qcom tree state

Compare this hash against the previous nightly run to determine whether to execute the build, test, and publish stages.

Fixes: https://github.com/qualcomm-linux/meta-qcom/issues/1662

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -7,10 +7,18 @@ on:
         required: false
         type: string
         default: "full" # values: pr | full
+      skip_if_github_cached:
+        required: false
+        type: boolean
+        default: false
+        description: "Skip the build matrix if the github actions cache already records a previously-successful run for the same inputs"
     outputs:
       artifacts_url:
         description: "URL to retrieve build artifacts"
         value: ${{ jobs.create-output.outputs.url }}
+      build_skipped:
+        description: "True when the compile matrix was skipped because the github cache recorded a previously-successful build for the same inputs"
+        value: ${{ jobs.github-cache-check.outputs.cache_hit }}
 
 # Concurrency is used to prevent multiple workflows from running for the same PR
 concurrency:
@@ -54,9 +62,62 @@ jobs:
           name: kas-container
           path: ${{ env.KAS_CONTAINER }}
 
-  yocto-run-checks:
+  github-cache-check:
     needs: kas-setup
     if: github.repository_owner == 'qualcomm-linux'
+    runs-on: ubuntu-latest
+    outputs:
+      cache_hit: ${{ steps.restore.outputs.cache-hit == 'true' }}
+      run_hash: ${{ steps.hash.outputs.run_hash }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+          ref: ${{ inputs.git_ref }}
+
+      - name: Download kas lockfile
+        uses: actions/download-artifact@v7
+        with:
+          name: kas-lockfile
+          path: ci
+
+      - name: Compute build input hash
+        id: hash
+        run: |
+          # Hash inputs that fully determine the build:
+          # - the resolved base.lock.yml produced by `kas lock` (pins every layer)
+          # - the meta-qcom revision being built
+          LOCK_HASH=$(sha256sum ci/base.lock.yml | cut -d' ' -f1)
+          SRC_HASH=$(git rev-parse HEAD)
+          RUN_HASH=$(printf '%s\n%s' "$LOCK_HASH" "$SRC_HASH" | sha256sum | cut -d' ' -f1)
+          echo "base.lock.yml hash: $LOCK_HASH"
+          echo "meta-qcom HEAD: $SRC_HASH"
+          echo "Combined run hash: $RUN_HASH"
+          echo "run_hash=$RUN_HASH" >> "$GITHUB_OUTPUT"
+
+      - name: Look up previous successful build marker in github cache
+        id: restore
+        uses: actions/cache/restore@v5
+        with:
+          path: /tmp/build-success-marker
+          key: build-success-${{ github.repository }}-${{ github.workflow }}-${{ github.ref_name }}-${{ steps.hash.outputs.run_hash }}
+          lookup-only: true
+
+      - name: Report github-cache-check decision
+        run: |
+          if [ "${{ inputs.skip_if_github_cached }}" != "true" ]; then
+            echo "skip_if_github_cached not set; build will run"
+          elif [ "${{ steps.restore.outputs.cache-hit }}" = "true" ]; then
+            echo "Run cache hit on key build-success, hash ${{ steps.hash.outputs.run_hash }}: skipping build matrix"
+          else
+            echo "No previous successful build run recorded in github cache for this hash: build will run"
+          fi
+
+  yocto-run-checks:
+    needs: [kas-setup, github-cache-check]
+    if: |
+      github.repository_owner == 'qualcomm-linux' &&
+      (inputs.skip_if_github_cached != true || needs.github-cache-check.outputs.cache_hit != 'true')
     runs-on: [self-hosted, qcom-u2404, amd64]
     steps:
       - uses: actions/checkout@v6
@@ -90,8 +151,10 @@ jobs:
           ci/kas-container-shell-helper.sh ci/yocto-check-layer.sh
 
   compile_warm_up:
-    needs: kas-setup
-    if: github.repository_owner == 'qualcomm-linux'
+    needs: [kas-setup, github-cache-check]
+    if: |
+      github.repository_owner == 'qualcomm-linux' &&
+      (inputs.skip_if_github_cached != true || needs.github-cache-check.outputs.cache_hit != 'true')
     runs-on: [self-hosted, qcom-u2404, amd64]
     strategy:
       fail-fast: true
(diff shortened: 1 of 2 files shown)


ricardosalveti:
Not yet fully validated, will validate on https://github.com/qualcomm-linux-stg/meta-qcom earlier next week.

ricardosalveti:
Validated with https://github.com/qualcomm-linux-stg/meta-qcom/actions/runs/25411207741 and https://github.com/qualcomm-linux-stg/meta-qcom/actions/runs/25411404458.

ndechesne:
My first reaction is that it's adding some fairly complicated logic into a CI that's already fairly complicated.. 
In the last year, we had 254 days with a merge commit, and 112 days with no merge commits, in meta-qcom. How much saving are we doing overall , given that the CI is most likely largely made of PR builds (and rebuilds) , I am not sure.

Also, you are not taking meta-qcom-distro into account here, is it intentional?

ndechesne:
> My first reaction is that it's adding some fairly complicated logic into a CI that's already fairly complicated.. In the last year, we had 254 days with a merge commit, and 112 days with no merge commits, in meta-qcom. How much saving are we doing overall , given that the CI is most likely largely made of PR builds (and rebuilds) , I am not sure.
> 
> Also, you are not taking meta-qcom-distro into account here, is it intentional?

Looking at current usage, nightly builds are 20% of our usage, so the savings would be ~6.5% (from a very high level estimate)

ricardosalveti:
> My first reaction is that it's adding some fairly complicated logic into a CI that's already fairly complicated.. In the last year, we had 254 days with a merge commit, and 112 days with no merge commits, in meta-qcom. How much saving are we doing overall , given that the CI is most likely largely made of PR builds (and rebuilds) , I am not sure.

Not only builds, nightlys are also the main target that is used internally for validation with axiom, so any run we not do we would be saving on build, storage, transfer to hyderabad and testing in axiom, saving a total of 15h or more, based on our current runs.

> Also, you are not taking meta-qcom-distro into account here, is it intentional?

Every layer is taken into account as I'm hashing the final base.lock file which is generated after the lock step, and after https://github.com/qualcomm-linux/meta-qcom/pull/2112 meta-qcom-distro is also part of it.

.github/workflows/build-yocto.yml
```diff
@@ -54,9 +62,62 @@ jobs:
...
+          path: ci
+
+      - name: Compute build input hash
+        id: hash
+        run: |
+          # Hash inputs that fully determine the build:
+          # - the resolved base.lock.yml produced by `kas lock` (pins every layer)
+          # - the meta-qcom revision being built
+          LOCK_HASH=$(sha256sum ci/base.lock.yml | cut -d' ' -f1)
+          SRC_HASH=$(git rev-parse HEAD)
```
ndechesne, line 91: you should have meta-qcom SHA in base.lock, no?

.github/workflows/build-yocto.yml
```diff
@@ -327,10 +394,36 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create build success marker
+        run: date -u +%FT%TZ > /tmp/build-success-marker
+
+      - name: Save successful build marker to github cache
+        uses: actions/cache/save@v5
+        with:
+          path: /tmp/build-success-marker
+          key: build-success-${{ github.repository }}-${{ github.workflow }}-${{ github.ref_name }}-${{ needs.github-cache-check.outputs.run_hash }}
```
ndechesne, line 413: one key per hash/build, it will be lots of keys. how about you put the hash as file and keep the key to be build-success-${{ github.repository }}-${{ github.workflow }}-${{ github.ref_name }} ?

.github/workflows/build-yocto.yml
ricardosalveti: No, because it is the base layer, the only one out of the lock.

.github/workflows/build-yocto.yml
ricardosalveti: The idea is to check for the existence of the cache, so using the hash is the easiest way to just verify if it exists or not without having to check the content.

Also, if we just use repo + workflow + ref_name we can easily have one run creating a cache that is not really unique.

Cache is automatically removed after 7 days if not really consumed by the build. Also, it is only executed when skip_if_github_cached is set to true, which is only set on nightly.

ricardosalveti:
> Not only builds, nightlys are also the main target that is used internally for validation with axiom, so any run we not do we would be saving on build, storage, transfer to hyderabad and testing in axiom, saving a total of 15h or more, based on our current runs.

To add a bit more, we also have nightly for both master and wrynose now, so anything we can do to remove duplicated runs the better.

ndechesne:
> > Not only builds, nightlys are also the main target that is used internally for validation with axiom, so any run we not do we would be saving on build, storage, transfer to hyderabad and testing in axiom, saving a total of 15h or more, based on our current runs.
> 
> To add a bit more, we also have nightly for both master and wrynose now, so anything we can do to remove duplicated runs the better.

that's ok, you convinced me :)

anujm1 (approved):

quaresmajose:
Wouldn't it make sense to move the `kas-setup` to a new workflow and perform the hash validation here? Then we wouldn't need to modify the `build-yocto`. 

It only makes sense to validate the hash in the scheduled builds because there are always differences in all the others. So, in the `.github/workflows/nightly-build.yml` schedule, we would have the first step with the new `uses: ./.github/workflows/kas-setup.yml` that would give us information on whether not there are differences in the hash, and this would condition the second step where the build would be done with `uses: ./.github/workflows/build-yocto.yml`.

However, I don't know if this would work, and I would need to do some testing to validate what I'm suggesting.

ricardosalveti:
> Wouldn't it make sense to move the `kas-setup` to a new workflow and perform the hash validation here? Then we wouldn't need to modify the `build-yocto`.

I thought about that, but I think it would complicate the setup a bit more, and more files to take care when backporting to wrynose. One issue is that while we could separate the kas-setup for verifying the cache, the saving of the cache needs to happen after a successful compile step from build-yocto, so in order to save that outside we would need to add another parameter and bring the logic outside it.

quaresmajose:
> > Wouldn't it make sense to move the `kas-setup` to a new workflow and perform the hash validation here? Then we wouldn't need to modify the `build-yocto`.
> 
> I thought about that, but I think it would complicate the setup a bit more, and more files to take care when backporting to wrynose. One issue is that while we could separate the kas-setup for verifying the cache, the saving of the cache needs to happen after a successful compile step from build-yocto, so in order to save that outside we would need to add another parameter and bring the logic outside it.

We can use our persistent storage to save the cache. The file would be placed there temporarly at the beginning by the kas-setup and will be marked ok and moved at the end of the build-yocto. 

This way we would always have a cache file for each branch in persistent storage that was updated with each build of the master branch. At the start of the nightly build, we check if there are any differences between what we're about to start with and what we have in persistent storage, which was updated in the last build.

quaresmajose:
> > > Wouldn't it make sense to move the `kas-setup` to a new workflow and perform the hash validation here? Then we wouldn't need to modify the `build-yocto`.
> > 
> > 
> > I thought about that, but I think it would complicate the setup a bit more, and more files to take care when backporting to wrynose. One issue is that while we could separate the kas-setup for verifying the cache, the saving of the cache needs to happen after a successful compile step from build-yocto, so in order to save that outside we would need to add another parameter and bring the logic outside it.
> 
> We can use our persistent storage to save the cache. The file would be placed there temporarly at the beginning by the kas-setup and will be marked ok and moved at the end of the build-yocto.
> 
> This way we would always have a cache file for each branch in persistent storage that was updated with each build of the master branch. At the start of the nightly build, we check if there are any differences between what we're about to start with and what we have in persistent storage, which was updated in the last build.

I can try to implement this instead of blocking your work.

ricardosalveti:
> We can use our persistent storage to save the cache. The file would be placed there temporarly at the beginning by the kas-setup and will be marked ok and moved at the end of the build-yocto.
> 
> This way we would always have a cache file for each branch in persistent storage that was updated with each build of the master branch. At the start of the nightly build, we check if there are any differences between what we're about to start with and what we have in persistent storage, which was updated in the last build.

I think this brings more complexity because we would rely on our persistent storage for everything related to the lock status, and that is why I moved to check for github cache variable instead (using the hash as part of the key).

ricardosalveti:
> I can try to implement this instead of blocking your work.

Sure, we can merge this one if you are OK and try to optimize it further later.

quaresmajose:
> > I can try to implement this instead of blocking your work.
> 
> Sure, we can merge this one if you are OK and try to optimize it further later.

Sure, I will experiment the approach in a followup to see if works

quaresmajose (approved):

PR #2099, opened 2026-05-03
ci: pr: allow committers to trigger CI via /run-ci comment

ricardosalveti:
PRs opened by the backport workflow do not trigger pr.yml because github blocks workflow runs that are created by GITHUB_TOKEN- authored pull_request events. As a result, required PR checks never run on backport PRs and they cannot be merged.

Add a comment-driven pr-comment-ci.yml workflow that listens for issue_comment events, validates that the commenter has write or admin access to the repository, resolves the PR head SHA, and either reruns the most recent pr.yml run for that SHA (failed runs) or dispatches pr.yml against the PR head branch when no prior run exists.

Rerun is preferred because it preserves the original pull_request event context, so checks attach to the PR head commit for both same-repo and fork PRs with the same job names branch protection already requires. The dispatch fallback handles the backport-action case where github blocked the initial trigger; backport PRs are always same-repo so the dispatch ref is valid.

Extend pr.yml to accept workflow_dispatch and gate the event-file and pr-to-workflow-dependency jobs on pull_request, since they rely on pull_request event context that is not available under dispatch (which is OK since the current CR update workflow does not cover backports).

### DIFF
--- a/.github/workflows/pr-comment-ci.yml
+++ b/.github/workflows/pr-comment-ci.yml
@@ -1,0 +1,76 @@
+name: Run PR CI on /run-ci comment
+
+# A committer (write/admin) can comment `/run-ci` on a pull request to
+# (re-)run pr.yml against the PR head. Intended for PRs created by the
+# backport workflow, which github does not auto-CI because they are opened
+# with GITHUB_TOKEN.
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  contents: read
+  pull-requests: read
+  actions: write   # required to dispatch pr.yml
+
+jobs:
+  dispatch:
+    if: >
+      github.event.issue.pull_request != null &&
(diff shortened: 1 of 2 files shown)


ricardosalveti:
@mynameistechno one possible workaround for running CI on prs created by the backport workflow (due limitation in github), which also ends up being useful for re-running failed jobs via comments.

Looks like using PAT would get backport PRs to run normally, but it requires a service account (we do have one but with higher privileges), so creating this one fist for discussion.

ricardosalveti:
> Looks like using PAT would get backport PRs to run normally, but it requires a service account (we do have one but with higher privileges), so creating this one fist for discussion.

Looking a bit further the only solution that would work properly is via a PAT, as at that point it generates a valid github.pull_request event.

Will experiment and evaluate that one first.

PR #2103, opened 2026-05-04
qps615-dlkm: fall back to random MAC address when none is predefined

ayaan-anwar:
Rb3Gen2 lacks an EEPROM chip and the bootloader does not append a MAC address into the DTB. As a result, all boards running QLI 2.0 end up with the same static fallback MAC addresses on both ports, which can cause DHCP lease conflicts across the network.

Fix this by assigning a random MAC address (with NET_ADDR_RANDOM) when both of_get_mac_address() and parse_config_file() fail to provide one. This allows a userspace network manager (e.g. NetworkManager) to detect the DEV_ADDR_RANDOM flag and persist a stable, board-unique address for subsequent boots.

### DIFF
--- a/recipes-kernel/qps615-module/qps615-dlkm/0002-net-ethernet-tc956x-read-MAC-address-from-DT-NVMEM-o.patch
+++ b/recipes-kernel/qps615-module/qps615-dlkm/0002-net-ethernet-tc956x-read-MAC-address-from-DT-NVMEM-o.patch
@@ -1,7 +1,7 @@
-From 1d10902d43db9320fa51dfeef24ab0672950c56f Mon Sep 17 00:00:00 2001
+From 585d7632a2a72d16264a71e240d552e0bf167de9 Mon Sep 17 00:00:00 2001
 From: Mohd Ayaan Anwar <mohd.anwar@oss.qualcomm.com>
 Date: Mon, 30 Mar 2026 01:24:43 +0530
-Subject: [PATCH 02/12] net: ethernet: tc956x: read MAC address from DT/NVMEM
+Subject: [PATCH 02/13] net: ethernet: tc956x: read MAC address from DT/NVMEM
  on platform
 
 Attempt to read the MAC address from the device tree or NVMEM via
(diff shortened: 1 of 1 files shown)


lumag (changes requested):
> A better method to
handle such scenarios

Why is it declared to be better?

ayaan-anwar:
> > A better method to
> > handle such scenarios
> 
> Why is it declared to be better?

This was something that was recently discussed on an internal thread (+ @ricardosalveti).
Rb3Gen2 doesn't have an EEPROM chip, and we are not appending the MAC address into the DTB from the bootloader. As a result, all the boards running QLI 2.0 end up having the same MAC address for both ports which causes DHCP lease issues.

With a random MAC address (and the DEV_ADDR_RANDOM flag), we can at least have a userspace network management entity create a semi-permanent address and use that for the interfaces.

ayaan-anwar:
Converting to draft. I think we should revisit this once we have a fork of the driver available.

lumag:
@ayaan-anwar this need to be a part of the commit message.

ricardosalveti:
> Converting to draft. I think we should revisit this once we have a fork of the driver available.

Still fine to include it here, just please add the justification as part of the commit.

recipes-kernel/qps615-module/qps615-dlkm/0002-net-ethernet-tc956x-read-MAC-address-from-DT-NVMEM-o.patch
```diff
@@ -30,9 +33,38 @@ index a4d8ca8..30031a9 100644
...
+@@ -15302,8 +15304,12 @@ static void parse_config_file(uint8_t port_id, uint8_t dev_id)
+ 	ret = kernel_read_file_from_path("config.ini", &data, &size, 3000, READING_POLICY);
+ #endif
+ 	if (ret < 0) {
+-		KPRINT_ERR("Mac configuration file not found\n");
+-		KPRINT_INFO("Using Default MAC Address\n");
++		/* There's no predefined MAC address so let's set a random
++		 * address and let userland take care of handling it.
++		 */
++		KPRINT_INFO("Falling back to random MAC address since there's no predefined address found");
```
lumag, line 56: Drop or switch to `dev_dbg()`. There is no need to be noisy about the known good and default setup

ricardosalveti:
Merged as part of https://github.com/qualcomm-linux/meta-qcom/pull/2385.

PR #2104, opened 2026-05-04
workflows: trigger nightly on wrynose via dispatch

abinjose-qualcomm:
The current nightly workflow for wrynose dynamically calculates the tree to check out at runtime. However, the GitHub Actions metadata still shows the workflow as being triggered from a commit on the master branch.

This prevents any per-branch customization of the nightly build workflow and disrupts downstream automations that depend on GitHub actions metadata to derive the correct workflow context.

To address this, trigger the workflow via a dispatch explicitly passing the wrynose ref.

### DIFF
--- a/.github/workflows/nightly-build-wrynose.yml
+++ b/.github/workflows/nightly-build-wrynose.yml
@@ -1,4 +1,4 @@
-name: Nightly Build (wrynose)
+name: Nightly Build Dispatcher (wrynose)
 
 on:
   schedule:
@@ -12,43 +12,14 @@ permissions:
   pull-requests: write
   contents: read
   packages: read
+  actions: write
 
 jobs:
-  # Resolve wrynose to a single SHA so all downstream jobs build the same commit,
-  # and so test results are attributed to the correct wrynose commit (not master's
-  # github.sha, which is what scheduled workflows expose by default).
-  resolve-ref:
+  trigger-build:
     runs-on: ubuntu-latest
-    outputs:
-      sha: ${{ steps.get-sha.outputs.sha }}
     steps:
-      - name: Resolve wrynose SHA
-        id: get-sha
+      - name: Trigger Nightly Build
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          SHA=$(gh api repos/${{ github.repository }}/git/ref/heads/wrynose --jq '.object.sha')
-          echo "sha=$SHA" >> $GITHUB_OUTPUT
-
-  build-nightly:
-    needs: resolve-ref
-    uses: ./.github/workflows/build-yocto.yml
-    with:
-      git_ref: ${{ needs.resolve-ref.outputs.sha }}
-
-  test-nightly:
-    uses: ./.github/workflows/test.yml
-    needs: [resolve-ref, build-nightly]
-    secrets: inherit
-    with:
-      build_id: ${{ github.run_id }}
-
-  publish-test-results:
-    uses: ./.github/workflows/publish-results.yml
-    needs: [resolve-ref, test-nightly]
-    secrets: inherit
-    with:
-      workflow_id: ${{ github.run_id }}
-      event_name: ${{ github.event_name }}
-      event_file: ${{ github.event_path }}
-      commit: ${{ needs.resolve-ref.outputs.sha }}
+          gh workflow run --repo ${{ github.repository }} --ref wrynose nightly-build.yml
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -1,4 +1,5 @@
-name: Nightly Build (master)
+name: Nightly Build
+run-name: "Nightly Build (${{ github.ref_name }})"
 
 on:
   workflow_dispatch:
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -1,6 +1,7 @@
 name: Nightly Build (master)
 
 on:
+  workflow_dispatch:
   schedule:
   # NOTE - changes to the cron spec should be pushed by https://github.com/quic-yocto-ci
   # so that build notification emails will be sent out properly.
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -7,10 +7,6 @@ on:
         required: false
         type: string
         default: "full" # values: pr | full
-      git_ref:
-        required: false
-        type: string
-        default: ""
     outputs:
       artifacts_url:
         description: "URL to retrieve build artifacts"
@@ -41,7 +37,6 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
-          ref: ${{ inputs.git_ref }}
 
       - name: Run kas lock
         if: ${{ hashFiles('ci/base.lock.yml') == '' }}
@@ -68,7 +63,6 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
-          ref: ${{ inputs.git_ref }}
 
       - name: Download kas lockfile
         uses: actions/download-artifact@v7
@@ -121,7 +115,6 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
-          ref: ${{ inputs.git_ref }}
 
       - name: Run kas build
         uses: ./.github/actions/compile
@@ -308,7 +301,6 @@ jobs:
         uses: actions/checkout@v6
         with:
           persist-credentials: false
-          ref: ${{ inputs.git_ref }}
 
       - name: Run kas build
         if: inputs.profile != 'pr' || (matrix.distro.name != 'debug' && matrix.distro.name != 'performance')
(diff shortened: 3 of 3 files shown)


.github/workflows/push.yml
```diff
@@ -1,5 +1,4 @@
 name: Build on push
-run-name: "${{ github.ref_name }}: ${{ github.event.head_commit.message }}"
```
ricardosalveti, line 2: Why are you reverting this change?

abinjose-qualcomm: I'll remove it. I only intended to revert the changes to the nightly workflow.

.github/workflows/nightly-build.yml
```diff
@@ -1,4 +1,4 @@
-name: Nightly Build (master)
+name: Nightly Build
```
lumag, line 1: This will squash all nightly builds into a single workflow on GH.

Instead of having using dispatch to trigger the nightly builds, would it work if the configuration for the nightly build is a part of the `wrynose` branch? You can reuse custom actions / workflows from the master branch by directly referencing those, but the triggering commit will be a part of the branch.

lumag (changes requested):

.github/workflows/nightly-build.yml
```diff
@@ -1,6 +1,7 @@
-name: Nightly Build (master)
```
ricardosalveti, line 1: The reason why we had completely separated workflows was to easily identify the nightly in the actions UI, here we are basically merging both again (the actual run will be from the main nightly).

.github/workflows/nightly-build.yml
abinjose-qualcomm: > This will squash all nightly builds into a single workflow on GH.

Yes, you can however differentiate the builds on the UI based on branch. Sample run is available on staging. https://github.com/qualcomm-linux-stg/meta-qcom/actions/workflows/nightly-build.yml

> Instead of having using dispatch to trigger the nightly builds, would it work if the configuration for the nightly build is a part of the wrynose branch? You can reuse custom actions / workflows from the master branch by directly referencing those, but the triggering commit will be a part of the branch.

I think you are suggesting this change in the currently nightly-build-wrynose workflow:

```yaml
  build-nightly:
    needs: resolve-ref
    uses: qualcomm-linux/meta-qcom/.github/workflows/build-yocto.yml@wrynose
```

This could work for the first issue - allow per-branch customization of the nightly build workflow. We'll still have an issue with the triggering commit being part of the master branch. Coral relies on the GitHub actions metadata to derive the correct workflow context and wouldn't really understand the internal steps used here to fetch and checkout the correct ref.

.github/workflows/nightly-build.yml
abinjose-qualcomm: Yes, you can however differentiate the builds on the UI based on branch. Sample run is available on staging. https://github.com/qualcomm-linux-stg/meta-qcom/actions/workflows/nightly-build.yml

.github/workflows/nightly-build.yml
lumag: No. I suggested moving the whole nightly-build-wrynose.yml to the wrynose branch. Will it work?

abinjose-qualcomm: No, GitHub doesn't support creating scheduled jobs from workflows in non-default branches. The parent workflow nightly-build-wrynose.yml needs to be in the master branch.

https://github.com/orgs/community/discussions/26374

lumag: Thanks for the pointer. Then, can we instead fix the GH <-> Coral integration?

abinjose-qualcomm: I don't think there is a generic way to do that. Coral gets its inputs from the GitHub Actions API, which is misleading in this case. Happy to initiate a discussion with Coral maintainers if that would help.

lumag: We already populate extra information for the CR / tests. Would that help Coral too?

.github/workflows/nightly-build.yml
ricardosalveti: You still have to open the run to see the branch that triggered it, is there a way to customize the run-name at least based on what triggered the run?

ricardosalveti:
Please change the order of the patches as well, the git_ref revert should be the last commit.

.github/workflows/nightly-build.yml
abinjose-qualcomm: That could work, however I see that as a workaround. The correct solution here would be to fix the data emitted by GitHub.

.github/workflows/nightly-build.yml
abinjose-qualcomm: I've added a branch identifier to the run name.

Sample runs on staging:
https://github.com/qualcomm-linux-stg/meta-qcom/actions/runs/25363107277
https://github.com/qualcomm-linux-stg/meta-qcom/actions/runs/25363091639

abinjose-qualcomm:
> Please change the order of the patches as well, the git_ref revert should be the last commit.

Makes sense. Updated.

.github/workflows/nightly-build-dispatcher.yml
```diff
@@ -0,0 +1,25 @@
+name: Nightly Build Dispatcher
```
ricardosalveti: I don't think we need to rename this one, just reuse nightly-build-wrynose and have it as a simple dispatcher.

We will have other branches later.

.github/workflows/nightly-build.yml
ricardosalveti: Dispatcher worked fine, but if you check https://github.com/qualcomm-linux-stg/meta-qcom/actions/runs/25363107277/job/74367137704 it is still building from the master branch, since checkout wasn't updated (and git_ref was removed).

We need to trigger the workflow and pass ref_name along.

abinjose-qualcomm: Not sure what you mean. I see this in the actions/checkout logs.

```bash
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin +622d9fb7447f6b77e00d3f309dabe6ad569b80aa:refs/remotes/origin/wrynose
  From https://github.com/qualcomm-linux-stg/meta-qcom
   * [new ref]         622d9fb7447f6b77e00d3f309dabe6ad569b80aa -> origin/wrynose
  /usr/bin/git branch --list --remote origin/wrynose
    origin/wrynose
  /usr/bin/git rev-parse refs/remotes/origin/wrynose
  622d9fb7447f6b77e00d3f309dabe6ad569b80aa
```

.github/workflows/nightly-build-dispatcher.yml
abinjose-qualcomm: I was thinking this can be a reusable workflow which can be extended in case other branches are added later.

.github/workflows/nightly-build.yml
ricardosalveti: Yup, missed the entry which checks out to wrynose, so it looks fine.

.github/workflows/nightly-build-dispatcher.yml
ricardosalveti: But we do expect each branch to use a different schedule, to avoid a runner surge.

How would we reuse this workflow for multiple branches and different schedules?

abinjose-qualcomm: > But we do expect each branch to use a different schedule, to avoid a runner surge.

Makes sense. I'll keep it simple. I'll still keep the name as `Nightly Build Dispatcher (wrynose)` to avoid confusing it with the actual nightly builds.

ricardosalveti (approved):

lumag (changes requested):
> This reverts commit https://github.com/qualcomm-linux/meta-qcom/commit/938add91ca4a1f2b8f1c0aa2700109af9a251dad.

Why?

abinjose-qualcomm:
> > This reverts commit [938add9](https://github.com/qualcomm-linux/meta-qcom/commit/938add91ca4a1f2b8f1c0aa2700109af9a251dad).
> 
> Why?

Added an explanation in the commit message.

lumag (approved):

abinjose-qualcomm:
We'll need to add these changes in wrynose branch as well.

.github/workflows/nightly-build-wrynose.yml
```diff
@@ -12,43 +12,14 @@ permissions:
...
-  publish-test-results:
-    uses: ./.github/workflows/publish-results.yml
-    needs: [resolve-ref, test-nightly]
-    secrets: inherit
-    with:
-      workflow_id: ${{ github.run_id }}
-      event_name: ${{ github.event_name }}
-      event_file: ${{ github.event_path }}
-      commit: ${{ needs.resolve-ref.outputs.sha }}
+          gh workflow run --repo ${{ github.repository }} --ref wrynose nightly-build.yml
```
quaresmajose, line 25: This is a fantastic feature. Thanks for sharing!

quaresmajose (approved):

PR #2107, opened 2026-05-04
Generalize U-Boot configuration

lumag:
Provide generic way to handle U-Boot configs, letting those be build-tested in the qcom-armv8a machine.


conf/machine/iq-9075-evk.conf
```diff
@@ -41,7 +41,6 @@ QCOM_RCU_NOCBS     = "7"
 QCOM_RCU_EXPEDITED = "1"
 QCOM_CPUIDLE_OFF   = "1"
 
-UBOOT_CONFIG_DEFAULT = "iq-9075-evk"
 UBOOT_CONFIG = "iq-9075-evk"
-UBOOT_CONFIG[iq-9075-evk] = "qcom_lemans_defconfig"
-BOARD_MBN_HEADER[iq-9075-evk] = "v6"
+
+require conf/machine/include/qcom-u-boot-common.inc
```
ricardosalveti: Since the u-boot-common.inc is just adding the infra for the multiple configs used (and not forcing any selection), should we instead include by default via machine/include/qcom-common.inc?

We could even include based on a conditional if needed (virtual/bootloader).

ricardosalveti (approved):

anujm1 (approved):

PR #2112, opened 2026-05-05
ci: build-yocto: always run kas lock and include meta-arm by default

ricardosalveti:
Having base.lock.yml present is not sufficient to guarantee that all
repositories are pinned (for example, meta-qcom-distro may be missing).
Always run 'kas lock', even when a lock file already exists, so missing
repositories are populated while existing pins are preserved. This also
behaves correctly when no lock file is present, equivalent to generating
a new lock.

Also include meta-arm.yml by default when generating the lock file as it
currently not included by base or qcom-distro.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -45,7 +45,7 @@ jobs:
 
       - name: Run kas lock
         run: |
-          ${KAS_CONTAINER} lock ci/base.yml:ci/qcom-distro.yml
+          ${KAS_CONTAINER} lock ci/base.yml:ci/qcom-distro.yml:ci/meta-arm.yml
 
       - name: Upload kas lockfile
         uses: actions/upload-artifact@v6
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -46,7 +46,7 @@ jobs:
       - name: Run kas lock
         if: ${{ hashFiles('ci/base.lock.yml') == '' }}
         run: |
-          ${KAS_CONTAINER} lock --update ci/base.yml:ci/qcom-distro.yml
+          ${KAS_CONTAINER} lock ci/base.yml:ci/qcom-distro.yml
 
       - name: Upload kas lockfile
         uses: actions/upload-artifact@v6
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -44,7 +44,6 @@ jobs:
           ref: ${{ inputs.git_ref }}
 
       - name: Run kas lock
-        if: ${{ hashFiles('ci/base.lock.yml') == '' }}
         run: |
           ${KAS_CONTAINER} lock ci/base.yml:ci/qcom-distro.yml
 


ndechesne (approved):
thanks for spotting/fixing that!

.github/workflows/build-yocto.yml
```diff
@@ -44,9 +44,8 @@ jobs:
           ref: ${{ inputs.git_ref }}
 
       - name: Run kas lock
-        if: ${{ hashFiles('ci/base.lock.yml') == '' }}
         run: |
-          ${KAS_CONTAINER} lock --update ci/base.yml:ci/qcom-distro.yml
+          ${KAS_CONTAINER} lock ci/base.yml:ci/qcom-distro.yml
```
quaresmajose: Nice catch! However, we should continue to condition the lock update on the existence of the ci/base.lock.yml. Something like:
```nash
KAS_ARGS="ci/base.yml:ci/qcom-distro.yml"
[ -f ci/base.lock.yml ] || KAS_ARGS="--update ${KAS_ARGS}"
${KAS_CONTAINER} lock ${KAS_ARGS}
```
So this could also be backported and used outside of the master branch.

ndechesne: I thought about it.. but when the lock file does not exist, at this stage (kas-setup) that means the workspace is empty, so --update is not needed. no? I think the --update was a left over from when we had mirrors (KAS_REPO_REF_DIR)

quaresmajose: Yes, you're right, the --update is no longer necessary. It updates the lock file in place, but updating the file, when it exists, is always done manually; so the --update is not necessary.

It was really necessary when we had the KAS_REPO_REF_DIR, and in reality, this seems more like a bug in kas. Because the result of the lock file is different depending on the presence of the KAS_REPO_REF_DIR, these have priority over the upstream repositories defined in the yml configuration.

I think we should remove the `--update` in a separate commit and reference that it was a leftover from KAS_REPO_REF_DIR, that way the reason for the change would be clearer and it could be useful to us when we implement the automation of the `--update`.

lumag (commented):
I think we need to add some checks on top. The meta-qcom-distro should be the only repo which is allowed to be not pinned.

ndechesne:
> I think we need to add some checks on top. The meta-qcom-distro should be the only repo which is allowed to be not pinned.

I agree, that would be a nice improvement, but I would not want to block this PR. it can be done separately.

.github/workflows/build-yocto.yml
ricardosalveti: Yeah, it is a leftover and not required anymore, I can add this to the commit.

ricardosalveti: Separated the commit removing update and pointing out that KAS_REPO_REF_DIR was removed.

ricardosalveti:
> I think we need to add some checks on top. The meta-qcom-distro should be the only repo which is allowed to be not pinned.

Maybe we should evaluate this outside the build-yocto workflow, because this same workflow is also reused by other repos.

quaresmajose (approved):

PR #2114, opened 2026-05-05
add phone capability via MACHINE_FEATURES for iq-x7181-evk and iq-x5121-evk

hsharm-bit:
This change adds the 'phone' capability to MACHINE_FEATURES for the
iq-x7181-evk and iq-x5121-evk machines having external M.2 attach modems.

The intent is to advertise telephony support at the machine level,
allowing higher layers to conditionally enable cellular-related components based on MACHINE_FEATURES.

### DIFF
--- a/conf/machine/iq-x5121-evk.conf
+++ b/conf/machine/iq-x5121-evk.conf
@@ -4,7 +4,7 @@
 
 require conf/machine/include/qcom-purwa.inc
 
-MACHINE_FEATURES += "efi pci"
+MACHINE_FEATURES += "efi pci phone"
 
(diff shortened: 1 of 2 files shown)


lumag (changes requested):
Please fix your editor settings. Your lines are too short.

Also:

> the SDX modem is attached via the platform configuration

No, the modem can't be attached via the platform configuration. It can be PCIe slot, M.2 slot, wired to the board, plugged into the USB socket, etc. But you can't attach the modem via the configuration.

lumag:
@sbanerjeeqc @knaveen-qc JFYI, the review was provided more than 30 days ago, the developer neither responded nor updated the code.

PR #2124, opened 2026-05-06
[backport wrynose] workflows: trigger nightly on wrynose via dispatch

abinjose-qualcomm:
Adds changes to the nightly-build.yml workflow to support runs on non-default branches.

Backport of https://github.com/qualcomm-linux/meta-qcom/pull/2104 to wrynose.


lumag (changes requested):
Why do we need to backport changes to the unused files? I'd prefer to get nightly.yml to be removed completely in this branch.

abinjose-qualcomm:
> Why do we need to backport changes to the unused files? I'd prefer to get nightly.yml to be removed completely in this branch.

nightly-build-wrynose.yml in master branch invokes the nightly.yml from wrynose branch, so we'll still need it. We can remove the nightly-build-wrynose.yml in the wrynose branch.

quaresmajose:
It would be preferable to tag the master PR trigger the backport workflow. This will use the `git commit -x` command which makes it easier to see where the changes are coming from.

abinjose-qualcomm:
> It would be preferable to tag the master PR trigger the backport workflow. This will use the `git commit -x` command which makes it easier to see where the changes are coming from.

I have updated one commit to drop an unused workflow now, so I'll remove the backport tag and update the description.

ricardosalveti:
When backporting use git cherry-pick -x so it adds the original commit that was used in master.

abinjose-qualcomm:
> When backporting use git cherry-pick -x so it adds the original commit that was used in master.

Thanks for the tip. I've updated the commits.

ricardosalveti (approved):

ricardosalveti:
> nightly-build-wrynose.yml in master branch invokes the nightly.yml from wrynose branch, so we'll still need it. We can remove the nightly-build-wrynose.yml in the wrynose branch.

So basically we will have to keep most of the workflows in sync then, including build-yocto, due the new nightly workflow.

abinjose-qualcomm:
> So basically we will have to keep most of the workflows in sync then, including build-yocto, due the new nightly workflow.

The reusable actions can still point to the master branch if we expect them to be the same. The build-yocto workflow should be separate and its intended. Not all targets declared in mainline are supported on wrynose. This approach would allow configuring a reduced set of CI targets for wrynose branch, which was not possible with the previous approach.

ricardosalveti:
> This approach would allow configuring a reduced set of CI targets for wrynose branch, which was not possible with the previous approach.

Yeah, target list should be the same but the permutation of jobs will soon be different, valid point.

quaresmajose (approved):

abinjose-qualcomm:
Can we merge this before next nightly schedule? We need these changes for the nightly build on wrynose to be triggered.

lumag (approved):

PR #2132, opened 2026-05-07
Switch debug and performance builds to catchall distro

vkraleti:
Update the build-yocto GitHub CI workflow to run debug and performance builds against qcom-distro-catchall instead of the default qcom-distro.

The catchall distro enables all Qualcomm-supported DISTRO_FEATURES and provides broader feature coverage, making it more suitable for debug and performance validations.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -223,9 +223,9 @@ jobs:
           - name: qcom-distro-catchall
             yamlfile: ':ci/qcom-distro-catchall.yml'
           - name: debug
-            yamlfile: ':ci/qcom-distro.yml:ci/debug.yml'
+            yamlfile: ':ci/qcom-distro-catchall.yml:ci/debug.yml'
           - name: performance
-            yamlfile: ':ci/qcom-distro.yml:ci/performance.yml'
+            yamlfile: ':ci/qcom-distro-catchall.yml:ci/performance.yml'
         kernel:
           - type: default
             dirname: ""


.github/workflows/build-yocto.yml
```diff
@@ -157,10 +157,10 @@ jobs:
             yamlfile: ':ci/qcom-distro.yml'
           - name: qcom-distro-catchall
             yamlfile: ':ci/qcom-distro-catchall.yml'
-          - name: debug
-            yamlfile: ':ci/qcom-distro.yml:ci/debug.yml'
-          - name: performance
-            yamlfile: ':ci/qcom-distro.yml:ci/performance.yml'
+          - name: debug_qcom-distro-catchall
```
ricardosalveti: Do we need / want to change the label here?

vkraleti: Label change makes it explicit that the build corresponds to catch‑all distro, avoiding ambiguity about which distro is being used.

vkraleti: I'll check with Test org one more time if they are fine calling it just debug/performance build without mentioning the distro name.

vkraleti: Got confirmation, no need to update distro name. Uploaded new commit.

quaresmajose (approved):

.github/workflows/build-yocto.yml
```diff
@@ -223,9 +223,9 @@ jobs:
           - name: qcom-distro-catchall
             yamlfile: ':ci/qcom-distro-catchall.yml'
           - name: debug
-            yamlfile: ':ci/qcom-distro.yml:ci/debug.yml'
+            yamlfile: ':ci/qcom-distro-catchall.yml:ci/debug.yml'
           - name: performance
-            yamlfile: ':ci/qcom-distro.yml:ci/performance.yml'
+            yamlfile: ':ci/qcom-distro-catchall.yml:ci/performance.yml'
```
lumag, line 228: Is the performance image being tested in the CI?

ricardosalveti: Not in lava, not sure if tested with axiom internally.

ricardosalveti (approved):

.github/workflows/build-yocto.yml
lumag: If not, should we start testing it at least for some usecases / devices? Otherwise we might be shipping a recipe to build the image which undergoes no testing at all.

lumag (changes requested):
Please enable testing of performance / catchatll in the CI.

.github/workflows/build-yocto.yml
vkraleti: To test in CI, need to check with @mwasilew and @smuppand about devices availability. Anyways nightly images of these flavors are going to consumed in post-merge testing for GA acceptance.

PR #2137, opened 2026-05-07
ci: workflows/push: drop run-name

ricardosalveti:
Run-name was added as a way to easily identify if the push job was done on master or wrynose, but github.event.head_commit.message is way too verbose since it includes the entire commit message, so remove it entirely as there is no way to use just the commit subject.


ricardosalveti:
Canceled the workflow as it has no impact to it.

mwasilew (approved):

quaresmajose (approved):

PR #2145, opened 2026-05-08
iq-9075-evk: Add support for Open Boot firmware (TF-A, OP-TEE and U-Boot) build

b49020:
Add OP-TEE support for Lemans EVK platform also know as iq-9075-evk. The QTISECLIB port for lemans is still work under progress. Without QTISECLIB it is able to boot OP-TEE and U-Boot but crashes in kernel due to some access permissions issues. Our end mission is to enable functional Qcom platforms without QTISECLIB only.

However, with an internal development version of QTISECLIB, Linux boot to shell with multiple cores up and xtest running successfully.

### DIFF
--- a/conf/machine/rb3gen2-core-kit-open-fw.conf
+++ b/conf/machine/rb3gen2-core-kit-open-fw.conf
@@ -9,4 +9,4 @@ EXTRA_IMAGEDEPENDS += "trusted-firmware-a-qcom"
 MACHINE_FEATURES += "optee"
 MACHINE_EXTRA_RRECOMMENDS += "${@bb.utils.contains('BBFILE_COLLECTIONS', 'meta-arm', 'packagegroup-optee', '', d)}"
 
-UBOOT_CONFIG = "qcs6490-rb3gen2"
+PREFERRED_PROVIDER_virtual/bootloader = "u-boot-qcom"
--- a/dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
+++ b/dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
@@ -1,4 +1,4 @@
-DEPENDS += "qtestsign-native optee-os-qcom"
+DEPENDS += "qtestsign-native optee-os-qcom u-boot-qcom"
 
 SRC_URI += "git://github.com/coreboot/qc_blobs.git;protocol=https;name=qc_blobs;subdir=qc_blobs;branch=main"
 SRCREV_qc_blobs = "16207f367ddbf280a0c2c8a3d3ee454a59710a25"
@@ -9,17 +9,19 @@ LIC_FILES_CHKSUM:qcm6490 += "file://${UNPACKDIR}/qc_blobs/sc7280/qtiseclib/LICEN
 COMPATIBLE_MACHINE:qcm6490 = "qcm6490"
 TFA_PLATFORM:qcm6490 = "rb3gen2"
 
-TFA_UBOOT ?= "1"
 TFA_BUILD_TARGET = "bl2 fip"
 TFA_SPD = "opteed"
 
 EXTRA_OEMAKE:append = " \
     BL32=${RECIPE_SYSROOT}/${nonarch_base_libdir}/firmware/tee-raw.bin \
+    BL33=${DEPLOY_DIR_IMAGE}/u-boot.bin \
    "
 
 EXTRA_OEMAKE:append:qcm6490 = " \
     QTISECLIB_PATH=${UNPACKDIR}/qc_blobs/sc7280/qtiseclib/libqtisec.a \
    "
+
+do_compile[depends] += " u-boot-qcom:do_deploy"
 
 do_install:append:qcm6490() {
     export CRYPTOGRAPHY_OPENSSL_NO_LEGACY=1
--- a/dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
+++ b/dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
@@ -6,8 +6,9 @@ SRCREV_qc_blobs = "16207f367ddbf280a0c2c8a3d3ee454a59710a25"
 LICENSE += "& LICENSE.qcom"
 LIC_FILES_CHKSUM:qcm6490 += "file://${UNPACKDIR}/qc_blobs/sc7280/qtiseclib/LICENSE;md5=fa83f30385e617b56ef0934f13645621"
 
-COMPATIBLE_MACHINE:qcm6490 = "qcm6490"
+COMPATIBLE_MACHINE = "qcm6490|qcs9100"
 TFA_PLATFORM:qcm6490 = "rb3gen2"
+TFA_PLATFORM:qcs9100 = "lemans_evk"
 
 TFA_BUILD_TARGET = "bl2 fip"
 TFA_SPD = "opteed"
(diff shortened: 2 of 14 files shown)


conf/machine/rb3gen2-core-kit-open-fw.conf
```diff
@@ -9,4 +9,4 @@ EXTRA_IMAGEDEPENDS += "trusted-firmware-a-qcom"
 MACHINE_FEATURES += "optee"
 MACHINE_EXTRA_RRECOMMENDS += "${@bb.utils.contains('BBFILE_COLLECTIONS', 'meta-arm', 'packagegroup-optee', '', d)}"
 
-UBOOT_CONFIG = "qcs6490-rb3gen2"
```
quaresmajose, line 12: Can you plase add some explanation on the commit for the UBOOT_CONFIG removing?

dynamic-layers/meta-arm/recipes-security/optee/optee-test_%.bbappend
```diff
@@ -1,4 +1,5 @@
-# Machine specific configurations
+DEPENDS:remove = "optee-os-tadevkit"
+DEPENDS:append = " optee-os-tadevkit-qcom"
```
ricardosalveti: Add qcom override on both.

recipes-bsp/u-boot/files/tfa-optee.cfg
```diff
@@ -0,0 +1,4 @@
+# Enables support for TF-A based OP-TEE as the open
+# source TrustZone stack on Qcom platforms
+CONFIG_TEE=y
+CONFIG_OPTEE=y
```
ricardosalveti, line 4: Would it be a problem to have this enabled by default in the upstream config or do we really need to enable both only when available at runtime?

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -9,18 +9,20 @@ LIC_FILES_CHKSUM:qcm6490 += "file://${UNPACKDIR}/qc_blobs/sc7280/qtiseclib/LICEN
 COMPATIBLE_MACHINE:qcm6490 = "qcm6490"
 TFA_PLATFORM:qcm6490 = "rb3gen2"
 
-TFA_UBOOT ?= "1"
```
ricardosalveti, line 12: Why was this removed? Not part of the commit message.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -9,18 +9,20 @@ LIC_FILES_CHKSUM:qcm6490 += "file://${UNPACKDIR}/qc_blobs/sc7280/qtiseclib/LICEN
...
 EXTRA_OEMAKE:append = " \
     BL32=${RECIPE_SYSROOT}/${nonarch_base_libdir}/firmware/tee-raw.bin \
+    BL33=${DEPLOY_DIR_IMAGE}/u-boot.bin \
    "
 
 EXTRA_OEMAKE:append:qcm6490 = " \
     QTISECLIB_PATH=${UNPACKDIR}/qc_blobs/sc7280/qtiseclib/libqtisec.a \
    "
 
+do_compile[depends] += " u-boot-qcom:do_deploy"
```
ricardosalveti, line 25: Also explain this in the commit message.

conf/machine/rb3gen2-core-kit-open-fw.conf
ricardosalveti: Don't we also need to keep the config?

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -6,8 +6,9 @@ SRCREV_qc_blobs = "16207f367ddbf280a0c2c8a3d3ee454a59710a25"
 LICENSE += "& LICENSE.qcom"
 LIC_FILES_CHKSUM:qcm6490 += "file://${UNPACKDIR}/qc_blobs/sc7280/qtiseclib/LICENSE;md5=fa83f30385e617b56ef0934f13645621"
 
-COMPATIBLE_MACHINE:qcm6490 = "qcm6490"
+COMPATIBLE_MACHINE = "qcm6490|qcs9100"
```
ricardosalveti, line 9: Your commit message is saying that you are adding OP-TEE support while you are really changing TF-A here.

conf/machine/rb3gen2-core-kit-open-fw.conf
ricardosalveti: Not needed anymore because it is included via conf/machine/rb3gen2-core-kit.conf, please add that to the commit message.

b49020: Sure, I will update the commit message.

recipes-bsp/u-boot/files/tfa-optee.cfg
b49020: There isn't a standard way to determine if a platform supports QTEE or OP-TEE at this point at runtime. The SCM/SMC calls are all custom implemented ones. So it is rather based on this OP-TEE config option that we add the OP-TEE DT node in U-Boot for driver probe.

Although in parallel, the OP-TEE SMC call is also used to verify OP-TEE presence but given the many variants of QTEE we have on many different platforms, we can't rely on that only.

However, we can rather add this config fragment in platform specific defconfig where the OP-TEE SMC based detection is verified to work. I will follow-up with a patch for that upstream and see how it goes. So let's use this fragment for now.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
b49020: `TFA_UBOOT` adds a hard dependency on `u-boot` which we don't wan't but rather `u-boot-qcom`. I will update commit message accordingly.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
b49020: Ah, I will fix the commit message.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
lumag: Please fix the recipe in meta-arm to let it use alternative u-boot implementations.

recipes-bsp/u-boot/files/tfa-optee.cfg
ricardosalveti: Makes sense, fragment is ok for now, thanks for following this up with upstream.

ricardosalveti:
> The QTISECLIB port for lemans is still work under progress.

I would prefer to wait for the qtiseclib release, but I know that it might still take more time, so OK to merge it now as it will at least help folks using this environment for validating the open boot stack.

ricardosalveti (approved):

dynamic-layers/meta-arm/recipes-security/optee/optee-os-tadevkit-qcom_git.bb
```diff
@@ -0,0 +1,25 @@
+require optee-os-qcom_git.bb
+
+SUMMARY = "OP-TEE Trusted OS TA devkit"
+DESCRIPTION = "OP-TEE TA devkit for build TAs"
+HOMEPAGE = "https://www.op-tee.org/"
```
lumag, line 5: For the followup: please extract the common include file for tadevkit in meta-arm and use it here.

recipes-bsp/u-boot/files/tfa-optee.cfg
lumag, line 4: Can we enable these two unconditionally for Qcom platforms?

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
lumag, line 12: For the followup: plese make TFA_UBOOT accept configurable U-Boot package name.

lumag (changes requested):

recipes-bsp/u-boot/files/tfa-optee.cfg
b49020: See discussion above: https://github.com/qualcomm-linux/meta-qcom/pull/2145#discussion_r3223976646, right now it's not possible to reliably distinguish among OP-TEE and QTEE presence for all platforms including legacy ones. However, I will try to push for addition to platform specific defconfig upstream where we can test that SMC calls don't break the system.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
b49020: Sure, I will submit follow-up patches to meta-arm

dynamic-layers/meta-arm/recipes-security/optee/optee-os-tadevkit-qcom_git.bb
b49020: Sure, I will submit follow-up patches to meta-arm

lumag (approved):

lumag:
@b49020 please submit a backport for wrynose.

anujm1:
> Git push to origin failed for wrynose with exitcode 1

Do changes to workflows need extra permissions for backport?

> [remote rejected] backport/2145-to-wrynose -> backport/2145-to-wrynose (refusing to allow a GitHub App to create or update workflow `.github/workflows/build-yocto.yml` without `workflows` permission)

ricardosalveti:
Yeah, looks like it, we would need to extend the backports for that.

PR #2148, opened 2026-05-08
weston: don't version the weston bbappend

lumag:
Having versioned bbappends makes it harder to uprade layers independently of each other. Replace version in the weston's bbappend with the wildcard, letting the layer build both with the locked master revision and with the tip of OE-Core.


ricardosalveti (approved):

anujm1:
> Having versioned bbappends makes it harder to uprade layers independently of each other. Replace version in the weston's bbappend with the wildcard, letting the layer build both with the locked master revision and with the tip of OE-Core.

I didn't understand. Are we expecting that the patches in bbappend will continue to apply for newer versions? Won't it continue to fail anyway just at a later step?

lumag:
> > Having versioned bbappends makes it harder to uprade layers independently of each other. Replace version in the weston's bbappend with the wildcard, letting the layer build both with the locked master revision and with the tip of OE-Core.
> 
> I didn't understand. Are we expecting that the patches in bbappend will continue to apply for newer versions? Won't it continue to fail anyway just at a later step?

They apply to e.g. 15.0.1

anujm1 (approved):

PR #2149, opened 2026-05-08
Add and enable oe-selftest for multi-dtb (fit)

ricardosalveti:
Add and enable both unit and integration tests for the QCOM multi-dtb FIT image generation, to make sure the format of the ITS/FIT is generated correctly in order to avoid regressions at the UEFI level.

Most of the code generated with claude/opus-4.6 (high).

### DIFF
--- a/lib/oeqa/selftest/cases/qcom_fitimage.py
+++ b/lib/oeqa/selftest/cases/qcom_fitimage.py
@@ -541,3 +541,279 @@ class QcomFitImageTests(OESelftestTestCase):
         self.assertGreaterEqual(conf_count, 2,
             "Expected at least 2 configuration entries in dumpimage output")
 
+
+class QcomFitImageIntegrationTests(OESelftestTestCase):
+    """Integration tests that build a real FIT image from the kernel recipe.
+
+    These tests validate that dtb-fit-image.bbclass, fit-dtb-compatible.inc
+    and QcomItsNodeRoot work together end-to-end to produce a FIT image that
+    the UEFI firmware will be able to parse at boot.
+
+    A real kernel build is triggered so that DTB files, the metadata blob and
+    the FIT binary are all produced by the same tooling as in production.
+    """
+
+    # Cache build vars across helper calls within the same test run.
+    _cached_bb_vars = None
+
+    # Metadata DTS node names we extract once (class-level cache).
+    _meta_nodes = None
+
+    # Suffixes allowed by the metadata-check blacklist
+    COMPAT_SKIP_PATTERNS = {"camx", "el2kvm", "staging"}
+
+    # ------------------------------------------------------------------
+    # Helpers
+    # ------------------------------------------------------------------
+
+    def _get_bb_vars(self):
+        """Retrieve bitbake variables needed by integration tests."""
+        if self.__class__._cached_bb_vars is None:
+            self.__class__._cached_bb_vars = get_bb_vars([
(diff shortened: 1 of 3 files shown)


ricardosalveti:
Validated as part of https://github.com/qualcomm-linux-stg/meta-qcom/actions/runs/25582113664.

vkraleti:
Thanks for the change. It is long pending. I'll validate it once at my end and approve.

lib/oeqa/selftest/cases/qcom_fitimage.py
```diff
@@ -0,0 +1,819 @@
...
+    def _get_test_dir(self):
+        topdir = os.environ['BUILDDIR']
+        d = os.path.join(topdir, 'qcom-fitimage-test', self._testMethodName)
+        if os.path.exists(d):
+            shutil.rmtree(d)
+        os.makedirs(d, exist_ok=True)
+        return d
+
+    def _build_qcom_fitimage(self, kernel_devicetree, fit_dtb_compatible):
+        """Replicate dtb-fit-image.bbclass logic and produce an ITS file.
```
kraghava-qti, line 75: * Is there a way to directly test a `bbclass` by injecting its dependencies (task stubs & more maybe) as stubs to setup the execution env ? I do not know how oe test infra code can be used here to aid this. This can be helpful to find out implementation drift between the class `QcomItsNodeRoot` and its consumption in `dtb-fit-image.bbclass`, when changes are made independently to them. In an ideal world one would depend on the interface contract of the class not impl behavior. Should this need a test recipe ?

* It looks like we are testing the class `QcomItsNodeRoot` using oe test harness code integration style. Why? 
Python std library provided [`unittest`](https://docs.python.org/3/library/unittest.html) module would allow testing the same class independently of the availability of oe test harness code. I am assuming that was not one of the driving forces, hence not considered. The code here (it becomes a bridge to activate/enable) can then drive those test definitions.

ricardosalveti: > * Is there a way to directly test a `bbclass` by injecting its dependencies (task stubs & more maybe) as stubs to setup the execution env ? I do not know how oe test infra code can be used here to aid this. This can be helpful to find out implementation drift between the class `QcomItsNodeRoot` and its consumption in `dtb-fit-image.bbclass`, when changes are made independently to them. In an ideal world one would depend on the interface contract of the class not impl behavior. Should this need a test recipe ?

The integration tests (as part of the second commit) was created to validate the actual usage of the class, because once you add dependencies and also depend on the build environment, it is better to just simulate the build instead.

This is also how u-boot fit is validated upstream: https://git.openembedded.org/openembedded-core/tree/meta/lib/oeqa/selftest/cases/fitimage.py#n849

That is why I added both type of tests here, but there is no easy way to simulate both scenarios and compare directly, as quite often you end up depending on multiple parts of the build.

> * It looks like we are testing the class `QcomItsNodeRoot` using oe test harness code integration style. Why?
>   Python std library provided [`unittest`](https://docs.python.org/3/library/unittest.html) module would allow testing the same class independently of the availability of oe test harness code. I am assuming that was not one of the driving forces, hence not considered. The code here (it becomes a bridge to activate/enable) can then drive those test definitions.

oe-selftest is based on unittest, so it can be tested somehow isolated (not fully due the rest of the OE environment).

kraghava-qti (approved):

vkraleti (approved):

lumag (approved):

PR #2155, opened 2026-05-11
debug.yml: enable FTrace settings in kernel cmdline

apateriy-qcom:
Add FTrace kernel cmdline options to ci/debug.yml for debug builds to enable tracing from early boot.

  ## Problem
  FTrace is off by default. On debug builds, timing-sensitive events (scheduler latency, IRQ storms, thermal throttling, clock/regulator transitions) happen during early init — before userspace can enable tracing — and are lost.

  ## Changes
  - Adds a cmdline section to local_conf_header in ci/debug.yml that appends FTrace options to KERNEL_CMDLINE_EXTRA for debug builds only.
  - The event list is stored in KERNEL_CMDLINE_EXTRA_FTRACE and referenced via ${KERNEL_CMDLINE_EXTRA_FTRACE} to keep the assignment readable.

  ## FTrace Options
  | Option | Reason |
  |---|---|
  | ftrace=tracing_on | Enables the tracer at boot so early-init events are not missed |
  | trace_event | Captures timer, IRQ, workqueue, scheduler, clock, CPU frequency, regulator, thermal, and RPMh events — the subsystems most relevant to platform debug |
  | trace_buf_size=5M | Increases per-CPU ring buffer from 1 MB to 5 MB to reduce data loss under high-frequency event bursts |

### DIFF
--- a/ci/debug.yml
+++ b/ci/debug.yml
@@ -6,3 +6,7 @@ local_conf_header:
     DEBUG_BUILD = "1"
     IMAGE_GEN_DEBUGFS = "1"
     IMAGE_FSTYPES_DEBUGFS = "tar.bz2"
+
+  cmdline: |
+    KERNEL_CMDLINE_EXTRA_FTRACE = "ftrace=tracing_on trace_buf_size=5M trace_event=timer:timer_expire_entry,timer:timer_expire_exit,timer:hrtimer_cancel,timer:hrtimer_expire_entry,timer:hrtimer_expire_exit,timer:hrtimer_init,timer:hrtimer_start,irq:*,workqueue:*,sched:sched_migrate_task,sched:sched_pi_setprio,sched:sched_switch,sched:sched_wakeup,sched:sched_wakeup_new,power:clock_set_rate,power:clock_enable,power:clock_disable,power:cpu_frequency,regulator:*,thermal:thermal_zone_trip,thermal:thermal_temperature,thermal:cdev_update,rpmh:rpmh_send_msg"
+    KERNEL_CMDLINE_EXTRA:append = " ${KERNEL_CMDLINE_EXTRA_FTRACE}"


vkraleti:
Please check and follow the same commit message pattern followed in this repository  `<file name>: <high level description>`
Also commit message doesn't really mention the reason for the change.

ricardosalveti:
Please justify why you need these options in your commit message, and explain them as well.

ci/debug.yml
```diff
@@ -4,3 +4,8 @@ header:
 local_conf_header:
   debug-build: |
     DEBUG_BUILD = "1"
+
+  cmdline: |
+    KERNEL_CMDLINE_EXTRA:append = " trace_event=timer:timer_expire_entry,timer:timer_expire_exit,timer:hrtimer_cancel,timer:hrtimer_expire_entry,timer:hrtimer_expire_exit,timer:hrtimer_init,timer:hrtimer_start,irq:*,workqueue:*,sched:sched_migrate_task,sched:sched_pi_setprio,sched:sched_switch,sched:sched_wakeup,sched:sched_wakeup_new,power:clock_set_rate,power:clock_enable,power:clock_disable,power:cpu_frequency,regulator:*,thermal:thermal_zone_trip,thermal:thermal_temperature,thermal:cdev_update,rpmh:rpmh_send_msg trace_buf_size=5M ftrace=tracing_on"
```
quaresmajose: Given that the line is very long, use a new variable for it.
```bash
KERNEL_CMDLINE_EXTRA_FTRACE = "trace_event=..."
KERNEL_CMDLINE_EXTRA:append = " ${KERNEL_CMDLINE_EXTRA_FTRACE}"
```

apateriy-qcom:
> Please justify why you need these options in your commit message, and explain them as well.

added the reason in commit message and PR description.

ci/debug.yml
apateriy-qcom: yes. Done. Thanks.

ricardosalveti (dismissed):

lumag (commented):
I understand the intention, but for me it feels like having too special command line in a generic build.

Also, some (older) bootloaders might cut the command line. Please move the enablement options to the beginning of the line so that they are never cut off.

lumag (approved):

vkraleti (approved):

ricardosalveti (approved):

PR #2159, opened 2026-05-11
Drop SoC version suffixes from FIT DTB compatible strings

vkraleti:
With the new DTB compatible string format, SoC version and board revision are expressed as explicit suffixes. When these suffixes are not present, FIT matching falls back to the best available compatible string. As a result, explicitly listing SoC-versioned compatible entries is optional when there is no difference in DTB or DTBO selection.

To simplify compatible string listings, remove redundant SoC version entries from FIT_DTB_COMPATIBLE and rely on base compatibles for matching.


ricardosalveti:
Not everything is being validated at https://github.com/qualcomm-linux/meta-qcom/pull/2149 yet, but should we try to get that merged first before proposing additional fit-dtb changes?

vkraleti:
Sure, this PR can wait.

lumag:
The #2149 is now merged. Please rebase and merge.

lumag (approved):

lumag (approved):

ricardosalveti (approved):

PR #1900, opened 2026-04-08
ci: add testimage

quaresmajose:
The build system has the ability to run a series of automated tests for qemu images.
All the tests are actually commands run on the target system over ssh.
The tests themselves are written in Python, making use of the unittest module.
The class that enables this is testimage.bbclass (which handles loading the tests and starting the qemu image)

For now it runs the default tests definded upstream in oe-core, The "SKIPPED" tests are due to missing packages in the image. I would prefer to start with the default definition and then adapt it to our needs.

```
INFO     - RESULTS:
INFO     - RESULTS - date.DateTest.test_date: PASSED (4.00s)
INFO     - RESULTS - df.DfTest.test_df: PASSED (0.59s)
INFO     - RESULTS - oe_syslog.SyslogTest.test_syslog_running: PASSED (0.60s)
INFO     - RESULTS - oe_syslog.SyslogTestConfig.test_syslog_logger: PASSED (1.95s)
INFO     - RESULTS - oe_syslog.SyslogTestConfig.test_syslog_restart: PASSED (0.97s)
INFO     - RESULTS - pam.PamBasicTest.test_pam: PASSED (2.26s)
INFO     - RESULTS - parselogs.ParseLogsTest.test_get_context: PASSED (0.00s)
INFO     - RESULTS - parselogs.ParseLogsTest.test_parselogs: PASSED (2.77s)
INFO     - RESULTS - perl.PerlTest.test_perl_works: PASSED (0.57s)
INFO     - RESULTS - ping.PingTest.test_ping: PASSED (0.00s)
INFO     - RESULTS - python.PythonTest.test_python3: PASSED (0.60s)
INFO     - RESULTS - scp.ScpTest.test_scp_file: PASSED (1.12s)
INFO     - RESULTS - ssh.SSHTest.test_ssh: PASSED (0.56s)
INFO     - RESULTS - systemd.SystemdBasicTests.test_systemd_basic: PASSED (0.55s)
INFO     - RESULTS - systemd.SystemdBasicTests.test_systemd_failed: PASSED (1.12s)
INFO     - RESULTS - systemd.SystemdBasicTests.test_systemd_list: PASSED (1.24s)
INFO     - RESULTS - systemd.SystemdJournalTests.test_systemd_boot_time: PASSED (0.57s)
INFO     - RESULTS - systemd.SystemdJournalTests.test_systemd_journal: PASSED (0.56s)
[...]

PR #2173, opened 2026-05-13
workflows: ensure build_successful checkrun runs if all builds pass

abinjose-qualcomm:
build_successful checkrun indicates to downstream automations that all builds passed and the build artifacts are uploaded to S3. The current criteria for evaluating this condition doesn't hold up if fail-fast is disabled and jobs fail in the compile_warm_up step.

Observed Axiom trigger failures as part of https://github.com/qualcomm-linux/meta-qcom/pull/1900/ since `fail-fast` was disabled for the compile_warm_up step.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -448,9 +448,9 @@ jobs:
   build_successful:
     if: |
       always() && github.repository_owner == 'qualcomm-linux' &&
-      (needs.compile.result == 'success' ||
+      ((needs.compile.result == 'success' && needs.compile_warm_up.result == 'success') ||
        (inputs.skip_if_github_cached == true && needs.github-cache-check.outputs.cache_hit == 'true'))
-    needs: [github-cache-check, compile]
+    needs: [github-cache-check, compile_warm_up, compile]
     runs-on: ubuntu-latest
     steps:
       - name: Build completed successfully


shivshah3400 (approved):

koenkooi:
Did you mean one or more issues? The use of plural 'criteria' indicates more than one, but I don't see that in the commit.

abinjose-qualcomm:
> Did you mean one or more issues? The use of plural 'criteria' indicates more than one, but I don't see that in the commit.

There are multiple criteria evaluated before the build_successful checkrun is published. I'm trying to add one more check to these criteria.

```bash
always() && github.repository_owner == 'qualcomm-linux' &&
(needs.compile.result == 'success' ||
(inputs.skip_if_github_cached == true && needs.github-cache-check.outputs.cache_hit == 'true'))
```

I've updated the commit message to make this clearer.

abinjose-qualcomm:
I think my assertion here is incorrect. The reason for failure in #1900 was due to the compile step being dropped from the workflow for debugging/speedup. I'll mark this as draft, while I check if another solution would help here or not.

quaresmajose:
Hi @abinjose-qualcomm, sorry about that. Would it be possible to check if the PR is in draft state and skip it if so?

anshulg-qti:
Agree with @quaresmajose, we should skip the test trigger if PR is in draft state

abinjose-qualcomm:
> Agree with @quaresmajose, we should skip the test trigger if PR is in draft state

In case someone raises a PR in "Draft" state, it would skip the test cycle. If it is set as "Ready for Review" without any changes, there wouldn't be any axiom test triggers for the PR.

quaresmajose:
> > Agree with @quaresmajose, we should skip the test trigger if PR is in draft state
> 
> In case someone raises a PR in "Draft" state, it would skip the test cycle. If it is set as "Ready for Review" without any changes, there wouldn't be any axiom test triggers for the PR.

In this situation, the PR will not be ready for merging because the validation is not fully complete.

When the PR is marked with "Ready for Review", it will be reviewed by the maintainers, and one of the conditions is that the changes have been tested. Therefore, if the changes in question have a valid test, it makes no difference whether the test was done while it was in "draft", because the code tested was the same code being reviewed.

abinjose-qualcomm:
> In this situation, the PR will not be ready for merging because the validation is not fully complete.

I agree. This was an unusual case where the workflow file was heavily modified and should be considered an outlier. We don't want to disrupt test cycles for valid PRs to accommodate this scenario.

I'll close this PR since the happy path seems to be working fine.

PR #2183, opened 2026-05-13
qcom-partition-conf: build qcom-ptool natively

EmbeddedAndroid:
qcom-ptool's new layout exposes a single qcom-ptool CLI installed via pip, and the Makefile now invokes that CLI. Add a qcom-ptool-native recipe so the CLI lands on the recipe-sysroot-native PATH, DEPENDS on it from qcom-partition-conf (dropping INHIBIT_DEFAULT_DEPS), and switch do_deploy's `cd platforms` to absolute now that ${B} != ${S}. Bump SRCREV to pick that refactor up; the new tip also adds the qcs6490-thundercomm-rubikpi3 machine.

Note one deviation from the suggested wording in the review thread: used python_setuptools_build_meta rather than setuptools3, since setuptools3 still drives setup.py and this repo is pyproject-only.

### DIFF
--- a/recipes-bsp/partition/qcom-ptool-native.bb
+++ b/recipes-bsp/partition/qcom-ptool-native.bb
@@ -1,0 +1,6 @@
+SUMMARY = "Qualcomm partition tool (build-time CLI)"
+DESCRIPTION = "Python CLI used by qcom-partition-conf to generate per-platform GPT partition tables and QDL flashing scripts."
+
+require qcom-ptool.inc
+
+inherit python_setuptools_build_meta native
--- a/recipes-bsp/partition/qcom-partition-conf_git.bb
+++ b/recipes-bsp/partition/qcom-partition-conf_git.bb
@@ -1,19 +1,16 @@
 SUMMARY = "Partition configuration for Qualcomm devices"
 DESCRIPTION = "GPT partition binaries and QDL scripts for Qualcomm reference devices"
-LICENSE = "BSD-3-Clause"
-LIC_FILES_CHKSUM = "file://LICENSE;md5=b0a8acd90d872086b279ead88af03369"
 
-SRC_URI = "git://github.com/qualcomm-linux/qcom-ptool.git;branch=main;protocol=https"
-SRCREV = "7b3365284e2c19f1457e335eb25d545e9a2c08a6"
+require qcom-ptool.inc
 
-INHIBIT_DEFAULT_DEPS = "1"
+DEPENDS = "qcom-ptool-native"
 
 inherit deploy allarch
 
 do_install[noexec] = "1"
 
 do_deploy() {
-    cd platforms
+    cd ${S}/platforms
     for gpt in `find . -name gpt_main0.bin` ; do
         QCOM_PLATFORM_SUBDIR=${gpt%%/gpt_main0.bin}
         install -d ${DEPLOYDIR}/partitions/${QCOM_PLATFORM_SUBDIR}
--- a/recipes-bsp/partition/qcom-ptool.inc
+++ b/recipes-bsp/partition/qcom-ptool.inc
@@ -1,0 +1,6 @@
+HOMEPAGE = "https://github.com/qualcomm-linux/qcom-ptool"
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=b0a8acd90d872086b279ead88af03369"
+
+SRC_URI = "git://github.com/qualcomm-linux/qcom-ptool.git;branch=main;protocol=https"
+SRCREV = "0909c6ea6e143e94beb8780b64172643d093d3d4"


lumag (changes requested):
This wasn't tested, it doesn't even build.

EmbeddedAndroid:
Sorry my fault, I wasn't expected a breaking change in ptool, but I'm happy to fix it now.

Is the right approach to actually build the Python package? i.e. drop INHIBIT_DEFAULT_DEPS, add a qcom-ptool-native (inherit setuptools3 native) using the same SRC_URI/SRCREV, and DEPENDS = "qcom-ptool-native" so qcom-ptool is on the recipe-sysroot-native PATH when make runs?

lumag:
@EmbeddedAndroid yes

ricardosalveti:
Yeah, since you will have to use qcom-ptool to generate the files from this same repo, and as that is native you would have to get native to work and use the native tool to then generate all the partition files.

lumag (changes requested):
This needs to be split in two commits. One adding qcom-ptool-native and another one bumping the qcom-ptool's SRCREV (and using the -native package).

recipes-bsp/partition/qcom-ptool-native.bb
```diff
@@ -0,0 +1,10 @@
+SUMMARY = "Qualcomm partition tool (build-time CLI)"
+DESCRIPTION = "Python CLI used by qcom-partition-conf to generate per-platform GPT partition tables and QDL flashing scripts."
+HOMEPAGE = "https://github.com/qualcomm-linux/qcom-ptool"
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=b0a8acd90d872086b279ead88af03369"
+
+SRC_URI = "git://github.com/qualcomm-linux/qcom-ptool.git;branch=main;protocol=https"
+SRCREV = "0909c6ea6e143e94beb8780b64172643d093d3d4"
+
+inherit python_setuptools_build_meta native
```
ricardosalveti, line 6: I wonder if we do need a new recipe or if we could just use BBCLASSEXTEND = "native" on the main recipe instead.

And if we need to duplicate for native, it should be minimal and just include the normal recipe instead (since everything is shared, including SRC_URI and SRCREV).

lumag: I like the current approach. It brings in a separate recipes for qcom-ptool and separate recipe for partitions. The fact that they share the same repo is just an upstream detail.
(FWIW, I don't think BBCLASSEXTEND will work here, the native and non-native parts are too distinct).

ricardosalveti: Hm, but I would prefer both using and staying on the same SRCREV, as in the end they the same project. License, SRC_URI and other variables are also all shared, so it should be in a common file (or just include the base recipe directly).

> FWIW, I don't think BBCLASSEXTEND will work here, the native and non-native parts are too distinct.

Well, it is based on the same source, the only difference is that the tool is provided in native.

ndechesne: > The fact that they share the same repo is just an upstream detail.

upstream details is exactly what we need to use when we package things. why is it a good idea to make 2 recipes for the same repo? we kind of had them in the past and got everything into a single recipe..

EmbeddedAndroid: How about something like this?

Same two recipes but LICENSE/SRC_URI/SRCREV/etc. moved to a shared qcom-ptool.inc.

https://github.com/EmbeddedAndroid/meta-qcom/commit/82ac477eff2d217f544acb50f0c7c78cd3399635
https://github.com/EmbeddedAndroid/meta-qcom/commit/695bf227cf2f52930987449809effaf027254cc7

rahujosh:
@lumag @ricardosalveti 
Could we have some feedback here to review and consume the latest qcom-ptool revision? Last update was 5 days ago.
There are some Kaanapali changes in qcom-ptool which needed in meta-qcom, to unblock the axiom workflow.

lumag:
> @lumag @ricardosalveti Could we have some feedback here to review and consume the latest qcom-ptool revision? 

There is nothing to comment on the current version. @EmbeddedAndroid needs to push an update anyways (following the received feedback).

Last update was 5 days ago. There are some Kaanapali changes in qcom-ptool which needed in meta-qcom, to unblock the axiom workflow.

EmbeddedAndroid:
@lumag @ricardosalveti can you confirm which direction you want to go, I made this comment https://github.com/qualcomm-linux/meta-qcom/pull/2183#discussion_r3244003679 and didn't get a reply. It's not clear to me there was a consensus on direction. Let me know which way and i'll push an update.

lumag:
> @lumag @ricardosalveti can you confirm which direction you want to go, I made this comment [#2183 (comment)](https://github.com/qualcomm-linux/meta-qcom/pull/2183#discussion_r3244003679) and didn't get a reply. It's not clear to me there was a consensus on direction. Let me know which way and i'll push an update.

I didn't have a strong opinion. Nico does. So, let's go with the shared include file.

EmbeddedAndroid:
> > @lumag @ricardosalveti can you confirm which direction you want to go, I made this comment [#2183 (comment)](https://github.com/qualcomm-linux/meta-qcom/pull/2183#discussion_r3244003679) and didn't get a reply. It's not clear to me there was a consensus on direction. Let me know which way and i'll push an update.
> 
> I didn't have a strong opinion. Nico does. So, let's go with the shared include file.

ack. rebased and updated.

ricardosalveti (approved):

lumag (approved):

PR #2127, opened 2026-05-07
ci: base.lock: update meta-openembedded to 39321ae90b

ricardosalveti:
Relevant changes:
- 39321ae90b nginx: upgrade 1.30.1 -> 1.30.2
- 11b95aa5fd imagemagick: upgrade 7.1.2-22 -> 7.1.2-23
- 9666b61fbd python3-ldap: upgrade 3.4.5 -> 3.4.7
- 76bb679f42 procmail: Add -std=gnu17 to fix error of do_compile
- 99f4962fa3 tigervnc: Fix do_configure Error
- bfd1582c46 xdg-desktop-portal-gnome: remove gnome-shell runtime dependency
- 0ef03e2f75 libjcat: upgrade 0.2.3 -> 0.2.6
- c967ae5edc freerdp3: upgrade 3.25.0 -> 3.26.0
- a2b90ccd7b nginx: upgrade 1.30.0 -> 1.30.1
- 2f64c27b18 python3-django: upgrade 6.0.4 -> 6.0.5
- ef903fca83 python3-django: upgrade 5.2.13 -> 5.2.14
- 143d18269c nodejs: upgrade 22.22.3 -> 24.16.0 (LTS)
- 9f1a858701 nodejs-oe-cache-native: upgrade to version 24.16
- 72154f38fc kmscon: fix zlib cross-compiling errors
- c6abe6c8b0 python3-orjson: Upgrade to 3.11.9
- 83cad3966c postgresql: upgrade 17.8 -> 17.10
- c5e90eff6c lvm2: Do not install sbin/dmvdostats to sysroot
- f90ccad49a openjpeg: add ptest support
- 4009dea868 jemalloc: fix version string
- f2973002c9 xfsdump: upgrade 3.2.0 -> 3.3.0
- a0589ae5a4 uriparser: upgrade 1.0.1 -> 1.0.2
- a86ee0d4f6 swagger-ui: upgrade 5.32.5 -> 5.32.6
- 2413c8b267 python3-virtualenv: upgrade 21.3.1 -> 21.3.3
- 537a4c4484 python3-typeguard: upgrade 4.5.1 -> 4.5.2
- 8ae0b005d2 python3-twisted: upgrade 25.5.0 -> 26.4.0
- d2656956d9 python3-tox: upgrade 4.53.1 -> 4.54.0
- b71b3e83fd python3-tomlkit: upgrade 0.14.0 -> 0.15.0
[...]

PR #2186, opened 2026-05-14
ci: base.lock: update layers to latest (all but meta-oe).

ricardosalveti:
Just holding meta-oe as there is a known test regression with the updated pipewire, as discussed at https://github.com/qualcomm-linux/meta-qcom/pull/2127.

### DIFF
--- a/recipes-bsp/u-boot/u-boot-tools_%.bbappend
+++ b/recipes-bsp/u-boot/u-boot-tools_%.bbappend
@@ -1,0 +1,4 @@
+# workaround due oe-core 5e97f3c1e2, to be fixed upstream via
+# https://lists.openembedded.org/g/openembedded-core/message/236619
+DEPENDS:append:class-native = " libyaml-native"
+DEPENDS:append:qcom = " libyaml-native"
--- a/ci/base.lock.yml
+++ b/ci/base.lock.yml
@@ -3,19 +3,19 @@ header:
 overrides:
     repos:
         oe-core:
-            commit: 42fa856a00ac16b2a7a83d7ecfa60a5be192b16c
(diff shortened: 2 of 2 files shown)


anujm1 (approved):

recipes-bsp/u-boot/u-boot-tools_%.bbappend
```diff
@@ -0,0 +1,4 @@
+# workaround due oe-core 5e97f3c1e2, to be fixed upstream via
+# https://lists.openembedded.org/g/openembedded-core/message/236619
+DEPENDS:append:class-native = " libyaml-native"
```
quaresmajose, line 3: I've been here doing several tests to try and bypass it, but I haven't been able to. It's an important point to keep in mind that `:qcom` only exist in the `class-target`, and that this isn't fully clear when we use it.

quaresmajose (approved):

recipes-bsp/u-boot/u-boot-tools_%.bbappend
ricardosalveti: Yeah, also something I discovered as I was fixing this issue.

lumag: A better (easier) fix would be to apply the patch in question.

ricardosalveti: Then once merged in oe-core (or once u-boot is updated), we will break the autobuilder again.

lumag (approved):

recipes-bsp/u-boot/u-boot-tools_%.bbappend
lumag: ok

PR #2195, opened 2026-05-14
ci: convert compile action in reusable workflow

quaresmajose:
The actions is composed by several named steps but it appears as a single step in the ci output log and because of this, interpreting the different steps is very difficult.

This will allow all the steps to be displayed in the ci output logs.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -20,6 +20,9 @@ on:
         required: true
         type: string
 
+env:
+  KAS_CLONE_DEPTH: 1
+
 jobs:
   reusable_compile_job:
     runs-on: [self-hosted, qcom-u2404, amd64]
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -1,6 +1,10 @@
 on:
   workflow_call:
     inputs:
+      enabled:
+        required: false
+        type: boolean
+        default: true
       machine:
         required: true
         type: string
@@ -25,6 +29,7 @@ env:
 
 jobs:
   reusable_compile_job:
+    if: inputs.enabled
     runs-on: [self-hosted, qcom-u2404, amd64]
     name: ${{ inputs.machine }}/${{ inputs.distro_name }}${{ inputs.kernel_dirname }}
     steps:
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -1,20 +1,34 @@
-  name: Run kas for a specific build configuration
-  inputs:
-    machine:
-      required: true
-    distro_yaml:
-      required: true
-    distro_name:
-      required: true
-    kernel_yaml:
-      required: true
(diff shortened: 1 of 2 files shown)


bhargav0610:
we are seeing IMSDK audio playback failure tracking with https://github.com/qualcomm-linux/meta-qcom/issues/2223

quaresmajose:
> we are seeing IMSDK audio playback failure tracking with #2223

Could you please clarify the relationship between the issues #2223 and this PR https://github.com/qualcomm-linux/meta-qcom/pull/2195?

lumag (commented):
What would be the job name for skipped jobs after this PR?

quaresmajose:
> What would be the job name for skipped jobs after this PR?

This is not what I expected! I'll drop it.

```
inputs.enabled && format('{0}/{1}{2}', inputs.machine, inputs.distro_name, inputs.kernel_dirname)
```

The PR was already ready, I'll leave that detail for later because it's more of a cosmetic thing.

.github/workflows/compile.yml
```diff
@@ -1,6 +1,10 @@
 on:
   workflow_call:
     inputs:
+      enabled:
+        required: false
+        type: boolean
+        default: true
```
ricardosalveti, line 7: Missing commit message.

.github/actions/compile/action.yml
```diff
@@ -1,153 +0,0 @@
-name: Run kas for a specific build configuration
```
ricardosalveti, line 1: Removing this file will break meta-qcom-distro and meta-qcom-3rdparty:

`.github/workflows/build-yocto.yml:        uses: qualcomm-linux/meta-qcom/.github/actions/compile@master`

We should only remove this once the other 2 repos are aligned with the updated workflow.

.github/workflows/compile.yml
quaresmajose: Fixed

.github/actions/compile/action.yml
quaresmajose: I discarded it to integrate later.

quaresmajose: It doesn't seem like a very good idea to have a floating-point `@master` CI; this has reproducibility problems and it's not a good security practice. It would be better to be pinned with sha everywhere, and perhaps add a job to update the pinning sha's afterward.

ricardosalveti: Using a tag/branch is fine here since we control both repositories.

quaresmajose: However, reproducibility problems still persist because the dependencies of the github/actions/workflows are not deterministic. I'll move this to an issue.

lumag (approved):

ricardosalveti (approved):

ricardosalveti:
@quaresmajose please propose a backport to wrynose.

PR #2199, opened 2026-05-14
ci/world: add qcom-distro 

quaresmajose:
We also need to validate that everything in qcom-distro continues to
function whenever changes are submitted.

### DIFF
--- a/ci/world.yml
+++ b/ci/world.yml
@@ -7,7 +7,7 @@ local_conf_header:
   world_build: |
     EXCLUDE_FROM_WORLD = "1"
     EXCLUDE_FROM_WORLD:layer-qcom = "0"
-    #EXCLUDE_FROM_WORLD:layer-qcom-distro = "0"
+    EXCLUDE_FROM_WORLD:layer-qcom-distro = "0"
 
 target:
   - world


lumag (changes requested):
Doesn't build:
```
ERROR: Nothing PROVIDES 'Qualcomm-XFCE-demo-image'
ERROR: Required build target 'qcom-xfce-demo-image' has no buildable providers.
Missing or unbuildable dependency chain was: ['qcom-xfce-demo-image', 'Qualcomm-XFCE-demo-image']
```

But, frankly speaking, I think this belongs only to meta-qcom-distro.

quaresmajose:
Yeah, fix proposed in https://github.com/qualcomm-linux/meta-qcom-distro/pull/409

lumag (changes requested):
This should be a part of qcom-distro rather than meta-qcom.

quaresmajose:
> This should be a part of qcom-distro rather than meta-qcom.

I disagree. Since the qcom-distro layer is also under our control and in our responsibility, we must prevent regressions originating from the BSP layer and also from all the other other layers which are controlled here by kas.

lumag:
Yes. However it should be done other way, like we discussed in Torino: moving integration CI to qcom-distro.

If we merge this, then merging anything in case of major breakage affecting several layers would become very troublesome.

lumag:
In my opinion, CI should exactly follow the layer dependencies. Meta-qcom depends only on OE-core, so it should also build and verify against OE-core only. The distro has multiple dependencies, binding then together. Then the distro should build and test the integration.

quaresmajose:
I also agree with that, but currently there are already some discrepancies between the CIs, and the one from meta-com is the one in the best condition.

quaresmajose:
> Yes. However it should be done other way, like we discussed in Torino: moving integration CI to qcom-distro.

We keep adding more things, but they still don't go in that direction, I should change my priorities for this, but new things keep popping up.

I'll put it in draft mode just for myself while I fix the pending build issues.

quaresmajose:
Proproposed in qcom-distro  https://github.com/qualcomm-linux/meta-qcom-distro/pull/469

PR #2193, opened 2026-05-14
ci: pin qcom-linux-testkit to a tagged revision for traceability

qcom-anilyada:
LAVA jobs were cloning qcom-linux-testkit at tip of tree, making it impossible to reproduce a test run or know exactly which test scripts produced a given result.

Introduce TESTKIT_REF in test.yml as the single place to record the pinned CalVer tag (testkit-YYYY.MM.DD). Thread it through test-distro.yml and the lava-test-plans composite action so it lands in TEST_DEFINITIONS_REVISION inside gh-variables.ini, which the lava-test-plans Jinja2 templates already consume to emit a revision: field on every test definition in the rendered LAVA job YAML.

Set to testkit-2026.05.11 as the initial pinned tag.

Signed-off-by: Anil Yadav <anilyada@qti.qualcomm.com>

PR #2207, opened 2026-05-15
Backport PR 2193 to wrynose

lumag:
Backport #2193 and its dependencies to wrynose branch.


mwasilew (approved):
This all works on `master`. I' don't think we need a separate test run in next.

ricardosalveti (approved):

PR #2209, opened 2026-05-15
workflows: move matrix-output-read action to qualcomm-linux

mwasilew:
Move source of github-action-matrix-outputs-read to qualcomm-linux org. The action is used in the test-distro workflow. It was previously hosted on the private github org.

### DIFF
--- a/.github/workflows/test-distro.yml
+++ b/.github/workflows/test-distro.yml
@@ -128,7 +128,7 @@ jobs:
     outputs:
       boot_result: "${{ steps.print-boot-result.outputs.boot_result }}"
     steps:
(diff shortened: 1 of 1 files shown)


mwasilew:
@mynameistechno I think it will require update in the action whitelist in qualcomm-linux org. I'll push it to `next` once the current run completes.

mynameistechno:
> @mynameistechno I think it will require update in the action whitelist in qualcomm-linux org. I'll push it to `next` once the current run completes.

I believe actions hosted within the org don't require whitelisting.

mynameistechno (approved):

ricardosalveti (approved):

anujm1 (approved):

PR #2220, opened 2026-05-18
Test pipewire upgrade

anujm1:
Test pipewire 1.6.5 upgrade to see if that helps with audio issues.


anujm1:
Doesn't work, closing.

ricardosalveti:
It could be an issue with our tests as well, need to investigate what is going on.

PR #2221, opened 2026-05-18
camx-dlkm: import fixes for camera driver

cjha01:
Add fixes required to resolve build and compatibility issues observed during kernel upgrade to 7.1-rc2 for camera drivers.

- Fix dma_fence lock access changes by using dma_fence_spinlock() instead of direct lock usage.

- Handle removal of of_get_named_gpio() and <linux/of_gpio.h> by introducing cam_of_get_named_gpio() wrapper for GPIO descriptor API migration.

These changes are introduced to unblock kernel upgrade and stabilize camera drivers against recent kernel updates. They can be dropped once equivalent fixes are available in the new camera driver tag.


vkraleti:
@cjha01 do send these changes to github.com/qualcomm-linux/camera-driver. Once merged update the SHA in recipe here.

lumag (changes requested):
Upstream-Status is not correct. There is nothing Yocto-specific in these patches. 

Additional demerit to camera team for not following linux-next.

cjha01:
Closing this PR since the patch is not acceptable in meta-qcom for downstream drivers. We will proceed with merging via github.com/qualcomm-linux/camera-driver instead. These changes were intended to unblock the kernel upgrade, which has been pending for a long time.

PR #2231, opened 2026-05-19
workflows: ensure build_successful checkrun runs only if all builds pass

abinjose-qualcomm:
The build_successful checkrun indicates to downstream automations that all builds passed and the build artifacts are uploaded to S3. At present, it reports success even if the builds didn't run because of caching. This reverts the change made to the build_successful checkrun in #2098.

Observed as part of the nightly workflow
https://github.com/qualcomm-linux/meta-qcom/actions/runs/26008497553

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -460,17 +460,10 @@ jobs:
           key: build-success-${{ github.repository }}-${{ github.workflow }}-${{ github.ref_name }}-${{ needs.github-cache-check.outputs.run_hash }}
 
   build_successful:
-    if: |
-      always() && github.repository_owner == 'qualcomm-linux' &&
-      (needs.compile.result == 'success' ||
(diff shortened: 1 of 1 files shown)


quic-viskuma (approved):

lumag:
So, what is broken now? Why do you need another build automation notification if the build didn't happen?

abinjose-qualcomm:
> So, what is broken now? Why do you need another build automation notification if the build didn't happen?

Yes, we don't want the notification to be triggered if the build didn't happen. Currently, that is not the case. It is triggering a notification even when builds are skipped because of the caching check.

See this workflow - https://github.com/qualcomm-linux/meta-qcom/actions/runs/26008497553

lumag (approved):
I'd say that the commit message is a bit cryptic, but.. at least I could understand it.

ricardosalveti (approved):

ricardosalveti:
@abinjose-qualcomm please propose a backport of this change to the wrynose branch.

PR #2240, opened 2026-05-19
ci: set console TERM via kernel cmdline to avoid boot delay

ctheegal:
By default, systemd attempts to auto-detect the terminal type of /dev/console in order to set the TERM variable appropriately.

On IQ-9075, the terminal capability query for /dev/console blocks for ~0.5s and eventually fails with "Operation not supported". This introduces unnecessary delay in the boot path.

Setting "systemd.tty.term.console=linux" kernel command line parameter avoids the probing step and ensures a deterministic TERM configuration, improving boot time.

On IQ-9075, this reduces console detection latency by ~0.5s per boot.

### DIFF
--- a/ci/performance.yml
+++ b/ci/performance.yml
@@ -3,4 +3,4 @@ header:
 
 local_conf_header:
   cmdline: |
-    KERNEL_CMDLINE_EXTRA:append = " quiet"
+    KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=linux"


lumag:
>  "systemd.tty.term.console=linux"

Does it depend on the type of the terminal emulator used?

ricardosalveti:
Missing sign-off-by from `Wenwen Fu <wenwfu@qti.qualcomm.com>`, the author of the patch.

ricardosalveti:
> On IQ-9075, the terminal capability query for /dev/console
> blocks for ~0.5s and eventually fails with "Operation not supported".
> This introduces unnecessary delay in the boot path.

Can't we identify why this is happening? I would rather debug systemd / kernel here.

Also, your commit message starts with 'systemd' but you are changing the performance.yml kas file.

lumag (changes requested):
Please respond to review comments.

ctheegal:
> > "systemd.tty.term.console=linux"
> 
> Does it depend on the type of the terminal emulator used?

No, it does not depend on the user's terminal emulator.

The `systemd.tty.term.console` parameter strictly applies only to `/dev/console` (the physical kernel framebuffer/UART console). The Linux kernel's internal Virtual Terminal (VT) subsystem inherently implements the `linux` terminfo standard, so hardcoding `linux` here is universally correct for the physical console.

If a user connects remotely using a terminal emulator (e.g., via SSH or ADB), they are allocated a pseudo-terminal (`/dev/pts/*`), not `/dev/console`. In those cases, the SSH/ADB daemon dynamically negotiates the `$TERM` variable with the user's host emulator (e.g., setting `TERM=xterm-256color`), completely bypassing this parameter.

Exact 333ms (CONSOLE_ANSI_SEQUENCE_TIMEOUT_USEC) timeout is in the terminal_get_terminfo_by_dcs() function (file:  systemd/src/basic/terminal-util.c )  which should be avoided.

We are passing this explicitly to bypass a known ~333ms bottleneck in early systemd boot. Without it, systemd's `fixup_environment()` attempts to auto-detect the console's color capabilities by sending a DCS (Device Control String) query to `/dev/console`. Because the raw hardware console does not respond to DCS queries, systemd blocks until a 333ms timeout expires. By hardcoding `systemd.tty.term.console=linux`, systemd skips the DCS probe entirely, saving us a critical ~333ms in early boot latency.

ctheegal:
> > On IQ-9075, the terminal capability query for /dev/console
> > blocks for ~0.5s and eventually fails with "Operation not supported".
> > This introduces unnecessary delay in the boot path.
> 
> Can't we identify why this is happening? I would rather debug systemd / kernel here.

systemd's fixup_environment() attempts to auto-detect the console's color capabilities by sending a DCS (Device Control String) query to /dev/console. Because the raw hardware console does not respond to DCS queries, systemd blocks until a 333ms timeout expires. By hardcoding systemd.tty.term.console=linux, systemd skips the DCS probe entirely, saving us a critical ~333ms in early boot latency.

> 
> Also, your commit message starts with 'systemd' but you are changing the performance.yml kas file.

Yes, I am updating the commit title and re-uploading now.

lumag (approved):

ricardosalveti:
> Yes, I am updating the commit title and re-uploading now.

Please change the commit message to 'ci/performance: ....' to make it clear it only affects the performance fragment.

lumag (approved):

ricardosalveti (approved):

koenkooi (approved):
I think this is worth considering for all configurations as well.

PR #2241, opened 2026-05-19
display-modules-load: Add DRM modules preload and udev rules

ctheegal:
Install DRM modules-load, modprobe, and udev rule configs to ensure early module loading via systemd and reduce udev-trigger latency.

Changes:
- Install modules-load configs to /etc/modules-load.d/
- Install modprobe config to /etc/modprobe.d/
- Install udev rules to /etc/udev/rules.d/

Motivation:
Ensures DRM/display modules are loaded early via systemd-modules-load instead of udev-triggered modprobe, reducing device enumeration latency and enabling earlier display initialization.

Impact:
- Improves display readiness during boot
- Reduces dependency on udev-triggered module loading

Tested:
- Verified files present in rootfs (/etc/modules-load.d, /etc/modprobe.d, /etc/udev/rules.d)
- Confirmed modules loaded early via systemd-modules-load


lumag (changes requested):
Drop the AI splat from the commit message. Use simple English text and describe, what is currently wrong or needs improvement, then describe (in imperative language) actions to be done to fix that.

Also, please tell your team to stop using this AI prompt, it is awful.

recipes-core/display-modules-load/display-modules-load/00-drm-modprobe.conf
```diff
@@ -0,0 +1 @@
+options drm skip_gpu=1
```
lumag, line 1: Definite no, this is only valid for the builds with the kgsl module in it.

recipes-core/display-modules-load/display-modules-load/01-display-modules.conf
```diff
@@ -0,0 +1,17 @@
+mdt_loader
+drm_kms_helper
+drm_core_init
+socinfo
```
lumag, line 4: This is a random drop of the modules, completely not related to the display.

recipes-core/display-modules-load/display-modules-load/01-display-modules.conf
```diff
@@ -0,0 +1,17 @@
...
+display_connector
+gpucc_sa8775p
```
lumag, line 6: Definte NAK. It's very platform-specific.

recipes-core/display-modules-load/display-modules-load.bb
```diff
@@ -0,0 +1,30 @@
+
+SUMMARY = "Display DRM modules preload and udev rules"
+DESCRIPTION = "Installs DRM module load configs and udev rules"
+LICENSE = "CLOSED"
```
ricardosalveti, line 4: ??

ricardosalveti:
This looks like a real ugly hack to me, and it is not even talking about the actual gains here.

PR #2253, opened 2026-05-20
ci/qcom-distro: fix building MariaDB for Glymur

lumag:
Building mariadb for ARMv8.3-A and later, fails with:

Assembler messages:
{standard input}:169: Error: selected processor does not support `retaa' {standard input}:271: Error: selected processor does not support `retaa'

It happens because the pmem_cvap() funciton manually inserts the `.arch armv8-2.a` clause, making GAS believe that retaa instruction (inserted by GCC) is invalid. Bump the manually inserted clause to armv8.3-a, to prevent GAS from choking on the retaa instructions.

### DIFF
--- a/ci/qcom-distro.yml
+++ b/ci/qcom-distro.yml
@@ -16,6 +16,10 @@ repos:
 
   meta-openembedded:
     url: https://github.com/openembedded/meta-openembedded
+    patches:
+      plain:
+        repo: meta-qcom
+        path: patches/meta-oe/0001-mariadb-fix-building-for-the-ARMv8.3-A-and-later-sys.patch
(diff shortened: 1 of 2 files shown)


quaresmajose (dismissed):

quaresmajose:
Why did this only appear now? It's strange because the block wasn't modified.

lumag:
> Why did this only appear now? It's strange because the block wasn't modified.

https://github.com/qualcomm-linux/meta-qcom-distro/pull/297#issuecomment-4492762780

quaresmajose (approved):

ricardosalveti (approved):

PR #2254, opened 2026-05-20
ci: workflows/build-yocto: move the sdk build to the second stage 

quaresmajose:
Currently, the heavy lifting of the build is divided into two stages where the
compilation takes place. The first one we call it "warmup", and in this stage
we intend to generate the sstate-cache building different architectures so that
it can be used as much as possible in the second stage.

The SDK is built in this first stage, more specifically on machine "qcom-armv8a".
Since it's an additional step, it considerably increases the time required to
complete this machine. The SDK is taking too long to build, reaching 3 times
the build time of the target image. The SDK main pain point it's really in the
last image:do_populate_sdk task which is quite difficult to reduce.
It currently takes approximately 30 minutes minutes to complete in the best case.

One solution to remove the 30 minutes delay introduced by the SDK build would be
moving it from the first "warmup" stage to the second stage of the build.
Thus, the first one finishes earlier, causing the second one to also start earlier.
In the second stage the SDK build will reuse all the sstate-cache from the "warmup"
and will jump directly image:do_populate_sdk task.

In the reusable workflow, control parameters were added for the internal sub-blocks
to allows us to skip it if necessary:
    - inputs.world
    - inputs.target
    - inputs.sdk

buildstats.log example for the best case (when everything comes from sstate-cache):
```
0:00:01    gcc-cross-canadian-aarch64:do_package_write_rpm_setscene
0:00:02    nativesdk-clang:do_package_write_rpm_setscene
0:00:02    nativesdk-llvm:do_package_write_rpm_setscene
0:00:02    nativesdk-spirv-tools:do_package_write_rpm_setscene
0:00:04    nativesdk-glibc-locale:do_package_write_rpm_setscene
0:00:05    nativesdk-glibc-locale:do_packagedata_setscene
0:00:08    nativesdk-qemu:do_package_write_rpm_setscene
0:34:07    qcom-multimedia-image:do_populate_sdk
0:35:56    qcom-multimedia-proprietary-image:do_populate_sdk
```

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -192,6 +192,7 @@ jobs:
             yamlfile: ""
     uses: ./.github/workflows/compile.yml
     with:
+      name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
       machine: ${{matrix.machine}}
       distro_yaml: ${{matrix.distro.yamlfile}}
       distro_name: ${{matrix.distro.name}}
@@ -206,6 +207,7 @@ jobs:
       (inputs.skip_if_github_cached != true || needs.github-cache-check.outputs.cache_hit != 'true')
     uses: ./.github/workflows/compile.yml
     with:
+      name: qcom-armv8a/qcom-distro-sdk
       sdk: true
       target: false
       world: false
@@ -401,6 +403,7 @@ jobs:
                 yamlfile: ""
     uses: ./.github/workflows/compile.yml
     with:
+      name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
       enabled: ${{inputs.profile != 'pr' || (matrix.distro.name != 'debug' && matrix.distro.name != 'performance')}}
       machine: ${{matrix.machine}}
       distro_yaml: ${{matrix.distro.yamlfile}}
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -199,6 +199,23 @@ jobs:
       kernel_dirname: ${{matrix.kernel.dirname}}
       cache_dir: ${{ needs.compile-env.outputs.CACHE_DIR }}
 
+  compile_sdk:
+    needs: [github-cache-check, compile_warm_up, compile-env]
+    if: |
+      github.repository_owner == 'qualcomm-linux' &&
+      (inputs.skip_if_github_cached != true || needs.github-cache-check.outputs.cache_hit != 'true')
+    uses: ./.github/workflows/compile.yml
+    with:
+      sdk: true
+      target: false
+      world: false
+      machine: 'qcom-armv8a'
+      distro_yaml: ':ci/qcom-distro.yml'
+      distro_name: 'qcom-distro-sdk'
+      kernel_yaml: ''
+      kernel_dirname: ''
+      cache_dir: ${{ needs.compile-env.outputs.CACHE_DIR }}
+
   compile:
     needs: [github-cache-check, compile_warm_up, compile-env]
     if: |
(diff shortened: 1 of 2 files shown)


.github/workflows/compile.yml
```diff
@@ -89,16 +102,15 @@ jobs:
...
         shell: bash
         run: |
           $KAS_CONTAINER build ${KAS_YAMLS}
           ci/kas-container-shell-helper.sh ci/yocto-buildstats.sh
           mv $KAS_WORK_DIR/build/buildstats* .
 
       - name: Kas build SDK
-        if:  inputs.machine == 'qcom-armv8a' &&
-             inputs.distro_yaml == ':ci/qcom-distro.yml' &&
-             inputs.kernel_yaml == ''
```
ricardosalveti: You just changed this in the previous commit, just squash instead.

quaresmajose: fixed

lumag (dismissed):

ricardosalveti (dismissed):

lumag:
@quaresmajose @mwasilew did we test it at next?

mwasilew:
I pushed it to `next` just now: https://github.com/qualcomm-linux/meta-qcom/actions/runs/26576552463

quaresmajose:
> I pushed it to `next` just now: https://github.com/qualcomm-linux/meta-qcom/actions/runs/26576552463

There one falilure on qcs615-ride-pre-merge but not related with this PR

.github/workflows/build-yocto.yml
```diff
@@ -200,6 +200,25 @@ jobs:
       kernel_dirname: ${{matrix.kernel.dirname}}
       cache_dir: ${{ needs.compile-env.outputs.CACHE_DIR }}
 
+  compile_sdk:
```
ndechesne, line 203: overall PR looks good, however this patch and commit in particular is hard to understand. I don't think the commit message explains well the purpose of the change.

.github/workflows/compile.yml
```diff
@@ -5,6 +5,18 @@ on:
         required: false
         type: boolean
         default: true
+      sdk:
```
ndechesne, line 8: this should be better explained in the commit message.

.github/workflows/build-yocto.yml
quaresmajose: I rewrote the commit and added more context.

.github/workflows/compile.yml
quaresmajose: Describe briefly the rationale behind the parameters.

ndechesne (dismissed):

.github/workflows/build-yocto.yml
```diff
@@ -192,13 +192,34 @@ jobs:
...
+    needs: [github-cache-check, compile_warm_up, compile-env]
+    if: |
+      github.repository_owner == 'qualcomm-linux' &&
+      (inputs.skip_if_github_cached != true || needs.github-cache-check.outputs.cache_hit != 'true')
+    strategy:
+      fail-fast: true
```
ricardosalveti: fail-fast only works with matrix jobs, which is not the case here anymore

.github/workflows/build-yocto.yml
```diff
@@ -192,13 +192,34 @@ jobs:
...
     with:
+      name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
       machine: ${{matrix.machine}}
       distro_yaml: ${{matrix.distro.yamlfile}}
       distro_name: ${{matrix.distro.name}}
       kernel_yaml: ${{matrix.kernel.yamlfile}}
       kernel_dirname: ${{matrix.kernel.dirname}}
       cache_dir: ${{ needs.compile-env.outputs.CACHE_DIR }}
 
+  compile_sdk:
```
ricardosalveti, line 203: build_successful, cache-save and publish_summary should now require dependency on compile_sdk, otherwise a failure in this step won't result in a failed run

.github/workflows/build-yocto.yml
quaresmajose: Fixed, it was together in copypast.

.github/workflows/build-yocto.yml
quaresmajose: good point

ricardosalveti (approved):

ricardosalveti:
Merging based on the previous approvals.

@quaresmajose this will require a manual backport to wrynose.

quaresmajose:
> Merging based on the previous approvals.
> 
> @quaresmajose this will require a manual backport to wrynose.

Done https://github.com/qualcomm-linux/meta-qcom/pull/2424

PR #2259, opened 2026-05-21
rb3gen2-core-kit: add paho-mqtt-c as machine essential

hariraj34:
Add Eclipse Paho MQTT C client library to rb3gen2-core-kit
machine configuration to enable MQTT connectivity with SSL/TLS
support out of the box.

Validated on QLI20_May8 build on RB3 Gen2 Core Kit:
- Paho MQTT C client SSL publish/subscribe verified
- mTLS with client certificate verification working

### DIFF
--- a/conf/machine/rb3gen2-core-kit.conf
+++ b/conf/machine/rb3gen2-core-kit.conf
@@ -29,6 +29,7 @@ MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     qairt-sdk-hexagon-v68 \
     qcom-fastcv-binaries-thundercomm-rb3gen2-dsp \
     qps615-dlkm \
+    paho-mqtt-c \
 "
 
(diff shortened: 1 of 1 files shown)


quaresmajose:
This is not an essential package for the machine to boot. Things like that you should do in meta-qcom-distro, but you need to have tests to verify the functionality in question.

Please look at:

https://docs.yoctoproject.org/ref-manual/variables.html#term-MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS

A list of recommended machine-specific packages to install as part of the image being built. The build process does not depend on these packages being present. However, because this is a “machine-essential” variable, the list of packages are essential for the machine to boot. The impact of this variable affects images based on packagegroup-core-boot, including the core-image-minimal image.

lumag:
I agree, this is not essential for the machine to boot. Closing.

PR #2236, opened 2026-05-19
fit-dtb-compatible: remove two repeated entries for Kodiak

jiegan0107:
Remove two invalid entries which are added by mistake.

Fixes: https://github.com/qualcomm-linux/meta-qcom/commit/12d9b826facf5051da25081f02bace285c14c478 ("fit-dtb-compatible: add SoC staging FIT DTB compatible entries")

PR #2273, opened 2026-05-23
selftests: add matrix validation for dtb and detect duplicated compatibles 

ricardosalveti:
Add QcomFitImageMatrixTests class with two test methods to validate DTB coverage without full image builds and two tests to QcomFitImageMatrixTests that catch mistakes like those fixed in PR https://github.com/qualcomm-linux/meta-qcom/pull/2236.


lumag:
I think instead we need to invert the logic. The `FIT_DTB_COMPATIBLE` (or a new variable) should be indexed by the FIT configuration values rather than the lists of DTBs being used.

ricardosalveti:
> I think instead we need to invert the logic. The `FIT_DTB_COMPATIBLE` (or a new variable) should be indexed by the FIT configuration values rather than the lists of DTBs being used.

But you still need to have validate that the DTBs are indeed valid, as you still need to include the dtb / dtbos in the end.

This PR is not done to change the behavior, this was just to catch possible issues with the current file, we can for sure change it, but it would be out of the scope from this PR I would say.

lumag (approved):

koenkooi (approved):
LGTM, the more checks for things like this, the better

PR #2275, opened 2026-05-23
ci/base : Limit swiotlb to 128

bibekpatro:
Add kernel command line parameter to set minimum
limit of 128 for bounce buffer as this will be used for DMA use cases on kmalloc addresses which are not page aligned.


bibekpatro:
Supposed to be taken care in URM, abandoning the PR

PR #2276, opened 2026-05-24
Several fixes for the CI scripts

lumag:
Fix several small issues, accumulated through the course of the CI development.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -73,7 +73,6 @@ jobs:
       - uses: actions/checkout@v6
         with:
           persist-credentials: false
-          ref: ${{ inputs.git_ref }}
 
       - name: Download kas lockfile
         uses: actions/download-artifact@v7
--- a/.github/workflows/test-distro.yml
+++ b/.github/workflows/test-distro.yml
@@ -43,7 +43,7 @@ on:
 
 jobs:
   prepare-boot-jobs:
-    name: "Prepare boot jobs for ${{ inputs.distro }}"
+    name: "Prepare boot jobs for ${{ inputs.distro_name }}"
     runs-on: ubuntu-latest
     outputs:
       jobmatrix: ${{ steps.lava-test-plans.outputs.jobmatrix }}
@@ -152,7 +152,7 @@ jobs:
           echo "boot_result=$BOOT_RESULT" >> "$GITHUB_OUTPUT"
 
   prepare-premerge-jobs:
-    name: "Prepare pre-merge jobs for ${{ inputs.distro }}"
+    name: "Prepare pre-merge jobs for ${{ inputs.distro_name }}"
     runs-on: ubuntu-latest
     needs: boot-job-result
     # run only if boot jobs are successful


lumag:
Colleagues,

Without the last fix backported to the wrynose branch, the tests are not being executed / reported correctly.

anujm1 (dismissed):

vkraleti (dismissed):

koenkooi (dismissed):

mwasilew (commented):
I think this might open the door for running workflows in the context of different repository than meta-qcom. TBH I don't know how to test this properly and make sure it doesn't allow to run tests from random forks in our LAB. I'll try to figure something out today.

.github/workflows/test.yml
```diff
@@ -6,6 +6,11 @@ on:
       build_id:
         required: true
         type: string
+      git_ref:
```
mwasilew: this is unused. I suspect it should be passed to `test-distro.yml`

mwasilew:
I tried this change and I'm still getting the version of the workflow from `master` executed on PRs to `wrynrose`. Investigating further.

mwasilew:
It doesn't seem to be possible to trigger the branch specific workflow with `workflow_run` trigger. Github always calls a workflow version from `master` branch. Would it be acceptable to use the event file (which we already save) to pass the target branch to the workflow? This means all branch use cases would need to be addressed in the `master` workflow. This is the key that can be used:
```
{
  "pull_request": {
    "base": {
      "ref": "wrynrose",
...
```
I skipped the redundant keys.

lumag:
> It doesn't seem to be possible to trigger the branch specific workflow with `workflow_run` trigger. Github always calls a workflow version from `master` branch. Would it be acceptable to use the event file (which we already save) to pass the target branch to the workflow? This means all branch use cases would need to be addressed in the `master` workflow. This is the key that can be used:

Do you have a contents of `github.event` somewhere for the comparison?

mwasilew:
> 
> Do you have a contents of `github.event` somewhere for the comparison?

We're saving `Event File` for each build that is triggered by PR. You can download it from the workflow page. Example:
https://github.com/qualcomm-linux/meta-qcom/actions/runs/26557002069/artifacts/7258518073
from this workflow:
https://github.com/qualcomm-linux/meta-qcom/actions/runs/26557002069

lumag:
> > Do you have a contents of `github.event` somewhere for the comparison?
> 
> We're saving `Event File` for each build that is triggered by PR. You can download it from the workflow page. Example: https://github.com/qualcomm-linux/meta-qcom/actions/runs/26557002069/artifacts/7258518073 from this workflow: https://github.com/qualcomm-linux/meta-qcom/actions/runs/26557002069

I don't see events for the 'Test PR build' workflows.

mwasilew:
The event file is saved in build workflow as it's called in the context of PR. Test workflow is executed in the context of main repository so it needs to learn about the PR from this file.

lumag:
> The event file is saved in build workflow as it's called in the context of PR. Test workflow is executed in the context of main repository so it needs to learn about the PR from this file.

I'm thinking about the `${{ github.event.workflow_run.foo }}`

ricardosalveti (approved):

koenkooi (approved):

mwasilew (approved):

PR #2277, opened 2026-05-24
LocationClientAPI: Add location HAL header recipe 

quic-vthuniki:
Add location-headers recipe for location HAL headers.
This recipe provides required definitions and APIs to interface with location HAL layer

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
+++ b/dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
@@ -1,0 +1,33 @@
+SUMMARY = "Qualcomm Location HAL libraries"
+
+DESCRIPTION = "Provides the Qualcomm location hardware abstraction \
+layer (HAL) libraries used by upper-layer location services and \
+applications. Includes platform abstraction, common utilities, core \
+location engine, batching, geofencing, and GNSS interface components, \
+along with the runtime accounts and groups required to access location \
+sockets and configuration."
+
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+SRC_URI = "git://github.com/qualcomm-linux/location-hal-qcom.git;protocol=https;branch=location.lnx.0.0;tag=v${PV}"
+SRCREV  = "96b21c90da387747155debb9f482fa11eed9bf74"
+
+inherit autotools pkgconfig useradd
+
+DEPENDS = "glib-2.0 sqlite3 libxml2"
+
+EXTRA_OECONF = " --with-glib"
+CPPFLAGS:append = " -DLOC_QCLINUX_TARGET "
+
+GROUPADD_PACKAGES = "${PN}"
+# locclient group gates access to the location HAL daemon's IPC socket;
+# any process that needs to talk to the daemon must belong to this group.
+GROUPADD_PARAM:${PN} = "--system locclient"
+
+USERADD_PACKAGES = "${PN}"
+USERADD_PARAM:${PN} = "--system --shell /sbin/nologin --groups locclient --no-create-home --user-group gps"
+
+do_install:append() {
+    install -m 0644 -D ${S}/etc/gps.conf ${D}${sysconfdir}/gps.conf
+}


recipes-location/location/include/common-location-defines.inc
```diff
@@ -0,0 +1,4 @@
+# Common bitbake recipe information for location meta layer.
+# Below are common values, statements and functions.
+
+PACKAGE_ARCH    ?= "${TUNE_PKGARCH}"
```
lumag: This is definitely not worth having a separate include. Fold it in annd provide an explanation, why it is required. BTW, why?

recipes-location/location/include/location-license-bsd3-clause.inc
```diff
@@ -0,0 +1,3 @@
+# include for recipe files which require BSD-3-Clause license
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/BSD-3-Clause-Clear;md5=7a434440b651f4a472ca93716d01033a"
```
lumag: NAK. Inline it and checksum the actual license file in the source tree.

lumag (changes requested):
1. It's `recipes-navigation` rather than `recipes-location`.
2. What are the users for these headers?

recipes-location/location/loc-pla-hdr_git.bb
```diff
@@ -0,0 +1,19 @@
+inherit allarch
+require include/common-location-defines.inc
+require include/location-license-bsd3-clause.inc
+
+DESCRIPTION = "GPS Loc Platform Library Abstraction"
```
anujm1: Please provide a proper SUMMARY and DESCRIPTION in both recipes.

recipes-location/location/loc-pla-hdr_git.bb
```diff
@@ -0,0 +1,19 @@
...
+
+SRC_URI = "git://git.codelinaro.org/clo/le/platform/hardware/qcom/gps.git;protocol=https;branch=location.lnx.0.0;destsuffix=hardware/qcom/gps"
+SRCREV  = "18cec4cd877f758292f5252d47c755b50f2838b3"
```
anujm1: What is the version here?

recipes-location/location/loc-pla-hdr_git.bb
```diff
@@ -0,0 +1,19 @@
+inherit allarch
+require include/common-location-defines.inc
+require include/location-license-bsd3-clause.inc
+
+DESCRIPTION = "GPS Loc Platform Library Abstraction"
+
+SRC_URI = "git://git.codelinaro.org/clo/le/platform/hardware/qcom/gps.git;protocol=https;branch=location.lnx.0.0;destsuffix=hardware/qcom/gps"
```
anujm1: Why do we need to clone to `destsuffix` here if it's a separate recipe?

recipes-location/location/location-api-iface_git.bb
```diff
@@ -0,0 +1,18 @@
+inherit allarch
+require include/common-location-defines.inc
+require include/location-license-bsd3-clause.inc
+
+DESCRIPTION = "location api interface"
+
+SRC_URI = "git://git.codelinaro.org/clo/le/platform/hardware/qcom/gps.git;protocol=https;branch=location.lnx.0.0;destsuffix=hardware/qcom/gps"
```
anujm1: Why do we need two recipes for packing headers for this same project? Why do we need to manually copy them?

recipes-location/location/include/common-location-defines.inc
quic-vthuniki: >why?

This seem redundant, I will fix it.
From my initial experiments, PACKAGE_ARCH was influencing the naming of the generated tar files for proprietary modules. I was informed that the tar artifacts needed to be aligned with the armv8 architecture. Hence i explicitly set PACKAGE_ARCH to ${TUNE_PKGARCH}.

However, after further validation, I confirmed that when not explicitly set, the default PACKAGE_ARCH already resolves to the appropriate armv8 value. So this assignment is redundant, and I’ll remove it in the next update.

>definitely not worth having a separate include

My baseline code is common between Android and LE.
In my follow-on PRs, there are recipes where we need to define a few CFLAGS that are used in the source code to differentiate parts of the code between Android and LE, and to enable some inlines/macros specific to the LE baseline. For that, we are using this .inc file to define those configurations.

example: 
`EXTRA_OECONF = " --with-glib"
`
code ref: 
https://git.codelinaro.org/clo/le/platform/hardware/qcom/gps/-/blob/location.lnx.0.0/utils/log_util.h?ref_type=heads#L78

Without the GLIB macro defined, I ran into an undefined macro error for ALOGE.

```
# compiler flag to identify QClinux target
CFLAGS:append = " -DLOC_QCLINUX_TARGET "
CPPFLAGS:append = " -DLOC_QCLINUX_TARGET "
```
There are some code compile decisions based on these CFLAGS. 
https://git.codelinaro.org/clo/le/platform/vendor/qcom-opensource/location/-/blob/location.lnx.8.5.r1-rel/client_api_testapp/main.cpp?ref_type=heads#L2655

Hence, I wanted to retain this file in the initial phase.

recipes-location/location/loc-pla-hdr_git.bb
quic-vthuniki: Sure, i will update it

recipes-location/location/loc-pla-hdr_git.bb
quic-vthuniki: I had taken these recipes as base from QLI 1.0. If the suggestion is to avoid revision based repo clone and always work on tip, we are okay. 
Please help confirm if we need to always work on tip of the repo?

recipes-location/location/include/common-location-defines.inc
lumag: No. Add that file when actually necessary rather than when thinking that it will be necessary in future (which might never come). Also be sure that you correctly split flags between `CFLAGS` and `CPPFLAGS`. Having the same flag in both is a definite mistake. `CFLAGS` should be used for the flags specific to the C compiler. All preprocessor options (`-I`, `-D`, etc.) should go to CPPFLAGS.

lumag:
The major question from my side would be, why are we bring the whole HAL from Android? Can't we use a native implementation in style of https://github.com/qualcomm-linux/meta-qcom/blob/master/dynamic-layers/openembedded-layer/recipes-navigation/gpsd/gpsd/0001-Introduce-Qualcomm-PDS-service-support.patch ?

recipes-location/location/location-api-iface_git.bb
quic-vthuniki: Reason for having separate recipe files comes from the Android build system, which was used as a reference in our legacy code. The same approach has been carried forward by the team into LE as well.
I agree that we can consolidate some of these and reduce the number of recipe files overall. However, I would prefer to take that up as a follow-up improvement/optimization.
For now, the goal is to enable the baseline with minimal changes to the existing recipe structure.

recipes-location/location/loc-pla-hdr_git.bb
quic-vthuniki: This was added mainly for readability and to follow a consistent convention. If you prefer to avoid it, I can remove it. Alternatively, can i take this up as part of a separate cleanup activity as all of our recipes have the similiar pattern

recipes-location/location/include/location-license-bsd3-clause.inc
quic-vthuniki: I checked how this is handled in other modules (e.g., recipes-multimedia/camx/camx-dlkm) and will update this accordingly.
My only concern is that github repository is currently undergoing LOST scanning, and it may take some time before the repo is created and license file can be added to the source tree.
In the meantime, I would appreciate your guidance on how you would prefer me to proceed, so that the follow-on recipes can continue to be reviewed without delay.

recipes-location/location/loc-pla-hdr_git.bb
lumag: The suggestion is to have tags, organizing the repo contents and repo changes.

recipes-location/location/loc-pla-hdr_git.bb
lumag: Please avoid the destdir unless required.

recipes-location/location/location-api-iface_git.bb
lumag: > prefer to take that up as a follow-up improvement/optimization.

Why? Please make it good from the beginning.

quic-vthuniki:
>The major question from my side would be, why are we bring the whole HAL from Android

The HAL repository is a common codebase shared between Android and LE, and we have maintained it this way to maximize code reuse across different software platforms.
While the QLI repo itself is not currently shared with Android, its baseline is derived from a common branch that supports both Android and LE.
From a source code perspective, the repository is not strictly Android-specific—it contains a significant amount of native code. For LE, we selectively build only the required modules from this repository.


>Can't we use a native implementation in style of https://github.com/qualcomm-linux/meta-qcom/blob/master/dynamic-layers/openembedded-layer/recipes-navigation/gpsd/gpsd/0001-Introduce-Qualcomm-PDS-service-support.patch 

The referenced approach appears to use the QRTR module directly. However, we have certain proprietary features that depend on shared modules from the HAL. These features cannot be supported purely using QRTR- or QMI-based implementations alone.

lumag:
> > Can't we use a native implementation in style of https://github.com/qualcomm-linux/meta-qcom/blob/master/dynamic-layers/openembedded-layer/recipes-navigation/gpsd/gpsd/0001-Introduce-Qualcomm-PDS-service-support.patch
> 
> The referenced approach appears to use the QRTR module directly. However, we have certain proprietary features that depend on shared modules from the HAL. These features cannot be supported purely using QRTR- or QMI-based implementations alone.

Features, such as...? How are they controlled / implemented? Can they be supported separately? Why do they depend on HAL modules?

quic-vthuniki:
>It's recipes-navigation rather than recipes-location.
What are the users for these headers?

GPSd is part of recipes-navigation.
In contrast, recipes-location brings in the LCA(Location Client API) SDK, which offers a parallel interface for location capabilities, similar to GPSd.


>Features, such as...? How are they controlled / implemented? Can they be supported separately? Why do they depend on HAL modules?

These are some of the high-level features supported through the LCA SDK:

- GNSS assistance data download (proprietary feature)
- Multiplexing of multiple client requests
- Location batching
- Geofencing
- Network-based positioning^
- ATL call bring-up for AGPS^

^these features are not part of the current POR. However, we do receive PdM/customer requests to enable them, and we plan to bring in only the required modules accordingly.

Although both GPSd(with [0001-Introduce-Qualcomm-PDS-service-support.patch](https://github.com/qualcomm-linux/meta-qcom/blob/master/dynamic-layers/openembedded-layer/recipes-navigation/gpsd/gpsd/0001-Introduce-Qualcomm-PDS-service-support.patch)) and LCA use QMI interface, our exisiting customers currently rely on the LCA SDK. Therefore, we are prioritizing LCA enablement in QLI 2.0.

Support for applicable features through GPSd is also planned, but will be handled as a separate effort.

recipes-location/location/location-api-iface_git.bb
quic-vthuniki: Sure, we are looking at the options to reduce the number of recipes which are header only and combine them into a single recipe. I will get back with proposal

lumag:
> > It's recipes-navigation rather than recipes-location.
> > What are the users for these headers?
> 
> GPSd is part of recipes-navigation. In contrast, recipes-location brings in the LCA(Location Client API) SDK, which offers a parallel interface for location capabilities, similar to GPSd.

It's still navigation. As such, use recipes-navigation subdir.

> 
> > Features, such as...? How are they controlled / implemented? Can they be supported separately? Why do they depend on HAL modules?
> 
> These are some of the high-level features supported through the LCA SDK:
> 
> * GNSS assistance data download (proprietary feature)

It's implemented in other software tools too.

> * Multiplexing of multiple client requests

Gpsd provides broadcasting of the NMEA data to multiple clients, if that's what you mean here.

> * Location batching
> * Geofencing
> * Network-based positioning^
> * ATL call bring-up for AGPS^
> 
> ^these features are not part of the current POR. However, we do receive PdM/customer requests to enable them, and we plan to bring in only the required modules accordingly.
> 
> Although both GPSd(with [0001-Introduce-Qualcomm-PDS-service-support.patch](https://github.com/qualcomm-linux/meta-qcom/blob/master/dynamic-layers/openembedded-layer/recipes-navigation/gpsd/gpsd/0001-Introduce-Qualcomm-PDS-service-support.patch)) and LCA use QMI interface, our exisiting customers currently rely on the LCA SDK. Therefore, we are prioritizing LCA enablement in QLI 2.0.
> 
> Support for applicable features through GPSd is also planned, but will be handled as a separate effort.

Let's schedule a meeting, discussing possible development plans and pitfalls.

recipes-location/location/location-api-iface_git.bb
quic-vthuniki: We identified a set of header-only recipes that currently package only .h files from the same source repository.
Proposed approach:

Consolidate these header-only recipes into a single recipe per source repository. The unified recipe will install all required headers from the repo. To avoid impacting existing dependencies, we will retain the original package names using PACKAGES and PROVIDES

This reduces recipe duplication while preserving backward compatibility with the current dependency structure.

Please let me know if this approach is inline with your guidance?

recipe file: dynamic-layers\openembedded-layer\recipes-navigation\location-client-api\location-headers_git.bb
```
PROVIDES += "loc-pla-hdr location-api-iface"
PACKAGES = "loc-pla-hdr location-api-iface"

do_install() {
    install -d ${D}${includedir}
    install -m 644 ${S}/location/*.h ${D}${includedir}
    install -m 644 ${S}/pla/oe/*.h ${D}${includedir}
}

FILES:loc-pla-hdr        = "${includedir}/loc_pla*.h"
FILES:location-api-iface = "${includedir}/ILocationAPI.h \
                            ${includedir}/LocationAPIClientBase.h \
                            ${includedir}/LocationAPI.h \
                            ${includedir}/LocationAPIUtils.h \
                            ${includedir}/LocationDataTypes.h \
                            ${includedir}/location_interface.h \
"

```

quic-vthuniki:
Hi Reviewers,
I’m currently working through a few logistical items (GitHub repo creation and adding the license file to the existing repo), which are taking a bit longer than expected.
In the meantime, I’d like to get your feedback on the following points to ensure the PR aligns with the expected guidelines.

>It's still navigation. As such, use recipes-navigation subdir.

Based on this feedback, I’m planning to place all LCA SDK-related recipes under a dedicated location-client-api directory within recipes-navigation.
We are following this folder structure as we plan to support both LCA SDK and GPSd under the same recipes-navigation hierarchy (with GPSd support to be added/extended as needed).
Could you please confirm if the following directory hierarchy looks appropriate?

```
recipes-navigation/
├── gpsd
│   ├── gpsd
│   │   └── 0001-Introduce-Qualcomm-PDS-service-support.patch
│   └── gpsd_%.bbappend
└── location-client-api
    ├── files
    │   ├── 0001-loc-remove-properties.h-usage.patch
    │   └── 0002-Fix-requiredlibs-in-msgproto-makefile.patch
    ├── include
    │   ├── common-location-defines.inc
    │   └── location-license-bsd3-clause.inc
    ├── location-api_git.bb
    ├── location-api-msg-proto_git.bb
    ├── location-client-api-hdr_git.bb
    ├── location-headers_git.bb

```

lumag:
If it is not obvious from the previous feedback:

>     ├── include
>     │   ├── common-location-defines.inc
>     │   └── location-license-bsd3-clause.inc

Drop those two.

quic-vthuniki:
>Drop those two.

I agree with this; the intent was only to illustrate the folder structure, not to include these in the final changes.
I will drop them and incorporate all feedback in the next update. The include directory will be introduced only if required; otherwise, it will be removed while addressing the LICENSE and LIC_FILES_CHKSUM related comments.

recipes-location/location/loc-pla-hdr_git.bb
quic-vthuniki: Hi reviewers,
I am experimenting with the tags section and am looking for few clarifications on the same.

For reference, I am using the example below:
[camx-dlkm_1.0.3.bb](https://github.com/qualcomm-linux/meta-qcom/blob/master/recipes-multimedia/camx/camx-dlkm_1.0.3.bb)  

```
SRC_URI = " \
git://github.com/qualcomm-linux/camera-driver.git;protocol=https;branch=camera-kernel.qclinux.0.0;tag=v${PV}
```

Is it always recommended to use repository tags as the recipe version i.e tag=v${PV} ? If the tag changes in the repository, should the recipe file also be renamed to reflect the new tag/revision since both are aligned?

In my case, I have multiple recipe files using the same repository. Would you suggest following the same approach and keeping the recipe version aligned with the repository tag?

For example, if the repository tag is 1.0.0, should all recipes be named as <recipe>_1.0.0.bb? If a new tag 1.0.1 is created, should all recipe files then be renamed to <recipe>_1.0.1.bb?

lumag: Yes.

lumag: Please use semantic versioning for the tags.

recipes-location/location/include/common-location-defines.inc
quic-vthuniki: I have removed this file in my updated PR. I will add back only if necessary

recipes-location/location/include/location-license-bsd3-clause.inc
quic-vthuniki: Addressed this issue in updated recipe. Added License and checksum inline in the recipe wit actual license file from source repo

recipes-location/location/loc-pla-hdr_git.bb
quic-vthuniki: Updated in new recipe file

recipes-location/location/loc-pla-hdr_git.bb
quic-vthuniki: Added tag field to SRC_URI path as suggested in updated recipe file

recipes-location/location/loc-pla-hdr_git.bb
quic-vthuniki: Removed destdir usage in our recipe files(current PR and all future PRs that are coming).

recipes-location/location/location-api-iface_git.bb
lumag: Why do you need separate packages? Just drop all of them and provide a single recipe and a single package with the headers.

lumag:
Drop the merge commit

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.0.bb
```diff
@@ -0,0 +1,23 @@
+SUMMARY = "Header-only package for Location API and PLA interfaces"
+
+DESCRIPTION = "This recipe provides header files for the Qualcomm Location      \
```
lumag: The extra whispaces will crawl into the description of the package, making it look weird. please drop them (and drop the indentation). See how other packages in OE-Core are written.

recipes-location/location/location-api-iface_git.bb
quic-vthuniki: Idea was to keep the recipes similar to other baselines so developers are familiar. But as suggested, updated recipe to single package

quic-vthuniki:
>Drop the merge commit

Fixed

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.0.bb
```diff
@@ -0,0 +1,23 @@
...
+               API and platform abstraction layer (PLA) interfaces. It includes \
+               definitions required for applications and libraries to integrate \
+               with the location stack"
+
+inherit allarch
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=8ea33b4e44af458236377788d0c61379"
```
lumag: Drop trailing whitespaces from the licence text. The md5 sum should be `2998c54c288b081076c9af987bdf4838`.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.0.bb
```diff
@@ -0,0 +1,23 @@
...
+
+S = "${UNPACKDIR}/${BP}"
+
+do_install() {
+    install -d ${D}${includedir}
+    install -m 644 ${S}/location/*.h ${D}${includedir}
+    install -m 644 ${S}/pla/oe/*.h ${D}${includedir}
+}
+
+FILES:${PN} = "${includedir}/"
```
lumag: Drop, headers belong to the -dev package.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.1.bb
```diff
@@ -0,0 +1,23 @@
...
+               with the location stack"
+
+inherit allarch
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=8ea33b4e44af458236377788d0c61379"
+
+SRC_URI = "git://github.com/qualcomm-linux/location-hal-qcom.git;protocol=https;branch=location.lnx.0.0;tag=v${PV}"
+SRCREV  = "4a35b86db087f7b002ccce6edd2ac6da2b1f9b8d"
+
+S = "${UNPACKDIR}/${BP}"
```
lumag:
```
do_configure[noexec] = "1"
do_compile[noexec] = "1"
```

lumag (changes requested):
Tidy everything up. Squash the commits. Remove Gerrit hooks from your Git clone. They will never be used here. Remove Gerrit tags from the commit mesage. Write a proper one: English text, no unnececssary bullet lists, proper capitalisation of the words (HAL not hal, etc). Write a commit subject that follows the current practice.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.0.bb
quic-vthuniki: Am confused on the proposed md5 sum. 

I have copied the exact marking that LOST team has given me for which the md5 in my patch is correct. 

> vthuniki@hu-vthuniki-hyd:/local/mnt/workspace/builds/qli_20_experiment/gitpush/test1_md5_hal_project$ curl -sL https://raw.githubusercontent.com/qualcomm-linux/location-hal-qcom/location.lnx.0.0/
> LICENSE.txt | md5sum
> `8ea33b4e44af458236377788d0c61379`  -
> vthuniki@hu-vthuniki-hyd:/local/mnt/workspace/builds/qli_20_experiment/gitpush/test1_md5_hal_project$

After removing trailing whitespaces, am getting a different md5 than what you have proposed. Could you share the exact license file that resulted in `2998c54c288b081076c9af987bdf4838`


> vthuniki@hu-vthuniki-hyd:/local/mnt/workspace/builds/qli_20_experiment/gitpush/test1_md5_hal_project/location-hal-qcom$ md5sum LICENSE.txt
> `7e50fb71aa022627cbbc82d17793a8c9`  LICENSE.txt
>

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.0.bb
quic-vthuniki: I observed varying conventions in OE-Core for recipe descriptions (single-line and multi-line, with and without indentation). However, trailing whitespace is indeed an issue and I have addressed it, and aligned the formatting accordingly.

https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-multimedia/sbc/sbc_2.2.bb 

https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-multimedia/ffmpeg/ffmpeg_8.1.1.bb 

https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-graphics/images/core-image-weston-sdk.bb

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.0.bb
quic-vthuniki: Done

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.1.bb
quic-vthuniki: Looks like these were unintentionally dropped during rebase or subsequent updates. I have added them back.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.0.bb
lumag: You can push back to the LOST team, stating that the maintainer asked to use the same copy of the licence as before. See:
```
$ grep CHKSUM `git grep -l BSD-3-Clause-Clear` | sed -e 's/.*=//g' | sort | uniq -c
      1  "\
      5 2998c54c288b081076c9af987bdf4838"
      2 7a434440b651f4a472ca93716d01033a"
      1 8c54773149e04ded5c0c3e293bb13509"
      1 8e1eb38e3de3966193d29f31f5d7e684"
      1 9701d0ef17353f1d05d7b74c8712ebbd"
      1 e877b35748195c6ab87bf2d1ebed9a89"

$ git grep 2998c54c288b081076c9af987bdf4838
dynamic-layers/openembedded-layer/recipes-multimedia/camera-service/camera-service_1.0.2.bb:LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
recipes-kernel/linux/qcom-dtb-metadata_0.5.bb:LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc:LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
recipes-support/abl2esp/abl2esp_1.0.bb:LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
recipes-support/userspace-resource-manager-extensions/userspace-resource-manager-extensions_0.3.6.bb:LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.2.bb:LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
```

There is clearly a preferred version of the text.

lumag (changes requested):
> Add location HAL header recipe

See `git log`. This is not the preferred form of the subject

> Add common header only recipe for location HAL headers
under recipes-navigation

And the commit message should not just describe the changes, you can read it from the commit itself. It should be describing, why the changes are performed.

lumag:
Please rebase on the tip to pick up build fixes.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.0.bb
quic-vthuniki: The observed differences are primarily due to line ending variations (LF vs CRLF), with no changes to the actual license text.
Aligned the copyright notice and updated the checksum to match the
version consistently used across existing recipes.

quic-vthuniki:
>This is not the preferred form of the subject

I initially assumed that the PR description would serve as the commit message in GitHub, which was incorrect.
I have now rebased and updated the commit message to follow the expected format and aligned it with existing commits, focusing on capturing the intent and rationale of the change.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.1.bb
```diff
@@ -0,0 +1,23 @@
...
+
+do_install() {
+    install -d ${D}${includedir}
+    install -m 644 ${S}/location/*.h ${D}${includedir}
+    install -m 644 ${S}/pla/oe/*.h ${D}${includedir}
+}
+
+do_configure[noexec] = "1"
```
lumag: It's not the logical order. Configure and compile happen before installation

lumag (changes requested):

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.1.bb
quic-vthuniki: Agreed. Updated ordering to reflect configure, compile, install sequence.

lumag (dismissed):

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-headers_1.0.1.bb
```diff
@@ -0,0 +1,23 @@
...
+
+S = "${UNPACKDIR}/${BP}"
+
+do_configure[noexec] = "1"
+do_compile[noexec] = "1"
+
+do_install() {
+    install -d ${D}${includedir}
+    install -m 644 ${S}/location/*.h ${D}${includedir}
+    install -m 644 ${S}/pla/oe/*.h ${D}${includedir}
```
ricardosalveti: Why are we just installing the headers for this repository? We do have https://github.com/qualcomm-linux/location-hal-qcom/blob/location.lnx.0.0/configure.ac and https://github.com/qualcomm-linux/location-hal-qcom/blob/location.lnx.0.0/Makefile.am, couldn't it just be installed as part of that instead?

quic-vthuniki: Although the repository contains a top-level `configure.ac` and `Makefile.am`, the current build only covers the sources under the `gnss` directory:

https://github.com/qualcomm-linux/location-hal-qcom/blob/location.lnx.0.0/Makefile.am#L6

This PR only introduces the header recipe, so the broader dependency context is not yet visible. I have additional recipes from the same repository that will be submitted separately.

Dependency example on why header only recipe is needed:

- `qcom-gps-utils.bb` [ref link](https://review-android.quicinc.com/plugins/gitiles/meta-qcom-extras/+/refs/heads/metalayers.qclinux.1.0/recipes-location/location/qcom-gps-utils_git.bb), which requires `LocationAPI.h` and related interfaces during compilation.
- `location-api.bb` [ref link](https://review-android.quicinc.com/plugins/gitiles/meta-qcom-extras/+/refs/heads/metalayers.qclinux.1.0/recipes-location/location/location-api_git.bb), which in turn depends on `qcom-gps-utils.bb`.

While `LocationAPI.h` is logically associated with the location api , providing it from the `location-api` recipe would introduce a dependency cycle between `location-api` and `qcom-gps-utils`.

To avoid that dependency cycle and make the common interfaces available to all consumers at build time, the shared headers are packaged in a separate header-only recipe.

Although the sources reside in a single repository, they are built and packaged through multiple recipes, each representing a separate build and dependency unit.

ricardosalveti: Don't we control this repository? The way you are proposing here is quite non standard, and I don't like the idea of having multiple recipes covering the same repository.

Also, this won't really work nicely once we have to create debian packages for this same component (fyi @basak-qcom / @gagath / @lool), so I would prefer us to just fix our upstream repository to just do and install the right thing without having additional logic in this recipe, or even multiple recipes covering the same source tree.

Just make it behave like a normal project.

quic-vthuniki: I see. If the preferred approach is to have a single recipe manage the source tree and produce the appropriate libraries, rather than splitting the repository across multiple recipes, I'll discuss possible approaches with the team and come back with a proposal for discussion.

koenkooi: @ricardosalveti agreed

quic-vthuniki: Hi  Reviewers,
As suggested, i was able to consolidate multiple recipes into single recipe per source repo. I have updated the PR with new recipe. Kindly help review it.

We have two open source repos

- [location-hal-qcom](https://github.com/qualcomm-linux/location-hal-qcom): This provies HAL implementation for our QC location services. This repo will have single recipe location-hal.
In our source repo, now the top level makefile.am includes the subdirs that are required for this recipe and generates the required libraries. 

- [location-apis-qcom](https://github.com/qualcomm-linux/location-apis-qcom): This provides the client facing APIs to  interact with QC location services. This repo will have single recipe location-apis. This recipe has depends on location-hal recipe
I will create a new PR for location-apis recipe as we have some dependency on qmi-framework recipe.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
```diff
@@ -0,0 +1,35 @@
...
+DEPENDS = "glib-2.0 sqlite3 libxml2"
+
+EXTRA_OECONF = " --with-glib"
+CPPFLAGS:append = " -DLOC_QCLINUX_TARGET "
+
+GROUPADD_PACKAGES = "${PN}"
+GROUPADD_PARAM:${PN} = "locclient"
+
+USERADD_PACKAGES = "${PN}"
+USERADD_PARAM:${PN} = "--groups locclient --no-create-home --user-group gps"
```
quic-vthuniki: Our processes runs as gps user group hence we needed to create the user grup here

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
```diff
@@ -0,0 +1,35 @@
...
+
+inherit autotools pkgconfig useradd
+
+DEPENDS = "glib-2.0 sqlite3 libxml2"
+
+EXTRA_OECONF = " --with-glib"
+CPPFLAGS:append = " -DLOC_QCLINUX_TARGET "
+
+GROUPADD_PACKAGES = "${PN}"
+GROUPADD_PARAM:${PN} = "locclient"
```
quic-vthuniki: Location daemon chowns its client sockets to locclient group so clients whose group is locclient will have read write access to socket files

quaresmajose:
Build fails. Please try it in your side
```
kas-container build ci/qcom-armv8a.yml --target location-hal
```

quic-vthuniki:
>Build fails. Please try it in your side
kas-container build ci/qcom-armv8a.yml --target location-hal

Build error seems to be random. I could not reproduce the error but updated the recipe to fix the error 

Error: 
_ERROR: location-hal-1.0.2-r0 do_package_qa: QA Issue: location-hal: /etc/gps.conf is owned by uid 1000, which is the same as the user running bitbake. This may be due to host contamination [host-user-contaminated]_

Fix: 
added --system option to GROUPADD_PARAM and USERADD_PARAM in recipe file.

Verified locally with 
 `kas-container build meta-qcom/ci/qcom-armv8a.yml:meta-qcom/ci/qcom-distro.yml --target location-hal`

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/90295a16-7d47-4988-b7e4-975141dbc5ee)**
---
**Test jobs for commit [`92b15f4`](https://github.com/qualcomm-linux/meta-qcom/commit/92b15f414dbcbf685c2e5304ca697c900554ac16)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 39 | Fail: 8 | Total: 47

quic-vthuniki:
In the latest compilation run all variants passed.
Test execution failed for only one variant [tests:QCM6490_PROP](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/699c4643-cbc4-4d40-99c9-5a43867797dc) for the test [Boot_Systemd_networkddisable_QLI](https://axiom.qualcomm.com/#/planner/testcases/2942738/23). 

The current recipe should not cause any bootup issue as this recipe is not yet enabled in meta-qcom-distro.
Please help suggest how to proceed further on the test result

lumag:
> In the latest compilation run all variants passed. Test execution failed for only one variant [tests:QCM6490_PROP](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/699c4643-cbc4-4d40-99c9-5a43867797dc) for the test [Boot_Systemd_networkddisable_QLI](https://axiom.qualcomm.com/#/planner/testcases/2942738/23).
> 
> The current recipe should not cause any bootup issue as this recipe is not yet enabled in meta-qcom-distro. Please help suggest how to proceed further on the test result

Ignore coral tests, only LAVA tests matter at this point.

ricardosalveti:
> In the latest compilation run all variants passed. Test execution failed for only one variant [tests:QCM6490_PROP](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/699c4643-cbc4-4d40-99c9-5a43867797dc) for the test [Boot_Systemd_networkddisable_QLI](https://axiom.qualcomm.com/#/planner/testcases/2942738/23).
> 
> The current recipe should not cause any bootup issue as this recipe is not yet enabled in meta-qcom-distro. Please help suggest how to proceed further on the test result

Infra / unrelated error, just ignore it for now.

koenkooi:
What's the status on this?

quic-vthuniki:
>What's the status on this?

Dmitry has rerun the LAVA tests twice, but the observed failures appear to be random. I am also waiting for reviewers guidance on the next steps.

quic-vthuniki:
Hi Reviewers,
Could you let me know the next step here? Do you think that these failures are intermitent setup issues or anything need to be checked from my side?

Run#4
failed test:
pre merge iq-8275-evk

Run#3:
failed tests:
pre merge iq-8275-evk
pre merge iq-x7181-evk

Run#2:
failed tests:
pre merge qcs615-ride
boot qcs615-ride

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
```diff
@@ -0,0 +1,35 @@
...
+GROUPADD_PARAM:${PN} = "--system locclient"
+
+USERADD_PACKAGES = "${PN}"
+USERADD_PARAM:${PN} = "--system --groups locclient --no-create-home --user-group gps"
+
+do_install:append() {
+    #Install default gps.conf file
+    install -m 0644 -D ${S}/etc/gps.conf ${D}${sysconfdir}/gps.conf
+    ## Ownership to gps:gps for gps.conf
+    chown gps:gps ${D}${sysconfdir}/gps.conf
```
lumag: Why? Then the user can write to a system config. It sould be `root:root 0644`.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
```diff
@@ -0,0 +1,35 @@
...
+
+inherit autotools pkgconfig useradd
+
+DEPENDS = "glib-2.0 sqlite3 libxml2"
+
+EXTRA_OECONF = " --with-glib"
+CPPFLAGS:append = " -DLOC_QCLINUX_TARGET "
+
+GROUPADD_PACKAGES = "${PN}"
+GROUPADD_PARAM:${PN} = "--system locclient"
```
lumag, line 26: What is it used for? Is it used to limit access to some socket?

lumag, line 26: What is it used for? Is it used to limit access to some socket?

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
```diff
@@ -0,0 +1,35 @@
...
+
+USERADD_PACKAGES = "${PN}"
+USERADD_PARAM:${PN} = "--system --groups locclient --no-create-home --user-group gps"
+
+do_install:append() {
+    #Install default gps.conf file
+    install -m 0644 -D ${S}/etc/gps.conf ${D}${sysconfdir}/gps.conf
+    ## Ownership to gps:gps for gps.conf
```
lumag: Drop obvous comments.

lumag (changes requested):

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
```diff
@@ -0,0 +1,35 @@
...
+location engine, batching, geofencing, and GNSS interface components, \
+along with the runtime accounts and groups required to access location \
+sockets and configuration."
+
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+SRC_URI = "git://github.com/qualcomm-linux/location-hal-qcom.git;protocol=https;branch=location.lnx.0.0;tag=v${PV}"
+SRCREV  = "96b21c90da387747155debb9f482fa11eed9bf74"
+S = "${UNPACKDIR}/${BP}"
```
anujm1: This shouldn't be needed as it's the default value.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
```diff
@@ -0,0 +1,35 @@
...
+DEPENDS = "glib-2.0 sqlite3 libxml2"
+
+EXTRA_OECONF = " --with-glib"
+CPPFLAGS:append = " -DLOC_QCLINUX_TARGET "
+
+GROUPADD_PACKAGES = "${PN}"
+GROUPADD_PARAM:${PN} = "--system locclient"
+
+USERADD_PACKAGES = "${PN}"
+USERADD_PARAM:${PN} = "--system --groups locclient --no-create-home --user-group gps"
```
anujm1: I think we should set `shell` to `/sbin/nologin` as well for system accounts.

anujm1 (changes requested):

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
quic-vthuniki: gps.conf is only read, never written, by the location HAL daemon, which runs as user/group gps (not root). We chown it to gps:gps so the daemon can read it via owner permissions rather than relying on the world-read bit.  We think this matters if SELinux/hardening policy on this baseline ever restricts "other" access to /etc files. We're not aware of an SELinux enforcement in place for this baseline. If this baseline does not enforce SELinux policy on /etc configs ever, we're open to root:root 0644 if you'd prefer.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
quic-vthuniki: sure

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
quic-vthuniki: Yes, that is correct. locclient controls access to the location HAL daemon's client socket (/run/loc_client). The socket directory is created via tmpfiles.d as gps:locclient 0775 (d /run/loc_client 0775 gps locclient - -), so both the daemon and its clients need access through this directory. The daemon runs as gps, and client apps need to be in group locclient, so gps is also added to locclient to guarantee the daemon side has access too.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
quic-vthuniki: You're right that this matches BitBake's default, so it's technically redundant.  Will fix it

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
quic-vthuniki: sure, will fix it

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
lumag: Yes. don't be fancy, the default will work for you.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
lumag: Brief comment in the recipe, so that we know what it is used for.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
lumag: Isn't it already a default for `--system`? I don't remember.

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
quic-vthuniki: Added comment

dynamic-layers/openembedded-layer/recipes-navigation/location-client-api/location-hal_1.0.2.bb
quic-vthuniki: Based on my checking, `--system` will only affect the UID and GID range selection but does not set the default useradd's shell variable.  I have set `--shell` to `/sbin/nologin` in updated PR

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/93b67d72-dd90-4e3d-b4d6-7d62a1ef4a3d)**
---
**Test jobs for commit [`2ffa5df`](https://github.com/qualcomm-linux/meta-qcom/commit/2ffa5dfd11c09fa7237693cc0e3d36b07b888c4e)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 48 | Fail: 0 | Total: 48

lumag (approved):

anujm1 (approved):

koenkooi (approved):

PR #2285, opened 2026-05-25
Update Targets specific to wrynose/2.0

phanendra-qc:
Update targets to build against respective kernel for wrynose/2.0

remove below 
qcs9100-ride-sx
qcs8300-ride-sx

conditoinalize below malinline targets to build against qcom->next kernel
glymur-crd
kaanapali-mtp
sm8750-mtp

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -160,12 +160,9 @@ jobs:
       fail-fast: true
       matrix:
         machine:
-          - glymur-crd
           - qcom-armv8a
           - rb3gen2-core-kit
         distro:
-          - name: nodistro
-            yamlfile: ""
           - name: qcom-distro
             yamlfile: ':ci/qcom-distro.yml'
         kernel:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-yocto.yml
```diff
@@ -160,12 +160,9 @@ jobs:
       fail-fast: true
       matrix:
         machine:
-          - glymur-crd
           - qcom-armv8a
           - rb3gen2-core-kit
         distro:
-          - name: nodistro
```
ndechesne, line 167: why? this is not listed in the commit message.

ndechesne (commented):
there is no commit message, only a comment on the PR. and this message does not explain why you are proposing this change. This message mentions 'mainline targets' but is against wrynose branch, i don't understand what you mean.

ricardosalveti:
Why should we disable CI for these targets in wrynose? They are part of the wrynose branch, so they should be supported there.

PR #2286, opened 2026-05-25
initramfs: Move GPUCC and UFS drivers to initramfs

Kavinaya99:
Move UFS drivers from static kernel build to initramfs modules to improve boot initialization timing


recipes-kernel/linux/linux-qcom-6.18/configs/bsp-additions.cfg
```diff
@@ -1,12 +1,17 @@
 # for UFS enumeration
 CONFIG_PHY_QCOM_QMP=y
-CONFIG_SCSI_UFS_QCOM=y
+CONFIG_SCSI_UFS_QCOM=m
+CONFIG_SCSI_UFSHCD=m
+CONFIG_SCSI_UFSHCD_PLATFORM=m
```
lumag: Every change related to a different recipe should go to a separate commit. Be sure to order them correctly.

recipes-kernel/images/initramfs-rootfs-image.bb
```diff
@@ -5,6 +5,11 @@ PACKAGE_INSTALL = " \
     initramfs-module-copy-modules \
     initramfs-module-rootfs \
     initramfs-module-udev \
+    kernel-module-governor-simpleondemand \
+    kernel-module-ufshcd-core \
+    kernel-module-ufshcd-pltfrm \
```
lumag: Aren't those pulled in by module dependencies?

lumag (changes requested):
> probe ordering issues and race conditions with other subsystems on some platforms.

Which probe issues? Which race conditions? Which platforms are affected and how? Please be exact.

Also drop the template or AI prompt, describing the changes. It's pretty obvious from the commit itself. Focus on something which is not obvious - reasons, errors, affected devices.

recipes-kernel/images/initramfs-rootfs-image.bb
```diff
@@ -5,6 +5,11 @@ PACKAGE_INSTALL = " \
...
+    kernel-module-ufs-qcom \
+    kernel-module-sdhci-msm \
```
quaresmajose: All of the kernel modules should go into the variable MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS, which is more appropriate for the purpose. In its current form, we will get an error if any of the kernel modules are built-in.

recipes-kernel/linux/linux-qcom-6.18/configs/bsp-additions.cfg
Kavinaya99: Sure, updated in the new pacth

recipes-kernel/images/initramfs-rootfs-image.bb
Kavinaya99: Without adding these modules to the recipe, we are getting "unable to mount root fs" errors.

Kavinaya99:
> > probe ordering issues and race conditions with other subsystems on some platforms.
> 
> Which probe issues? Which race conditions? Which platforms are affected and how? Please be exact.
> 
> Also drop the template or AI prompt, describing the changes. It's pretty obvious from the commit itself. Focus on something which is not obvious - reasons, errors, affected devices.

Updated the commit message

recipes-kernel/images/initramfs-rootfs-image.bb
Kavinaya99: Moving kernel modules to MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS ensures they are included in the image when built as modules, but it does not guarantee that they will be loaded during boot.
I have tried using MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS and saw bootup failures as modules were not loaded.
The error: root '/dev/disk/by-partlabel/rootfs' doesn't exist or does not contain a /dev.

lumag (changes requested):
> On QCS8300 (Monaco), the kernel experiences a race condition
during boot where UFS storage driver, ARM-SMMU, and GPUCC (GPU
Clock Controller) all initialize at the same device_initcall
level (level 6). This creates a dependency chain where UFS
requires SMMU, SMMU requires GPUCC clocks, but GPUCC may not
finish initialization before SMMU times out waiting for it.

Hmm, no. The GPU SMMU and the UFS SMMU are two different SMMU instances. So, the fact that `3da0000.iommu` has not probed should not affect probing of 15000000.iommu and the UFS.

Please continue and find the actual root cause. What exactly is causing UFS to not to probe?

> The race condition manifests as:
gcc-qcs8300 100000.clock-controller: sync_state() pending
due to 3d90000.clock-controller
arm-smmu 3da0000.iommu: deferred probe timeout, ignoring
dependency
arm-smmu 3da0000.iommu: probe with driver arm-smmu failed with
error -110
Kernel panic in iommu_domain_free() at gmu_core_iommu_init()

recipes-kernel/images/initramfs-rootfs-image.bb
lumag: > Moving kernel modules to MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS ensures they are included in the image when built as modules

Well. No. Packages listed in `MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS` don't get included into the `initramfs-rootfs-image`. The recipe overrides `PACKAGE_INSTALL` (purposedly), so `packagegroup-core-boot` doesn't get included into the image. That's why you've observed errors.

At the same time, no, you can't list modules here. The image should be generic. Also, it should not fail to build if one changes the kernel config. If you need modules, you need to have a packagegroup which would recommend necessary packages. I'd have said that you should resurrect `packagegroup-qcom-boot` and `initramfs-qcom-image`, see commit 05b73a175cfa ("initramfs-qcom-image: remove the recipe and packagegroup").

Also, this commit should be the first one, otherwise booting of the image would be broken between commits moving the drivers to the modules and this one (and thus breaking `git bisect`, which is a bad idea).

shashim-quic:
> > On QCS8300 (Monaco), the kernel experiences a race condition
> > during boot where UFS storage driver, ARM-SMMU, and GPUCC (GPU
> > Clock Controller) all initialize at the same device_initcall
> > level (level 6). This creates a dependency chain where UFS
> > requires SMMU, SMMU requires GPUCC clocks, but GPUCC may not
> > finish initialization before SMMU times out waiting for it.
> 
> Hmm, no. The GPU SMMU and the UFS SMMU are two different SMMU instances. So, the fact that `3da0000.iommu` has not probed should not affect probing of 15000000.iommu and the UFS.
> 
> Please continue and find the actual root cause. What exactly is causing UFS to not to probe?
> 

The commit msg need to be re-written. The issue is not GPU SMMU blocking UFS , but actually few configs which are part of static image however their dependencies are configured as modules. This causes repeated probe deferrals sometime delaying ufs bring up and in some other cases exceeding probe deferral timeout that blocks other re-probes.

Few cases from last debug that were identified are below:
- SMMU configured as 'y' while GPUCC as 'm' leading to repeated probe deferral of adreno smmu
- USB controller configured as 'y' while usb phy as 'm'

I prefer such modules (which are dependencies for drivers that are part of static kernel image) be moved to initramfs so boot delays (and deferrable timeout) can be better managed.

> > The race condition manifests as:
> > gcc-qcs8300 100000.clock-controller: sync_state() pending
> > due to 3d90000.clock-controller
> > arm-smmu 3da0000.iommu: deferred probe timeout, ignoring
> > dependency
> > arm-smmu 3da0000.iommu: probe with driver arm-smmu failed with
> > error -110
> > Kernel panic in iommu_domain_free() at gmu_core_iommu_init()

lumag:
> SMMU configured as 'y' while GPUCC as 'm' leading to repeated probe deferral of adreno smmu

Is this verified? Doesn't fw_devlink take care of this? There should be no probe_deferral's (as in probe() callback returning -EPROBE_DEFER).

> USB controller configured as 'y' while usb phy as 'm'

Ideally fw_devlink should take care of it and prevent extra probes and extra deferrals.

Anyway, I don't see changes related to either of your points in this PR.

Kavinaya99:
> > SMMU configured as 'y' while GPUCC as 'm' leading to repeated probe deferral of adreno smmu
> 
> Is this verified? Doesn't fw_devlink take care of this? There should be no probe_deferral's (as in probe() callback returning -EPROBE_DEFER).
> 
> > USB controller configured as 'y' while usb phy as 'm'
> 
> Ideally fw_devlink should take care of it and prevent extra probes and extra deferrals.
> 
> Anyway, I don't see changes related to either of your points in this PR.

Updated the contents and commit message accordingly.

recipes-kernel/images/initramfs-rootfs-image.bb
Kavinaya99: Updated the new patch which has resurrected packagegroup-qcom-boot and initramfs-qcom-image

lumag:
> > > SMMU configured as 'y' while GPUCC as 'm' leading to repeated probe deferral of adreno smmu
> > 
> > 
> > Is this verified? Doesn't fw_devlink take care of this? There should be no probe_deferral's (as in probe() callback returning -EPROBE_DEFER).
> > > USB controller configured as 'y' while usb phy as 'm'
> > 
> > 
> > Ideally fw_devlink should take care of it and prevent extra probes and extra deferrals.
> > Anyway, I don't see changes related to either of your points in this PR.
> 
> Updated the contents and commit message accordingly.

Please respond to the questions.

recipes-kernel/packagegroups/packagegroup-qcom-boot.bb
```diff
@@ -0,0 +1,10 @@
+SUMMARY = "Qualcomm boot requirements"
+DESCRIPTION = "Storage and boot-critical kernel modules required for Qualcomm platforms"
+
+inherit packagegroup
+
+RRECOMMENDS:${PN} += " \
+    kernel-module-gpucc-sa8775p \
```
lumag, line 7: This is a platform-specific module. Why are we including only this gpucc driver?

lumag (changes requested):
> 1.CONFIG_ARM_SMMU_QCOM=y (for GPU) depends on
  CONFIG_SA_GPUCC_8775P=m
2.CONFIG_USB_DWC3_QCOM=y depends on
  CONFIG_PHY_QCOM_USB_SNPS_FEMTO_V2=m

> Due to this configuration mismatch, built-in drivers
attempt to probe before their module dependencies are
loaded from rootfs, causing repeated probe deferrals.
Randomly, the kernel deferred probe timeout exceeds
leading to probe failures.

I stopped reviewing after this. There are two SMMU instances. Only one of them needs the GPUCC. The system SMMU entry is probing and working fine. Also this applies to all platforms. I don't see what is so specific to Lemans / Monaco.

Next, these are not module dependencies, they are driver dependencies which should be cared by fw_devlink. As such, there should be no probe deferrals, so the commit message is also inacccurate at best.

If SMMU or USB errors out because of the device being not probed, please fix the driver instead.

Kavinaya99:
> > 1.CONFIG_ARM_SMMU_QCOM=y (for GPU) depends on
> > CONFIG_SA_GPUCC_8775P=m
> > 2.CONFIG_USB_DWC3_QCOM=y depends on
> > CONFIG_PHY_QCOM_USB_SNPS_FEMTO_V2=m
> 
> > Due to this configuration mismatch, built-in drivers
> > attempt to probe before their module dependencies are
> > loaded from rootfs, causing repeated probe deferrals.
> > Randomly, the kernel deferred probe timeout exceeds
> > leading to probe failures.
> 
> I stopped reviewing after this. There are two SMMU instances. Only one of them needs the GPUCC. The system SMMU entry is probing and working fine. Also this applies to all platforms. I don't see what is so specific to Lemans / Monaco.

As highlighted: CONFIG_ARM_SMMU_QCOM=y (for GPU), the issue is observed for GPU SMMU only.
You are right. This applies to all the platforms, the debugging was done only for lemans and monaco.
> 
> Next, these are not module dependencies, they are driver dependencies which should be cared by fw_devlink. As such, there should be no probe deferrals, so the commit message is also inacccurate at best.
> 
I will get back with logs related fw_devlink.

apateriy-qcom:
> > > > SMMU configured as 'y' while GPUCC as 'm' leading to repeated probe deferral of adreno smmu
> > > 
> > > 
> > > Is this verified? Doesn't fw_devlink take care of this? There should be no probe_deferral's (as in probe() callback returning -EPROBE_DEFER).
> > > > USB controller configured as 'y' while usb phy as 'm'
> > > 
> > > 
> > > Ideally fw_devlink should take care of it and prevent extra probes and extra deferrals.
> > > Anyway, I don't see changes related to either of your points in this PR.
> > 
> > 
> > Updated the contents and commit message accordingly.
> 
> Please respond to the questions.


Hi @lumag ,


You are right that actual driver->probe is not called. However Linux device framework keeps checking device readiness (and suppliers dependency) before calling probe and that is wasting cpu cycles.
 
The logs(added above) clearly establish that packaging GPUCC in ramdisk shifts GPU SMMU probe success by 3+ secs. 


**Code Flow:**

Before attempting driver probe, it check if supplier is available, if not then it return -EDEFER
(if device_links_check_suppliers return EDEFER),and puts the device on deferred_probe_pending list.
 
Linux framework will keep checking on the deferred_probe_pending_list and try to probe once suppliers are available. 
After successful probe it is moved to active list. 
This will continue until supplier is available or deferred probe timeout happens.

Now for ardeno smmu waiting gpucc, without moving gpucc to ramdisk, we are deferred till 19 seconds 
whereas with movement to ramdisk, it comes up in 15 seconds.
 
**Logs :**

**Without moving to ramdisk:**
```  
       Line 2221: [   15.222719]   active:  3da0000.iommu: platform: supplier 3d90000.clock-controller not ready
  Line 2236: [   15.339958]   pending: 3da0000.iommu: platform: supplier 3d90000.clock-controller not ready
  Line 2249: [   15.397583]   pending: 3da0000.iommu: platform: supplier 3d90000.clock-controller not ready
  Line 2260: [   15.524319]   active:  3da0000.iommu: platform: supplier 3d90000.clock-controller not ready
  Line 2447: [   19.004209]   pending: 3da0000.iommu: platform: supplier 3d90000.clock-controller not ready
  Line 2459: [   19.096994]   active:  3da0000.iommu: platform: supplier 3d90000.clock-controller not ready
  Line 2470: [   19.197913]   active:  3da0000.iommu: platform: supplier 3d90000.clock-controller not ready
  Line 2481: [   19.290581]   active:  3da0000.iommu: platform: supplier 3d90000.clock-controller not ready
  Line 2492: [   19.383305]   active:  3da0000.iommu: platform: supplier 3d90000.clock-controller not ready 
```


**after moving to ramdisk:**


**ramdisk loads here**

```
  Line 3226: [   14.246883] Freeing unused kernel memory: 2112K


  Line 3283: [   15.182426] platform 3da0000.iommu:   fwnode suppliers: none unresolved (check device_links)
  Line 3289: [   15.228371]   active:  3da0000.iommu: platform: supplier 3d90000.clock-controller not ready
  Line 3290: [   15.236975] platform 3da0000.iommu:   fwnode suppliers: none unresolved (check device_links)
  Line 3301: [   15.324499]   active:  3da0000.iommu: platform: supplier 3d90000.clock-controller not ready
  Line 3302: [   15.333100] platform 3da0000.iommu:   fwnode suppliers: none unresolved (check device_links)
  Line 3354: [   15.707398] arm-smmu 3da0000.iommu: probing hardware configuration...
  Line 3374: [   15.714034] arm-smmu 3da0000.iommu: SMMUv2 with:
  Line 3375: [   15.714061] arm-smmu 3da0000.iommu:  stage 1 translation`

```

lumag:
> The logs(added above) clearly establish that packaging GPUCC in ramdisk shifts GPU SMMU probe success by 3+ secs.

This is not our goal. Does it affect booting time?

Kavinaya99:
> > The logs(added above) clearly establish that packaging GPUCC in ramdisk shifts GPU SMMU probe success by 3+ secs.
> 
> This is not our goal. Does it affect booting time?

Before moving the modules to ramdisk, boot time is:  Startup finished in 3.194s (firmware) + 181ms (loader) + 7.725s (kernel) + 16.323s (userspace) = 27.424s multi-user.target reached after 14.725s in userspace.

After moving the modules to ramdisk, boot time is: Startup finished in 3.234s (firmware) + 172ms (loader) + 16.610s (kernel) + 1min 30.985s (userspace) = 1min 51.003s multi-user.target reached after 1min 30.861s in userspace.

lumag:
So, do I read it correctly that the proposed change makes the boot time worse?

koenkooi (changes requested):
No, once boot critical drivers are builtin, we shouldn't change them to modules.

shashim-quic:
> No, once boot critical drivers are builtin, we shouldn't change them to modules.

our reference is upstream arm64 defconfig (CONFIG_SCSI_UFS_QCOM=m as an example ).

koenkooi:
> > No, once boot critical drivers are builtin, we shouldn't change them to modules.
> 
> our reference is upstream arm64 defconfig (CONFIG_SCSI_UFS_QCOM=m as an example ).

That's nice, but in this repo it *is* a builtin, modifying that is a huge change that should not be taken lightly.

shashim-quic:
> > The logs(added above) clearly establish that packaging GPUCC in ramdisk shifts GPU SMMU probe success by 3+ secs.
> 
> This is not our goal. Does it affect booting time?

Pasting one of the boot time sequence for Kodiak logs that failed to probe gpu because of late arrival of dependency

```
# Bootchart: adreno / 3d00000  (RB3gen2  QCS6490/SC7280)

  T(s)  COMPONENT                      EVENT / DEPENDENCY
  ─────────────────────────────────────────────────────────────────────────

   0.0  [power-on]                     SBL1 → UEFI → Linux (QCS6490 SC7280)

   0.66 arm-smmu / 15000000            ──► OK  (system SMMU ready)

  ≈1.5  rpmhpd / 18200000.rsc          ──► OK  (RPMh power domain controller)
        │
  ≈2.0  gcc-sc7280 / 100000            ──► OK  ← GPLL0 / CFG AHB clocks ready
             │
             ├─ GCC_GPU_CFG_AHB_CLK ────────────────────────────────────┐
             ├─ GCC_GPU_GPLL0_CLK_SRC ────────────────────────────────── ┤  gpucc deps
             └─ GCC_GPU_GPLL0_DIV_CLK_SRC ─────────────────────────────  ┤
                                                                          │
  ≈2.1  arm-smmu / 3da0000 (GPU SMMU)  ──► -517  (gpucc not registered)  │
             needs: GPU_CC_HLOS1_VOTE_GPU_SMMU_CLK  ◄─────────────────── ┘
        │
        …  (<snip for ~4 s>)
        │
   6.58 [storage] UFS/SCSI             sda..sdf enumerated  (6 LUNs)
   6.84 [storage] sda1/sda2            partitions visible

   9.01 [init]                         Run /init as init process

        …  (<snip for ~5 s>)

  14.59 [rootfs]  sda2                 EXT4 mounted r/w
  15.88 [systemd] PID 1                systemd 259.5 running

        …  (unit-file processing + service startup,  ~4 s)

  19.46 [coldplug]                     udev trigger fires ────────────► gpucc (MODULE) load
        │                                                                │
  21.99 ⚠ arm-smmu / 3da0000 (GPU SMMU)──► -110 TIMEOUT ✗              │
             (GPU_CC_HLOS1_VOTE_GPU_SMMU_CLK still unavailable —        │
              gpucc_sc7280 module not yet initialised)                  │
        │                                                                │
  22.21 gcc-sc7280 / 100000            sync_state pending:              │
                                        3d90000 (gpucc) not yet probed  │
  22.22 gcc-sc7280 / 100000            sync_state pending:              │
                                        3da0000 (GPU SMMU) not yet probed
        │                                                                │
 ≈24.0  gpucc_sc7280 / 3d90000 (MODULE)◄────────────────────────────── ┘
             ──► OK                    GPU_CC_HLOS1_VOTE_GPU_SMMU_CLK
             │                        now available  (≈2.0 s after GPU SMMU gave up)

 ≈25.99 [coldplug]                     Finished Coldplug All udev Devices
  26.62 remoteproc modem/adsp/cdsp     ──► OK

        …  (<snip for ~4 s>)

  31.14 ⚠ adreno / 3d00000            ──► -ENODEV TIMEOUT ✗
             (SMMU domain unavailable — GPU SMMU at 3da0000 already timed out)
```

### Key Observations
- Deferred probe timeout : arm-smmu (3da0000.iommu)    at T=21.99 s
- Late dependency      : gpucc_sc7280 (3d90000)      arrived at T≈24 s
- Root bottleneck        : gpucc_sc7280 is CONFIG_QCOM_GPUCC_SC7280=m (module);
                           loads only after udev coldplug fires at T=19.46 s
- Summary:
The module only loads after udev coldplug at T=19.46 s, but the GPU SMMU's deferred-probe timeout fires at T=21.99 s — just ~2 s into the coldplug window, before modprobe completes.

shashim-quic:
> > > The logs(added above) clearly establish that packaging GPUCC in ramdisk shifts GPU SMMU probe success by 3+ secs.
> > 
> > 
> > This is not our goal. Does it affect booting time?
> 
> Pasting one of the boot time sequence for Kodiak logs that failed to probe gpu because of late arrival of dependency

Considering roughly ~2 sec window where the probe timed out , for now I am thinking to go with setting a higher _deferred_probe_timeout_ (say 60sec) for release.

Let me know if you see any issue with that. Will raise a separate PR to confine it to address probe timeout case.

lumag:
> Considering roughly ~2 sec window where the probe timed out , for now I am thinking to go with setting a higher _deferred_probe_timeout_ (say 60sec) for release.

SGTM. I think Debian images use 30 sec now.

lumag:
@shashim-quic but please note. Ideally the boards should be able to boot without any extra arguments. If we need to increase the timeout, it means we have a more generic problem. Maybe we need to rework the kernel code instead, to change the event triggering the timeout to start counting. Udev start? Last driver probe? Last module load?
I think starting the timeout timer from the module load might be the best option.

shashim-quic:
> @shashim-quic but please note. Ideally the boards should be able to boot without any extra arguments. If we need to increase the timeout, it means we have a more generic problem. Maybe we need to rework the kernel code instead, to change the event triggering the timeout to start counting. Udev start? Last driver probe? Last module load? I think starting the timeout timer from the module load might be the best option.

Agree, it needs a better solution. For now backporting Hans patch [1] to configure timeout and set it to 60sec from kernel config. It is part of upstream now.

1. https://lore.kernel.org/all/20260314084916.10868-1-johannes.goede@oss.qualcomm.com/#r

lumag:
@shashim-quic sounds good to me, thanks!

lumag:
I assume, this one can be closed now. Please reopen if I'm wrong

PR #2251, opened 2026-05-20
Enable KVM by default on Lemans and Monaco Platforms

vkraleti:
All PILs are functioning correctly and KVM can be enabled by default on Lemans and Monaco platforms. 
Update FIT_DTB_COMPATIBLE entries and machine configurations to switch to default KVM for these boards.

PR #2289, opened 2026-05-25
Define QCOM_XBL_CONFIG in qcom-common.inc for XBL config selection

vkraleti:
Avoid hardcoding XBL configuration selection in image_types_qcom.bbclass, which limits flexibility to enable KVM selectively across boards.  Define QCOM_XBL_CONFIG to select the appropriate XBL config based on COMBINED_FEATURES. This enables per-board control of XBL configuration when KVM is enabled.

### DIFF
--- a/classes-recipe/image_types_qcom.bbclass
+++ b/classes-recipe/image_types_qcom.bbclass
@@ -112,18 +112,13 @@ create_qcomflash_pkg() {
         done
 
         # xbl_config
-        xbl_config="xbl_config.elf"
-        if ${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}; then
-            xbl_config="xbl_config_kvm.elf"
-        fi
-
         # Prefer the OEM-cert-injected xbl_config deployed by the capsule recipe
         # when available.
         if [ -n "${QCOM_CAPSULE_FIRMWARE}" ] && \
                 [ -f "${DEPLOY_DIR_IMAGE}/xbl_config-with-oem-cert.elf" ]; then
             install -m 0644 "${DEPLOY_DIR_IMAGE}/xbl_config-with-oem-cert.elf" xbl_config.elf
-        elif [ -f "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${xbl_config}" ]; then
-            install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${xbl_config}" xbl_config.elf
+        elif [ -f "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${QCOM_XBL_CONFIG}" ]; then
+            install -m 0644 "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}/${QCOM_XBL_CONFIG}" xbl_config.elf
         fi
 
         # bootloader selection
--- a/conf/machine/include/qcom-common.inc
+++ b/conf/machine/include/qcom-common.inc
@@ -36,6 +36,9 @@ PREFERRED_PROVIDER_android-tools-conf = "android-tools-conf-configfs"
 IMAGE_FSTYPES ?= "ext4.gz"
 IMAGE_ROOTFS_ALIGNMENT ?= "4096"
 
+# XBL Config selection
+QCOM_XBL_CONFIG ?= "${@bb.utils.contains("COMBINED_FEATURES", "kvm", "xbl_config_kvm.elf", "xbl_config.elf", d)}"
+
 # Android boot image settings
 QCOM_BOOTIMG_KERNEL_BASE ?= "0x80000000"
 QCOM_BOOTIMG_PAGE_SIZE ?= "4096"


ci/kvm.yml
```diff
@@ -5,4 +5,6 @@ header:
   includes:
    - ci/qcom-distro.yml
 
-distro: qcom-distro-kvm
+local_conf_header:
+  kvm: |
+      MACHINE_FEATURES:append = " kvm"
```
lumag: No. Just drop qcom-distro-kvm completely, enabling kvm in distro features of qcom-distro.

lumag (changes requested):

ci/kvm.yml
vkraleti: That breaks CI. Removal has to happen in 2 steps.

lumag: Then make qcom-distro-kvm noop for now.

lumag: DISTRO should not be touching MACHINE_FEATURES at all.

vkraleti: Renamed the KAS fragment as `ci/kvm.yml` to avoid this ambiguity.

lumag (changes requested):
> Replace ci/qcom-distro-kvm.yml with a new ci/kvm.yml KAS

Why? What is the issue that you are trying to solve? I was under assumption that the machine configs are going to have the feature enabled. Now you are adding a config fragment to enable it.

vkraleti:
> > Replace ci/qcom-distro-kvm.yml with a new ci/kvm.yml KAS
> 
> Why? What is the issue that you are trying to solve? I was under assumption that the machine configs are going to have the feature enabled. Now you are adding a config fragment to enable it.

Not all machine are ready to switch to KVM. Only Lemans and Monaco are tested and ready to switch. For Kodiak, Talos, Hamoa, Purwa meta-qcom still need to provide a parallel build for KVM verifications while continuing regular testing with Gunyah.

lumag:
> > > Replace ci/qcom-distro-kvm.yml with a new ci/kvm.yml KAS
> > 
> > 
> > Why? What is the issue that you are trying to solve? I was under assumption that the machine configs are going to have the feature enabled. Now you are adding a config fragment to enable it.
> 
> Not all machine are ready to switch to KVM. Only Lemans and Monaco are tested and ready to switch. For Kodiak, Talos, Hamoa, Purwa meta-qcom still need to provide a parallel build for KVM verifications while continuing regular testing with Gunyah.

Commit message.

vkraleti:
Updated commit messages. 

<https://github.com/qualcomm-linux/meta-qcom/pull/2251> is the follow‑up PR to enable KVM by default for Lemans. Changes for Monaco, Kodiak, and Talos will be submitted shortly in separate PRs.

ci/kvm.yml
```diff
@@ -2,7 +2,7 @@
 
 header:
   version: 14
-  includes:
-   - ci/qcom-distro.yml
 
-distro: qcom-distro-kvm
+local_conf_header:
+  kvm: |
+      MACHINE_FEATURES:append = " kvm"
```
koenkooi: Having non-machines poking at MACHINE_FEATURES is a very bad idea

koenkooi (changes requested):

ci/kvm.yml
vkraleti: Yeah. But is there a better way to have two builds with and without a feature?  We must have two builds with and without KVM till all SoCs are ready to migrate.

vkraleti: @koenkooi @lumag @ricardosalveti are you aligned to this approach? If not, is it fine to restore https://github.com/qualcomm-linux/meta-qcom-distro/commit/32e749249929b5dd8940f27d35945eaed4c91614 `

```
# Enable KVM distro feature if corresponding machine feature is set.
# This allows users to enable KVM either via MACHINE_FEATURES or DISTRO_FEATURES.
DISTRO_FEATURES:append = "${@bb.utils.contains('MACHINE_FEATURES', 'kvm', ' kvm', '', d)}"
```

Test teams are expecting KVM builds in CI for boards that are fully not ready.

lumag: No, setting DISTRO_FEATURES based on MACHINE_FEATURES is a clear NAK.

vkraleti: Ok, how about adding a check in [image_types_qcom.bbclass](https://github.com/qualcomm-linux/meta-qcom/blob/master/classes-recipe/image_types_qcom.bbclass#L114-L118) to check for `kvm` either in machine or distro features?

```
if ${@bb.utils.contains('MACHINE_FEATURES', 'kvm', 'true', 'false', d)} || \
   ${@bb.utils.contains('DISTRO_FEATURES', 'kvm', 'true', 'false', d)}; then
    xbl_config="xbl_config_kvm.elf"
fi
```

lumag: What for? What does it logically mean? Should it be instead `if ${bb.utils.contains('COMBINED_FEATURES', 'gunyah', 'true', 'false'} ; then ... `

vkraleti: After exploring additional options, seems this transition needs to happen in incremental steps as listed below:

1. Move XBL config selection logic from image_types_qcom.bbclass to the respective machine .conf files.
 This PR #2289 is updated for this change.
 
2. Introduce a Gunyah-specific distro to clearly model the alternate hypervisor use case.
Updated https://github.com/qualcomm-linux/meta-qcom-distro/pull/306

3. For KVM-ready machines, update XBL config selection based on the Gunyah distro. Updated PR #2251

4. Once all machines are migrated, drop the KVM distro feature and standardize on the Gunyah distro for alternate hypervisor builds.

ricardosalveti:
I find this all a bit confusing, and we are not really improving the current state of things here.

The configuration nobs we have in the end:
- Machine specific decision to support either KVM or gunyah (based on the DTBs available and also based on the XBL config used by the machine, both are BSP specific decisions)
- Distro specific decision that enables KVM support (as found in meta-virtualization)

Now for gunyah, do we need any distro specific policy to support it to the same way we support KVM? If indeed the case, then we should indeed have multiple distros and pick one by default (KVM). Here COMBINED_FEATURES would work better when deciding the XBL to flash (and done in a common file, not in every machine conf), but for this to work we would have to define a KVM/Gynuah machine feature.

vkraleti:
Yes, Gunyah needs to be another distro feature,  otherwise managing builds becomes really hard. https://github.com/qualcomm-linux/meta-qcom-distro/pull/306 is defining this. 

Once this distro feature is available, machine configurations can start making decisions like
https://github.com/qualcomm-linux/meta-qcom/pull/2251/changes/6da7c1000f205ff6bae7393b899047e079271296#diff-d0fec212e122859b520a519af2bcefb7be12afc7148c6df0f991115c362ea85cR38 
`QCOM_XBL_CONFIG ?= "${@bb.utils.contains("DISTRO_FEATURES", "gunyah", "xbl_config_gunyah.elf", "xbl_config_kvm.elf", d)}"`

To convert KVM to a combined feature, first machine configs need to have the control on XBL config installations. Current PR (#2289) provides this control. Once this and gunyah distro feature are available,  `kvm' can be added to DISTRO_FEATURES by default to facilitate combined feature checks. 

Post that machine confs can define 'kvm' as a machine feature and convert the XBL config installation checks to use 'kvm' COMBINED_FEATURE.

All this juggling is to avoid CI breakages and to continue providing a secondary build for alternate hypervisor testing.

vkraleti:
If you agree for a brief CI downtime, I can make changes to add `kvm` unconditionally to DISTRO_FEATURES and defining new qcom-distro-gunyah conf in one meta-qcom-distro PR,  equivalent machine conf changes in one meta-qcom PR. meta-qcom-distro PR needs to merge by skipping CI. meta-qcom PR can validate e2e in CI.

lumag:
@vkraleti I'd rather not. If something fails, we end up with broken CI or broken config.

vkraleti:
Ok. PRs should be merged in following order to make the switch. 

1. https://github.com/qualcomm-linux/meta-qcom-distro/pull/306 (Defines qcom-distro-gunyah)
2. https://github.com/qualcomm-linux/meta-qcom/pull/2289 (Move XBL config selection to machine.conf & stop using qcom-distro-kvm)
3. https://github.com/qualcomm-linux/meta-qcom-distro/pull/314 (Drop qcom-distro-kvm and make `kvm` default part of  DISTRO_FEATURES)
4. https://github.com/qualcomm-linux/meta-qcom/pull/2251 (Switch Leamans & Monaco to default KVM and introduce Gunyah builds for 2nd Hypervisor testing)
5. Similar changes to above for Kodaik, Talos and any other machine that needs to switch.
6. Drop all KVM references in CI KAS files.

vkraleti:
@ricardosalveti @lumag if there are no concerns with this approach, can you review this PR? This don't have any functional impact, as it is just reordering things.

ricardosalveti:
> Yes, Gunyah needs to be another distro feature, otherwise managing builds becomes really hard. [qualcomm-linux/meta-qcom-distro#306](https://github.com/qualcomm-linux/meta-qcom-distro/pull/306) is defining this.

I don't see the need to have a gunyah distro features if all that is needed is the selection of the xbl binary to use, which could be done as part of the if / else check based on KVM. Distro features for kvm has a direct impact in meta-virtualization, but I don't see that as a direct conflict for gunyah.

ci/qcom-distro-kvm.yml
```diff
@@ -5,4 +5,7 @@ header:
   includes:
    - ci/qcom-distro.yml
 
-distro: qcom-distro-kvm
+local_conf_header:
+  kvm: |
+        MACHINE_FEATURES:append = " kvm"
+        DISTRO_FEATURES:append = " kvm"
```
ricardosalveti: We can set kvm in distro features by default and change image_types_qcom.bbclass to use combined for kvm, and when not available use gunyah by default, then you add the kvm machine feature to the machine you want to switch to kvm by default.

conf/machine/glymur-crd.conf
```diff
@@ -23,3 +23,5 @@ QCOM_PARTITION_FILES_SUBDIR ?= "partitions/glymur-crd/nvme"
 QCOM_PARTITION_FILES_SUBDIR_SPINOR ?= "partitions/glymur-crd/spinor"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-glymur"
+
+QCOM_XBL_CONFIG ?= "${@bb.utils.contains("COMBINED_FEATURES", "kvm", "xbl_config_kvm.elf", "xbl_config.elf", d)}"
```
ricardosalveti: This is duplicating the same check in every machine, you can move to common or just let the check in classes-recipe/image_types_qcom.bbclass.

ci/qcom-distro-kvm.yml
vkraleti: For machines that have KVM by default, we also need to provide an alternate build in CI with Gunyah, just like how we are providing a KVM build as of today. If gunyah distro feature is not defined, KAS configuration needs to have `DISTRO_FEATURES:remove = "kvm"` which is objected https://github.com/qualcomm-linux/meta-qcom/pull/2289#discussion_r3302166910.

vkraleti: Current PR is still modifying MACHINE_FEATURES & DISTRO_FEATURES in KAS config but this is a temporary change to facilitate switching. IT'll will be reverted as soon as meta-qcom-distro supports KVM by default.

conf/machine/glymur-crd.conf
vkraleti: Done.

vkraleti:
@ricardosalveti @lumag Thank you for the offline discussion. As suggested, moved all changes to #2251. This PR can be dropped for now.

PR #2300, opened 2026-05-27
qcom-capsule: enable capsule generation for hamoa (iq-x7181-evk)

xueqnie:
Enable UEFI FMP capsule generation for the Hamoa platform (iq-x7181-evk).
  Hamoa uses SPINOR flash and only updates the DTB partition via capsule.

  This extends the qcom-capsule class with a CAPSULE_ENTRIES mechanism for
  dynamic FvUpdate.xml generation and guards xbl_config.elf cert injection
  for platforms that lack this component.

 Built and verified capsule generation produces a valid .cap file targeting
 the SPINOR dtb/dtb_BACKUP main/backup partitions.

### DIFF
--- a/classes-recipe/qcom-capsule.bbclass
+++ b/classes-recipe/qcom-capsule.bbclass
@@ -200,6 +200,57 @@ python generate_fvupdate() {
 
 do_compile[prefuncs] += "generate_fvupdate"
 
+# Inject the OEM root certificate into xbl_config.elf.
+# Dumps the config sections, auto-detects the post-DDR DTB (or uses
+# XBLCONFIG_DTB / XBLCONFIG_DTB_SECTION overrides), patches QcCapsuleRootCert
+# in that DTB, and repacks the updated DTB back into xbl_config.elf in place.
+# $1 - path to xbl_config.elf (modified in place on success)
+patch_xblconfig_cert() {
+    local xbl_config="$1"
+    local staged_dir
+    staged_dir=$(dirname "${xbl_config}")
+
+    XBL_DUMP_LOG="${CAPSULE_DIR}/xbl_dump.log"
+    qcom-capsule-tool parse-config \
+        "${xbl_config}" dump \
+        --out-dir "${staged_dir}" | tee "${XBL_DUMP_LOG}"
+
+    DTB_PATCH="${XBLCONFIG_DTB}"
+    DTB_SECTION="${XBLCONFIG_DTB_SECTION}"
+    if [ -z "${DTB_PATCH}" ]; then
+        # Parse a line like:
+        #   [+] config_item[6] -> PH# 8 -> './post-ddr-kodiak-1.0.dtb' (90280 bytes)
+        POST_DDR_LINE=$(grep -m1 "post-ddr.*\.dtb" "${XBL_DUMP_LOG}" || true)
+        if [ -n "${POST_DDR_LINE}" ]; then
+            DTB_PATCH=$(echo "${POST_DDR_LINE}" | sed "s|.* -> '||;s|'.*||" | xargs basename)
+            DTB_SECTION=$(echo "${POST_DDR_LINE}" | sed "s/.*PH# \([0-9]*\).*/\1/")
+        fi
+    fi
+
+    if [ -n "${DTB_PATCH}" ]; then
+        ORIG_DTB="${staged_dir}/${DTB_PATCH}"
+        UPDATED_DTB="${staged_dir}/${DTB_PATCH%.dtb}-updated.dtb"
+
+        qcom-capsule-tool set-dtb-property \
+            "${ORIG_DTB}" \
+            /sw/uefi/uefiplat \
+            QcCapsuleRootCert \
+            "@list:${ROOT_INC}" \
+            "${UPDATED_DTB}"
+
+        qcom-capsule-tool parse-config \
+            "${xbl_config}" replace \
+            "${DTB_SECTION}" \
+            "${UPDATED_DTB}" \
+            "${staged_dir}/xbl_config_patched.elf"
+
+        mv "${staged_dir}/xbl_config_patched.elf" \
+           "${xbl_config}"
+
+        touch "${CAPSULE_DIR}/.xbl_with_oem_cert"
+    fi
+}
+
 do_compile() {
     CBSP_DATA="${STAGING_DATADIR_NATIVE}/cbsp-boot-utilities"
     EDK2_BASETOOLS="${STAGING_DATADIR_NATIVE}/edk2-basetools"
@@ -234,47 +285,10 @@ do_compile() {
     mkdir -p "${BOOTBINS_STAGED}"
     cp -r "${BOOTBINS_DIR}/." "${BOOTBINS_STAGED}/"
 
-    # Dump xbl_config.elf sections; parse output to auto-detect the post-DDR
-    # DTB and its section index.  XBLCONFIG_DTB / XBLCONFIG_DTB_SECTION
-    # override auto-detection when set explicitly.
-    XBL_DUMP_LOG="${CAPSULE_DIR}/xbl_dump.log"
-    qcom-capsule-tool parse-config \
-        "${BOOTBINS_STAGED}/xbl_config.elf" dump \
-        --out-dir "${BOOTBINS_STAGED}" | tee "${XBL_DUMP_LOG}"
-
-    DTB_PATCH="${XBLCONFIG_DTB}"
-    DTB_SECTION="${XBLCONFIG_DTB_SECTION}"
-    if [ -z "${DTB_PATCH}" ]; then
-        # Parse a line like:
-        #   [+] config_item[6] -> PH# 8 -> './post-ddr-kodiak-1.0.dtb' (90280 bytes)
-        POST_DDR_LINE=$(grep -m1 "post-ddr.*\.dtb" "${XBL_DUMP_LOG}" || true)
-        if [ -n "${POST_DDR_LINE}" ]; then
-            DTB_PATCH=$(echo "${POST_DDR_LINE}" | sed "s|.* -> '||;s|'.*||" | xargs basename)
-            DTB_SECTION=$(echo "${POST_DDR_LINE}" | sed "s/.*PH# \([0-9]*\).*/\1/")
-        fi
-    fi
-
-    if [ -n "${DTB_PATCH}" ]; then
-        ORIG_DTB="${BOOTBINS_STAGED}/${DTB_PATCH}"
-        UPDATED_DTB="${BOOTBINS_STAGED}/${DTB_PATCH%.dtb}-updated.dtb"
-
-        qcom-capsule-tool set-dtb-property \
-            "${ORIG_DTB}" \
-            /sw/uefi/uefiplat \
-            QcCapsuleRootCert \
-            "@list:${ROOT_INC}" \
-            "${UPDATED_DTB}"
-
-        qcom-capsule-tool parse-config \
-            "${BOOTBINS_STAGED}/xbl_config.elf" replace \
-            "${DTB_SECTION}" \
-            "${UPDATED_DTB}" \
-            "${BOOTBINS_STAGED}/xbl_config_patched.elf"
-
-        mv "${BOOTBINS_STAGED}/xbl_config_patched.elf" \
-           "${BOOTBINS_STAGED}/xbl_config.elf"
-
-        touch "${CAPSULE_DIR}/.xbl_with_oem_cert"
+    # Inject OEM root cert into xbl_config.elf when present.  Platforms
+    # without xbl_config.elf (e.g. hamoa) skip this step.
+    if [ -f "${BOOTBINS_STAGED}/xbl_config.elf" ]; then
+        patch_xblconfig_cert "${BOOTBINS_STAGED}/xbl_config.elf"
     fi
 
     qcom-capsule-tool sysfw-version-create \
--- a/classes-recipe/qcom-capsule.bbclass
+++ b/classes-recipe/qcom-capsule.bbclass
@@ -59,13 +59,30 @@ XBLCONFIG_DTB_SECTION ?= ""
 BOOTBINS_DIR ?= "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}"
 
 # ---------------------------------------------------------------------------
-# Custom FvUpdate.xml (optional)
+# Custom / generated FvUpdate.xml
 # ---------------------------------------------------------------------------
 # To provide a board/project-specific capsule layout, append your file to
 # SRC_URI and name it FvUpdate.xml, e.g. in a .bbappend or local.conf:
 #   SRC_URI:append = " file://my-board-FvUpdate.xml;subdir=fvupdate"
 # The class detects a custom FvUpdate.xml placed in ${WORKDIR} and uses
 # it in place of the upstream default.
+#
+# Alternatively, set CAPSULE_ENTRIES to a space-separated list of entry
+# names to generate FvUpdate.xml at build time.  For each name FOO define
+# the following flags on CAPSULE_ENTRY_FOO:
+#
+#   [binary]           - input filename resolved relative to BOOTBINS_STAGED
+#   [dest_disk]        - destination DiskType  (e.g. SPINOR, UFS_LUN1)
+#   [dest_partition]   - destination PartitionName
+#   [dest_guid]        - destination PartitionTypeGUID
+#   [backup_disk]      - backup DiskType       (optional)
+#   [backup_partition] - backup PartitionName  (optional)
+#   [backup_guid]      - backup PartitionTypeGUID (optional)
+#
+# When CAPSULE_ENTRIES is empty the class falls back to a static FvUpdate.xml
+# provided via SRC_URI or the default bundled in cbsp-boot-utilities.
+CAPSULE_FLASH_TYPE ?= "UFS"
+CAPSULE_ENTRIES    ?= ""
 
 inherit python3native deploy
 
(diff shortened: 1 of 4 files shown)


lumag (changes requested):
Split the second commit into atomical logical chunks.

lumag:
Also please drop the description of the changes from the PR description. If you want to keep the 'testing' part, please rewrite it to be an English text rather than an unreadable bullet list.

xueqnie:
> Also please drop the description of the changes from the PR description. If you want to keep the 'testing' part, please rewrite it to be an English text rather than an unreadable bullet list.
>done

conf/machine/iq-x7181-evk.conf
```diff
@@ -29,6 +29,8 @@ QCOM_PARTITION_FILES_SUBDIR_SPINOR ?= "partitions/iq-x7181-evk/spinor"
 
 QCOM_BOOT_FIRMWARE = "firmware-qcom-boot-iq-x7181"
 
+CAPSULE_GUID       = "0F6D58FC-2258-4D27-9E23-D77219B0897C"
```
lumag, line 32: Why does it need a separate GUID?

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_%.bbappend
```diff
@@ -0,0 +1,10 @@
+CAPSULE_FLASH_TYPE:iq-x7181-evk = "NORUFS"
```
lumag, line 1: This is useless before the next commit. And I guess that the previous commit is also useless. Could you order them?

classes-recipe/qcom-capsule.bbclass
```diff
@@ -59,13 +59,24 @@ XBLCONFIG_DTB_SECTION ?= ""
 BOOTBINS_DIR ?= "${DEPLOY_DIR_IMAGE}/${QCOM_BOOT_FILES_SUBDIR}"
```
lumag, line 59: > CAPSULE_ENTRIES variable flag system generates 

No, the flag can't generate anything. Also you are still describing the contents. Start by describing the issue. What is not working?

classes-recipe/qcom-capsule.bbclass
```diff
@@ -59,13 +59,24 @@ XBLCONFIG_DTB_SECTION ?= ""
...
 
 # ---------------------------------------------------------------------------
-# Custom FvUpdate.xml (optional)
+# Firmware entries (generates FvUpdate.xml at build time)
 # ---------------------------------------------------------------------------
-# To provide a board/project-specific capsule layout, append your file to
-# SRC_URI and name it FvUpdate.xml, e.g. in a .bbappend or local.conf:
-#   SRC_URI:append = " file://my-board-FvUpdate.xml;subdir=fvupdate"
```
lumag, line 66: Why are you dropping this comment?

classes-recipe/qcom-capsule.bbclass
```diff
@@ -109,6 +120,82 @@ do_configure[noexec] = "1"
 # Ensure boot binaries are deployed before we try to consume them
 do_compile[depends] += "${@'${QCOM_BOOT_FIRMWARE}:do_deploy' if d.getVar('QCOM_BOOT_FIRMWARE') else ''}"
 
+# Pull in the kernel DTB vfat image when capsule includes a dtb entry.
+do_compile[depends] += "${@'virtual/kernel:do_deploy' if 'dtb' in (d.getVar('CAPSULE_ENTRIES') or '').split() else ''}"
```
lumag: You don't need `or ''` here, There is a default value.

classes-recipe/qcom-capsule.bbclass
```diff
@@ -109,6 +120,82 @@ do_configure[noexec] = "1"
...
+                '      <PartitionName>%s</PartitionName>' % bkup_part,
+                '      <PartitionTypeGUID>%s</PartitionTypeGUID>' % bkup_guid,
+                '    </Backup>',
+            ]
+
+        lines += ['  </FwEntry>', '']
+
+    lines.append('</FVItems>')
+
+    out = os.path.join(workdir, 'FvUpdate.xml')
```
lumag: What will clean this file? Can we use `${B}` instad?

classes-recipe/qcom-capsule.bbclass
```diff
@@ -140,47 +229,56 @@ do_compile() {
...
-    DTB_SECTION="${XBLCONFIG_DTB_SECTION}"
-    if [ -z "${DTB_PATCH}" ]; then
-        # Parse a line like:
-        #   [+] config_item[6] -> PH# 8 -> './post-ddr-kodiak-1.0.dtb' (90280 bytes)
-        POST_DDR_LINE=$(grep -m1 "post-ddr.*\.dtb" "${XBL_DUMP_LOG}" || true)
-        if [ -n "${POST_DDR_LINE}" ]; then
-            DTB_PATCH=$(echo "${POST_DDR_LINE}" | sed "s|.* -> '||;s|'.*||" | xargs basename)
-            DTB_SECTION=$(echo "${POST_DDR_LINE}" | sed "s/.*PH# \([0-9]*\).*/\1/")
-        fi
+    # Stage kernel DTB vfat image as dtb.bin so FVCreation.py can find it
```
lumag, line 291: This looks like a separate commit.

lumag (changes requested):

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
```diff
@@ -5,9 +5,8 @@ LICENSE = "BSD-3-Clause-Clear"
 LIC_FILES_CHKSUM = "file://uefi_capsule_generation/LICENSE;md5=8e1eb38e3de3966193d29f31f5d7e684"
 
 SRC_URI = "git://github.com/quic/cbsp-boot-utilities.git;protocol=https;branch=main \
-           file://0001-FVCreation-fall-back-to-PATH-when-locating-GenFfs-Ge.patch \
           "
-SRCREV = "c25b41a49595aca2a3480db050dee78edd2a3bff"
+SRCREV = "b5c7d77383b0090f62daeea9bf059a2bbbef6376"
```
ricardosalveti: Needs coordination with https://github.com/qualcomm-linux/meta-qcom/pull/2288.

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
```diff
@@ -2,7 +2,7 @@ DESCRIPTION = "UEFI FMP capsule for Qualcomm platforms"
 LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
-COMPATIBLE_MACHINE = "qcm6490|qcs9100|qcs8300|qcs615"
+COMPATIBLE_MACHINE = "qcm6490|qcs9100|qcs8300|qcs615|hamoa"
```
ricardosalveti: This should only be done after all the other changes, reorder the commits logically.

conf/machine/iq-x7181-evk.conf
xueqnie: Each platform requires a unique ESRT GUID so the OS firmware update stack can identify which device a capsule targets. The IQ-X7181 GUID is defined in cbsp-boot-utilities documentation.

classes-recipe/qcom-capsule.bbclass
xueqnie: fiixed

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_%.bbappend
xueqnie: Done, commits have been reordered

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
xueqnie: Done, commits have been reordered

classes-recipe/qcom-capsule.bbclass
xueqnie: Done

classes-recipe/qcom-capsule.bbclass
xueqnie: Fixed, the comment is preserved

classes-recipe/qcom-capsule.bbclass
xueqnie: Fixed. Commit message now starts with the problem: the bundled FvUpdate.xml only contains a sample XBL entry and does not cover Hamoa's DTB partition layout.

dynamic-layers/meta-arm/recipes-devtools/cbsp-boot-utilities/cbsp-boot-utilities-native_1.0.bb
xueqnie: done

classes-recipe/qcom-capsule.bbclass
```diff
@@ -233,47 +233,51 @@ do_compile() {
...
-    if [ -z "${DTB_PATCH}" ]; then
-        # Parse a line like:
-        #   [+] config_item[6] -> PH# 8 -> './post-ddr-kodiak-1.0.dtb' (90280 bytes)
-        POST_DDR_LINE=$(grep -m1 "post-ddr.*\.dtb" "${XBL_DUMP_LOG}" || true)
-        if [ -n "${POST_DDR_LINE}" ]; then
-            DTB_PATCH=$(echo "${POST_DDR_LINE}" | sed "s|.* -> '||;s|'.*||" | xargs basename)
-            DTB_SECTION=$(echo "${POST_DDR_LINE}" | sed "s/.*PH# \([0-9]*\).*/\1/")
+    # Inject OEM root cert into xbl_config.elf when present.  Platforms
+    # without xbl_config.elf (e.g. hamoa) skip this step.
+    if [ -f "${BOOTBINS_STAGED}/xbl_config.elf" ]; then
```
lumag, line 303: The function is already long and complex. As you are touching this piece of code, just extract it to a function, processing `xbl_config.elf` passed as a param.

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_%.bbappend
```diff
@@ -0,0 +1,10 @@
...
+CAPSULE_ENTRIES:iq-x7181-evk    = "dtb"
+
+CAPSULE_ENTRY_dtb[binary]           = "dtb.bin"
+CAPSULE_ENTRY_dtb[dest_disk]        = "SPINOR"
+CAPSULE_ENTRY_dtb[dest_partition]   = "dtb_a"
+CAPSULE_ENTRY_dtb[dest_guid]        = "{2A1A52FC-AA0B-401C-A808-5EA0F91068F8}"
+CAPSULE_ENTRY_dtb[backup_disk]      = "SPINOR"
+CAPSULE_ENTRY_dtb[backup_partition] = "dtb_b"
```
lumag: I'd say, there is a difference between 'a/b' system and 'main/backup'. The former expects that we update the 'non-current slot', while the latter is expected to always update the first one and then either revert from a backup or update the backup too on success. Could we please make sure that our capsule updates are coherent between the naming and the purpose.

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
lumag: please keep it sorted.

lumag (changes requested):

classes-recipe/qcom-capsule.bbclass
xueqnie: Done . extracted the xbl_config cert injection logic into a new patch_xblconfig_cert() shell function that takes the
  xbl_config.elf path as $1

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_%.bbappend
xueqnie: Confirmed from bootup log — Hamoa uses a main/backup model.

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_1.0.bb
xueqnie: done

dynamic-layers/meta-arm/recipes-firmware/firmware/firmware-qcom-capsule_%.bbappend
lumag: Then we need to get rid of the _a / _b naming. I think some of the laptops were explicitly using the `BACKUP_foo` name for the backup partitions. Can we do this?

xueqnie: The platform partition layout is inconsistent — some partitions use _BACKUP suffix (CPUCP_DTB_BACKUP,
  ADSP_UEFI_DTB_BACKUP) while others use _b (dtb_b, uefi_dtb_b). The DTB partition was defined with _b naming in the firmware.Renaming to BACKUP_dtb would require a firmware partition table change.

lumag: PLease fix those in qcom-ptool repo.

xueqnie: done

classes-recipe/qcom-capsule.bbclass
```diff
@@ -234,47 +285,10 @@ do_compile() {
...
-            "${UPDATED_DTB}" \
-            "${BOOTBINS_STAGED}/xbl_config_patched.elf"
-
-        mv "${BOOTBINS_STAGED}/xbl_config_patched.elf" \
-           "${BOOTBINS_STAGED}/xbl_config.elf"
-
-        touch "${CAPSULE_DIR}/.xbl_with_oem_cert"
+    # Inject OEM root cert into xbl_config.elf when present.  Platforms
+    # without xbl_config.elf (e.g. hamoa) skip this step.
+    if [ -f "${BOOTBINS_STAGED}/xbl_config.elf" ]; then
```
igoropaniuk, line 303: Where exactly is the root certificate injected for capsule verification on Hamoa, and do you plan to add support for that?

classes-recipe/qcom-capsule.bbclass
```diff
@@ -109,6 +126,80 @@ do_configure[noexec] = "1"
 # Ensure boot binaries are deployed before we try to consume them
 do_compile[depends] += "${@'${QCOM_BOOT_FIRMWARE}:do_deploy' if d.getVar('QCOM_BOOT_FIRMWARE') else ''}"
 
+python generate_fvupdate() {
```
igoropaniuk, line 132: I'd like to flag a longer-term direction for this work before it lands, so we have a clear migration target - not as a blocker for this PR.

The CAPSULE_ENTRIES mechanism gets us Hamoa capsule support today, which is the right call and great. But I think the structural home for `FvUpdate.xml` generation is `qcom-ptool`, alongside the other XMLs it emits (partitions.xml, rawprogram*.xml, patch*.xml). The reason is single source of truth.

Right now we have to declare each capsule-updatable partition twice:
* Once in platforms/<board>/<storage>/partitions.conf - name, GUID, disk type, filename
* Again in a bbappend via CAPSULE_ENTRY_<name>[dest_partition|dest_guid|dest_disk|binary]

These will drift. Bump a GUID in partitions.conf, forget the bbappend, and the capsule deploys cleanly but fails on the target - the kind of bug you only find by bricking a board (the worst case - during capsule updates). `qcom-ptool` already has every field `FvUpdate.xml` needs (name, type-guid, storage.type, filename); adding capsule output means declaring "this partition is capsule-updatable" once and getting both XMLs from the same spec.

There's an ongoing private discussion of transition of qcom-ptool's source format from the current line-based partitions.conf to a structured format (YAML), which will make FvUpdate.xml a natural additional emitter target, I'll soon create an issue directly on qcom-ptool's github. With the planned extends mechanism for shared layouts, common capsule patterns (the SPINOR dtb/dtb_BACKUP pair, secure-boot HMAC quadruplets, etc.) would compose the same way partition groups will - declared once in a shared file, reused across boards. The bbappend approach can't factor that out.

That said, the YAML migration on qcom-ptool will take time (if we have common ground about how that should look like and approved). I don't think this PR should wait - I guess Hamoa needs capsule support now, and the fallback chain (${B}/FvUpdate.xml -> ${WORKDIR}/FvUpdate.xml -> ${CBSP_DATA}/FvUpdate.xml) already gives us a clean migration path: when qcom-ptool supports the new format, it just becomes another source feeding the same ${B} slot.

So: ack the approach for landing today, with the intent to track the qcom-ptool YAML transition as the longer-term home.

igoropaniuk: FYI https://github.com/qualcomm-linux/qcom-ptool/issues/124

classes-recipe/qcom-capsule.bbclass
xueqnie: You're right that there's an inconsistency: QCS615 cert injection is handled by the bbclass (via xbl_config.elf), while Hamoa relies on QDTE externally (via uefi_dtbs.elf).  I've raised an issue on cbsp-boot-utilities ([quic/cbsp-boot-utilities#](https://github.com/quic/cbsp-boot-utilities/issues/42)) to track uefi_dtbs.elf cert injection support to
  qcom-capsule-tool. Once that tooling is available, the bbclass can handle cert injection for both paths consistently.

classes-recipe/qcom-capsule.bbclass
xueqnie: thanks for reviewing
Agree this is a problem. My plan is to have **generate_fvupdate(**) look up GUIDs directly from the deployed partitions.xml (already produced by qcom-partition-conf), eliminating the [dest_guid]/[backup_guid] flags from bbappends entirely. This keeps FvUpdate.xml generation in the capsule class (since it's capsule-specific) and avoids changing qcom-ptool for all platforms.

classes-recipe/qcom-capsule.bbclass
igoropaniuk: FYI, there has been a PR adding support for `uefi_dtbs.elf` published (https://github.com/quic/cbsp-boot-utilities/pull/43) by @GuanquanTian

Could you take a look as well and provide comments? thanks!

igoropaniuk:
I’m overall happy with this PR. Given the ongoing discussion about whether to extend the in-house scripts in cbsp-boot-utilities to support also `uefi_dtbs.elf` or integrate the QDTE tool into `meta-qcom` - which may take some time - we can add certificate-embedding support for `uefi_dtbs.elf` later.

So, LGTM.

lumag (approved):

ricardosalveti (approved):

PR #2302, opened 2026-05-27
qcom-common.inc: reduce SWIOTLB allocation to 2 MB

jpagadal:
The default SWIOTLB allocation of 64 MiB leads to unnecessary kernel reserved memory.

Limit the SWIOTLB pool to 2 MiB by setting swiotlb=128 in the kernel command line.
Validation confirms the allocation is reduced from 64 MiB to 2 MiB.

This change reduces the kernel reserved memory footprint and
improves available RAM for user space.

### DIFF
--- a/conf/machine/include/qcom-common.inc
+++ b/conf/machine/include/qcom-common.inc
@@ -95,3 +95,5 @@ RT_KERNEL_CMDLINE = "${RT_ARGS_ISOL} ${RT_ARGS_IRQ} ${RT_ARGS_NOCBS} \
 KERNEL_CMDLINE_EXTRA:append = "${@bb.utils.contains_any('PREFERRED_PROVIDER_virtual/kernel',\
                               'linux-qcom-rt linux-qcom-next-rt', \
                               ' ${RT_KERNEL_CMDLINE}', '', d)}"
+
+KERNEL_CMDLINE_EXTRA:append = " swiotlb=128"


lumag (changes requested):
Is 2 MiB enough? Do we have CONFIG_SWIOTLB_DYNAMIC enabled to get more pools if required?

Also please drop the bullet lists from the PR description. Write English text instead.

jpagadal:
> Is 2 MiB enough? Do we have CONFIG_SWIOTLB_DYNAMIC enabled to get more pools if required?
> 
RB3Gen2, IQ8, and IQ9 devices show that /sys/kernel/debug/swiotlb/io_tlb_used is 0 and /sys/kernel/debug/swiotlb/io_tlb_used_hiwater is 4 (8 KiB). The 2 MiB value is proposed based on observations from the previous QLI release.
 
Additionally, a separate change has been proposed (https://github.com/qualcomm-linux/kernel/pull/642) to enable CONFIG_SWIOTLB_DYNAMIC and it is merged.

> Also please drop the bullet lists from the PR description. Write English text instead.

Done.

jpagadal:
@lumag, can you please review.

ci/performance.yml
```diff
@@ -3,4 +3,4 @@ header:
 
 local_conf_header:
   cmdline: |
-    KERNEL_CMDLINE_EXTRA:append = " quiet"
+    KERNEL_CMDLINE_EXTRA:append = " quiet swiotlb=128"
```
ndechesne: why are we making this change in a KAS yaml file? How are we expecting customers to find it and use it? Why is it not the default?

jpagadal: Initially, I added this change in ci/performance.yaml assuming it to be the
central place for performance-related optimizations. Based on your review, I realized that placing such changes here is not ideal.

To address this, I have moved the change to qcom-common.inc so it becomes part
of the default configuration and is consistently applied across builds.

ricardosalveti (changes requested):
NAK for custom kernel command line changes that are not strictly required.

lumag (changes requested):
> This aligns with the configuration used in QLI1.0.

This is not the reason to reduce the defaults. Please explain why it is a reasonable change, and why it can't be handled elsewhere.

jpagadal:
> NAK for custom kernel command line changes that are not strictly required.

This is required to meet critical product memory KPI and influences the RAM availability.

jpagadal:
> > This aligns with the configuration used in QLI1.0.
> 
> This is not the reason to reduce the defaults. Please explain why it is a reasonable change, and why it can't be handled elsewhere.

This change is based on observations that SWIOTLB is significantly under‑utilized on Qualcomm platforms.
In our validation with display, camera, and audio workloads active, the maximum observed usage was 8KB compared to the default 64MB reservation.
This indicates that the default allocation is overly conservative and results in unnecessary kernel memory reservation.
I initially placed this change in ci/performance.yml as it can help performance when applied. After it was rejected, I moved it to qcom-common.inc‎, but that also got NACK.

For the final end prodct commercializations, maintaining these kind of optimizations at a code place like performance.yml would be very benificial for the customer and easy to track.

Could we move this to performance.yml atleast.

ricardosalveti:
> This is required to meet critical product memory KPI and influences the RAM availability.

Then we should fix the default behavior to be able to do so, remember not everyone will use meta-qcom.

ci/performance.yml
ndechesne: ok.. there is more context now. And I think my previous comment above probably was not correct.. 

in fact this change brings some interesting question.. like do we really want such a chance in our build. QLI 1.x was a highly optimized/customized product/build. With QLI 2.x we are looking for a generic Linux product, and this SWIOTLB is a product specific configuration. and reducing it could be a problem for some customers / some use cases. 

So the real question is why are we doing it in the first place?

jpagadal:
> > This is required to meet critical product memory KPI and influences the RAM availability.
> 
> Then we should fix the default behavior to be able to do so, remember not everyone will use meta-qcom.

Thanks, could you please suggest the appropriate place to implement this?

lumag:
> Thanks, could you please suggest the appropriate place to implement this?

Kernel configuration?

jpagadal:
> > Thanks, could you please suggest the appropriate place to implement this?
> 
> Kernel configuration?

The kernel allocates a default SWIOTLB buffer size of 64MB, defined by IO_TLB_DEFAULT_SIZE right at the early boot itself.
This value is hardcoded and cannot be tuned via defconfig; it can only be overridden using the kernel command-line parameter (swiotlb=).
While CONFIG_SWIOTLB_DYNAMIC is already enabled, it does not change the initial default size and relies on runtime conditions to adjust allocations.

https://github.com/torvalds/linux/blob/2d3090a8aeb596a26935db0955d46c9a5db5c6ce/include/linux/swiotlb.h#L36

/* default to 64MB */
#define IO_TLB_DEFAULT_SIZE (64UL << 20)

lumag:
> The kernel allocates a default SWIOTLB buffer size of 64MB, defined by IO_TLB_DEFAULT_SIZE right at the early boot itself.

Implement Kconfig option to tune `IO_TLB_DEFAULT_SIZE`, send it upstream.

ricardosalveti:
Kernel config would then be available for any distro and user to set, better than setting it here.

jpagadal:
Thanks for the feedback.
The corresponding Kconfig support has already been submitted upstream and has been accepted by the DMA mapping maintainer.

Patch: https://lore.kernel.org/all/20260617-swiotlb-v1-1-abfee3faf4ea@oss.qualcomm.com/

lumag:
Good, then get it merged to the qcom-next kernel.

PR #2304, opened 2026-05-27
workflows: only run qcom-distro with 6.18 kernel on master

mwasilew:
There are 2 tracks for running PR tests in meta-qcom branches: master and wrynrose. The build matrix in these branches diverged and qcom-distro with 6.18 kernel is not a separate build on wrynrose. When submitting a PR against wrynrose test workflow from master branch is used. For this reason the workflow from master must implement the logic to select branch specific test plan.

### DIFF
--- a/.github/workflows/test-pr-wrynose.yml
+++ b/.github/workflows/test-pr-wrynose.yml
@@ -27,17 +27,18 @@ jobs:
           BRANCH=$(echo "$PULL_REQUESTS_JSON" | jq -r '.[0].base.ref // "master"')
           echo "value=$BRANCH" >> "$GITHUB_OUTPUT"
 
-  test:
+  test-wrynose:
     needs: determine-target-branch
-    if: github.event.workflow_run.conclusion == 'success' && needs.determine-target-branch.outputs.branch == 'master'
-    uses: ./.github/workflows/test.yml
+    if: github.event.workflow_run.conclusion == 'success' && needs.determine-target-branch.outputs.branch == 'wrynose'
+    uses: qualcomm-linux/meta-qcom/.github/workflows/test.yml@wrynose
     secrets: inherit
     with:
       build_id: ${{ github.event.workflow_run.id }}
 
   comment-on-pr:
     name: "Comment on PR"
-    needs: test
+    needs: [test-wrynose]
+    if: always() && (needs.test-wrynose.result == 'success')
     runs-on: ubuntu-latest
     steps:
 
@@ -45,7 +46,7 @@ jobs:
         id: download-result-summary
         uses: actions/download-artifact@v7
         with:
-          artifact-ids: ${{ needs.test.outputs.summary_ids }}
+          artifact-ids: ${{ needs.test-wrynose.outputs.summary_ids }}
           path: results_summary
 
       - name: Download event file
@@ -86,7 +87,8 @@ jobs:
   publish-test-results:
     uses: ./.github/workflows/publish-results.yml
     secrets: inherit
-    needs: [test]
+    needs: [test-wrynose]
+    if: always() && (needs.test-wrynose.result == 'success')
     with:
       commit: ${{ github.event.workflow_run.sha }}
       event_file: artifacts/Event File/event.json


.github/workflows/test.yml
```diff
@@ -55,6 +81,7 @@ jobs:
       testkit_ref: "${{ needs.prepare-env.outputs.testkit-ref }}"
   test-qcom-distro_linux-qcom-6-18:
     needs: [prepare-env]
+    if: ${{ needs.prepare-env.outputs.target-branch == 'master' }}
```
lumag: This becomes too branch-specific. For the wrynose branch we also need to schedule tests using the `qcom-next` kernels. I'd really prefer to find a way to schedule the tests from the correct branch. Can we use workflow dispatching from here?

lumag (changes requested):

.github/workflows/test.yml
mwasilew: Workflows on `push` event will run using the code from the branch. The only issue we're trying to solve here is pull requests. When using `pull_request` trigger in the workflow, it runs in the context of forked repository. This means it doesn't have access to the tokens in `meta-qcom` so the LAVA jobs can't be submitted. When the workflow is triggered by `workflow_run` or `pull_request_trigger` it runs in the context of main repository (meta-qcom) and has access to the secrets. The trade-off is that the source code of the workflow is taken from the default branch of the repository, in our case `master`. I'm reading the docs, but I don't see an option to grant access to the secrets and execute the workflow code from the target branch.

mwasilew:
@lumag this is another approach (pasting here to gather initial comments)

```
diff --git a/.github/workflows/test-pr.yml b/.github/workflows/test-pr.yml
index ef6d92d4..e5437426 100644
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -15,16 +15,38 @@ permissions:
   packages: read

 jobs:
-  test:
-    if: ${{ github.event.workflow_run.conclusion == 'success' }}
-    uses: ./.github/workflows/test.yml
+  determine-target-branch:
+    runs-on: ubuntu-latest
+    outputs:
+      branch: ${{ steps.branch.outputs.value }}
+    steps:
+      - id: branch
+        env:
+          PULL_REQUESTS_JSON: ${{ toJSON(github.event.workflow_run.pull_requests) }}
+        run: |
+          BRANCH=$(echo "$PULL_REQUESTS_JSON" | jq -r '.[0].base.ref // "master"')
+          echo "value=$BRANCH" >> "$GITHUB_OUTPUT"
+
+  test-master:
+    needs: determine-target-branch
+    if: github.event.workflow_run.conclusion == 'success' && needs.determine-target-branch.outputs.branch == 'master'
+    uses: qualcomm-linux/meta-qcom/.github/workflows/test.yml@master
+    secrets: inherit
+    with:
+      build_id: ${{ github.event.workflow_run.id }}
+
+  test-wrynose:
+    needs: determine-target-branch
+    if: github.event.workflow_run.conclusion == 'success' && needs.determine-target-branch.outputs.branch == 'wrynose'
+    uses: qualcomm-linux/meta-qcom/.github/workflows/test.yml@wrynose
     secrets: inherit
     with:
       build_id: ${{ github.event.workflow_run.id }}

   comment-on-pr:
     name: "Comment on PR"
-    needs: test
+    needs: [test-master, test-wrynose]
+    if: always() && (needs.test-master.result == 'success' || needs.test-wrynose.result == 'success')
     runs-on: ubuntu-latest
     steps:

@@ -32,7 +54,7 @@ jobs:
         id: download-result-summary
         uses: actions/download-artifact@v7
         with:
-          artifact-ids: ${{ needs.test.outputs.summary_ids }}
+          artifact-ids: ${{ needs.test-master.outputs.summary_ids || needs.test-wrynose.outputs.summary_ids }}
           path: results_summary

       - name: Download event file
@@ -73,7 +95,8 @@ jobs:
   publish-test-results:
     uses: ./.github/workflows/publish-results.yml
     secrets: inherit
-    needs: [test]
+    needs: [test-master, test-wrynose]
+    if: always() && (needs.test-master.result == 'success' || needs.test-wrynose.result == 'success')
     with:
       commit: ${{ github.event.workflow_run.sha }}
       event_file: artifacts/Event File/event.json
```

I didn't test it yet. It allows to call `test.yml` from the target branch but still hardcodes the branch names. From what I understand github doesn't allow to dynamically call 
```
uses: qualcomm-linux/meta-qcom/.github/workflows/test.yml@${{ env.BRANCH }}
```

Would something like this work from your point of view?

lumag:
@mwasilew that looks ok.

mwasilew:
If you're happy with that let me change this PR implementation and see if it works.

mwasilew:
I updated the patch with the branch specific `test.yml`. This doesn't look great in the actions tab. Here are the examples:
PR against master (or main) in my test repository:
main_branch

PR against wrynose in my test repository:
wrynose_branch

The workflow graph looks "incomplete" to me, however I don't think there is a better solution.

lumag:
Yes, that's pretty bad. Can we instead dispatch a testing workflow instead? I.e. detect the branch in the workflow_run trigger and then use that branch to dispatch a new workflow?

mwasilew:
> Yes, that's pretty bad. Can we instead dispatch a testing workflow instead? I.e. detect the branch in the workflow_run trigger and then use that branch to dispatch a new workflow?

that's what this patch is doing. `test.yml` works as "reusable workflow" and it's called from `test-pr.yml` that is triggered by `workflow_run`. Ideally it would be possible to pass variables to the call as proposed above:
```
qualcomm-linux/meta-qcom/.github/workflows/test.yml@${{ env.BRANCH }}
```
But github doesn't support that. I don't see any other way of dealing with different branches in the PR. IMHO there are 2 options: 
 - a "know all" workflow in the main branch that disables (skips) parts based on the PR target brach
 - the "dispatcher" workflow (like the current patch) that calls relevant workflows from the proper branches.

I guess only other option is a github app that is implements all the test plan/branch logic.

lumag:
@mwasilew I have been thinking about it for a while. Can we get rid of using the `workflow_run` event completely and just make the `build-pr.yml` trigger (or include) the test.yml? In the end, we know exactly, which branch we are building and testing.

mwasilew:
The problem with your proposal is that the PR runs in the context of forked repository (unless it's a branch in qualcomm-linux/meta-qcom). This means the workflow doesn't have access to the LAVATOKEN to submit the jobs to our LAVA instance.

lumag (approved):

vkraleti (approved):

PR #2305, opened 2026-05-27
fit-dtb-compatible: invert key direction and split linux-qcom-only entries

ricardosalveti:
Redesigns FIT_DTB_COMPATIBLE so each encoded compatible string is the flag
key and the DTB+overlay combo is the value, inverting the previous layout
that buried the firmware target in the value field and made duplicate
compatible detection impossible at parse time.

Linux-qcom-only entries (el2, camx, staging, ifp-mezzanine overlays) are
moved into a new fit-dtb-compatible-linux-qcom.inc, loaded via
LINUX_QCOM_FIT_DTB_COMPATIBLE_INC only for linux-qcom* kernel variants,
keeping the base file valid for linux-yocto builds.  Selftests are updated
to cover both files.

Assisted-by: GitHub Copilot:claude-opus-4-6

### DIFF
--- a/classes-recipe/dtb-fit-image.bbclass
+++ b/classes-recipe/dtb-fit-image.bbclass
@@ -7,7 +7,12 @@
 inherit kernel-arch
 
 require conf/image-fitimage.conf
-require conf/machine/include/fit-dtb-compatible.inc
+
+# Selects which FIT_DTB_COMPATIBLE include to load.  Defaults to the base
+# file; set to fit-dtb-compatible-linux-qcom.inc in linux-qcom* kernel
+# recipes to extend with linux-qcom-only overlay entries.
+LINUX_QCOM_FIT_DTB_COMPATIBLE ?= "conf/machine/include/fit-dtb-compatible.inc"
+require ${LINUX_QCOM_FIT_DTB_COMPATIBLE}
 
 DEPENDS += "\
     u-boot-tools-native \
--- a/classes-recipe/dtb-fit-image.bbclass
+++ b/classes-recipe/dtb-fit-image.bbclass
@@ -53,30 +53,45 @@ python do_generate_qcom_fitimage() {
     # Collect DTB/DTBO names selected in KERNEL_DEVICETREE to validate declarative FIT_DTB_COMPATIBLE combinations
     dtb_keys_list  = {os.path.splitext(f)[0].replace(',', '_') for f in files_set}
 
-    # Parse composite compatible keys :
-    # FIT_DTB_COMPATIBLE[base+ovl1+ovl2] = "..."
+    # Parse FIT_DTB_COMPATIBLE[<encoded-compat>] = "<dtb-stem> [<overlay-stem>...]"
+    # Flag keys encode commas as underscores (BitBake syntax constraint); decode
+    # with replace('_', ',') to recover the actual compatible string.
+    base_compats = {}   # base_dtb_id -> space-separated compat string(s)
     overlay_groups  = {}
     overlay_compats = {}
+    seen_combos = set()  # track overlay combos already added to overlay_groups
 
     compat_flags = d.getVarFlags("FIT_DTB_COMPATIBLE") or {}
-    for key, compat_val in compat_flags.items():
-        if '+' not in key:
-            continue
+    for encoded_key, combo_val in compat_flags.items():
+        compat_str = encoded_key.replace('_', ',')
 
-        parts = [os.path.basename(p) for p in key.split('+')]
+        parts = [os.path.basename(p) for p in combo_val.split()]
         if not parts:
             continue
 
-        # Skip base+overlay combinations not present in KERNEL_DEVICETREE to avoid generating invalid FIT configs
-        # from declarative FIT_DTB_COMPATIBLE metadata
+        # Skip combinations not present in KERNEL_DEVICETREE to avoid generating
+        # invalid FIT configs from declarative FIT_DTB_COMPATIBLE metadata.
         if not all(dtb in dtb_keys_list for dtb in parts):
             continue
 
         base = parts[0] + ".dtb"
+        base_dtb_id = base.replace(',', '_')
         overlays = [ovl + ".dtbo" for ovl in parts[1:]]
 
-        overlay_groups.setdefault(base, []).append(overlays)
-        overlay_compats[key] = compat_val
+        if overlays:
+            # Encode commas as underscores so combo_key matches the lookup_key
+            # built in fitimage_emit_section_qcomconfig (dtb_only_fitimage.py),
+            # which applies the same replacement; keep overlay_groups keyed by
+            # the encoded base id for the same reason.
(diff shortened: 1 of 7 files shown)


ricardosalveti:
@lumag @kraghava-qti @vkraleti as discussed at https://github.com/qualcomm-linux/meta-qcom/pull/2273#issuecomment-4528312212, I decided to get copilot to invert the logic as it makes more sense, and it is also aligned with qcom-metadata. I also separated the the list of entries based on the kernel used, so we can easily maintain the two sets separately, making it clear about the origin of the DTS.

This is a major change (the inverted logic), but I believe it is easier to understand and maintain this way.

Opened as draft until it is properly validated with all targets.

conf/machine/include/fit-dtb-compatible.inc
```diff
@@ -1,238 +1,242 @@
...
+FIT_DTB_COMPATIBLE[qcom_sm8750-mtp] = "sm8750-mtp"
+
+# ---------- hamoa ----------
+FIT_DTB_COMPATIBLE[qcom_hamoa-evk] = \
+    "hamoa-iot-evk+hamoa-iot-evk-camera-imx577"
+FIT_DTB_COMPATIBLE[qcom_hamoa-evk-camx] = "hamoa-iot-evk+hamoa-evk-camx"
+
+# ---------- lemans ----------
+FIT_DTB_COMPATIBLE[qcom_qcs9075-iot] = \
+    "lemans-evk+lemans-evk-camera-csi1-imx577"
```
lumag: Could you please now replace '+' with whitespaces here? It would make the compats more readable (and would also allow splitting them to multiple lines).

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -9,6 +9,17 @@ inherit kernel-arch
 require conf/image-fitimage.conf
 require conf/machine/include/fit-dtb-compatible.inc
 
+# Kernel-variant-specific FIT_DTB_COMPATIBLE overrides (mirrors LINUX_QCOM_KERNEL_DEVICETREE).
+# Entries in fit-dtb-compatible-linux-qcom.inc can add overlay combos that include
+# LINUX_QCOM_KERNEL_DEVICETREE DTBOs and can override base-file entries for the same
+# key (BitBake last-write-wins).
+LINUX_QCOM_FIT_DTB_COMPATIBLE_INC ?= ""
```
lumag: Would it be better to default to `fit-dtb-compatible.inc` here and make all new includes include that file by default?

lumag (changes requested):

conf/machine/include/fit-dtb-compatible.inc
ricardosalveti: Yup, good suggestion, updated.

classes-recipe/dtb-fit-image.bbclass
ricardosalveti: Yup, updated.

classes-recipe/dtb-fit-image.bbclass
```diff
@@ -7,7 +7,16 @@
...
+
+# Selects which FIT_DTB_COMPATIBLE include to load.  Defaults to the base
+# file; linux-qcom* kernels load fit-dtb-compatible-linux-qcom.inc which
+# extends the base via its own require statement.
+LINUX_QCOM_FIT_DTB_COMPATIBLE_INC ?= "conf/machine/include/fit-dtb-compatible.inc"
+LINUX_QCOM_FIT_DTB_COMPATIBLE_INC:pn-linux-qcom         = "conf/machine/include/fit-dtb-compatible-linux-qcom.inc"
+LINUX_QCOM_FIT_DTB_COMPATIBLE_INC:pn-linux-qcom-next    = "conf/machine/include/fit-dtb-compatible-linux-qcom.inc"
+LINUX_QCOM_FIT_DTB_COMPATIBLE_INC:pn-linux-qcom-rt      = "conf/machine/include/fit-dtb-compatible-linux-qcom.inc"
+LINUX_QCOM_FIT_DTB_COMPATIBLE_INC:pn-linux-qcom-next-rt = "conf/machine/include/fit-dtb-compatible-linux-qcom.inc"
+include ${LINUX_QCOM_FIT_DTB_COMPATIBLE_INC}
```
quaresmajose: The change from 'require' to 'include' is not clear, the *.inc file will always be present so I think 'require' fits better.
Another point was the 'LINUX_QCOM_FIT_DTB_COMPATIBLE_INC:pn-' that maybe fits better inside the kernel recipe. We can also call it LINUX_QCOM_FIT_DTB_COMPATIBLE

ricardosalveti: Right, this is indeed fine as it is included via a kernel class, moved to the kernel recipe and moved back to required.

lib/qcom/dtb_only_fitimage.py
```diff
@@ -118,7 +118,7 @@ def fitimage_emit_section_qcomconfig(self, overlay_groups, overlay_compats):
                 dt_list = [dtb_id] + ovl_list
 
                 fdtentries = [f"fdt-{dt}" for dt in dt_list]
-                lookup_key = "+".join([os.path.splitext(dt)[0].replace(',', '_') for dt in dt_list])
+                lookup_key = " ".join([os.path.splitext(dt)[0].replace(',', '_') for dt in dt_list])
```
lumag, line 121: Hmm. Does it mean that it requires exactly one whitespace the value? It would be really inconvenient.

Vkenjam:
bootup failure on Rb8 and RB4 :https://axiom.qualcomm.com/#/reports/job/36984663 
[2026-05-29 05:00:48.260] BDS: LogFs sync skipped, Unsupported
[2026-05-29 05:00:48.260] App Log Flush : 0 ms
[2026-05-29 05:00:48.260]  RemoteObj_invoke  count 1 
[2026-05-29 05:00:48.260]  RemoteObj_invoke  count 1 
[2026-05-29 05:00:48.260] Remote object table is not empty. 4  
[2026-05-29 05:00:48.260] Local object table is not empty. 2  
[2026-05-29 05:00:48.310] ScmArmV8ExitBootServicesHandler, Status = 0x0. 
[2026-05-29 05:00:48.310] Gunyah based bootup
[2026-05-29 05:00:48.310] SmcInvokeInit ret 0
[2026-05-29 05:00:48.310] SmcInvokeCompatInit ret 0
[2026-05-29 05:00:48.310] Exit EBS        [ 3478] UEFI End

lib/qcom/dtb_only_fitimage.py
ricardosalveti: Here it is just building a key to index the overlay_compats dict, and it is derived out of a .split, which drops the extra whitespace that could be part of the variable, so this is safe.

ricardosalveti:
Lava and Axiom tests all functional now.

lumag (approved):

lumag (approved):

vkraleti (approved):

PR #2283, opened 2026-05-25
ci: base.lock: update layers to latest

vkraleti:
Update meta-arm, meta-audioreach, meta-security, meta-updater, oe-core
and bitbake layers to the latest available SHAs. Skip meta-openembedded
update as audio issues with pipewire upgrade are still not addressed.

PR #2309, opened 2026-05-28
ci: base.lock.yml: update oe-core to latest

jinwli-qc:
Update oe-core layer to the latest.


ricardosalveti:
See https://github.com/qualcomm-linux/meta-qcom/pull/2283

PR #2333, opened 2026-06-01
conf: qcom-common: append efi=runtime to kernel cmdline for RT builds

apateriy-qcom:
When PREEMPT_RT is enabled, EFI_DISABLE_RUNTIME defaults to 'y'
(drivers/firmware/efi/Kconfig), which disables EFI runtime services to
avoid the large latencies EFI calls can introduce under RT workloads.

EFI runtime services are required on Qualcomm platforms for several
post-boot operations. KVM hypervisor mode (el2kvm) is activated by
writing the VendorDtbOverlays EFI variable from userspace to select
the EL2 DTB overlay before rebooting; without EFI runtime services
this variable write fails, blocking KVM enablement. Watchdog and ramdump
collection, and remoteproc/PIL firmware loading, similarly require EFI
runtime services to be active.

Append efi=runtime to RT_KERNEL_CMDLINE in qcom-common
to re-enable EFI runtime services. Non-RT builds are unaffected as
EFI_DISABLE_RUNTIME is not set for them.

### DIFF
--- a/conf/machine/include/qcom-common.inc
+++ b/conf/machine/include/qcom-common.inc
@@ -92,7 +92,7 @@ RT_ARGS_IDLE  = "${@oe.utils.conditional('QCOM_CPUIDLE_OFF',   '', '', \
 
 # Combine
 RT_KERNEL_CMDLINE = "${RT_ARGS_ISOL} ${RT_ARGS_IRQ} ${RT_ARGS_NOCBS} \
-                     ${RT_ARGS_EXP} ${RT_ARGS_IDLE}"
+                     ${RT_ARGS_EXP} ${RT_ARGS_IDLE} efi=runtime"
 
 # Append only when RT kernel is selected
 KERNEL_CMDLINE_EXTRA:append = "${@bb.utils.contains_any('PREFERRED_PROVIDER_virtual/kernel',\


lumag (changes requested):
> From: Anurag Pateriya <anurag.pateriya@oss.qualcomm.com>
> Signed-off-by: Anurag Pateriya <apateriy@qti.qualcomm.com>

Which breaks DCO check.

Also, you are just stating that EFI Runtime services require a bootparam on RT kernels. Please point out why does it affect only RT kernels.

lumag (changes requested):
> RT kas configs were missing the efi=runtime boot parameter that are
preesnt in non-RT builds

I don't see `efi=runtime` being set in any of the non-RT builds. Pleae point out, where they are set.

apateriy-qcom:
> > RT kas configs were missing the efi=runtime boot parameter that are
> > preesnt in non-RT builds
> 
> I don't see `efi=runtime` being set in any of the non-RT builds. Pleae point out, where they are set.


EFI_DISABLE_RUNTIME is the readon why runtime efi services is disabled with preempt kernel
config EFI_DISABLE_RUNTIME
        bool "Disable EFI runtime services support by default"
        default y if PREEMPT_RT

The EFI runtime services are disabled by default when PREEMPT_RT is
          enabled, because measurements have shown that some EFI functions calls
          might take too much time to complete, causing large latencies which is
          an issue for Real-Time kernels.

shashim-quic:
> > RT kas configs were missing the efi=runtime boot parameter that are
> > preesnt in non-RT builds
> 
> I don't see `efi=runtime` being set in any of the non-RT builds. Pleae point out, where they are set.

Other (non-RT) configs do not need it as it gets enabled by kernel config. However for RT, we have following
case which needs explicit enablement.

Referring drivers/firmware/efi/Kconfig
```
config EFI_DISABLE_RUNTIME
        bool "Disable EFI runtime services support by default"
        default y if PREEMPT_RT
        help
          Allow to disable the EFI runtime services support by default. This can
          already be achieved by using the efi=noruntime option, but it could be
          useful to have this default without any kernel command line parameter.

          The EFI runtime services are disabled by default when PREEMPT_RT is
          enabled, because measurements have shown that some EFI functions calls
          might take too much time to complete, causing large latencies which is
          an issue for Real-Time kernels.

          This default can be overridden by using the efi=runtime option.
```

lumag:
> > > RT kas configs were missing the efi=runtime boot parameter that are
> > > preesnt in non-RT builds
> > 
> > 
> > I don't see `efi=runtime` being set in any of the non-RT builds. Pleae point out, where they are set.
> 
> Other (non-RT) configs do not need it as it gets enabled by kernel config. However for RT, we have following case which needs explicit enablement.

Sure. So this needs to be described in the commit message instead of it having the quoted phrase.

ricardosalveti:
> > > RT kas configs were missing the efi=runtime boot parameter that are
> > > preesnt in non-RT builds
> > 
> > 
> > I don't see `efi=runtime` being set in any of the non-RT builds. Pleae point out, where they are set.
> 
> EFI_DISABLE_RUNTIME is the readon why runtime efi services is disabled with preempt kernel config EFI_DISABLE_RUNTIME bool "Disable EFI runtime services support by default" default y if PREEMPT_RT
> 
> The EFI runtime services are disabled by default when PREEMPT_RT is enabled, because measurements have shown that some EFI functions calls might take too much time to complete, causing large latencies which is an issue for Real-Time kernels.

Please improve the commit description how explaining the current kernel behavior and why this is needed just for the RT configurations.

apateriy-qcom:
> > > > RT kas configs were missing the efi=runtime boot parameter that are
> > > > preesnt in non-RT builds
> > > 
> > > 
> > > I don't see `efi=runtime` being set in any of the non-RT builds. Pleae point out, where they are set.
> > 
> > 
> > EFI_DISABLE_RUNTIME is the readon why runtime efi services is disabled with preempt kernel config EFI_DISABLE_RUNTIME bool "Disable EFI runtime services support by default" default y if PREEMPT_RT
> > The EFI runtime services are disabled by default when PREEMPT_RT is enabled, because measurements have shown that some EFI functions calls might take too much time to complete, causing large latencies which is an issue for Real-Time kernels.
> 
> Please improve the commit description how explaining the current kernel behavior and why this is needed just for the RT configurations.

Thanks @ricardosalveti . Modified the commit description informing EFI_DISABLE_RUNTIME  being default as y for RT builds and why efi=runtime is required only for RT and not required for non-RT builds.

ricardosalveti (dismissed):

lumag (changes requested):
> re-enable EFI runtime services, which are required for KVM boot

Required _why_? You are still not desribing the issue.

apateriy-qcom:
> > re-enable EFI runtime services, which are required for KVM boot
> 
> Required _why_? You are still not desribing the issue.

@lumag  Added full details now to commit message describing the issue resolved by enabling efi runtime services. Kindly review.

lumag (changes requested):
The commit message indicates that `efi=runtime` should be set for all users of RT kernels even if they don't use our KAS files. Set it in `qcom-common.inc` for RT kernels.

apateriy-qcom:
> The commit message indicates that `efi=runtime` should be set for all users of RT kernels even if they don't use our KAS files. Set it in `qcom-common.inc` for RT kernels.

Hi @lumag ,

Added the change in qcom-common.inc for RT kernel commandline. Changed the commit text according to latest change.

lumag (dismissed):

ricardosalveti (dismissed):

conf/machine/include/qcom-common.inc
```diff
@@ -92,7 +89,7 @@ RT_ARGS_IDLE  = "${@oe.utils.conditional('QCOM_CPUIDLE_OFF',   '', '', \
 
 # Combine
 RT_KERNEL_CMDLINE = "${RT_ARGS_ISOL} ${RT_ARGS_IRQ} ${RT_ARGS_NOCBS} \
-                     ${RT_ARGS_EXP} ${RT_ARGS_IDLE}"
+                     ${RT_ARGS_EXP} ${RT_ARGS_IDLE} efi=runtime"
```
ndechesne, line 95: It's disabled upstream for a reason, "to avoid the large latencies EFI calls can introduce under RT workloads.". Are we saying that we ignore that completely for all our builds/users/customers?

shashim-quic: > It's disabled upstream for a reason, "to avoid the large latencies EFI calls can introduce under RT workloads.". Are we saying that we ignore that completely for all our builds/users/customers?

For production customer can disable it. However for development needs without EFI runtime variable we cannot realize efi variable based boot time overlays.

ricardosalveti: Did we measure if this will indeed cause large latencies on our socs? If indeed the case, then it is not really a great change to enable by default.

shashim-quic: > Did we measure if this will indeed cause large latencies on our socs? If indeed the case, then it is not really a great change to enable by default.

latencies will be induced when runtime services are used while RT workloads are running. user need to take care of that (and disable based on his runtime service need).

ricardosalveti: We should measure with our stack to know for sure, and document as well.

shashim-quic: > We should measure with our stack to know for sure, and document as well.

ack, working on getting this evaluation in parallel.

lumag (approved):

ricardosalveti (approved):

PR #1659, opened 2026-02-27
gstreamer1.0-plugins-bad: enable WebRTC support

Raja-Ganapathi-Busam:
Enable WebRTC functionality by adding the required components (sctp, srt, srtp, webrtc) to PACKAGECONFIG. 
This brings real‑time audio/video and data‑channel capabilities to GStreamer pipelines, allowing seamless peer‑to‑peer media transport and modern browser interoperability.

PR #2343, opened 2026-06-03
gst-plugins-imsdk-common: Add required dependencies for WebRTC

cnhrist:
Add gstreamer1.0-plugins-bad and libsoup dependencies required for WebRTC libs.

### DIFF
--- a/recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
+++ b/recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
@@ -33,7 +33,7 @@ PACKAGECONFIG[ml]           = "-DENABLE_GST_ML_PLUGINS=1, -DENABLE_GST_ML_PLUGIN
 PACKAGECONFIG[python-apps]  = "-DENABLE_GST_PYTHON_EXAMPLES=1, -DENABLE_GST_PYTHON_EXAMPLES=0"
 PACKAGECONFIG[qairt]        = "-DENABLE_GST_QAIRT_PLUGINS=1 -DENABLE_GST_PLUGIN_MLTOOLS=1, -DENABLE_GST_QAIRT_PLUGINS=0 -DENABLE_GST_PLUGIN_MLTOOLS=0, qairt-sdk, qairt-sdk"
 PACKAGECONFIG[redissink]    = "-DENABLE_GST_PLUGIN_REDISSINK=1, -DENABLE_GST_PLUGIN_REDISSINK=0, hiredis"
-PACKAGECONFIG[sample-apps]  = "-DENABLE_GST_SAMPLE_APPS=1, -DENABLE_GST_SAMPLE_APPS=0, opencv cairo json-glib"
+PACKAGECONFIG[sample-apps]  = "-DENABLE_GST_SAMPLE_APPS=1, -DENABLE_GST_SAMPLE_APPS=0, opencv cairo json-glib libsoup gstreamer1.0-plugins-bad"
 PACKAGECONFIG[smartvencbin] = "-DENABLE_GST_PLUGIN_SMARTVENCBIN=1, -DENABLE_GST_PLUGIN_SMARTVENCBIN=0, smart-venc-ctrl-algo"
 PACKAGECONFIG[sw]           = "-DENABLE_GST_SOFTWARE_PLUGINS=1, -DENABLE_GST_SOFTWARE_PLUGINS=0, gstreamer1.0-rtsp-server"
 PACKAGECONFIG[tflite]       = "-DENABLE_GST_PLUGIN_MLTFLITE=1, -DENABLE_GST_PLUGIN_MLTFLITE=0, tensorflow-lite"
@@ -50,4 +50,5 @@ RDEPENDS:${PN}-qtismartvencbin += "smart-venc-ctrl-algo"
 # To ensure runtime availability, added runtime dependency on 'mosquitto'.
 RDEPENDS:libgstqtimqttadaptor += "mosquitto"
 
-RDEPENDS:${PN}-apps += "${@bb.utils.contains('PACKAGECONFIG', 'python-apps', 'python3-core gstreamer1.0-python python3-pygobject gtk+3 python3-opencv', '', d)}"
+RDEPENDS:${PN}-apps += "${@bb.utils.contains('PACKAGECONFIG', 'python-apps', 'python3-core gstreamer1.0-python python3-pygobject gtk+3 python3-opencv', '', d)} \
+                        ${@bb.utils.contains('PACKAGECONFIG', 'webrtc', 'gstreamer1.0-plugins-bad-webrtc', '', d)}"


lumag (dismissed):

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
```diff
@@ -16,6 +16,7 @@ SRCREV = "bf7adb22735769b028aff3b94de7a4cc41db57a7"
 DEPENDS += "\
     gstreamer1.0 \
     gstreamer1.0-plugins-base \
+    gstreamer1.0-plugins-bad \
```
Raja-Ganapathi-Busam: Add `gstreamer1.0-plugins-bad` in `sample-apps` packageconfig option, since it is required for WebRTC sample apps only.

ricardosalveti: Also, do we need all the bad here?

ndechesne: +1, given the variety of plugins in bad let's be selective. 
and also let's keep things sorted!

cnhrist: gstreamer1.0-plugins-bad was moved to sample-apps, but not sure how to enable only webrtc from it. Also added dependency to libnice with enable nicesrc and nicesink plugins required for webrtcbin plugin.

Raja-Ganapathi-Busam: `libnice` is already a dependency for webrtc packageconfig.

`PACKAGECONFIG[webrtc]          = "-Dwebrtc=enabled,-Dwebrtc=disabled,libnice"`
https://github.com/openembedded/openembedded-core/blob/master/meta/recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_1.28.2.bb#L101

If it is not a direct dependency for sample apps, then we can drop from here.

cnhrist: Agree, libnice is removed

recipes-support/libnice/libnice_0.1%.bbappend
```diff
@@ -0,0 +1,2 @@
+DEPENDS:append:qcom = " gstreamer1.0"
```
ndechesne: I don't understand why you have this bbappend.

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
```diff
@@ -33,7 +33,7 @@ PACKAGECONFIG[ml]           = "-DENABLE_GST_ML_PLUGINS=1, -DENABLE_GST_ML_PLUGIN
 PACKAGECONFIG[python-apps]  = "-DENABLE_GST_PYTHON_EXAMPLES=1, -DENABLE_GST_PYTHON_EXAMPLES=0"
 PACKAGECONFIG[qairt]        = "-DENABLE_GST_QAIRT_PLUGINS=1 -DENABLE_GST_PLUGIN_MLTOOLS=1, -DENABLE_GST_QAIRT_PLUGINS=0 -DENABLE_GST_PLUGIN_MLTOOLS=0, qairt-sdk, qairt-sdk"
 PACKAGECONFIG[redissink]    = "-DENABLE_GST_PLUGIN_REDISSINK=1, -DENABLE_GST_PLUGIN_REDISSINK=0, hiredis"
-PACKAGECONFIG[sample-apps]  = "-DENABLE_GST_SAMPLE_APPS=1, -DENABLE_GST_SAMPLE_APPS=0, opencv cairo json-glib"
+PACKAGECONFIG[sample-apps]  = "-DENABLE_GST_SAMPLE_APPS=1, -DENABLE_GST_SAMPLE_APPS=0, opencv cairo json-glib libsoup gstreamer1.0-plugins-bad"
```
ndechesne, line 36: if you only need webrtc, then you can install only this plugin, this is what we were asking. the gst packaging is such that each plugin is installed in its own package, see gstreamer1.0-plugins-packaging.inc.

anujm1: > if you only need webrtc, then you can install only this plugin, this is what we were asking. the gst packaging is such that each plugin is installed in its own package, see gstreamer1.0-plugins-packaging.inc.

I think if the code is looking at `gstreamer-webrtc-1.0.pc`, it will need the build dependency on recipe. The runtime dependency should be only for webrtc.

recipes-support/libnice/libnice_0.1%.bbappend
```diff
@@ -0,0 +1 @@
+PACKAGECONFIG:append:qcom = " gstreamer"
```
koenkooi: I think this should be a seperate commit

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
```diff
@@ -7,3 +7,5 @@ SRC_URI:append:qcom = " \
     file://0002-waylandsink-Release-pending-buffers-during-PAUSED-to.patch \
     file://0003-waylandsink-support-gap-buffers.patch \
 "
+
+PACKAGECONFIG:append:qcom = " webrtc sctp srtp"
```
koenkooi: And this should also be a separate commit.

koenkooi (changes requested):

recipes-support/libnice/libnice_0.1%.bbappend
cnhrist: This is needed because webrtcbin depends on plugins nicesrc and nicesink which are provided by libnice but not enabled by default:
libnice_0.1.23.bb - PACKAGECONFIG[gstreamer] = "-Dgstreamer=enabled,-Dgstreamer=disabled,gstreamer1.0"

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
cnhrist: Added runtime dependency but I think cannot skip build dependency on gstreamer1.0-plugins-bad.

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
Raja-Ganapathi-Busam: This was already enabled in meta-qcom-distro since this is not BSP related feature.
Please find recommendation below.
https://github.com/qualcomm-linux/meta-qcom/pull/1659#discussion_r2865675481

The changes are already present here.
https://github.com/qualcomm-linux/meta-qcom-distro/blob/main/recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%25.bbappend

recipes-support/libnice/libnice_0.1%.bbappend
cnhrist: Done

recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%.bbappend
cnhrist: Yes, removed from this patch

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
cnhrist: The runtime dependency on WebRTC is not required, since we only have a build-time dependency and use only the gst/webrtc/webrtc.h header. After that, the application uses the webrtcbin plugin directly, which is already installed (and not provided by the gst-plugins-imsdk-oss recipe).

lumag: > which is already installed 

So, there is a runtime dependency. Please capture the dependency on the exact plugin package.

cnhrist: I need to add the following runtime dependency only for qcom-distro:
RDEPENDS:${PN}-apps:append:qcom-distro = " gstreamer1.0-plugins-bad-webrtc"

This is because WebRTC support is currently enabled only in qcom-distro, as defined here:
https://github.com/qualcomm-linux/meta-qcom-distro/blob/main/recipes-multimedia/gstreamer/gstreamer1.0-plugins-bad_%25.bbappend

Given this, should I keep the runtime dependency scoped only to qcom-distro, or would it be better to enable WebRTC support in this layer?

lumag: The WebRTC is a feature which should be enabled on the distro layer. This package should be able to detect if WebRTC is enabled or not and adjust the configuration to follow it.

cnhrist: Got it, I’ve updated the recipe. Тhanks for the guidance.

lumag (changes requested):
> - Extend PACKAGECONFIG[sample-apps] with:

Please, don't describe patch contents in the commit message. It's useless and usually can be understood from the patch itself. Instead start your commit message by describing the issue or the problem.

lumag (approved):

ricardosalveti (approved):

ricardosalveti:
@koenkooi mind another look?

anujm1 (approved):

koenkooi (approved):

PR #2345, opened 2026-06-03
gst-plugins-imsdk-common: Inherit GI and add dependecy on PyGObject

c-petarg:
Inherit gobject-introspection and add depends on PyGObject in order enable Python bindings from IMSDK base.

This change needs to be merged after a source commit gets merged.


lumag (changes requested):
> This change needs to be merged after a source commit gets merged.

What does this mean?

c-petarg:
> > This change needs to be merged after a source commit gets merged.
> 
> What does this mean?

Several commits are in queue to be merged in IMSDK mainline which these recipes are building. This change is required to be merged after they are merged on mainline.

PR #2348, opened 2026-06-03
Enable Purwa camx DTB compilation for linux-qcom kernel

cjha01:
Downstream camera DTB is not included in the final image
when building linux-qcom, linux-qcom-next, linux-qcom-rt,
or linux-qcom-next-rt kernels for the purwa EVK board.
This prevents the downstream camera stack from being
enabled at boot, even though the kernel supports it.

Include the downstream camera DTB so it is packaged into
the final image and downstream camera functionality is
available for the IQ-x5121 EVK.

Add multi-DTB support for camx overlay to enable downstream
camera for IQ-x5121 (purwa-evk-camx).


lumag (approved):

ricardosalveti (changes requested):
Invert the order of the commits.

cjha01:
> Invert the order of the commits.

Updated.

ricardosalveti (approved):

PR #2356, opened 2026-06-04
Enable downstream camera for QCS5430 Kodiak lite FP1

kripal-qipl:
Updated FIT_DTB_COMPATIBLE entries to enable downstream camera for 
QCS5430 base kodiak lite FP1 device.

Include the downstream qcs5430-fps-camx.dtbo DTB, so it is packaged
into the final image and downstream camera functionality is available
for the QCS5430 base kodiak lite FP1 device.

qcs6490-rb3gen2-vision-mezzanine-camx targets full RB3 Gen2 (QCS6490)
hardware, while Kodiak Lite is a QCS5430 Soft SKU where camera resources
(IFE etc.) are restricted by hypervisor based on license.
Using qcs6490-rb3gen2-vision-mezzanine-camx DT will lead to mismatch in
camera topology and inaccessible hardware blocks, so a dedicated qcs5430 camx DTBO is required.

### DIFF
--- a/conf/machine/rb3gen2-core-kit.conf
+++ b/conf/machine/rb3gen2-core-kit.conf
@@ -20,6 +20,7 @@ LINUX_QCOM_KERNEL_DEVICETREE ?= " \
                       qcom/qcs6490-rb3gen2-industrial-mezzanine-staging.dtbo \
                       qcom/kodiak-el2.dtbo \
                       qcom/kodiak-staging.dtbo \
+                      qcom/qcs5430-fps-camx.dtbo \
                       "
 
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
--- a/conf/machine/include/fit-dtb-compatible-linux-qcom.inc
+++ b/conf/machine/include/fit-dtb-compatible-linux-qcom.inc
@@ -181,17 +181,17 @@ FIT_DTB_COMPATIBLE[qcom_qcs6490-iot-subtype2-staging] = \
     "qcs6490-rb3gen2 qcs6490-rb3gen2-vision-mezzanine kodiak-staging"
 
 FIT_DTB_COMPATIBLE[qcom_qcs5430-iot-camx] = \
-    "qcs6490-rb3gen2 qcs6490-rb3gen2-vision-mezzanine-camx"
+    "qcs6490-rb3gen2 qcs5430-fps-camx"
 FIT_DTB_COMPATIBLE[qcom_qcs5430-iot-subtype2-camx] = \
-    "qcs6490-rb3gen2 qcs6490-rb3gen2-vision-mezzanine-camx"
+    "qcs6490-rb3gen2 qcs5430-fps-camx"
 FIT_DTB_COMPATIBLE[qcom_qcs6490-iot-camx] = \
     "qcs6490-rb3gen2 qcs6490-rb3gen2-vision-mezzanine-camx"
 FIT_DTB_COMPATIBLE[qcom_qcs6490-iot-subtype2-camx] = \
     "qcs6490-rb3gen2 qcs6490-rb3gen2-vision-mezzanine-camx"
 FIT_DTB_COMPATIBLE[qcom_qcs5430-iot-camx-staging] = \
-    "qcs6490-rb3gen2 qcs6490-rb3gen2-vision-mezzanine-camx kodiak-staging"
+    "qcs6490-rb3gen2 qcs5430-fps-camx kodiak-staging"
 FIT_DTB_COMPATIBLE[qcom_qcs5430-iot-subtype2-camx-staging] = \
-    "qcs6490-rb3gen2 qcs6490-rb3gen2-vision-mezzanine-camx kodiak-staging"
+    "qcs6490-rb3gen2 qcs5430-fps-camx kodiak-staging"
 FIT_DTB_COMPATIBLE[qcom_qcs6490-iot-camx-staging] = \
     "qcs6490-rb3gen2 qcs6490-rb3gen2-vision-mezzanine-camx kodiak-staging"
 FIT_DTB_COMPATIBLE[qcom_qcs6490-iot-subtype2-camx-staging] = \


conf/machine/rb3gen2-core-kit.conf
```diff
@@ -20,6 +20,7 @@ LINUX_QCOM_KERNEL_DEVICETREE ?= " \
                       qcom/qcs6490-rb3gen2-industrial-mezzanine-staging.dtbo \
                       qcom/kodiak-el2.dtbo \
                       qcom/kodiak-staging.dtbo \
+                      qcom/qcs5430-fps-camx.dtbo \
```
ricardosalveti, line 23: Why can't we just use qcs6490-rb3gen2-vision-mezzanine-camx?

kripal-qipl: qcs6490-rb3gen2-vision-mezzanine-camx targets full RB3 Gen2 (QCS6490)
hardware, while Kodiak Lite is a QCS5430 Soft SKU where camera resources
(IFE etc.) are restricted by hypervisor based on license.
Using 6490 DT will lead to mismatch in camera topology and inaccessible
hardware blocks, so a dedicated qcs5430 camx DTBO is required.

ricardosalveti: Can you please add this to the commit message?

kripal-qipl: PR commit message or git commit?

ricardosalveti: Both.

kripal-qipl: Done

ricardosalveti (approved):

conf/machine/rb3gen2-core-kit.conf
lumag, line 23: Why can't the hypervisor use DT_FIXUP to disable DT blocks corresponding to unavailable devices?

lumag (changes requested):

conf/machine/rb3gen2-core-kit.conf
kripal-qipl: camera HW block are enable/disable from camera DT.
In QCS5430 softsku FP1 device, one IFE is less compared to QCS6490. Need to disable it from camera overlay DT

cjha01: > Why can't the hypervisor use DT_FIXUP to disable DT blocks corresponding to unavailable devices?

Here is our understanding — please correct us if we are wrong.

QCS5430 (Kodiak Lite) is a SoftSKU variant of QCS6490 where cam_csid2 and cam_vfe2 (IFE2) are physically present on the silicon but disabled via SoftSKU licensing (FP1/FP2). The SoftSKU Feature Pack is activated via a .pfm license file installed through QWES — this happens in userspace, after the kernel is fully up.
 
DT_FIXUP runs in the hypervisor, before the kernel even starts — long before any SoftSKU license is evaluated. The hypervisor has no visibility into which Feature Pack is installed or active on the device. These two mechanisms operate at completely different boot stages and cannot interact.
 
Therefore, at the time DT_FIXUP executes, there is no way for the hypervisor to know that cam_csid2 and cam_vfe2 should be disabled for this specific SoftSKU configuration. The DTSO overlay is the correct and stage-appropriate solution — it statically captures the intended hardware topology for QCS5430 at compile time,
independent of any runtime license state.

lumag: I assume that the licence is also evaluated before the kernel starts.

lumag: Okay, I missed the point about the licence being submitted from the userspace. Indeed, it's too late. But then we can't really disable the nodes here. How would we know if it is FP1 or FP2?

cjha01: > Okay, I missed the point about the licence being submitted from the userspace. Indeed, it's too late. But then we can't really disable the nodes here. How would we know if it is FP1 or FP2?

FP1 and FP2 differ only in CPU core count and GPU frequency — there is no change in the camera subsystem. Both FP1 and FP2 have identical camera HW topology with cam_vfe2 and cam_csid2 disabled. Therefore, the same camera DTBO used for FP1 can be reused for FP2 without any modification.

**Our Understanding: ** 

At boot time, ABL differentiates between FP1 and FP2 by reading the softsku-id from the SoftSKU EFI Protocol and matching it against the qcom,msm-id field encoded in each DTB (e.g., 0x1023F for FP1 and 0x2023F for FP2 on QCS5430). The board-id does not change between FP1 and FP2 for the same physical board — it only differentiates hardware board subtypes (e.g., Kodiak core kit vs Kodiak vision-mezz).

lumag: So, what is going to happen if tomorrow there is FP3 which disables  one more camera path?

cjha01: > So, what is going to happen if tomorrow there is FP3 which disables one more camera path?

If a future FP (e.g., FP3) introduces a new camera topology difference, our current understanding suggests that adding a new per-FP DTSO overlay would be the natural extension of this model — one additional file with the relevant camera nodes disabled, with zero impact on existing FP overlays.

Regarding DT_FIXUP — we acknowledge we do not have deep expertise in the UEFI/hypervisor layer, but based on our understanding, scaling DT_FIXUP independently for each new Feature Pack would likely require a new UEFI handler per FP transition. Since UEFI is owned by a different team, this would introduce a cross-team dependency every time a new Feature Pack is defined.

The DTSO/DTBO model, in contrast, keeps camera topology changes in the camera/kernel DT layer, owned and maintained entirely by the camera team, resolved at compile time. We believe this is a cleaner and more scalable ownership model — but we are open to guidance from you if there is a better or established way to handle this via DT_FIXUP that we may be unaware of. Please provide an example of DT_FIXUP where a small or partial modification has been implemented. We will check it to better understand the approach.

cjha01: The downstream CamX  overlay for FP1 is already planned and targeted for the GA release. We kindly request your approval to proceed with the current DTSO/DTBO approach for FP1 so we can meet the GA timeline.

In parallel, we genuinely appreciate your suggestion on DT_FIXUP and are very open to exploring it further. Since we have limited expertise in the UEFI/hypervisor layer, could you please share any references, examples, or pointers on how DT_FIXUP can be used to achieve per-FP camera topology differentiation (e.g., disabling cam_csid2 / cam_vfe2)? This will help us evaluate the feasibility and plan the DT_FIXUP based approach for FP2 or a future Feature Pack.

lumag: > The downstream CamX overlay for FP1 is already planned and targeted for the GA release. 

This is not a reason to merge anything. Plans can and should be changed if the solution is incorrect.

Could you please explain, what is the expected course of action for the users if the change the licence? I'd really expect that the only action is to check in the new licence and then maybe reboot.

cjha01: > Could you please explain, what is the expected course of action for the users if the change the licence? I'd really expect that the only action is to check in the new licence and then maybe reboot.

You are absolutely right — the expected and intended user experience is exactly:
 **Install new .pfm license → Reboot → Done** 

This works seamlessly because all per-FP DTBs (FP1, FP2 etc..) are pre-bundled into the FIT image at build time. On reboot, ABL reads the updated softsku-id from the newly installed license and automatically selects the correct per-FP DTB from the image — no reflash, no manual intervention required.

lumag: So, how will it work if FP3 enables or disables some other devices? Would we have to duplciate the entries in the metadata and the FIT image again and again?

lumag: I'm fine with merging this now, but first we would need an ETA and a commitment from the boot team that we can drop these entries. Device enablement must be handled by the UEFI only by using the DT_FIXUP protocol.

cjha01: > I'm fine with merging this now, but first we would need an ETA and a commitment from the boot team that we can drop these entries. Device enablement must be handled by the UEFI only by using the DT_FIXUP protocol.

We completely understand the concern and the long-term vision of having device enablement handled via DT_FIXUP protocol in UEFI. We will initiate the discussion with the boot team at the earliest, share the full context of this thread, and work with them to understand the feasibility, scope, and effort involved. Based on their inputs and assessment, we will come back with a concrete ETA and commitment.

@kripal-qipl — could you please take this forward and connect with the boot team? Please share this PR as a reference so they have the full context of the discussion and understand the ask clearly. Once we have their inputs, we will post the ETA here.

kripal-qipl:
qli-2.0 GA pull-request freeze

conf/machine/rb3gen2-core-kit.conf
kripal-qipl: > I'm fine with merging this now, but first we would need an ETA and a commitment from the boot team that we can drop these entries. Device enablement must be handled by the UEFI only by using the DT_FIXUP protocol.

We are currently evaluating the DT-FIX protocol with the Boot team since the implementation involves bootloader components, which are outside our ownership scope. From the camera side, this is a new requirement and needs alignment with the bootloader changes.
Given the dependency and limited visibility on the bootloader side, we plan to properly enable DT-FIX protocol support in the **upcoming QLI releases** once the approach is finalized.
For the current scope, this PR is a critical blocker for the GA release of Kodiak FP1.

Requesting your approval to proceed with this PR for the current release. We will follow up with required DT-FIX protocol support in the **upcoming QLI releases**.

lumag: Still, waiting for the ETA.

kripal-qipl: Since this is a new area for the camera team and the boot team will not own these changes, the camera team will need to take full ownership of the DT-FIX implementation.
This requires obtaining necessary access permissions, onboarding to the boot codebase and tools, and following the boot team’s code check and merge process.
Considering these dependencies, ramp-up effort, implementation, and validation, the overall ETA is estimated to be ~8 to 10 weeks (21Aug).

lumag: This was discussed today during the GA meeting. Please converge with PdMs regarding the SoftSKU support.

vkraleti:
@kripal-qipl as per guidance from PdM,  SoftSKU support isn't available any more in Kodiak. Hence revoked "qli-2.0 GA pull-request freeze" label.

conf/machine/rb3gen2-core-kit.conf
kripal-qipl: As per discussion with PDM, SoftSKU is removed in QLI 2.0 and 5430 transitions to a HW SKU–based model (default FP1). DT selection is handled via SOC/JTAG ID (Kodiak_lite DT), so SoftSKU infra is no longer required. This PR ensures correct default FP1 behavior and fixes camera functionality issues observed on 5430 boards.

lumag (approved):

PR #2366, opened 2026-06-05
Enable KVM by default on IQ-615-EVK

vkraleti:
All PILs are functioning correctly and KVM can be enabled by default on IQ-615-EVK.
Update FIT_DTB_COMPATIBLE entries and machine configurations to switch to default KVM.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -311,14 +311,6 @@ jobs:
                 type: default
                 dirname: ""
                 yamlfile: ""
-          - machine: iq-615-evk
-            distro:
-                name: qcom-distro-kvm
-                yamlfile: ':ci/qcom-distro-kvm.yml'
-            kernel:
-                type: default
-                dirname: ""
-                yamlfile: ""
           - machine: qcom-armv7a
             distro:
                 name: nodistro
--- a/conf/machine/include/fit-dtb-compatible-linux-qcom.inc
+++ b/conf/machine/include/fit-dtb-compatible-linux-qcom.inc
@@ -125,32 +125,32 @@ FIT_DTB_COMPATIBLE[qcom_qcs8300-adp-camx-staging] = \
     "qcs8300-ride qcs8300-ride-camx monaco-el2 monaco-camx-el2 monaco-staging"
 
 # ---------- talos ----------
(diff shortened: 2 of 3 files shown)


lumag (dismissed):

conf/machine/qcs615-ride.conf
```diff
@@ -4,7 +4,7 @@
 
 require conf/machine/include/qcom-qcs615.inc
 
-MACHINE_FEATURES += "efi pci"
+MACHINE_FEATURES += "efi kvm pci"
```
ricardosalveti: qcs615-ride is not yet using multi-dtb, wouldn't that be an issue?

vkraleti: Without multi-dtb, device is booting but PILs are not up. Similar behavior on qcs8300-ride-sx & qcs9100-ride-sx as well. Should they be reverted back to 'Gunyah'?

ricardosalveti: Either we disable kvm or migrate to multi-dtb, but I remember multi-dtb would also cause these boards to stop booting (due our current hw availability, not using the iot variants), so I guess our only option is to use gunyah by default.

vkraleti: I dropped qcs615-ride changes. I'll revert `kvm` enablement changes for qcs8300-ride-sx & qcs9100-ride-sx in another PR.

lumag: Or enable non-IoT devices as a temporal measure. Should we patch qcom-dtb-metadata inside the meta-qcom?

vkraleti: Maybe send a change to qcom-dtb-metadata itself. This is just a database update. Should not be any objections to accept it.

lumag: @vkraleti we have discussed it several times, I think. The kernel team doesn't want to support non-IoT SKUs.

vkraleti: For now, to avoid functionality breakages on Ride boards, let's switch back to Gunyah. If there is a general agreement for KVM on Ride, can make necessary changes. I'll start internal email regd. this.

sbanerjeeqc (approved):

lumag (approved):

PR #2370, opened 2026-06-05
weston:compositor: skip deferred repaint when no output is actively painting

yashgupta2126:
During MST (Multi-Stream Transport) display testing on RB8, we connected a DP MST dongle with two displays and observed that after an SST→MST transition(plugging in a second monitor while one is already active), weston would enable both MST virtual connectors correctly but one or both displays would remain black. The only recovery was restarting the weston service.

Root cause analysis traced the issue to weston_output_schedule_repaint() — when device recovery is triggered by a spurious HPD udev event for the first MST output, weston_backend_set_deferred() is called. Any newly-enabled second output that calls schedule_repaint() while deferred=true is unconditionally set to REPAINT_DEFERRED. Since no other output is actively painting at that moment, clear_deferred() is never triggered for it, the output never transitions to REPAINT_BEGIN_FROM_IDLE, and the display stays black indefinitely.

The fix checks whether any output on the same backend is actually in an active repaint cycle before deferring. If none are, the newly-enabled output skips the deferred path and enters the repaint loop immediately.

### DIFF
--- a/recipes-graphics/wayland/weston/0001-compositor-skip-deferred-repaint-when-no-output-is-a.patch
+++ b/recipes-graphics/wayland/weston/0001-compositor-skip-deferred-repaint-when-no-output-is-a.patch
@@ -1,0 +1,63 @@
+From 08dac965dd5c6962719828117ae8075065c0fa7f Mon Sep 17 00:00:00 2001
+From: Yash Gupta <yash.gupta@oss.qualcomm.com>
+Date: Fri, 5 Jun 2026 20:15:57 +0530
+Subject: [PATCH] compositor: skip deferred repaint when no output is actively
+ painting
+
+When weston_backend_set_deferred() is called, subsequent calls to
+weston_output_schedule_repaint() unconditionally set the output to
+REPAINT_DEFERRED.  This is correct when another output on the same
+backend is mid-repaint and the deferred state is meaningful.
+
+However, when a newly-enabled output calls schedule_repaint() while
+deferred=true but no other output is in an active repaint cycle
+(BEGIN_FROM_IDLE, SCHEDULED, or AWAITING_COMPLETION), setting
+REPAINT_DEFERRED is incorrect: the output will wait for a
+clear_deferred() that may never come, or may come too late after the
+output's repaint window has passed.
+
+Before setting REPAINT_DEFERRED, check whether any output on the same
+backend is actively painting.  If none are, skip the deferred path and
+proceed directly to REPAINT_BEGIN_FROM_IDLE so the output enters the
+repaint loop immediately.
+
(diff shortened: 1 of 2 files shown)


yashgupta2126:
Already raised merge request in weston gitlab repository

https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/2095

recipes-graphics/wayland/weston/0001-compositor-skip-deferred-repaint-when-no-output-is-a.patch
```diff
@@ -0,0 +1,63 @@
...
+REPAINT_DEFERRED is incorrect: the output will wait for a
+clear_deferred() that may never come, or may come too late after the
+output's repaint window has passed.
+
+Before setting REPAINT_DEFERRED, check whether any output on the same
+backend is actively painting.  If none are, skip the deferred path and
+proceed directly to REPAINT_BEGIN_FROM_IDLE so the output enters the
+repaint loop immediately.
+
+Upstream-Status: Pending
```
lumag: Nope

lumag (changes requested):
Instead of duplicating the message from the patch, the commit message should be explaining, why you are submitting this fix to the meta-qcom layer.

yashgupta2126:
> Instead of duplicating the message from the patch, the commit message should be explaining, why you are submitting this fix to the meta-qcom layer.

Updated the commit message with explanation for this patch

recipes-graphics/wayland/weston/0001-compositor-skip-deferred-repaint-when-no-output-is-a.patch
yashgupta2126: Updated the patch

quic-devanshi:
qli-2.0 GA Critical Fix

lumag:
@quic-devanshi ask somebody from the upstream community review the MR.

quic-devanshi:
> @quic-devanshi ask somebody from the upstream community review the MR.

Added comment in the MR to request review from the maintainers.

quic-devanshi:
> @quic-devanshi ask somebody from the upstream community review the MR.

@lumag , The MR has review from one of the maintainer Marius Vlad. Is it feasible to now merge change in meta-qcom?

lumag:
@quic-devanshi with the promise that you track future possible changes by the maintainers and update the patch if there are any.

lumag (approved):

quic-devanshi:
> @quic-devanshi with the promise that you track future possible changes by the maintainers and update the patch if there are any.

Sure @lumag, will keep a track of upstream change and update the patch in meta-qcom accordingly. Thanks.

PR #2372, opened 2026-06-06
recipes-bsp: u-boot: enable Gunyah exit support for KVM-enabled targets

aswinm94:
Add a new config fragment (gunyah-exit.cfg) to enable Gunyah exit handling and the ARM SoC boot0 hook required for KVM-based boot.

The fragment is conditionally included when MACHINE_FEATURES contains 'kvm'.

### DIFF
--- a/recipes-bsp/u-boot/files/gunyah-exit.cfg
+++ b/recipes-bsp/u-boot/files/gunyah-exit.cfg
@@ -1,0 +1,4 @@
+// Enable Gunyah EL2 exit handling and
+// ARM SoC boot0 hook in QCOM SOCs
+CONFIG_QCOM_EL2_GUNYAH_EXIT_SUPPORT=y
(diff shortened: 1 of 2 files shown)


ricardosalveti:
@b49020 would this be an issue for the open boot fw use case, which already jumps in el2?

aswinm94:
> @b49020 would this be an issue for the open boot fw use case, which already jumps in el2?

@ricardosalveti , these changes were verified by @b49020, when entering U-Boot at EL2 in the open boot firmware scenario.

b49020:
> > @b49020 would this be an issue for the open boot fw use case, which already jumps in el2?
> 
> @ricardosalveti , these changes were verified by @b49020, when entering U-Boot at EL2 in the open boot firmware scenario.

Yeah I have tested this change, it will be a no-op when entry in U-Boot is in EL2 mode. The Gunyah exit SMC call is only invoked when U-Boot enters in EL1 mode.

PR #2374, opened 2026-06-06
Switch to Gunyah based virtualization on Ride platforms

vkraleti:
Ride platforms lack multi-DTB support which needed for EL2 DTBO loading. Without EL2 DTBO, remoteproc won't be functional with KVM. Switch to Gunyah on Ride platforms until multi-DTB support is available.


lumag (approved):

lumag:
Merging without waitign for the second review to unblock the CI. All tests pass.

PR #2383, opened 2026-06-07
workflows: fix PR test branch detection for forked PRs

ricardosalveti:
github.event.workflow_run.pull_requests is empty for pull requests opened from forked repositories, so determine-target-branch always fell through to the "master" default. As nearly all PRs come from forks, PRs targeting wrynose ended up running the master test plan while the wrynose plan was skipped, so the branch-specific routing introduced for the two test tracks never worked for fork PRs.

Read the target branch from the Event File artifact uploaded by the build instead. It is the literal pull_request event that triggered the build, so it resolves the target branch reliably, forks included. Add the actions: read permission needed to download that artifact from the build run.

### DIFF
--- a/.github/workflows/test-pr-wrynose.yml
+++ b/.github/workflows/test-pr-wrynose.yml
@@ -13,6 +13,7 @@ permissions:
   pull-requests: write
   contents: read
   packages: read
+  actions: read
 
 jobs:
   determine-target-branch:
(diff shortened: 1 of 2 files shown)


ricardosalveti:
Validated with forked repos against master and wrynose on stg, worked as expected.

ricardosalveti:
@mwasilew @lumag should we merge and push validation further? I canceled the CI as there is no need for running it.

koenkooi (approved):

vkraleti (approved):

ricardosalveti:
Merging and will follow up in case of issues.

PR #2408, opened 2026-06-09
os-release: record GitHub build/run id in BUILD_ID

ricardosalveti:
Embed the CI run identifier in the image's /etc/os-release BUILD_ID field so an installed image can be traced back to the workflow run that produced it. The value matches the S3 artifact folder name (<run_id>-<run_attempt>), mapping an image directly to its artifacts.

ci/base.yml enables the field for every build and sets a clearly marked local default (BUILD_ID ?= "local-${DATETIME}"). The compile workflow generates ci/build-id.yml at build time with a hard BUILD_ID = "<run_id>-<run_attempt>" and appends it to KAS_YAMLS, so CI builds carry the run id while local builds keep the "local-" marker.

The value is baked into a generated kas fragment rather than passed via the environment because kas-container forwards only a fixed env allowlist into the container, so GITHUB_RUN_ID cannot reach bitbake otherwise. The local default uses a weak assignment so the CI override wins regardless of how kas orders the merged local.conf fragments.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -63,16 +63,29 @@ jobs:
           echo "DL_DIR=${INPUTS_CACHE_DIR}/downloads" >> $GITHUB_ENV
           echo "SSTATE_DIR=${INPUTS_CACHE_DIR}/sstate-cache-$(date '+%Y-%m')" >> $GITHUB_ENV
           echo "KAS_WORK_DIR=${PWD}/../kas" >> $GITHUB_ENV
-          echo "KAS_YAMLS=ci/ci.yml:ci/${INPUTS_MACHINE}.yml${INPUTS_DISTRO_YAML}${INPUTS_KERNEL_YAML}" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


mwasilew (approved):
This can be used in a LAVA test job to verify that the expected image is actually running.

lumag (approved):

quaresmajose (approved):
LGTM

However, I think this should be excluded from the initramfs images. I intend to save them in the sstate-cache, but I won't be doing it tomorrow. I'll see what to do at the time.

ricardosalveti:
> LGTM
> 
> However, I think this should be excluded from the initramfs images. I intend to save them in the sstate-cache, but I won't be doing it tomorrow. I'll see what to do at the time.

Our initrd images are also getting used externally (kernel related CI), so it is actually useful to have it there from that perspective.

quaresmajose:
> > LGTM
> > However, I think this should be excluded from the initramfs images. I intend to save them in the sstate-cache, but I won't be doing it tomorrow. I'll see what to do at the time.
> 
> Our initrd images are also getting used externally (kernel related CI), so it is actually useful to have it there from that perspective.

If this information is truly necessary, I think it should be deployed in a new artifact simply for that purpose. That way we could reuse the initramfs which are small but take a considerable amount of time to build, but I'll leave the pros and cons for later.

koenkooi (approved):

PR #2410, opened 2026-06-09
ci: pin kas-container to 5.3 and pull image from internal ECR mirror

ricardosalveti:
Fetch the kas-container script from the pinned 5.3 tag instead of the latest siemens/kas tag, and set KAS_CONTAINER_IMAGE so all kas jobs pull the container image from the internal ECR mirror at 418295714268.dkr.ecr.us-west-2.amazonaws.com/ghcr.io/kas-image:5.3.

All kas-container invocations run on the self-hosted runners, which are the only hosts with access to the mirror, so no GitHub-hosted job is affected. Pinning the script to the same version as the image avoids script/image skew when a new kas version is tagged upstream. The env var is set in both build-yocto.yml and compile.yml because env blocks do not propagate across workflow_call.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -28,6 +28,11 @@ concurrency:
 env:
   CACHE_DIR: /efs/qli/meta-qcom
   KAS_CLONE_DEPTH: 1
+  # Internal mirror of ghcr.io/siemens/kas/kas, only reachable from the
+  # self-hosted runners. Keep the image tag in sync with KAS_VERSION
+  # (kas-container script version) and with compile.yml.
+  KAS_VERSION: "5.3"
+  KAS_CONTAINER_IMAGE: 418295714268.dkr.ecr.us-west-2.amazonaws.com/ghcr.io/kas-image:5.3
 
 jobs:
   kas-setup:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-yocto.yml
```diff
@@ -38,8 +43,7 @@ jobs:
         run: |
           KAS_CONTAINER=$RUNNER_TEMP/kas-container
           echo "KAS_CONTAINER=$KAS_CONTAINER" >> $GITHUB_ENV
-          LATEST=$(git ls-remote --tags --refs --sort="v:refname" https://github.com/siemens/kas | tail -n1 | sed 's/.*\///')
-          wget -qO ${KAS_CONTAINER} https://raw.githubusercontent.com/siemens/kas/refs/tags/$LATEST/kas-container
+          wget -qO ${KAS_CONTAINER} https://raw.githubusercontent.com/siemens/kas/refs/tags/${KAS_VERSION}/kas-container
```
quaresmajose, line 46: You can strip the version from KAS_CONTAINER_IMAGE and so that's one less variable.

quaresmajose: Proposed the extraction of kas-container from the image in https://github.com/qualcomm-linux/meta-qcom/pull/2934.

quaresmajose:
@ricardosalveti can you please rebase this.

PR #2453, opened 2026-06-15
systemd-boot: align kernel image allocation to 64kb for arm64

vkraleti:
On arm64 systems using 64 KB pages, the kernel requires 64 KB-aligned load addresses. The current systemd-boot loader guarantees only 4 KB alignment, which can result in misaligned allocations on fragmented UEFI memory maps. This causes the kernel to abort at entry.

Bring in a proposed change[1] to honor PE SectionAlignment and allocate properly aligned memory, ensuring reliable kernel loading.

[1] https://github.com/systemd/systemd/pull/42463

### DIFF
--- a/recipes-core/systemd/systemd-boot_%.bbappend
+++ b/recipes-core/systemd/systemd-boot_%.bbappend
@@ -1,0 +1,5 @@
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/systemd:"
+
+SRC_URI:append:qcom = " \
+    file://0001-boot-stub-honor-PE-SectionAlignment-when-loading-inn.patch \
+"
--- a/recipes-core/systemd/systemd/0001-boot-stub-honor-PE-SectionAlignment-when-loading-inn.patch
+++ b/recipes-core/systemd/systemd/0001-boot-stub-honor-PE-SectionAlignment-when-loading-inn.patch
@@ -1,0 +1,167 @@
+From 2d6a7100adc2ab4a239ac31f094c1037c1a99d98 Mon Sep 17 00:00:00 2001
+From: Josh Hoffer <jhoffer@anduril.com>
+Date: Wed, 3 Jun 2026 17:19:33 -0700
+Subject: [PATCH] boot/stub: honor PE SectionAlignment when loading inner
+ kernel
+
+The stub copies the embedded kernel PE into an xmalloc_pages()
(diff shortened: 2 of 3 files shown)


lumag (dismissed):

ricardosalveti:
This should ideally land in oe-core instead.

ricardosalveti:
> This causes the kernel to abort at entry.

Also, is this really true? Doesn't affect the current builds we have.

vkraleti:
> This should ideally land in oe-core instead.

Sure, The original PR https://github.com/systemd/systemd/pull/42463 is still under review. Once it concludes I'll send a patch to oe-core master and wrynose. Since this is a GA blocker I request to accept here to unblock testing.

vkraleti:
> > This causes the kernel to abort at entry.
> 
> Also, is this really true? Doesn't affect the current builds we have.

Yes, this is reproducible on a few devices in the test bed. They are failing to boot with 64-bit alignment abort. Added you to the internal email where more details re shared.

ricardosalveti:
Please refresh to the latest revision of the patch from https://github.com/systemd/systemd/pull/42463.

ricardosalveti:
Commit message is also incorrect, you are not backporting the patch as it was not even merged in systemd upstream yet.

recipes-core/systemd/systemd-boot_%.bbappend
```diff
@@ -0,0 +1,5 @@
+FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/${PN}:"
+
+SRC_URI:append:qcom = " \
+    file://0001-boot-stub-honor-PE-SectionAlignment-when-loading-inn.patch \
```
ricardosalveti, line 4: I believe we will also have to patch systemd as well, since it also generates and installs systemd-boot at usr/lib/systemd/boot/efi/systemd-bootaa64.efi (which can also be updated automatically by systemd on boot) - yeah, an upstream recipe issue.

Having on both would avoid potential issues on image updates later on.

vkraleti: Done, applied same patch in both recipes.

vkraleti:
> Commit message is also incorrect, you are not backporting the patch as it was not even merged in systemd upstream yet.

Reworded commit message a bit to indicate this is a praposed change.

ricardosalveti (approved):

koenkooi (approved):

PR #2459, opened 2026-06-15
userspace-resource-manager-extensions: Update to 0.3.1

kartnema:


ricardosalveti:
Fails to build, converting to draft.

kartnema:
@ricardosalveti 
urm Extensions has a build dependency on URM, the issue is related to a missing header, which is being installed by the main urm recipe (from v0.4.1 onwards, for which I created a PR yesterday as well parallelly). We think the issue is happening because urm Extensions build was run with URM v0.4 and not v0.4.1

Since the URM recipe is merged now (v0.4.1), re-triggering the urm-extensions failing workflow should resolve the issue.
I have verified this locally.

lumag:
> @ricardosalveti urm Extensions has a build dependency on URM, the issue is related to a missing header, which is being installed by the main urm recipe (from v0.4.1 onwards, for which I created a PR yesterday as well parallelly). We think the issue is happening because urm Extensions build was run with URM v0.4 and not v0.4.1
> 
> Since the URM recipe is merged now (v0.4.1), re-triggering the urm-extensions failing workflow should resolve the issue. I have verified this locally.

Please fix the URM build scripts to check for the dependencies (and their versions).

lumag (approved):

ricardosalveti (approved):

PR #2461, opened 2026-06-15
ci: base.lock: update layers to latest 

ricardosalveti:
Update lock to the latest of every repository we depend on.

Draft as it is not yet built for every matrix combination we have.


koenkooi (approved):
Now we can race with Bruce for the kernel config fixes :)

lumag (approved):

quaresmajose (approved):

PR #2466, opened 2026-06-15
workflows/compile: fix deploy_dir artifacts for publishing

### LINKED ISSUE #1984: Skip populating SDK for non-nightly builds
**lumag:** _Should we skip populating SDK for non-nightly builds?_

**quaresmajose:** _Yes, we talked about this somewhere. But I'd prefer to do this without adding new imput entries to the action. If it were possible to know that we were running the nightly in another way, for example using/reading an environment variable. I would prefer to investigate how to do this and I propose in a new PR._

From [ci: actions/compile: disable SDK input control](https://github.com/qualcomm-linux/meta-qcom/pull/1601#top)#1601

quaresmajose:
To improve debug of the staging artifacts step:
- trace bash commands
- run cp with verbose

The (sdk_dir) path depends on component (deploy_dir) path, and
therefore it doesn't exist when the (deploy_dir) also doesn't exist.
However, this isn't true because the (sdk_dir) can exist even when the
(deploy_dir) path doesn't exist.

Set (deploy_dir) to the real bitbake deploy and add a new (images_dirs)
and (sdk_dir) and these without dependencies on each other.

Partially fix https://github.com/qualcomm-linux/meta-qcom/issues/1984

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -165,6 +165,7 @@ jobs:
       - name: Stage build artifacts for publishing
         shell: bash
         run: |
+          set -x
           # The upload-private-artifact-action runs from a container that
           # expects file to be relative to our PWD. deploy_dir is outside
           # that, so we move things around:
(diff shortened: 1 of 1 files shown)


.github/workflows/compile.yml
```diff
@@ -170,6 +170,7 @@ jobs:
           # expects file to be relative to our PWD. deploy_dir is outside
           # that, so we move things around:
           deploy_dir=../kas/build/tmp/deploy/images/${INPUTS_MACHINE}
+          sdk_dir=$(readlink -m $deploy_dir/../../sdk)
```
lumag: Should this be a raw path, `sdk_dir=../kas/build/tmp/deploy/sdk`?

lumag (changes requested):

.github/workflows/compile.yml
quaresmajose: I already have a new version like this where I added a new "$images_dir". Squashed and pushed.

lumag (approved):

koenkooi (approved):

ricardosalveti (approved):

PR #2479, opened 2026-06-16
rpmsgexport: add recipe for rpmsgexport

quic-vishsant:
Add a recipe for rpmsgexport [1], an upstream utility that creates rpmsg endpoint character devices by issuing RPMSG_CREATE_EPT_IOCTL on the rpmsg control device (/dev/rpmsg_ctrl*).

The recipe uses the meson build system, which properly inherits toolchain CFLAGS and LDFLAGS from the environment.

[1] https://github.com/linux-msm/rpmsgexport

### DIFF
--- a/recipes-support/rpmsg-export/rpmsgexport_git.bb
+++ b/recipes-support/rpmsg-export/rpmsgexport_git.bb
@@ -1,0 +1,11 @@
+SUMMARY = "Qualcomm RPMSGEXPORT application"
+HOMEPAGE = "https://github.com/linux-msm/rpmsgexport.git"
+SECTION = "devel"
+
+LICENSE = "BSD-3-Clause"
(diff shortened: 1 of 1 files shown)


lumag (changes requested):
> The utility is needed to create rpmsg
endpoint devices when remote processors boot.

It's definitely not essential to boot. And ideally there should be a kernel driver providing correct uAPI instead of exporting raw control device. The tool can go in, the enablement is very questionable. Please drop it.

quic-vishsant:
> > The utility is needed to create rpmsg
> > endpoint devices when remote processors boot.
> 
> It's definitely not essential to boot. And ideally there should be a kernel driver providing correct uAPI instead of exporting raw control device. The tool can go in, the enablement is very questionable. Please drop it.

Dropped the packagegroup-qcom commit as suggested.

lumag (approved):

ricardosalveti (approved):

PR #2450, opened 2026-06-13
shikra-evk: add initial changes for shikra-evk board

vivpuar:
Adding below changes for shikra-evk Board:

- machine configuration
- KAS configuration file
- firmware packagegroup for shikra
- boot-firmware recipe for shikra

**NOTE**: This Shikra target PR will be for sample engineering, it won't be of the GA quality as we will have for the other targets.

PR #2485, opened 2026-06-16
iris-video-dlkm: shikra: load qcom-iris instead of iris_vpu

gouravk-qualcomm:
On Shikra targets, the iris_vpu out-of-tree DLKM is not supported.
When iris-video-dlkm is installed (for example, in qcom-multimedia-proprietary-image),
the default update-alternatives priority activates `blacklist-video.conf.venus`,
which results in `qcom_iris` being blacklisted and prevents the in-kernel driver
from loading and functioning correctly.

Raise the update-alternatives priority of the `bl-vidc` sub-package to 150 for Shikra,
ensuring it takes precedence over `bl-venus` and becomes the active alternative.
This activates `blacklist-video.conf.vidc` , effectively blacklisting `iris_vpu` while
allowing `qcom_iris` to load and operate as expected on Shikra targets.


ricardosalveti (approved):

koenkooi:
Do we want to get this in before #2450 ?

koenkooi (approved):

gouravk-qualcomm:
> Do we want to get this in before #2450 ?

@koenkooi   Yes, this change can go independently.

lumag (approved):

PR #2497, opened 2026-06-17
qwes service: upgrade 1.0 -> 1.1

maniSankarJ:
The QWES daemon currently attempts to load the QWES TA from obsolete or legacy firmware locations. This change updates the TA loading logic to check for machine-specific paths under /lib/qtee-tas fo the TAs.
This ensures the daemon can correctly locate TA binaries.

Changes Made:
Updated QWES daemon TA search logic
Added support for machine-specific TA paths under /lib/qtee-tas
Upgraded the source tarball to include the latest path resolution changes

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.1.bb
+++ b/dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.1.bb
@@ -6,10 +6,10 @@ the store and optionally load QcWES TA to provide these features."
 LICENSE = "LICENSE.qcom-2"
 LIC_FILES_CHKSUM = "file://usr/share/doc/${BPN}/LICENSE.qcom-2;md5=165287851294f2fb8ac8cbc5e24b02b0"
 
-PBT_BUILD_DATE = "260409"
+PBT_BUILD_DATE = "260620"
 
-SRC_URI = "https://softwarecenter.qualcomm.com/nexus/generic/software/chip/component/sec-userspace.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qwes_${PV}_armv8a.tar.gz"
-SRC_URI[sha256sum] = "2fc38e033aa32e59a1749017526d29e9ad71f38f168ff652383eb06a05be3985"
+SRC_URI = "https://softwarecenter.qualcomm.com/nexus/generic/software/chip/component/sec-userspace.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qwes_1.0_armv8a.tar.gz"
+SRC_URI[sha256sum] = "7bb20daa916e4b13e54948ab90af92e9c2e6129cc22da1a23b924adb3f0610d4"
 
 S = "${UNPACKDIR}"
 
@@ -22,16 +22,24 @@ DEPENDS += "curl minkipc qmi-framework glibc"
 COMPATIBLE_MACHINE = "^$"
 COMPATIBLE_MACHINE:aarch64 = "(.*)"
 
+PACKAGES += "${PN}-ta"
+
 SYSTEMD_SERVICE:${PN} = "qwesd.service"
 
 do_install() {
     install -d ${D}${bindir}
     install -d ${D}${systemd_system_unitdir}
     install -d ${D}${docdir}/${BPN}
+    install -d ${D}${nonarch_base_libdir}/qtee-tas
 
     # Install binaries
     install -m 0755 ${S}/usr/bin/* ${D}${bindir}/
     install -m 0644 ${S}/usr/lib/systemd/system/qwesd.service ${D}${systemd_system_unitdir}/qwesd.service
     install -m 0644 ${S}/usr/share/doc/${BPN}/NOTICE.txt ${D}${docdir}/${BPN}
     install -m 0644 ${S}/usr/share/doc/${BPN}/LICENSE.qcom-2 ${D}${docdir}/${BPN}
+    cp -R ${S}/lib/qtee-tas/* ${D}${nonarch_base_libdir}/qtee-tas/
 }
+
+FILES:${PN}-ta += "${nonarch_base_libdir}/qtee-tas"
+RDEPENDS:${PN} = "${PN}-ta"
+INSANE_SKIP:${PN}-ta += "arch"


maniSankarJ:
qli-2.0 GA Critical Fix

dynamic-layers/openembedded-layer/recipes-security/qwes/qwes_1.1.bb
```diff
@@ -6,10 +6,10 @@ the store and optionally load QcWES TA to provide these features."
 LICENSE = "LICENSE.qcom-2"
 LIC_FILES_CHKSUM = "file://usr/share/doc/${BPN}/LICENSE.qcom-2;md5=165287851294f2fb8ac8cbc5e24b02b0"
 
-PBT_BUILD_DATE = "260409"
+PBT_BUILD_DATE = "260616.1"
 
-SRC_URI = "https://softwarecenter.qualcomm.com/nexus/generic/software/chip/component/sec-userspace.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qwes_${PV}_armv8a.tar.gz"
-SRC_URI[sha256sum] = "2fc38e033aa32e59a1749017526d29e9ad71f38f168ff652383eb06a05be3985"
+SRC_URI = "https://softwarecenter.qualcomm.com/nexus/generic/software/chip/component/sec-userspace.qclinux.0.0/${PBT_BUILD_DATE}/prebuilt_yocto/qwes_1.0_armv8a.tar.gz"
```
ricardosalveti, line 11: Why is PV 1.1 and here it is 1.0?

maniSankarJ: The version mismatch is due to a limitation in our internal tarball generation pipeline.
Although the source content has been updated with fixes, the pipeline did not increment the version as expected and generated a tarball with the same version (1.0). As the tarball published automatically after generation and it is not overwriting the previous tarball (${PBT_BUILD_DATE} is changing) we want to move forward with the updated source.
I have raised the concern internally to fix the versioning.

ricardosalveti: OK, please track it internally to get this fixed for the next update, thanks!

ricardosalveti (dismissed):

lumag (commented):
> Added support for machine-specific TA paths under /lib/qtee-tas

How does it detect the machine path? E.g. what would be the path for the X13s laptop?

maniSankarJ:
> > Added support for machine-specific TA paths under /lib/qtee-tas
> 
> How does it detect the machine path? E.g. what would be the path for the X13s laptop?

It reads the data from file "/sys/firmware/devicetree/base/compatible" which some machine specific(SOC/chipset specific information) ex: qcom,purwa-iot-evkqcom,purwa-iot-somqcom,x1p42100
Based on this information, it determines the machine name information and checks for the folder under "/lib/qtee-tas"

lumag:
> > > Added support for machine-specific TA paths under /lib/qtee-tas
> > 
> > 
> > How does it detect the machine path? E.g. what would be the path for the X13s laptop?
> 
> It reads the data from file "/sys/firmware/devicetree/base/compatible" which some machine specific(SOC/chipset specific information) ex: qcom,purwa-iot-evkqcom,purwa-iot-somqcom,x1p42100 Based on this information, it determines the machine name information and checks for the folder under "/lib/qtee-tas"

So, which path would be used on X13s laptop if I run QWES service on it?

lumag:
Or, I'd ask it differntly: do we have all the mapping in the QWES service itself? How would it work if the customer fuseds their CA PK in the fuses?

maniSankarJ:
We do not keep the mapping inside QWES. The path is found at runtime from the information that we get from the file by parsing the data in the file to extract the chipset names like purwa, qcm6490, qcs9100 etc. So, we look under "/lib/qtee-tas/purwa", "/lib/qtee-tas/qcm6490" and "/lib/qtee-tas/qcs9100".
For you scenario, as the information is dependent on the chipset the laptop is using if the device uses glymur, we will check under "/lib/qtee-tas/glymur".
Though customer fused their PK hash the chipset names will be the same.

lumag:
> For you scenario, as the information is dependent on the chipset the laptop is using if the device uses glymur, we will check under "/lib/qtee-tas/glymur".

Well, I asked about X13s especially because it is a fused device, produced by Lenovo. So ideally the path should be `/lib/qtee-tas/sc8280xp/LENOVO/21BX/` (similar to firmware). We have been through the similar questions when defining the firmware paths. You can't just use the device name without vendor or SoC part in it (there are devices with the same name, but produced by different vendors and for the same or for different SoCs).

maniSankarJ:
The TA loading logic here is for loading QWES TA qwes.mbn. It is qualcomm's TA which we are planning to keep them at paths like "/lib/qtee-tas/glymur".
In case, it is not clear this logic is to load QWES TA which we will push into the paths like "/lib/qtee-tas/glymur" and not applicable to loading TAs in general.

lumag:
> The TA loading logic here is for loading QWES TA qwes.mbn.

Why would qwes.mbn loading logic differentiate from the logic to load other TAs (BTW: why is it a part of the QWES service rather than one of minkipc libs?).

maniSankarJ:
I meant that this logic is tailored to enable our service to load the QWES TA's expected path on device. It will applicable to any such TAs only in the case where they are expected to be at the same path as qwes.mbn
As per guidance from the internal target team, it is deployed under SoC-specific paths:
/lib/qtee-tas/<soc-name>/

The current logic follows this deployment model. At runtime, we extract the SoC name from /sys/firmware/devicetree/base/compatible and construct the lookup path accordingly (e.g., /lib/qtee-tas/sc8280xp).
This is intentional because the QWES TA is SoC-specific and not vendor/device-specific, and is expected to be identical across devices (including fused devices) using the same chipset. Therefore, we do not include vendor/product-level paths (e.g., LENOVO/21BX) or maintain any static mappings in the service.
Also, this logic is scoped only to loading the QWES TA from this service as it is the only TA that this service needs and loads for the current requirement.
If the deployment model changes in the future (e.g., vendor/device-specific placement), the lookup logic can be extended accordingly.

And for your other query, it is agreed with minkipc team that TA owners need to take the responsibility of loading the TAs. In this case QWES service loads qwes.mbn as it needs the TA for its TEE functionality

qcom-anilyada:
Failures Observed: 

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>


<div style='direction:ltr'>


Workflow | Targets | Testcases | Error Logs
-- | -- | -- | --
[27711265107](https://github.com/qualcomm-linux/meta-qcom/actions/runs/27711265107) | iq-8275-evk     qcs6490     qcs9100-rb8 | BT_FW_KMD_Service | No   /sys/class/bluetooth/hci* found (HCI not up).
  |   | BT_SCAN | bluetoothd not   detected after 3 attempts.
  |   | BT_ON_OFF | Bluetooth daemon   not detected after 3 attempts.
  |   | WiFi_Firmware_Driver | FAIL - WiFi   firmware/driver validation encountered errors.
  |   | adsp_remoteproc | adsp present in DT   but no /sys/class/remoteproc entry found
  |   | cdsp_remoteproc | cdsp present in DT   but no /sys/class/remoteproc entry found
  |   | Weston_Runtime_Preflight | weston.service is   in failed state and passive runtime recovery did not make Weston usable
  |   | weston-simple-egl | No Wayland socket   found and not starting Weston on base build, failing weston-simple-egl.
  |   | weston-simple-shm | weston.service is   in failed state and passive runtime recovery did not make Weston usable



</div>


</body>

</html>

maniSankarJ:
@qcom-anilyada , these errors are not related to the change that I am promoting. We dont have any usecases where we are using these features.

lumag:
@maniSankarJ I've just checked, Lenovo ThinkPad T14s uses vendor-signed TAs. So, no, vendor-fused devices don't necessarily use the same binaries.

```
$ ~/Projects/Qcomm/Tools/sig-tools/target/debug/mbn-verify hdcp1.mbn  | grep Cert
Certificate: L=Yokohama,OU=Lenovo Japan Ltd.,O=Lenovo Ltd.,CN=Qualcomm Secure Boot Root CA,ST=Kanagawa,C=JP
    Certificate: L=Yokohama,OU=Lenovo Japan Ltd.,O=Lenovo Ltd.,ST=Kanagawa,C=JP,CN=Qualcomm Secure Boot Attestation CA N42
        Certificate: OU=Lenovo Japan Ltd.,L=Yokohama,O=Lenovo Ltd.,ST=Kanagawa,C=JP,CN=One Time Use Attestation Certificate
```

qcom-anilyada:
[Workflow ](https://github.com/qualcomm-linux/meta-qcom/actions/runs/27777656851)- [Failed Job](https://lava.infra.foundries.io/scheduler/job/278541): qcom-distro: glymur-crd: [Boot] Kernel panic - not syncing

arakshit011:
@ricardosalveti non of the tests failures are due to qwes update. Can you please help check?

qcom-anilyada:
Update on the Failures:
- [qcom-distro_linux-qcom-6.18][qcs8300][cdsp_remoteproc]: PASSED (showing pass due to malformed string)
- [qcom-distro_linux-qcom-6.18][qcs6490][boot]: [PASSED ](https://lava.infra.foundries.io/scheduler/job/286193) on rerunning manually [FAILED_JOB](https://lava.infra.foundries.io/scheduler/job/285342) in workflow 
             Error logs:
             `Kernel panic - not syncing: Asynchronous SError Interrupt`

lumag:
> @ricardosalveti non of the tests failures are due to qwes update. Can you please help check?

Please sort out TA loading.

maniSankarJ:
> @maniSankarJ I've just checked, Lenovo ThinkPad T14s uses vendor-signed TAs. So, no, vendor-fused devices don't necessarily use the same binaries.
> 
> ```
> $ ~/Projects/Qcomm/Tools/sig-tools/target/debug/mbn-verify hdcp1.mbn  | grep Cert
> Certificate: L=Yokohama,OU=Lenovo Japan Ltd.,O=Lenovo Ltd.,CN=Qualcomm Secure Boot Root CA,ST=Kanagawa,C=JP
>     Certificate: L=Yokohama,OU=Lenovo Japan Ltd.,O=Lenovo Ltd.,ST=Kanagawa,C=JP,CN=Qualcomm Secure Boot Attestation CA N42
>         Certificate: OU=Lenovo Japan Ltd.,L=Yokohama,O=Lenovo Ltd.,ST=Kanagawa,C=JP,CN=One Time Use Attestation Certificate
> ```

@lumag I have discussed about the signing TAs for IOT and compute platforms. There is a process called double-signing where Qualcomm provides the images that it signed with it's own keys. So, we may find images that signed by OEMs after Qualcomm signed them first.
QWES TA is always expected to be qualcomm signed(QTI) TA. OEM signing is not mandatory for such TAs. Even if OEMs signs the expectation is that the binaires are to be maintained at the same location "/lib/qtee-tas/<machine>" We are also updating our documentation to guide the OEMs not to change the location of the QWES binaries. 
While loading the QWES TA, the internal verification will validate the qti signing only. So, fusing the OEM PK doesn't have any effect on loading the QWES TA image.

lumag:
Will TZ accept those binaries?

lumag:
Note, the binaries which I quoted are not double-signed. They don't have Qualcomm signature at all.

maniSankarJ:
Yes these binaries are accepted by TZ.
Qualcomm internal TAs are expected to have Qualcomm signing. But, if they are OEM TAs, then we will have OEM signing only.
The images that you found may be OEM TAs.
QWES TA is qualcomm's internal TA. So, the image will be loaded and TZ will be able to verify.
These are the general guidlines, I got.
Qualcomm-internal TAs                       ->  Typically Qualcomm-signed only (single QC signature)
OEM TAs                                              -> OEM-signed
TAs requiring both QC + OEM trust    -> Double-signed

lumag (approved):
OK. Thanks for all explanations. For the future relase please converge with the mink and use a standard implementation to load the TA apps.

maniSankarJ:
@ricardosalveti could you please provide reviews again?

vkraleti (approved):

PR #2517, opened 2026-06-18
minkipc: Update tag to v1.2.8 to bring-in RPMB updates

harshaldev27:
This release brings the following updates:
- RPMB listener now detects and supports RPMB partition on eMMC storage.
- rpmb_client is now available to allow provisioning of RPMB keys.
- Fix for PKCS#11 regression caused by oe-core update.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.8.bb
+++ b/dynamic-layers/openembedded-layer/recipes-security/minkipc/minkipc_1.2.8.bb
@@ -23,7 +23,7 @@ LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2b1366ebba1ebd9ae25ad19626bbca93 \
                     file://optee-client/optee_client/LICENSE;md5=69663ab153298557a59c67a60a743e5b \
                     file://optee-test/optee_test/LICENSE.md;md5=daa2bcccc666345ab8940aab1315a4fa"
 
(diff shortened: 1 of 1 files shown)


lumag:
> Fix for PKCS#11 regression caused by oe-core update.

Please provide more information about this.

lumag (approved):

lumag:
Please don't open new PRs if there was an existing one. If you closed it by mistake you can usually ask it to be reopened.

harshaldev27:
> > Fix for PKCS#11 regression caused by oe-core update.
> 
> Please provide more information about this.

oe-core update brought some changes to the tool-chain linker. It appears that one of those was for the linker to give priority to libc symbols when looking for external symbols to link against. Instead of linking against the `recv` symbol defined in syscall.S, it linked against `recv` from libc. This caused a segmentation fault during program execution when libminkadaptor attempted to make a system call to the kernel via syscall.S.

Do let me know if I should update the commit message to explain this as well.

harshaldev27:
> Please don't open new PRs if there was an existing one. If you closed it by mistake you can usually ask it to be reopened.

I actually realized that I can re-open a closed PR by adding a comment. I will ensure this doesn't happen again. 😄

lumag:
> Do let me know if I should update the commit message to explain this as well.

No need to. It's captured here.

ricardosalveti (approved):

Issue #2457, opened 2026-06-15
[wrynose] meta-lts-mixin dependency broke Yocto autobuilder builds


ycongal-smile:
https://github.com/qualcomm-linux/meta-qcom/commit/bfb98caebbc3b961357a33c013cc7e672b7bf2d3 added a dependency on `meta-lts-mixin` which is not currently available on autobuilder so the build fails:
see for example: https://autobuilder.yoctoproject.org/valkyrie/#/builders/113/builds/927

pbrkr:
I would suggest making this an optional dependency. Note that if this is a required dependency then the linux-firmware LTS mixin layer will also need to be Yocto Project Compatible to preserve that status for meta-qcom. That means that the updated linux-firmware recipe should not be update by default and should require PREFERRED_VERSION to be set to enable it (this can be done via a config fragment in the layer that can be enabled to set PREFERRED_VERSION to the latest linux-firmware version).

lumag:
I'll take a look at it.

ndechesne:
Ouch. We should have caught that in our CI, check-layer should have failed for us. We need to understand why it was not the case.

lumag:
@ndechesne it didn't fail because a97b7fdbff385e7c2fac4d1aa2c7e12816aea068 added the dependency to the check-layer invocation.

ndechesne:
> [@ndechesne](https://github.com/ndechesne) it didn't fail because [a97b7fd](https://github.com/qualcomm-linux/meta-qcom/commit/a97b7fdbff385e7c2fac4d1aa2c7e12816aea068) added the dependency to the check-layer invocation.

yeah. i noticed too. but meta-qcom cannot depend on the mixin layer, we should move the dependency to meta-qcom-distro?

lumag:
> but meta-qcom cannot depend on the mixin layer, we should move the dependency to meta-qcom-distro?

The biggest issue would be the platform bringup. We wouldn't be able to depend on the firmware being available from linux-firmware. Also it would require us to support both older and newer sets of firmware, spreading the resources and increasing the pressure on the CI.

I'm slowly leaning towards keeping the firmware as a required layer and making it Yocto Project Compliant.

@pbrkr if we fix the layer to be compliant, what would it require from us to make it being used for autobuilder? Would that be possible at all?

ndechesne:
> > but meta-qcom cannot depend on the mixin layer, we should move the dependency to meta-qcom-distro?
> 
[...]

PR #2542, opened 2026-06-19
Update the meta-lts-mixins layer

quaresmajose:
In order to make the meta-lts-mixins/wrynose/linux-firmware layer
Yocto Project Compatible, the layer no more provide linux-firmware
by default unless that is explicitly stated in the variable DEFAULT_PREFERENCE.
The default setting in mxin layer is now DEFAULT_PREFERENCE to "-1".

Relevant changes for meta-lts-mixins:
- 6ad95d0 linux-firmware: upgrade 20260519 -> 20260622
- d4a9842 drop LTS_LINUX_FIRMWARE_MIXIN_DEFAULT_PREFERENCE and use just DEFAULT_PREFERENCE
- 7d003ac README: add LTS_LINUX_FIRMWARE_MIXIN_DEFAULT_PREFERENCE
- 2145371 linux-firmware: add LTS_LINUX_FIRMWARE_MIXIN_DEFAULT_PREFERENCE
- 6b8dd8b layer.conf: add LTS_LINUX_FIRMWARE_MIXIN_DEFAULT_PREFERENCE
- 11dcf7d SECURITY.md: describe security policy and pointers
- 9cab569 linux-firmware: set DEFAULT_PREFERENCE to "-1"
- b9d53c1 layer.conf: use the same priority as OE-Core

Fix https://github.com/qualcomm-linux/meta-qcom/issues/2457

### DIFF
--- a/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
+++ b/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
@@ -1,4 +1,8 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+# use latest version of linux-firmware from the lts-linux-firmware-mixin layer
+DEFAULT_PREFERENCE:layer-lts-linux-firmware-mixin:qcom = "1"
+
 # To make the layer pass yocto-check-layer only inherit update-alternatives when building for qualcomm
 ALTERNATIVES_CLASS = ""
 ALTERNATIVES_CLASS:qcom = "update-alternatives"
--- a/ci/base.lock.yml
+++ b/ci/base.lock.yml
@@ -5,7 +5,7 @@ overrides:
         oe-core:
             commit: 06dd66e6220e5ce4ed4b9af4d8231ae5f0a8ce80
         meta-lts-mixins:
-            commit: 348a9eaaf3991e2329d0cef14dce23fabaef9191
+            commit: 6ad95d00531b32ec01792ab496718943dc95277b
         bitbake:
             commit: 22021758e66737bcf68dfd2b74adc6a0cb1d42d9
         meta-arm:
--- a/recipes-bsp/hexagon-dsp-binaries/hexagon-dsp-binaries_20260519.bb
+++ b/recipes-bsp/hexagon-dsp-binaries/hexagon-dsp-binaries_20260519.bb
@@ -68,7 +68,6 @@ PACKAGE_BEFORE_PN =+ "\
     ${PN}-qcom-sa8775p-ride-gdsp \
     ${PN}-qcom-sdm845-hdk-adsp \
     ${PN}-qcom-sdm845-hdk-cdsp \
-    ${PN}-qcom-shikra-cqs-evk-cdsp \
     ${PN}-qcom-sm8750-mtp-adsp \
     ${PN}-qcom-sm8750-mtp-cdsp \
     ${PN}-radxa-dragon-q6a-adsp \
@@ -86,6 +85,12 @@ PACKAGE_BEFORE_PN =+ "\
     ${PN}-thundercomm-rb5-sdsp \
     ${PN}-thundercomm-rubikpi3-adsp \
     ${PN}-thundercomm-rubikpi3-cdsp \
+"
+# Dependent packages are only available from the lts-linux-firmware-mixins
+# layer and thus can't be enabled by default. Enable it for Qualcomm machines,
+# which also force enable the updated linux-firmware version.
+PACKAGES:append:qcom = " \
+    ${PN}-qcom-shikra-cqs-evk-cdsp \
 "
 
 LICENSE:${PN} = "dspso-WHENCE"
--- a/recipes-bsp/hexagon-dsp-binaries/hexagon-dsp-binaries_20260622.bb
+++ b/recipes-bsp/hexagon-dsp-binaries/hexagon-dsp-binaries_20260622.bb
@@ -13,7 +13,7 @@ LICENSE = " \
 LIC_FILES_CHKSUM = "\
     file://LICENSE.qcom;md5=56e86b6c508490dadc343f39468b5f5e \
     file://LICENSE.qcom-2;md5=165287851294f2fb8ac8cbc5e24b02b0 \
-    file://WHENCE;md5=7033ed3790855866a84dedb1a5a215d4 \
+    file://WHENCE;md5=ba0674c526f9fdc6f687dc13cb432f31 \
     file://conf.d/hexagon-dsp-binaries-qualcomm-iq9075-evk.yaml;endline=2;md5=077232564320a8fce4ea446daad3d726 \
 "
 NO_GENERIC_LICENSE[dspso-qcom] = "LICENSE.qcom"
@@ -24,7 +24,7 @@ SRC_URI = " \
     git://github.com/linux-msm/dsp-binaries;protocol=https;branch=trunk;tag=${PV} \
 "
 
-SRCREV = "f4021ba695bc1335666561d4f96fd3844cf88c45"
+SRCREV = "2ba83638b373c0a6bbb7ecb32f5e2b9dfca2c4ce"
 
 inherit allarch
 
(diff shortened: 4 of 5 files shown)


ricardosalveti:
```
AssertionError: Adding layer meta-qcom changed signatures.
19 signatures changed, initial differences (first hash before, second after):
   linux-firmware:do_create_recipe_spdx: bab1b2067febb9419099fa5f43f9f8393514d72d23db048ecc69510465282d5c -> 2b7f4315de17afadbc4bf8a353daadb8793ffe16dfec1f4977a4fe43b366a154
      bitbake-diffsigs --task linux-firmware do_create_recipe_spdx --signature bab1b2067febb9419099fa5f43f9f8393514d72d23db048ecc69510465282d5c 2b7f4315de17afadbc4bf8a353daadb8793ffe16dfec1f4977a4fe43b366a154
      NOTE: Reconnecting to bitbake server...
      basehash changed from 46cba62ccb5eff24221fabd763de8b77e724db7f6a8eeb77c4ffac3c381fc667 to 339c2ad62f98341d22c355d096046c3fdf7bccb0ed59759573b376a4ebd79a7f
      Variable FILE_LAYERNAME value changed from 'core' to 'lts-linux-firmware-mixin'
      Variable PV value changed from '20260410' to '20260519'

   linux-firmware:do_recipe_qa: 0a18db85941ae3add26ad234486a1b01b31b7b54ffb4633aff35a5fa2381386a -> 6641f9c12a0021a333d7c1c68977c65da63bc9ab155765a6de49f03b1b75e449
      bitbake-diffsigs --task linux-firmware do_recipe_qa --signature 0a18db85941ae3add26ad234486a1b01b31b7b54ffb4633aff35a5fa2381386a 6641f9c12a0021a333d7c1c68977c65da63bc9ab155765a6de49f03b1b75e449
      NOTE: Reconnecting to bitbake server...
      basehash changed from fa8bf6aaa007c88ff876d2fc147da2c6078f9d10f7d6f81c03814bb94c8f5f2e to fb1537cd0a14db2f65fd517807245c6f82c9118587c5c752fcdf6714836c0f03
      Variable PV value changed from '20260410' to '20260519'
      Variable SRC_URI[sha256sum] value changed from 'b7812ed6d59f6b09ecceddaa0be842a7e82a79cc0e46ca60478a4ebf02f1e178' to 'b14e7197a290a7e5569f5ef790cde289bddc47e32126f2eb262a8e677fc39727'

```

conf/layer.conf
```diff
@@ -17,6 +17,9 @@ LAYERDEPENDS_qcom = "core lts-linux-firmware-mixin"
 LAYERRECOMMENDS_qcom = "openembedded-layer meta-arm"
 LAYERSERIES_COMPAT_qcom = "wrynose"
 
+# Use the linux-firmware provided by the lts-mixins layer
+LTS_LINUX_FIRMWARE_MIXIN_DEFAULT_PREFERENCE = "1"
```
lumag: This should be set via setting the DEFAULT_PREFERENCE in the bbappend

lumag (changes requested):

lumag:
I've posted #2551 . Please consider that one.
Also, @quaresmajose ,  `LTS_LINUX_FIRMWARE_MIXIN_DEFAULT_PREFERENCE` doesn't fit the YP compatibility idea. Please revert those comments from the mixins layer.

quaresmajose:
I'm trying to find a solution to the problem.

quaresmajose:
> I've posted #2551 . Please consider that one.

The mixins is already YP compatible given that it is passing the check layer with your change `11dcf7d SECURITY.md: describe security policy and pointers`.  But we can't use it as a dependency of meta-qcom, because when we add the override in meta-qcom to disable the `DEFAULT_PREFERENCE = -1` of the mixins we break YP compat.

> Also, @quaresmajose , `LTS_LINUX_FIRMWARE_MIXIN_DEFAULT_PREFERENCE` doesn't fit the YP compatibility idea. Please revert those comments from the mixins layer.

Replied on the mailing list:

_I have merged these changes but this also breaks the Yocto Project Compatible check layer.
In reality, I think the concept of the mixin layer will always break the Yocto Project Compatible,
because the desire of the mixin layer is to introduce a more recent version of the one
provided by OE-core._

_I'll have to revert those commits, but I'm looking for a way to do this while maintaining
the layer Yocto Project Compatible. If anyone has any suggestions, I would appreciate it._

https://lists.yoctoproject.org/g/yocto-patches/message/4250

conf/layer.conf
quaresmajose: That's not the reason for the problem.

lumag: Yeah, fixed in #2606 . I'll rebase the #2551 on top of the first two patches of that PR.

quaresmajose:
Update the meta-lts-mixins layer to tip and add fixes from https://github.com/qualcomm-linux/meta-qcom/pull/2606

recipes-kernel/linux-firmware/linux-firmware_%.bbappend
```diff
@@ -1,4 +1,8 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+# use latest version of linux-firmware from the lts-linux-firmware-mixin layer
+DEFAULT_PREFERENCE:qcom = ""
```
lumag: This should come before the previous commit, otherwise the tree is broken between them.

lumag (changes requested):

recipes-kernel/linux-firmware/linux-firmware_%.bbappend
lumag: Also, I'd rather keep it in the versioned bbappend to make sure that we don't touch linux-firmware recipes possibly coming from other layers.

quaresmajose: With the layer-lts-linux-firmware-mixin override we will just change the mxins layer: 
```
DEFAULT_PREFERENCE:layer-lts-linux-firmware-mixin:qcom = "1"
```

_To verify I set the `DEFAULT_PREFERENCE = "1"` in the mixins `linux-firmware_20260519` use the `DEFAULT_PREFERENCE:layer-core:qcom = "2"` here and bitbake picks the core 20260410 version._

It works but will push to see the check-layer.

ricardosalveti (dismissed):

recipes-kernel/linux-firmware/linux-firmware_%.bbappend
lumag: I think the versioned bbappend is an easier and less fragile solution.

quaresmajose: Using the version will mean we'll have to modify yet another file when bumping the mixins layer. The _layer-lts-linux-firmware-mixin_ override does the same and we are just modifying the DEFAULT_PREFERENCE of the version in question, which is 20260519 currently.

I don't understand why you say it's fragile; I think it's a more robust and simpler way to do the same thing.

lumag: Consider other BSP layer having its own version of the `linux-firmware` recipe and using `DEFAULT_PREFERENCE` to disable it by default. With your approach the recipe will be forcebly reenabled too. With my suggestion it won't happen. Yes, it comes for the price of performing two actions when bumping lts-mixins (bump commit id, rename the recipe) instead of doing one (bumping the commit id).

quaresmajose: We only chose the version of the mixin layer for `:qcom` machines. Other BSP layers will not be affected by this patch, that would break the YP compatibility.

lumag: But assignment is happening within the wildcard bbappend, so it will be applied to all `linux-firmware` recipes.

quaresmajose: No, it will be applied only for the recipes in the mixins layer as it is guarded by `:layer-lts-linux-firmware-mixin`.

lumag: Ah, ok. I was thinking about the original setting your patch. With the layer override it should be fine.

lumag (dismissed):

lumag (commented):
Please also pick up https://github.com/qualcomm-linux/meta-qcom/pull/2647/commits/13af9c798f8bac5458204b40ab34e4bb7f9c90dd

quaresmajose:
> Please also pick up [13af9c7](https://github.com/qualcomm-linux/meta-qcom/commit/13af9c798f8bac5458204b40ab34e4bb7f9c90dd)

I'll wait for the https://github.com/qualcomm-linux/meta-qcom/pull/2647 merge on master to backport the https://github.com/qualcomm-linux/meta-qcom/commit/13af9c798f8bac5458204b40ab34e4bb7f9c90dd.

vkraleti:
@quaresmajose #2647 merged.

lumag (approved):

vkraleti (approved):

koenkooi (approved):

PR #2546, opened 2026-06-19
selinux: Add Container, Wayland, and Docker policy updates

kemalrasimsh:
Introduce a set of SELinux policy updates to improve support for containerized workloads requiring access to host configuration, audio, and graphical services.

The following changes are included:

  - Container: Allow access to /etc/cdi for CDI (Container Device Interface) configuration, enabling container runtimes to discover and use device plugins.

  - Wayland: Add a new interface, wayland_stream_connect(), to allow confined domains to connect to Wayland compositor sockets over UNIX stream sockets.

  - Docker: Add optional policy to allow dockerd_t to connect to PulseAudio and Wayland sockets via:
      * pulseaudio_stream_connect(dockerd_t)
      * wayland_stream_connect(dockerd_t)

    These permissions are wrapped in optional_policy blocks and are
    only enabled when the corresponding modules are present.

These changes collectively improve integration of containerized applications with desktop environments (audio and display) while preserving modular SELinux policy design.

Note that granting dockerd access to user session services may have security implications and should only be enabled for environments that require GUI or audio-enabled containers.


lumag:
There is nothing BSP-specific in these policy files. Please split them according to the target and push to corresponding layers. Also, please, don't submit the PRs against the wrynose branch, unless the change is actually wrynose-specific and doesn't apply to the master branch.
I'm going to close this PR.

ricardosalveti:
These are all distro policies.

PR #2551, opened 2026-06-19
Fix Yocto Project Compatibility requirements

### LINKED ISSUE #2457: [wrynose] meta-lts-mixin dependency broke Yocto autobuilder builds

lumag:
It's required that YP Compatible layers don't depend on non-compatible layers. Also the wrynose/linux-firmware branch of the mixins is not a part of the Yocto Autobuilder. Make sure that meta-qcom layer builds without the extra layers.

Closes #2457

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -13,8 +13,8 @@ BBFILE_COLLECTIONS += "qcom"
 BBFILE_PATTERN_qcom := "^${LAYERDIR}/"
 BBFILE_PRIORITY_qcom = "6"
 
-LAYERDEPENDS_qcom = "core lts-linux-firmware-mixin"
-LAYERRECOMMENDS_qcom = "openembedded-layer meta-arm"
+LAYERDEPENDS_qcom = "core"
+LAYERRECOMMENDS_qcom = "openembedded-layer meta-arm lts-linux-firmware-mixin"
 LAYERSERIES_COMPAT_qcom = "wrynose"
 
 LICENSE_PATH += "${LAYERDIR}/licenses"
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -24,6 +24,8 @@ BBFILES_DYNAMIC += " \
     meta-arm:${LAYERDIR}/dynamic-layers/meta-arm/*/*/*.bbappend \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bb \
     networking-layer:${LAYERDIR}/dynamic-layers/networking-layer/*/*/*.bbappend \
+    lts-linux-firmware-mixin:${LAYERDIR}/dynamic-layers/lts-linux-firmware-mixin/*/*/*.bb \
+    lts-linux-firmware-mixin:${LAYERDIR}/dynamic-layers/lts-linux-firmware-mixin/*/*/*.bbappend \
     openembedded-layer:${LAYERDIR}/dynamic-layers/openembedded-layer/*/*/*.bb \
     openembedded-layer:${LAYERDIR}/dynamic-layers/openembedded-layer/*/*/*.bbappend \
     qcom-distro:${LAYERDIR}/dynamic-layers/qcom-distro/*/*/*.bb \
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -299,6 +299,14 @@ jobs:
           # Additional builds for specific machines
           - machine: qcom-armv8a
             distro:
+                name: qcom-distro-nomixin
+                yamlfile: ':ci/qcom-distro-nomixin.yml'
+            kernel:
+                type: default
+                dirname: ""
+                yamlfile: ""
+          - machine: qcom-armv8a
+            distro:
                 name: qcom-distro
                 yamlfile: ':ci/qcom-distro.yml'
             kernel:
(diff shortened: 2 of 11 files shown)


conf/layer.conf
```diff
@@ -13,8 +13,8 @@ BBFILE_COLLECTIONS += "qcom"
 BBFILE_PATTERN_qcom := "^${LAYERDIR}/"
 BBFILE_PRIORITY_qcom = "6"
 
-LAYERDEPENDS_qcom = "core lts-linux-firmware-mixin"
-LAYERRECOMMENDS_qcom = "openembedded-layer meta-arm"
+LAYERDEPENDS_qcom = "core"
+LAYERRECOMMENDS_qcom = "openembedded-layer meta-arm lts-linux-firmware-mixin"
```
quaresmajose, line 17: Is this really what we intend? This way we'll have to support two firmware versions, and the version qcom prefer  will no longer be the one from the mixin layer.

With this change, you need to remove the meta-lts-mixins dependency in the `ci/yocto-check-layer.sh`.

lumag: This is what we must support until meta-lts-mixins / firmware branch is enable and available on the autobuilder.

lumag: > and the version qcom prefer will no longer be the one from the mixin layer.

See the bbappend. If the layer is available, it will be used.

quaresmajose: > This is what we must support until meta-lts-mixins / firmware branch is enable and available on the autobuilder.

Have you considered that meta-lts-mixins may never be YP compatible?

quaresmajose: > > and the version qcom prefer will no longer be the one from the mixin layer.
> 
> See the bbappend. If the layer is available, it will be used.

This will introduce a lot of complexity. We will now support two distinct versions of linux-firmware. For example https://github.com/qualcomm-linux/meta-qcom/pull/2450, that patch the linux-firmare may won't work, and similar ones will have the same problems.

lumag: @quaresmajose any other proposals? The worse case would be to drop the meta-qcom/wrynose from the autobuilder and from the YP compatibility program.

ndechesne: > @quaresmajose any other proposals? The worse case would be to drop the meta-qcom/wrynose from the autobuilder and from the YP compatibility program.

This is not possible. meta-qcom being YP compatible is a must have. and part of our 'product' definition.

lumag: > > @quaresmajose any other proposals? The worse case would be to drop the meta-qcom/wrynose from the autobuilder and from the YP compatibility program.
> 
> This is not possible. meta-qcom being YP compatible is a must have. and part of our 'product' definition.

Which means that (per definition) we should depend only on YP-compatible layers. Until the required branch of the mixins layer is not marked as compatible, we can not depend on it.

quaresmajose: I have no other proposal or solution to this problem. I think we need to discuss this more broadly. Moving this discussion to https://github.com/qualcomm-linux/meta-qcom/issues/2457 is more appropriate in my opinion because we can get opinions from other YP maintainers.

koenkooi:
This looks fairly good to me.

ricardosalveti:
I don't think we have any other option until the mixin layer is also compatible.

At the same time, we do need to have at least one qcom-armv8a build using qcom-distro and not enabling the mixin layer to make sure we keep things working as needed (as duplicating more builds will also explode our CI).

lumag:
@ricardosalveti just the build or the build+test?

ricardosalveti:
> @ricardosalveti just the build or the build+test?

At this point my main concern would be build testing as to test this effectively we would have to duplicate CI and validate with qcom-distro builds, which would be too much.

Build tests should allow us to at least easily identify packaging issues.

quaresmajose:
> I don't think we have any other option until the mixin layer is also compatible.

That's not the problem! Actually, the problem is that even if it's compatible, you won't be able to use it without your product losing compatibility.

quaresmajose:
The problem is the following:

- OE-core: YP Compatible
- Mixins-layer: YP Compatible (assuming they will be accepted because they currently pass the checklayer.)
- BSP-layer + OE-core: YP Compatible
- BSP-layer + OE-core + Mixins-layer: YP Compatibility is lost because the checklayer start failing.

lumag:
> * BSP-layer + OE-core + Mixins-layer: YP Compatibility is lost because the checklayer start failing.

I've run the yocto-check-layer on this combo and it still passes. Please point out the error.

quaresmajose:
> > * BSP-layer + OE-core + Mixins-layer: YP Compatibility is lost because the checklayer start failing.
> 
> I've run the yocto-check-layer on this combo and it still passes. Please point out the error.

It fails in https://github.com/qualcomm-linux/meta-qcom/pull/2542. Where do you say it passes through?

lumag:
> > > * BSP-layer + OE-core + Mixins-layer: YP Compatibility is lost because the checklayer start failing.
> > 
> > 
> > I've run the yocto-check-layer on this combo and it still passes. Please point out the error.
> 
> It fails in #2542. Where do you say it passes through?

See #2606

PR #2553, opened 2026-06-19
ci: performance: enable NM wait-online override

DaizhuangBai:
Enable the NetworkManager wait-online override only for performance builds by setting:

    NETWORKMANAGER_WAIT_ONLINE_WAIT_FOR_STARTUP = "0"

This keeps the generic distro behavior unchanged while applying the boot-time optimization in performance profile.

### DIFF
--- a/ci/performance.yml
+++ b/ci/performance.yml
@@ -4,3 +4,5 @@ header:
 local_conf_header:
   cmdline-perf: |
     KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=linux"
+    # Do not wait for NetworkManager startup-complete in wait-online.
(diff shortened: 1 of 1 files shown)


ricardosalveti:
Again, this is the wrong fix, we should fix the services depending on network-online.

https://networkmanager.dev/docs/api/latest/NetworkManager-wait-online.service.html

Optimally, all services on the host react dynamically to network changes and systemd services do not need to be configured to start after reaching the network-online target. In this case, NetworkManager-wait-online.service has no effect and does not delay the boot time. On the other hand, if you encounter a long boot time due to the delay of NetworkManager-wait-online, investigate the services that require network access and fix them.

PR #2554, opened 2026-06-19
Add Qcom SDX75 SoC and IDP board support

Mani-Sadhasivam:
Qcom SDX75 is the modem SoC supported in mainline kernel. This PR adds the SoC and IDP reference board support. This board features NAND as the storage medium as like its predecessor SDX55. Hence, added the necessary UBI configurations as well.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -417,6 +417,14 @@ jobs:
                 type: u-boot-qcom-fit-6.18
                 dirname: "_linux-qcom-6.18_u-boot-qcom-fit"
                 yamlfile: ":ci/linux-qcom-6.18.yml:ci/u-boot-qcom.yml:ci/kernel-fit-image.yml"
+          - machine: sdx75-idp
+            distro:
+                name: nodistro
+                yamlfile: ""
+            kernel:
+                type: default
+                dirname: ""
+                yamlfile: ""
     uses: ./.github/workflows/compile.yml
     with:
       name: ${{ matrix.machine }}/${{ matrix.distro.name }}${{ matrix.kernel.dirname }}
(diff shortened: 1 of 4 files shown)


ricardosalveti:
Is this IDP available in public somewhere?

We should also add in our CI to make sure we're validating the build. An extra build at https://github.com/qualcomm-linux/meta-qcom/blob/master/.github/workflows/build-yocto.yml#L299 using nodistro should be more than enough.

Mani-Sadhasivam:
> Is this IDP available in public somewhere?

No, this is an internal evaluation board, just like MTP, CRD...

> We should also add in our CI to make sure we're validating the build. An extra build at https://github.com/qualcomm-linux/meta-qcom/blob/master/.github/workflows/build-yocto.yml#L299 using nodistro should be more than enough.

Done

mwasilew:
Are there any plans to run boot tests on this board?

.github/workflows/build-yocto.yml
```diff
@@ -241,6 +241,7 @@ jobs:
           - qcs9100-ride-sx
           - rb1-core-kit
           - rb3gen2-core-kit
+          - sdx75-idp
```
ricardosalveti: This will add for all types of builds, just the block you added after should be fine.

Mani-Sadhasivam: Fixed

Mani-Sadhasivam:
> Are there any plans to run boot tests on this board?

Not at this point. This is added to let the NAND team easily flash and validate upstream stack on the board. But I do have plans to hook it up in CI and run boot tests. Due to board scarcity, it is not plugged to CI yet.

lumag (approved):

ricardosalveti (approved):

PR #2557, opened 2026-06-20
firmware-qcom: make split packages installable

### LINKED ISSUE #2504: sm8550-hdk firmware files generate empty package which blocks installation
Hi,

Adding
```
packagegroup-sm8550-hdk-firmware
```
in `MACHINE_ESSENTIAL_EXTRA_RECOMMENDS` together with a properly set `NHLOS_URI` for the firmware
does not result in installation of the firmware into the image.
The reason for that is that all firmware packages have an `RDEPENDS` on `firmware-qcom-sm8550-hdk`,
which is empty.

The firmware generation has a note about the license:
```
    for pkg in d.getVar("SPLIT_FIRMWARE_PACKAGES").split():
        # Depend on the main package to get the license file
        d.appendVar("RDEPENDS:" + pkg, " " + pn)
```
but how are we supposed to add the license to the main package?

For now I removed the `RDEPENDS` with a `.bbappend`:
```
# Remove REDEPENDS for linux-firware-* packages, the main package is empty and
# thus can't be depended on.

RDEPENDS:linux-firmware-qcom-sm8550-modem:remove = "firmware-qcom-sm8550-hdk"
RDEPENDS:linux-firmware-qcom-sm8550-audio:remove = "firmware-qcom-sm8550-hdk"
RDEPENDS:linux-firmware-qcom-sm8550-compute:remove = "firmware-qcom-sm8550-hdk"
RDEPENDS:linux-firmware-qcom-sm8550-ipa:remove = "firmware-qcom-sm8550-hdk"
```

lumag:
Since the commit 77e6f4504d9b ("firmware-qcom-nhlos: make sure that meta firmware package is empty") the main package in the split firmware doesn't have any files and thus is skipped by OE. All split firmware packages depend on PN, which can't be fond. Drop the manual dependency, making all split packages installable again.

Closes #2504

### DIFF
--- a/recipes-bsp/firmware/firmware-qcom.inc
+++ b/recipes-bsp/firmware/firmware-qcom.inc
@@ -43,9 +43,7 @@ python() {
     pn = d.getVar("PN")
     insanes = d.getVar("INSANE_SKIP:%s" % pn)
     for pkg in d.getVar("SPLIT_FIRMWARE_PACKAGES").split():
-        # Depend on the main package to get the license file
-        d.appendVar("RDEPENDS:" + pkg, " " + pn)
-        # and append the INSANE_SKIP of the main package to pass QA
+        # append the INSANE_SKIP of the main package to pass QA
         d.appendVar("INSANE_SKIP:" + pkg, " " + insanes)
     if d.getVar("FW_QCOM_NAME") == "unset" and d.getVar("SPLIT_FIRMWARE_PACKAGES") != "":
         bb.error("%s: split firmware-qcom packages engaged, but FW_QCOM_NAME is not defined" % pn)


vkraleti (dismissed):

ricardosalveti:
Is the license now part of the individual packages?

koenkooi (approved):
Nit: _fond_ → _found_

lumag:
> Is the license now part of the individual packages?

No, but we have a plenty of packages for which there is no binding license. For example, for the RBn devices Thundercomm doesn't define any licence for the files downloaded via the SDK manager.

ricardosalveti:
> > Is the license now part of the individual packages?
> 
> No, but we have a plenty of packages for which there is no binding license. For example, for the RBn devices Thundercomm doesn't define any licence for the files downloaded via the SDK manager.

So the license will basically stay as CLOSED, right? Which is not really valid, but not sure we can do much here if we want to keep these recipes around.

lumag:
> So the license will basically stay as CLOSED, right? Which is not really valid, but not sure we can do much here if we want to keep these recipes around.

Yes. I have a solution for some of the recipes, but I'll submit it separately. Let's fix the bug first.

ricardosalveti (approved):

PR #2579, opened 2026-06-22
Enable Glymur camx DTB compilation for linux-qcom-next kernel

cjha01:
Downstream camera DTB is not included in the final image
when building linux-qcom-next, kernel for the glymur crd 
board. This prevents the downstream camera stack from being
enabled at boot, even though the kernel supports it.

Include the downstream camera DTB so it is packaged into
the final image and downstream camera functionality is
available for the Glymur crd.

Add multi-DTB support for camx overlay to enable downstream
camera for Glymur crd (glymur-crd-camx).

### DIFF
--- a/conf/machine/glymur-crd.conf
+++ b/conf/machine/glymur-crd.conf
@@ -17,6 +17,11 @@ LINUX_QCOM_KERNEL_DEVICETREE ?= " \
     qcom/glymur-staging.dtbo \
 "
 
+# These DTs are not upstreamed and currently exist only in linux-qcom-next kernel
+KERNEL_DEVICETREE:append:pn-linux-qcom-next = " \
+    qcom/glymur-crd-camx.dtbo \
+"
+
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     packagegroup-glymur-crd-firmware \
     packagegroup-glymur-crd-hexagon-dsp-binaries \
--- a/conf/machine/include/fit-dtb-compatible-linux-qcom.inc
+++ b/conf/machine/include/fit-dtb-compatible-linux-qcom.inc
@@ -213,3 +213,4 @@ FIT_DTB_COMPATIBLE[qcom_qcs6490-iot-subtype2-camx-staging] = \
     "qcs6490-rb3gen2 qcs6490-rb3gen2-vision-mezzanine-camx kodiak-staging"
 # ---------- glymur ----------
 FIT_DTB_COMPATIBLE[qcom_glymur-crd-staging] = "glymur-crd glymur-staging"
+FIT_DTB_COMPATIBLE[qcom_glymur-crd-camx] = "glymur-crd glymur-crd-camx"


lumag (dismissed):

ricardosalveti (dismissed):

cjha01:
The build is failing on the 6.18 kernel. As per our understanding, camera enablement is required on Glymur-CRD for qcom-next; So, we did not add the glymur-crd-camx.dtso file to the 6.18.y kernel.

Camera enablement depends on clock, regulator, and PMIC changes, and these dependencies are not yet present in 6.18.y kernel. Due to this, we cannot introduce glymur-crd-camx.dtso in 6.18 kernel.

Additionally, we received input from PDM that camera needs to be enabled only on QLI mainline as of now, not for QLI 2.0.

However, the issue is that **LINUX_QCOM_KERNEL_DEVICETREE** expects DTSO change from both qcom-next and 6.18 kernels, which is causing the build failure on 6.18.

lumag:
@cjha01 you have written a lengthy message, explaining why it fails. What is not obvious from your message, how do you plan to fix it. Hint: stop using `LINUX_QCOM_KERNEL_DEVICETREE`.

lumag:
> camera needs to be enabled only on QLI mainline as of now, not for QLI 2.0.

How does it affect the build failure? Both kernels are a part of the QLI mainline.

cjha01:
> @cjha01 you have written a lengthy message, explaining why it fails. What is not obvious from your message, how do you plan to fix it. Hint: stop using `LINUX_QCOM_KERNEL_DEVICETREE`.

Our understanding is that QLI mainline enablement changes should be part of qcom-next. We will discuss this with PDM and request the dependent teams to merge the required changes into the 6.18.y kernel. Once those are in place, we will enable the downstream camera DTSI in 6.18.y kernel. Until then, this PR will remain in draft state.

cjha01:
> > camera needs to be enabled only on QLI mainline as of now, not for QLI 2.0.

Our understanding is that QLI mainline enablement changes should be part of qcom-next. We will discuss this with PDM and request the dependent teams to merge the required changes into the 6.18.y kernel. Once those are in place, we will enable the downstream camera DTSI in 6.18.y kernel. Until then, this PR will remain in draft state.

> How does it affect the build failure? Both kernels are a part of the QLI mainline.

Downstream camera DTSI is not available in the 6.18.y kernel. As a result, the glymur-crd-camx.dtbo is not listed in the Makefile for 6.18.y builds, leading to a compilation failure.

cjha01:
@lumag,  As Glymur CRD is not targeted for the 6.18.y kernel, we are not sure when the camera dependencies changes will be available on 6.18.y. So, we have a few queries:

1. Do we have any provision in meta-qcom to restrict compilation of DTBO only for the qcom-next kernel? Is this approach allowed?

2. Since the dependency changes are not yet available, do you expect us to add glymur-crd-camx.dtso and include it in the build with basic software camera-specific nodes to avoid build failures, or should we first get all dependency changes into the 6.18.y kernel and then add the camera-specific DTSI changes on top?

lumag:
> @lumag, As Glymur CRD is not targeted for the 6.18.y kernel, we are not sure when the camera dependencies changes will be available on 6.18.y. So, we have a few queries:
> 
> 1. Do we have any provision in meta-qcom to restrict compilation of DTBO only for the qcom-next kernel? Is this approach allowed?

Yes, it's allowed. Just use a proper variable assignment (`KERNEL_DEVICETREE:append:pn-linux-qcom-next`).

> 2. Since the dependency changes are not yet available, do you expect us to add glymur-crd-camx.dtso and include it in the build with basic software camera-specific nodes to avoid build failures, or should we first get all dependency changes into the 6.18.y kernel and then add the camera-specific DTSI changes on top?

It should be in qcom-next tree.

lumag (approved):

conf/machine/glymur-crd.conf
```diff
@@ -17,6 +17,11 @@ LINUX_QCOM_KERNEL_DEVICETREE ?= " \
     qcom/glymur-staging.dtbo \
 "
 
+# These DTs are not upstreamed and currently exist only in linux-qcom-next kernel
+KERNEL_DEVICETREE:append:pn-linux-qcom-next ?= " \
```
lumag: No need for ?=

lumag (changes requested):

lumag (approved):

ricardosalveti (approved):

PR #2591, opened 2026-06-23
Revert "mesa: Backport UBWC fix for PIPE_FORMAT_R8_G8B8_420_UNORM for…

ricardosalveti:
…mat"

This reverts commit 540437f2aadbcd34bb0c5eedb0eaac5ac6662de5.

Patch now available via mesa 26.1.2, currently available in oe-core.


ricardosalveti:
We are not yet using the oe-core rev that contains latest mesa, but that is part of https://github.com/qualcomm-linux/meta-qcom/pull/2461 (due the series of changes required, from multiple layers), but separated this change as it is required to unblock autobuilder, which is now failing with meta-qcom.

koenkooi (approved):
In the future we could start using versioned bbappends and discuss if we want to downgrade the dangle error to warning.

Anyway, approved

lumag:
That would mean that the master branch is broken if built using kas. I'm fine with landing this one as a desperate measure, but what does it take to land #2461 ?

lumag (approved):

ricardosalveti:
> That would mean that the master branch is broken if built using kas. I'm fine with landing this one as a desperate measure, but what does it take to land #2461 ?

Going over the multiple failures and bringing the fixes as part of https://github.com/qualcomm-linux/meta-qcom/pull/2461, taking a bit to validate the entire run due the lack of cache and constant build failures caused by long clang/rust builds.

The only actual known issue (linux-yocto-dev fails to build) will be fixed once https://lists.openembedded.org/g/openembedded-core/message/239324 is merged upstream. The current workaround breaks yocto-check-layer, I could bring the kas patch but hopefully it will be merged soon.

PR #2597, opened 2026-06-24
refpolicy_targeted: Enable the tunable flag tee_supplicant_qtee

wenjz-qualcomm:
In the upstream repository, OP-TEE and QTEE share the same set of SELinux policy files. QTEE-specific SELinux policies are therefore required by upstream to be placed in a tunable policy module. Upstream requires that all tunable policy modules must default to false.

However, the contional SELinux policies need applies to KLMT devices in meta-qcom. As such, this patch must be a part of meta-qcom.

CRs: 4567589
Tag: qli-2.0 GA pull-request freeze

### DIFF
--- a/dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted_git.bbappend
+++ b/dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted_git.bbappend
@@ -2,4 +2,5 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 SRC_URI:append:qcom = " \
     file://0001-Add-SELinux-policy-for-nhx.sh.patch \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'optee', '', 'file://0002-Enable-the-tunable-flag-tee_supplicant_qtee.patch', d)} \
 "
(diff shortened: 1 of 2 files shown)


lumag (approved):

koenkooi (approved):

dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted_git.bbappend
```diff
@@ -2,4 +2,5 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 SRC_URI:append:qcom = " \
     file://0001-Add-SELinux-policy-for-nhx.sh.patch \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'optee', '', 'file://0002-Enable-the-tunable-flag-tee_supplicant_qtee.patch', d)} \
```
ricardosalveti, line 5: Don't we want to apply for both optee and qtee?

lumag: Other way. It's optee-opt-out

ricardosalveti: Yup, updated my comment, was expecting it to be enabled for both.

ricardosalveti: Reading https://github.com/SELinuxProject/refpolicy/blob/main/policy/modules/services/tee_supplicant.te it is actually specific to qtee, the file name confused me as I was expecting both supplicants to be used there.

ricardosalveti: > Reading https://github.com/SELinuxProject/refpolicy/blob/main/policy/modules/services/tee_supplicant.te it is actually specific to qtee, the file name confused me as I was expecting both supplicants to be used there.

https://github.com/SELinuxProject/refpolicy/commit/694c913f851a332c6673b19cdfbeb3852a588d27 shows the pieces that are specific here, all good.

ricardosalveti (approved):

PR #2598, opened 2026-06-24
[RFC] qcom-firmware-sign: build-time secure-boot signing of Qualcomm firmware

igoropaniuk:
This RFC adds a build-time secure-boot signing pipeline for Qualcomm firmware. It introduces three -native recipes - `sectools-native`, `security-profiles-native`, and `qdl-native` - together with a `qcom-firmware-sign.bbclass` that any recipe can inherit to sign its deploy artefacts before they ship.

A development ECDSA test-key set and a ci/secure-boot.yml kas overlay are included so a fresh checkout can exercise the full pipeline end-to-end without sourcing OEM signing material.

Taking into account that `qualcomm/security-profiles` is still private, it can be tested with (so build likely to fail until the rep is opened):

```yaml
# ci/security-profiles-local.yml
header:
  version: 14

local_conf_header:
  security-profiles-local: |
    PREMIRRORS:prepend = "git://github.com/qualcomm/security-profiles.git.* git://${QCOM_SECURITY_PROFILES_LOCAL};protocol=file \n"
```

To test this build:
```bash
kas build ../ci/rb3gen2-core-kit.yml:../ci/secure-boot.yml:../ci/ecdsa-secure-boot-test-keys.yml:../ci/security-profiles-local.yml
```

When this feature is enabled it generates:
* Signed boot binaries in `deploy/.../<machine>/`, for example `deploy/images/rb3gen2-core-kit/qcm6490`
* The `qcomflash-vip.tar.gz`, which contains the full qcomflash artifact set under `<image>-<machine>/ plus a vip-tables/DigestsToSign.bin.mbn` OEM-signed via the same key chain. The package is ready to be flashed with QDL to the secured device with VIP engaged.


koenkooi (changes requested):
This completely removes qdl (non-native), please change that to one commit updating qdl_git to 2.7+ and adding `inherit native`. I use `qdl` on-device to flash other devices, it's more convenient than buying even more USB hubs 😎

igoropaniuk:
> This completely removes qdl (non-native), please change that to one commit updating qdl_git to 2.7+ and adding inherit native. I use qdl on-device to flash other devices, it's more convenient than buying even more USB hubs 😎

Sure, I thought it's no used anymore (as still points to 2.1, which is a bit obsolete). Also 2.7 release requires libzip, which is available in meta-oe, that's why I also moved the recipe to the dynamic layers

igoropaniuk:
FYI: the build fails because of security-profiles

recipes-devtools/security-profiles/security-profiles-native_git.bb
```diff
@@ -0,0 +1,22 @@
...
+do_compile[noexec]   = "1"
+
+# Install the XML profiles into a well-known native datadir location
+# so qcom-firmware-sign.bbclass can resolve them by filename via
+# QCOM_FIRMWARE_SIGN_SECPROFILE.
+do_install() {
+    install -d "${D}${datadir}/qcom-security-profiles"
+    install -m 0644 "${S}"/*_security_profile.xml \
+        "${D}${datadir}/qcom-security-profiles/"
+}
```
ricardosalveti, line 22: Missing EOL.

recipes-devtools/sectools/sectools-native_1.48.0.bb
```diff
@@ -0,0 +1,47 @@
+SUMMARY = "Qualcomm Sectools v2 -- image signing / verification tool"
+DESCRIPTION = "Qualcomm Security Tools v2: the host \
+binary used to sign, verify and inspect Qualcomm firmware images (XBL, \
+TZ, modem, etc.) according to a per-chipset Security Profile XML"
+HOMEPAGE = "https://softwarecenter.qualcomm.com/catalog/item/Qualcomm_Security_Tools"
+LICENSE = "LICENSE.qcom-2"
+
+LIC_FILES_CHKSUM = "file://${UNPACKDIR}/${ZIP_TOPDIR}/CHANGES.txt;md5=d2a0bb01dcd8befe660b832fbbe05900"
```
ricardosalveti, line 8: This is not really correct, started an internal thread to see if we can have a license inside the zip file.

recipes-devtools/sectools/sectools-native_1.48.0.bb
```diff
@@ -0,0 +1,47 @@
...
+ZIP_TOPDIR = "1.48"
```
ricardosalveti, line 9: PV?

ricardosalveti: Saw now that the directory is 1.48.0 but the zip is 1.48.zip, not great.

recipes-devtools/sectools/sectools-native_1.48.0.bb
```diff
@@ -0,0 +1,47 @@
...
+    install -m 0644 "${S}/CHANGES.txt" "${D}${datadir}/sectools/CHANGES.txt"
+}
+
+do_unpack[postfuncs] += "sectools_chmod_unpacked"
+sectools_chmod_unpacked() {
+    chmod -R u+w "${UNPACKDIR}"
+}
+
+FILES:${PN} += "${datadir}/sectools"
+INSANE_SKIP:${PN} += "already-stripped"
```
ricardosalveti, line 47: Missing EOL.

recipes-devtools/sectools/sectools-native_1.48.0.bb
```diff
@@ -0,0 +1,47 @@
...
+
+INHIBIT_DEFAULT_DEPS = "1"
+
+inherit native
+
+do_configure[noexec] = "1"
+do_compile[noexec]   = "1"
+
+# Pick the per-platform sectools binary that matches the build host.
+SECTOOLS_PLATFORM_DIR = "${@'Linux_aarch64' if d.getVar('BUILD_ARCH') == 'aarch64' else 'Linux'}"
```
ricardosalveti, line 24: We should also set COMPATIBLE_HOST.

classes-recipe/qcom-firmware-sign.bbclass
```diff
@@ -0,0 +1,257 @@
...
+    QCOM_FUSE_OEM_HW_ID \
+    QCOM_FUSE_OEM_PRODUCT_ID \
+    QCOM_FUSE_SEC_KEY_DERIVATION_KEY \
+"
+
+# Run before do_deploy (for recipes that ship signed blobs via DEPLOYDIR,
+# e.g. firmware-qcom-boot-common.inc) and before do_package (for recipes
+# that ship via package_install into /lib/firmware, e.g.
+# firmware-qcom.inc / firmware-qcom-hlosfw style consumers)
+addtask qcom_firmware_sign before do_deploy before do_package after do_install
```
ricardosalveti, line 257: Again :-)

classes-recipe/qcom-firmware-sign.bbclass
```diff
@@ -0,0 +1,257 @@
...
+    vpu30_4v_16mb.mbn:VENUS-FW \
+    wpss.mbn:WPSS \
+    xbl_config.elf:XBL-CONFIG \
+    xbl_config_gunyah.elf:XBL-CONFIG \
+    xbl_config_kvm.elf:XBL-CONFIG \
+    xbl.elf:XBL \
+    XblRamdump.elf:XBL-RAM-DUMP \
+    DigestsToSign.bin.mbn:VIP \
+    FD02C9DA-306C-48C7-A49C-BBD827AE86EE.mbn:TZ-APP-OEM \
+"
```
ricardosalveti, line 67: Can we move this to an inc file or similar? This is just metadata for the class.

ci/ecdsa-secure-boot-test-keys.yml
```diff
@@ -0,0 +1,14 @@
...
+# only and must NOT be used in production images.
+#
+# Usage: kas build ci/base.yml:ci/<machine>.yml:ci/secure-boot.yml:ci/ecdsa-secure-boot-test-keys.yml
+
+header:
+  version: 14
+
+local_conf_header:
+  ecdsa-secure-boot-test-keys: |
+    QCOM_FIRMWARE_SIGN_KEY_DIR = "${LAYERDIR_qcom}/ci/test-keys/ecdsa"
```
ricardosalveti, line 14: Same.

ci/test-keys/ecdsa/qpsa_attestca0.key
```diff
@@ -0,0 +1,9 @@
+-----BEGIN EC PARAMETERS-----
```
ricardosalveti, line 1: Split this commit in 2, have one dedicated just for the keys.

ci/secure-boot.yml
```diff
@@ -0,0 +1,57 @@
...
+      meta-oe:
+
+local_conf_header:
+  secure-boot: |
+    QCOM_FIRMWARE_SIGN_ENABLE = "1"
+
+    # Per-SoC Security Profile XML.  Bare filenames are resolved
+    # against the directory staged by security-profiles-native; an
+    # absolute path can be used to point at a custom profile.
+    QCOM_SECURITY_PROFILE       ?= "kodiak_security_profile.xml"
```
ricardosalveti, line 36: Profile should probably come from the board conf.

ci/secure-boot.yml
```diff
@@ -0,0 +1,57 @@
...
+    QCOM_FIRMWARE_SIGN_ANTI_ROLLBACK = "0x0"
+
+    # Enable the qcomflash-vip image type.  Produces an additional
+    # ${IMAGE_NAME}.qcomflash-vip.tar.gz next to the qcomflash tarball
+    # containing the signed VIP (Validated Image Programming) digest
+    # table -- intended to be flashed alongside qcomflash via
+    # `qdl --vip-table-path`.  See classes-recipe/image_types_qcom_vip.bbclass
+    # and recipes-devtools/qdl/qdl-native_2.7.bb.
+    IMAGE_CLASSES += "image_types_qcom_vip"
+    IMAGE_FSTYPES += "qcomflash-vip"
```
ricardosalveti, line 57: Same.

dynamic-layers/openembedded-layer/recipes-devtools/qdl/qdl-native_2.7.bb
```diff
@@ -0,0 +1,17 @@
...
+
+SRC_URI = "git://github.com/linux-msm/qdl.git;protocol=https;branch=master;tag=v${PV}"
+SRCREV = "f540e59dbeb462242239778ae0993549304f3abb"
+
+# qdl links against libusb-1.0, libxml-2.0 and libzip; all three are
+# packaged as -native in OE-core.  meson/ninja and pkgconfig provided
+# by the meson + pkgconfig classes.
+DEPENDS = "libusb1-native libxml2-native libzip-native"
+
+inherit native meson pkgconfig
```
ricardosalveti, line 17: Same.

dynamic-layers/openembedded-layer/recipes-devtools/qdl/qdl-native_2.7.bb
```diff
@@ -0,0 +1,17 @@
...
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=da6bfde9cb5bc5120a51775381f6edf1"
+
+SRC_URI = "git://github.com/linux-msm/qdl.git;protocol=https;branch=master;tag=v${PV}"
+SRCREV = "f540e59dbeb462242239778ae0993549304f3abb"
+
+# qdl links against libusb-1.0, libxml-2.0 and libzip; all three are
+# packaged as -native in OE-core.  meson/ninja and pkgconfig provided
+# by the meson + pkgconfig classes.
+DEPENDS = "libusb1-native libxml2-native libzip-native"
```
ricardosalveti, line 15: Why is it depending on libzip?

classes-recipe/image_types_qcom_vip.bbclass
```diff
@@ -0,0 +1,119 @@
...
+    # artefact, with the additional vip-tables/ subdir.
+    ${IMAGE_CMD_TAR} --sparse --numeric-owner \
+        --transform="s,^\./,${IMAGE_BASENAME}-${MACHINE}/," \
+        -cf- -C "${QCOMFLASH_VIP_DIR}" . \
+        | pigz -p ${BB_NUMBER_THREADS} -9 -n --rsyncable \
+        > "${IMGDEPLOYDIR}/${IMAGE_NAME}.qcomflash-vip.tar.gz"
+
+    ln -sf "${IMAGE_NAME}.qcomflash-vip.tar.gz" \
+        "${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.qcomflash-vip.tar.gz"
+}
```
ricardosalveti, line 117: I would prefer this to be done as part of the main tarball, to avoid confusion.

dynamic-layers/openembedded-layer/recipes-devtools/qdl/qdl-native_2.7.bb
igoropaniuk: Because of this feature https://github.com/linux-msm/qdl/pull/225/changes/fb3b622631a696fdf4fea463f579100721bce1ab

classes-recipe/image_types_qcom_vip.bbclass
igoropaniuk: will do

ci/secure-boot.yml
igoropaniuk: make sense, will do

recipes-devtools/sectools/sectools-native_1.48.0.bb
igoropaniuk: Yeah, that was actually the issue (no LICENSE file inside the archive )

dynamic-layers/openembedded-layer/recipes-devtools/qdl/qdl-native_2.7.bb
igoropaniuk: Maybe we can decouple this functionality and have a compile flag in meson.option  (it can be enabled by default, but in our case for VIP generation it's not needed)

ricardosalveti: Yeah, would be nice to have this as an optional dependency upstream, that way we won't have to depend on meta-oe for such a core BSP functionality.

lumag: If it needs to be a ZIP file, can we use a standard zlib for it?

recipes-devtools/sectools/sectools-native_1.48.0.bb
lumag, line 8: Drop UNPACKDIR. Also, can we work with the sectools authors to include the licence into the archive?

recipes-devtools/sectools/sectools-native_1.48.0.bb
```diff
@@ -0,0 +1,47 @@
...
+    install -d "${D}${bindir}"
+    ln -sf "../share/sectools/${SECTOOLS_PLATFORM_DIR}/sectools" \
+        "${D}${bindir}/sectools"
+
+    install -m 0644 "${S}/CHANGES.txt" "${D}${datadir}/sectools/CHANGES.txt"
+}
+
+do_unpack[postfuncs] += "sectools_chmod_unpacked"
+sectools_chmod_unpacked() {
+    chmod -R u+w "${UNPACKDIR}"
```
lumag, line 43: Why?

classes-recipe/qcom-firmware-sign.bbclass
lumag, line 257: Missing EOL

ci/secure-boot.yml
lumag, line 57: Do we need a separate class and separate image type?
Also the class hasn't been seen up to now, so this commit breaks the build. Please reoder them.

recipes-devtools/qdl/qdl_git.bb
```diff
@@ -1,23 +0,0 @@
-SUMMARY = "Qualcomm DownLoader flashing tool"
```
lumag, line 1: Would it be better to upgrade it instead of adding a native-only version?

classes-recipe/image_types_qcom_vip.bbclass
```diff
@@ -0,0 +1,119 @@
...
+
+QCOMFLASH_VIP_DIR = "${IMGDEPLOYDIR}/${IMAGE_NAME}.qcomflash-vip"
+
+do_image_qcomflash_vip[dirs]      = "${QCOMFLASH_VIP_DIR}"
+do_image_qcomflash_vip[cleandirs] = "${QCOMFLASH_VIP_DIR}"
+do_image_qcomflash_vip[depends]  += "qdl-native:do_populate_sysroot \
+                                     sectools-native:do_populate_sysroot \
+                                     security-profiles-native:do_populate_sysroot"
+
+create_qcomflash_vip_pkg() {
```
lumag, line 43: Okay, having a separate fstype kind of makes sense.

lumag (changes requested):

recipes-devtools/security-profiles/security-profiles-native_git.bb
```diff
@@ -0,0 +1,22 @@
+SUMMARY = "Qualcomm Security Profiles"
+DESCRIPTION = "Per-chipset Security Profile XML files consumed by Sectools"
+HOMEPAGE = "https://github.com/qualcomm/security-profiles"
+LICENSE = "BSD-3-Clause-Clear"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+
+SRC_URI = "git://github.com/qualcomm/security-profiles.git;protocol=https;branch=main"
+SRCREV = "122f1917af2b428880a7607ed705bb46bec66f5b"
+
+inherit native
```
quaresmajose, line 10: Look like this is just metadata so please add `allarch` and remove the default toolchain dependencies and friends

```
INHIBIT_DEFAULT_DEPS = "1"
```

recipes-devtools/sectools/sectools-native_1.48.0.bb
quaresmajose: the oe trim_version strips it
```
ZIP_TOPDIR="@oe.utils.trim_version("${PV}", 2)"
```

recipes-devtools/sectools/sectools-native_1.48.0.bb
```diff
@@ -0,0 +1,47 @@
...
+# Stage the per-platform sectools binary under ${datadir}/sectools/ and
+# drop a thin symlink into ${bindir} so consumers can invoke `sectools`
+# from PATH regardless of which subdir was selected.
+do_install() {
+    install -d "${D}${datadir}/sectools/${SECTOOLS_PLATFORM_DIR}"
+    install -m 0755 "${S}/${SECTOOLS_PLATFORM_DIR}/sectools" \
+        "${D}${datadir}/sectools/${SECTOOLS_PLATFORM_DIR}/sectools"
+
+    install -d "${D}${bindir}"
+    ln -sf "../share/sectools/${SECTOOLS_PLATFORM_DIR}/sectools" \
```
quaresmajose, line 35: I think you should use `${datadir}` for `../` and follow the installation's destination instead of assuming the destination is predefined.

koenkooi: @quaresmajose agreed

dynamic-layers/openembedded-layer/recipes-devtools/qdl/qdl-native_2.7.bb
ndechesne, line 15: why is it a -native only recipe? I would prefer qdl recipe that provides target and native support, like it was before.

igoropaniuk: @lumag @ricardosalveti zlib gives you `inflate()`/`deflate()` on a byte stream and nothing that understands a multi-file .zip container. There is aalso a `.zip` container library in oe-core - `libarchive`, it can read and write ZIP archives (plus tar, cpio, 7z, etc.). 

The problem is that `libzip` is random-access by name (zip_name_locate -> zip_fopen_index); `libarchive` is streaming (iterate archive_read_next_header, read in order). And QDL doesn't read the archive front-to-back - the device requests images on demand (Sahara/Firehose ask for specific members in device-driven order), which is exactly the random-access pattern libzip gives.

So I see three options how we can resolve this:
1. Carry the libzip recipe in `meta-qcom` layer (not sure we can easily move this recipe from meta-oe to oe-core, as nothing else depends on this recipe on oe-core)
2. Keep dependency on meta-oe layer
3. Completely re-implement QDL to use libarchive. I think we also want to support VIP flashing of zip-ed builds, so decoupling this functionality and making it optional is not the case either. @andersson any thoughts on this?

lumag: What about reusing [minizip](https://github.com/madler/zlib/tree/develop/contrib/minizip) ? I'd really like to prevent extra dependencies.

igoropaniuk: > What about reusing [minizip](https://github.com/madler/zlib/tree/develop/contrib/minizip) ? I'd really like to prevent extra dependencies.

I think pulling in an additional 10000 LOC into QDL just because we don’t want a dedicated recipe for libzip or want to avoid a dependency on meta-oe is overkill, tbh.

lumag: Then we need to justify bringing libzip to OE-Core.

igoropaniuk: @lumag made libzip support optional https://github.com/qualcomm-linux/meta-qcom/pull/2819

igoropaniuk:
Closing this in favor of https://github.com/qualcomm-linux/meta-qcom/pull/3168

PR #2606, opened 2026-06-24
Bump the lts-linux-firmware-mixin layer

lumag:
Bumpt the linux-firmware mixin layer to the YP compatible revision.

### DIFF
--- a/recipes-kernel/linux-firmware/linux-firmware_20260519.bbappend
+++ b/recipes-kernel/linux-firmware/linux-firmware_20260519.bbappend
@@ -1,0 +1,2 @@
+DEFAULT_PREFERENCE:qcom = "1"
+DEFAULT_PREFERENCE:class-devupstream:qcom = "-1"
--- a/ci/base.lock.yml
+++ b/ci/base.lock.yml
@@ -5,7 +5,7 @@ overrides:
         oe-core:
             commit: 06dd66e6220e5ce4ed4b9af4d8231ae5f0a8ce80
         meta-lts-mixins:
-            commit: 348a9eaaf3991e2329d0cef14dce23fabaef9191
+            commit: 11dcf7de2b3cf74fd11049684903b73206a69df6
         bitbake:
             commit: 22021758e66737bcf68dfd2b74adc6a0cb1d42d9
         meta-arm:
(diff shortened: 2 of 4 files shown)


quaresmajose:
For me, this fails locally. For some reason the workflow it's not running. Maybe because the target branch changed, can you please force push again?

lumag:
Inreresting. I now also can see the error with this version (there were no errors on top of the previous PR).

lumag:
Okay, I think I figured it out. If the test passes, I'll push it here.

lumag:
Okay, this PR passes the yocto-check-layer

quaresmajose:
Nice! let me play a bit with it to better understand the issue and the fix.

recipes-kernel/linux-firmware/linux-firmware_20260519.bbappend
```diff
@@ -0,0 +1,2 @@
+DEFAULT_PREFERENCE:qcom = "1"
+DEFAULT_PREFERENCE:class-devupstream:qcom = "-1"
```
quaresmajose, line 2: In my local tests, simply disabling the DEFAULT_PREFERENCE is enough.
```
--- a/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
+++ b/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
@@ -1,4 +1,8 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+# use latest version of linux-firmware from the lts-linux-firmware-mixin layer
+DEFAULT_PREFERENCE:qcom = ""
+
 # To make the layer pass yocto-check-layer only inherit update-alternatives when building for qualcomm
 ALTERNATIVES_CLASS = ""
 ALTERNATIVES_CLASS:qcom = "update-alternatives"

```

My idea with the variable `LTS_LINUX_FIRMWARE_MIXIN_DEFAULT_PREFERENCE` was to make it simpler to choose the version using only one variable, but it ends up being a new variable, so nobody knows it.
I'll revert it and suggest using `DEFAULT_PREFERENCE:override = ""` above in the readme.

lumag:
Closing, necessary patches are a part of #2542

PR #2624, opened 2026-06-25
ci: fix test result reporting after re-runs and PR updates

ricardosalveti:
Fix the "Test Results" check which can contain failures even after a fully green re-run by dropping stale re-run jobs and aborting older runs after the PR gets updated.

### DIFF
--- a/.github/actions/test-job-summary/action.yml
+++ b/.github/actions/test-job-summary/action.yml
@@ -45,7 +45,28 @@ runs:
           INPUTS_SUMMARY_TITLE: ${{ inputs.summary_title }}
         run: |
           set -x
-          INPUT=$(for TESTJOB in $(find "${GITHUB_WORKSPACE}" -name "${INPUTS_PREFIX}.json")
+          # When a failed LAVA job is re-run, GitHub keeps the original failed
+          # job's "<distro>-test-job-<id>" artifact AND adds a new artifact for the
+          # re-submitted job (LAVA assigns a fresh, higher id). Both get downloaded
+          # here, so without de-duplication the stale failed job is counted again
+          # and the summary reports phantom failures even after a green re-run.
+          # Keep only the newest LAVA job (highest id) per device + test, keyed by
+          # (requested_device_type, description) which is stable across re-runs of
+          # the same job. The fields are read from the saved job-detail JSON.
+          declare -A KEEP_FILE
+          declare -A KEEP_ID
+          for TESTJOB in $(find "${GITHUB_WORKSPACE}" -name "${INPUTS_PREFIX}.json")
+          do
+              KEY=$(jq -r '"\(.requested_device_type)@@\(.description)"' "${TESTJOB}")
+              JOB_ID=$(jq -r '.id' "${TESTJOB}")
+              if [ -z "${KEEP_ID[$KEY]:-}" ] || [ "${JOB_ID}" -gt "${KEEP_ID[$KEY]}" ]; then
+                  KEEP_ID["$KEY"]="${JOB_ID}"
+                  KEEP_FILE["$KEY"]="${TESTJOB}"
+              fi
+          done
+          TESTJOBS=$(printf '%s\n' "${KEEP_FILE[@]}")
+
+          INPUT=$(for TESTJOB in ${TESTJOBS}
           do
               JOB_ID=$(cat "$TESTJOB" | jq ".id")
               JOB_URL="https://lava.infra.foundries.io/results/$JOB_ID"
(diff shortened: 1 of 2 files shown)


ricardosalveti:
@mwasilew not sure if we have an easy way to validate this, we could push in next but that won't really validate the PR workflow.

ricardosalveti:
@lumag it seems you are validating another change in next, let me know once you are done there.

lumag:
> @lumag it seems you are validating another change in next, let me know once you are done there.

Please proceed, if you have something to test.

mwasilew:
The change for de-duplication looks OK. As for PR cancelling it's more tricky. I have a repository where I can test it. The main issue is that the cancellation doesn't propagate to LAVA jobs. If the jobs were submitted, they will run till the end. I would prefer cancelling them as well, but there seem to be no easy mechanism for doing so. I'll test it in my repository and then push to `next`

ricardosalveti:
@mwasilew were you able to give it a try?

mwasilew:
> Milosz Wasilewski (@mwasilew) were you able to give it a try?

not yet. I tried to mock the lava jobs part and it partly worked, but didn't reproduce the same behaviour as with one failed test. I think I'll need to use real jobs and devices in the LAB, but I didn't want to get in the way of potential GA testing. Will continue next week.

.github/actions/test-job-summary/action.yml
```diff
@@ -45,7 +45,28 @@ runs:
           INPUTS_SUMMARY_TITLE: ${{ inputs.summary_title }}
         run: |
           set -x
-          INPUT=$(for TESTJOB in $(find "${GITHUB_WORKSPACE}" -name "${INPUTS_PREFIX}.json")
+          # When a failed LAVA job is re-run, GitHub keeps the original failed
```
mwasilew, line 48: When testing this patch I noticed we might also need something like this
```
diff --git a/.github/actions/test-job-summary/action.yml b/.github/actions/test-job-summary/action.yml
index 263618d..b48a2b0 100644
--- a/.github/actions/test-job-summary/action.yml
+++ b/.github/actions/test-job-summary/action.yml
@@ -84,7 +84,7 @@ runs:
               TEST_RESULTS=$(for SUITE in $(echo "$JOB_SUITES" | jq -r -c ".results[]")
               do
                   if [ "${JOB_TYPE}" = "test" ]; then
-                      if [ "${JOB_STATE}" = "Finished" ] && [ ${JOB_HEALTH} = "Complete" ]; then
+                      if [ "${JOB_STATE}" = "Finished" ]; then
                           SUITE_NAME=$(echo "$SUITE" | jq -r ".name")
                           SUITE_ID=$(echo "$SUITE" | jq -r ".id")
                           SUITE_TESTS=$(curl -s "https://lava.infra.foundries.io/api/v0.2/jobs/$JOB_ID/suites/$SUITE_ID/tests/")
```
Without it all `Incomplete` jobs with tests (not boot jobs) are discarded

mwasilew (changes requested):
@ricardosalveti I think it works as expected. Newer PR cancels the workflow for older version. This doesn't cancel the test jobs that were already submitted and I don't think it's possible in the current setup. 

When testing I discovered one more reporting issue and posted a proposed patch as a comment. If you could add this patch here I think it will be ready to go to `next`

.github/actions/test-job-summary/action.yml
ricardosalveti: Added https://github.com/qualcomm-linux/meta-qcom/pull/2624/commits/b648deb7d139f029d4603b46b26ef859dc3ce13f to the set.

ricardosalveti:
> When testing I discovered one more reporting issue and posted a proposed patch as a comment. If you could add this patch here I think it will be ready to go to `next`

Updated, do you want to push to next or should I just do it?

mwasilew:
Pushed to `next` https://github.com/qualcomm-linux/meta-qcom/actions/runs/28923142631

mwasilew:
I'm very pleased to announce it worked as expected. I had to re-run one test job and reporting looks green now:
https://github.com/qualcomm-linux/meta-qcom/actions/runs/28972073378.

mwasilew (approved):

mwasilew:
I thought I approved it before. Done properly now. However the PR is still in "draft". @ricardosalveti could you transition it to a "ready for review" state?

ricardosalveti:
> I thought I approved it before. Done properly now. However the PR is still in "draft". Ricardo Salveti (@ricardosalveti) could you transition it to a "ready for review" state?

Done, thanks!

quaresmajose (approved):

PR #2629, opened 2026-06-26
ci/test-oe-nogl: drop the patch applied upstream

lumag:
The commit 29f2ed33a64a ("Revert "ci/qcom-distro: fix building MariaDB for Glymur"") dropped the patch
0001-mariadb-fix-building-for-the-ARMv8.3-A-and-later-sys.patch, but didn't update the separate test-oe-nogl.yml, which is also using that patch.

Fixes: 7a82ae52a531 ("ci: base.lock: update layers to latest  (#2461)")

### DIFF
--- a/ci/qcom-distro.yml
+++ b/ci/qcom-distro.yml
@@ -30,7 +30,7 @@ repos:
     branch: main
 
   meta-virtualization:
-    url: https://git.yoctoproject.org/git/meta-virtualization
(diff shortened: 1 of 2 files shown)


lumag:
Rebased on top of #2630 to let this PR pass the CI.

quaresmajose (dismissed):

lumag:
Force-merging to fix the CI

PR #2630, opened 2026-06-26
ci/qcom-distro: update the URL for meta-virtualization

lumag:
Using https://git.yoctoproject.org/git/meta-virtualization generates fetching errors from time to time. Follow other Yocto Project repositories and use https://git.yoctoproject.org/meta-virtualization as a URL.

PR #2632, opened 2026-06-26
CI: Fix meta-virtualization repository URL

vkraleti:
The "/git/" URL format used for meta-virtualization is deprecated  and causing fetcher failures. Fix this by updating the repository URL to the standard form.


lumag:
See #2629, #2630

PR #2634, opened 2026-06-26
Handle proprietary licences for the firmware / Heaxgon packages

lumag:
While it's not infrequent for the vendor to not define a proper licensing terms for the firmware files and Hexagon DSP binaries in some cases it is possible to establish those. Add infrastructure to declare the licence for firmware packages.

### DIFF
--- a/classes-recipe/nhlos-license.bbclass
+++ b/classes-recipe/nhlos-license.bbclass
@@ -1,0 +1,35 @@
+#
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+#
+# SPDX-License-Identifier: MIT
+#
+# If specified, NHLOS_LICENSE and NHLOS_LICENSE_MD5 specify the licence file
+# (and the checksum) for the NHLOS binaries.
+
+NHLOS_LICENSE ??= ""
+NHLOS_LICENSE_MD5 ??= ""
+
+# LICENSE needs to be set directly as base.bbclass verified it before the
+# anonymous python function has a chance to kick in.
+LICENSE = "${@ 'NHLOS-${BPN}' if d.getVar('NHLOS_LICENSE') else 'CLOSED'}"
+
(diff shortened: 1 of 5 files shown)


lumag:
See https://github.com/qualcomm-linux/meta-qcom-3rdparty/pull/29 for the example of the usage.

classes-recipe/nhlos-license.bbclass
```diff
@@ -0,0 +1,35 @@
...
+
+NHLOS_LICENSE ??= ""
+NHLOS_LICENSE_MD5 ??= ""
+
+# LICENSE needs to be set directly as base.bbclass verified it before the
+# anonymous python function has a chance to kick in.
+LICENSE = "${@ 'NHLOS-${BPN}' if d.getVar('NHLOS_LICENSE') else 'CLOSED'}"
+
+def handle_nhlos_license(d, pkgs):
+    if d.getVar("NHLOS_LICENSE"):
```
koenkooi, line 17: If someone sets `NHLOS_LICENSE`, but not `NHLOS_LICENSE_MD5`, do we want to error out here, or later when the offending recipe gets built?

lumag: With the empty `NHLOS_LICENSE_MD5` bitbake will print out the expected md5 sum, so I think it's worth keeping it as is.

koenkooi: That is a more graceful error than I'd expected. I expected `md5=` to trigger a parse error during the recipe build.

ricardosalveti (dismissed):

ricardosalveti (dismissed):

ricardosalveti (dismissed):

ricardosalveti (dismissed):

ricardosalveti (dismissed):

koenkooi (dismissed):

lumag:
Merging based on the previous approvals.

PR #2635, opened 2026-06-26
Update revision of rmtfs to latest commit

### LINKED ISSUE #2582: QCM6490/IQ9/IQ8:Boot delay due to rmtfs.service restart loop on non-modem platforms
boot-time regression issue caused by the rmtfs.service on non-MSS platforms, specifically IQ-9075 devices running QLI 6.18.y The service is designed for modem synchronization but is enabled unconditionally, leading to repeated failures and unnecessary CPU churn during early boot when no modem subsystem is present. Logs indicate failures in rproc_init() due to missing MSS-related entries in /sys/class/remoteproc/*/device/modalias, triggering a continuous restart loop and inflating boot times by ~10 seconds. Suggested improvements include conditional service activation using ConditionPathExistsGlob=/sys/class/remoteproc/*mpss* to avoid impacting non-modem platforms. The technical root cause lies in the rmtfs -s option invoking rproc_init() without platform capability checks, resulting in failed initialization and repeated service restarts.

Steps to reproduce: 

1. take the boot numbers

2. check the rmtfs.service time

roshs189:
Update recipe to bring in commits which execute the service only after
 /dev/qcom_rmtfs_mem1 is created and doesn't start it if the node doesn't exist(indicating modemless device).

Fixes #2582

### DIFF
--- a/recipes-support/rmtfs/rmtfs_1.3.bb
+++ b/recipes-support/rmtfs/rmtfs_1.3.bb
@@ -7,8 +7,8 @@ LIC_FILES_CHKSUM = "file://LICENSE;md5=ca25dbf5ebfc1a058bfc657c895aac2f"
 
 inherit systemd
 
-SRCREV = "44facf5694036ebda53fd09c9535774982df5247"
-SRC_URI = "git://github.com/linux-msm/${BPN}.git;branch=master;protocol=https;tag=v1.1.1"
+SRCREV = "b30a3eb38f9af283f18dbd3c7755653efc52c094"
+SRC_URI = "git://github.com/linux-msm/${BPN}.git;branch=master;protocol=https;tag=v${PV}"
 DEPENDS = "qmic-native qrtr udev"
 
 do_install () {


roshs189:
https://github.com/linux-msm/rmtfs/pull/35
https://github.com/linux-msm/rmtfs/pull/36
These are the original fixes.

ricardosalveti:
Commit message not following the expected pattern, see git log.

Also, I would prefer if we could track upstream releases, or backport changes that were already merged, instead of tracking git. @lumag do we expect to create a new release for rmtfs soon?

lumag:
> Commit message not following the expected pattern, see git log.
> 
> Also, I would prefer if we could track upstream releases, or backport changes that were already merged, instead of tracking git. @lumag do we expect to create a new release for rmtfs soon?

Done. Please use [v1.3](https://github.com/linux-msm/rmtfs/releases/tag/v1.3)

lumag (changes requested):
Switch to release 1.3

roshs189:
Done.

lumag (changes requested):
> Update revision to release 1.3

Update which recipe? Please use `git log` as an inspiration. Also please fix your editor to wrap commit messages at 72 char boundary.

roshs189:
Thanks for the suggestions, could you check now?

lumag (approved):

ricardosalveti (approved):

PR #2636, opened 2026-06-26
ci/base: apply a patch, fixing the IGT build failures

lumag:
Temporarily pick up a patch fixing the IGT build failure in debug builds.

Fixes: 7a82ae52a531 ("ci: base.lock: update layers to latest  (#2461)")
Link: https://lore.kernel.org/r/20260626201256.112352-1-dmitry.baryshkov@oss.qualcomm.com/

PR #2637, opened 2026-06-26
igt-gpu-tools: backport amd_basic spin-wait portability fix

ricardosalveti:
amd_basic.c hardcodes the x86 'pause' instruction as a spin-wait CPU hint, which is not a valid mnemonic on other architectures. The amdgpu tests therefore fail to cross-build for aarch64 whenever that code is actually emitted, which only happens in unoptimized (debug) builds: at -O2 the compiler discards the hint as dead code, but at -O0 it is assembled and the build breaks with:

  {standard input}: Error: unknown mnemonic `pause' -- `pause'

This breaks the GitHub CI debug builds, i.e. the ones using ci/debug.yml, which sets DEBUG_BUILD = "1" and therefore builds at -O0.

Carry a bbappend, scoped to the qcom override, that applies a backport making the hint architecture-portable. The fix has been submitted upstream but is not yet accepted/merged, so it is proposed here in meta-qcom first rather than in oe-core.


ricardosalveti:
Just saw https://github.com/qualcomm-linux/meta-qcom/pull/2636, will close this one.

lumag:
Your patch is more comprehensive. Would you send it to the IGT list?

ricardosalveti:
> Your patch is more comprehensive. Would you send it to the IGT list?

Sent to https://lists.freedesktop.org/archives/igt-dev/2026-June/112474.html.

PR #2528, opened 2026-06-18
qcom-rtss-mailbox: add KMD and UMD recipes for RTSS mailbox IPC

snegi-qti:
The RTSS (Real Time Subsystem) currently has no IPC path to the application processor, and userspace applications have no stable API to communicate with it. These two recipes  together establish the full mailbox IPC stack.

  - qcom-rtss-mailbox-kmd: Adds the DLKM recipe for the rtss_mailbox kernel module, which is the prerequisite IPC driver without which no communication between RTSS and the APSS is possible. .
  - qcom-rtss-mailbox-umd: Adds the middleware recipe that gives userspace applications a stable API over the mailbox channel — isolating the kernel UAPI from application code and enabling independent versioning of the mailbox and OTA SDK components. Builds librtss_mailbox, librtss_safemlib, and librtss_ota (shared + static), plus rtssdbg, rtss_console, and rtss_ota debug/utility binaries. Output is split into runtime, dev, staticdev, and utils subpackages for fine-grained install control.

PR #2644, opened 2026-06-28
qcom-rtss-can: add recipe for RTSS CAN userspace deamon

q-AnupKulkarni:
Add a BitBake recipe for the RTSS CAN userspace daemon. This daemon acts
as a gateway between SocketCAN applications and the RTSS subsystem using
the RTSS mailbox UMD libraries (qcom-rtss-mailbox-umd).

Depends on qcom-rtss-mailbox-umd for librtss_mailbox.so and on
linux-libc-headers for linux/can.h and linux/can/raw.h.

Tested on: iq-9075-evk

### DIFF
--- a/recipes-support/qcom-rtss-can/qcom-rtss-can_1.0.0.bb
+++ b/recipes-support/qcom-rtss-can/qcom-rtss-can_1.0.0.bb
@@ -1,0 +1,52 @@
+SUMMARY = "Qualcomm RTSS CAN userspace daemon"
+DESCRIPTION = "Userspace daemon acting as a gateway between SocketCAN \
+applications and RTSS using the RTSS mailbox UMD libraries."
+
+HOMEPAGE = "https://github.com/qualcomm/rtss-can"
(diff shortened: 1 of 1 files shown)


q-AnupKulkarni:
Depends on https://github.com/qualcomm-linux/meta-qcom/pull/2528

lumag (commented):
This is what we talked about earlier, uAPI. Don't invent extra API. Provide the standard uAPI. Linux has AF_CAN. As such, if you want to provide CAN support, hook into the existing subsytem instead of providing something new.

This is NAK from my side.

lumag (changes requested):
This is what we talked about earlier, uAPI. Don't invent extra API. Provide the standard uAPI. Linux has AF_CAN. As such, if you want to provide CAN support, hook into the existing subsytem instead of providing something new.

This is NAK from my side.

lumag:
To point out, provide AF_CAN by the kernel, there should be no need for the extra userspace bridges.

PR #2652, opened 2026-06-29
README.md: provide commit guidelines

lumag:
Based on the lots of similar feedback we've had to give to contributors, codify contributing guidelines, saving maintainers from typing the same text again and again. Especially frown upon the bullet lists, which some engineer prefer instead of writing the text.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -114,18 +114,14 @@ Use the helper scripts for CI parity whenever possible.
 
 ## 6) Pull request / contribution workflow
 
-Follow the repository `README.md` contribution flow:
+Follow the contribution workflow documented in
+[CONTRIBUTING.md](CONTRIBUTING.md):
 
 1. Target branch: **master**.
 2. Fork `qualcomm-linux/meta-qcom`, create a topic branch, implement changes.
 3. Rebase on latest upstream `master`.
 4. Open a GitHub pull request.
 5. Use PR discussion for review iteration.
-
-Important:
-
-- Follow Yocto submission guidance referenced in README:
-  [Preparing Changes for Submission](https://docs.yoctoproject.org/dev/contributor-guide/submit-changes.html#preparing-changes-for-submission)
 
 Before opening/updating a PR, run CI-equivalent checks in this order:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -151,6 +151,37 @@ For some useful guidelines when submitting patches, please refer to:
...
 
+### Commit messages
+
+Each commit must be atomic: it must contain exactly one logical change. Do not
+squash multiple features, fixes, or otherwise unrelated changes into a single
+commit — split them into separate commits, one per logical change.
+
+Each commit must contain a well-formed commit subject and message.
+
+The commit subject must follow the form `recipe-name: summary of the
```
ndechesne: it makes sense. However for this commit in particular, it ends up as

docs: xxx

while you are updating AGENTS.md. I am not sure this is right..

lumag: Changed to 'README, AGENTS'

AGENTS.md
```diff
@@ -185,7 +185,7 @@ Guidelines:
 - Avoid unnecessary bullet lists in the body; prefer prose paragraphs.
 - Wrap body lines for readability (~72 chars).
 - Use consistent recipe bump wording for version updates, e.g.
-  `recipe-name: Update to vX.Y.Z`.
+  `recipe-name: upgrade vX.Y.Z -> vA.B.C`.
```
koenkooi: If we're talking to AI agents, can we please use a proper, U+2192 UTF8 right arrow, →?

koenkooi:
I'm not sure about dropping PR numbers completely, would use the full URL, e.g. https://github.com/qualcomm-linux/meta-qcom/pull/2652 be better?

I do agree with the numbers alone not being as useful, I have been manually 'cherry-picking' patches between repos (e.g. tflite patches from meta-qcom-distro into meta-ai) and the PR numbers stop making sense.

AGENTS.md
lumag: That's would deviate from the format used by OE.

lumag:
It was talking about the PR number in the subject. I think that's pretty useless. And we don't do that. I think the previous iteration of the `AGENT.md` picked it up from the merge commits (which do have a PR number - of the merged PR).

README.md
```diff
@@ -151,6 +151,37 @@ For some useful guidelines when submitting patches, please refer to:
...
+- not merely restate *what* the diff changes line by line — the diff
+  already shows that;
+- avoid unnecessary bullet lists; prefer prose paragraphs.
+
+Every commit must also carry a `Signed-off-by` trailer matching the
+author identity from your local `git config` (use `git commit -s`).
+
+If an AI coding assistant or other advanced tool was used to help create the
+change, acknowledge that use by adding an `Assisted-by` trailer naming the
+tool, for example `Assisted-by: Claude <noreply@anthropic.com>`.
```
ricardosalveti: This will make the README a bit too long, I think it would be better to have a docs/contributing.md or similar and a link here instead.

AGENTS.md
```diff
@@ -161,14 +164,31 @@ Signed-off-by: $(git config user.name) <$(git config user.email)>
 
 Never fabricate a name or email; always read from `git config`.
 
+If an AI coding assistant or other advanced tool was used to help create the
+change, acknowledge that use by adding an `Assisted-by` trailer naming the
+tool, e.g.:
+
+```text
+Assisted-by: Claude <noreply@anthropic.com>
```
ricardosalveti: Can we use `Assisted-by: AGENT_NAME:MODEL_VERSION` or similar? We should not promote the tool / company here :-)

AGENTS.md
```diff
@@ -161,14 +164,31 @@ Signed-off-by: $(git config user.name) <$(git config user.email)>
...
+```
```
kraj: maybe add back reference to - https://docs.yoctoproject.org/contributor-guide/submit-changes.html#acceptance-of-ai-generated-code

lumag: Heh, I missed that document. Why couldn't major projects settle on a single way to mark AI contributions? The kernel enforces ` Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]`

ricardosalveti: Our internal guides for projects maintained in the open is also aligned with `Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]`, following the kernel guidelines. Maybe we should update yocto to also follow the kernel one?

README.md
lumag: Done

ricardosalveti (approved):

koenkooi (approved):

ricardosalveti:
@lumag mind backporting to wrynose?

PR #2665, opened 2026-07-01
shikra-evk: add phone to MACHINE_FEATURES

mdembla-bit:
shikra-evk does not support cellular data calls, as modemmanager is
not enabled for this machine.

modemmanager is conditionally enabled at the distro layer based on
'phone' being present in a machine's MACHINE_FEATURES. Add 'phone'
to MACHINE_FEATURES for shikra-evk so that modemmanager gets enabled
and cellular data connectivity is supported.

### DIFF
--- a/conf/machine/shikra-evk.conf
+++ b/conf/machine/shikra-evk.conf
@@ -4,7 +4,7 @@
 
 require conf/machine/include/qcom-shikra.inc
 
-MACHINE_FEATURES += "efi pci"
+MACHINE_FEATURES += "efi pci phone"
 
 KERNEL_DEVICETREE ?= " \
                       qcom/shikra-cqm-evk.dtb \


lumag (changes requested):
Pleasev start your commit message by describing the issue you are trying to solve.

mdembla-bit:
> Pleasev start your commit message by describing the issue you are trying to solve.

Hi @lumag , updated the commit text, as well as PR description. Please let me know if this is fine?

lumag:
> > Pleasev start your commit message by describing the issue you are trying to solve.
> 
> Please let me know if this is fine?

Did you start your commit message with the description of the issue? No, you didn't. So you can guess my answer. Also please drop the description of the change. "This change does this" is completely redundant, you can see it from the patch itself. Why do you need to repeat yourself?

mdembla-bit:
> > > Pleasev start your commit message by describing the issue you are trying to solve.
> > 
> > 
> > Please let me know if this is fine?
> 
> Did you start your commit message with the description of the issue? No, you didn't. So you can guess my answer. Also please drop the description of the change. "This change does this" is completely redundant, you can see it from the patch itself. Why do you need to repeat yourself?

I've updated the commit message based on your feedback to better explain the problem being solved and the role of phone in enabling ModemManager support.
Could you please take another look?

lumag (changes requested):
Why do we want to support cellular calls on Shikra?

mdembla-bit:
> Why do we want to support cellular calls on Shikra?

Shikra includes modem support and cellular data connectivity is an expected platform capability. This change enables the standard ModemManager userspace stack required to make use of the modem for WWAN data calls.

lumag:
> Shikra includes modem support and cellular data connectivity is an expected platform capability

This should be a part of the commit message. Which board has a modem? Is it a PCIe modem or a part of the SoC, etc.

mdembla-bit:
> > Shikra includes modem support and cellular data connectivity is an expected platform capability
> 
> This should be a part of the commit message. Which board has a modem? Is it a PCIe modem or a part of the SoC, etc.

Thanks. I've updated the commit message accordingly. It now explicitly states that shikra-evk requires WWAN connectivity support for the modem integrated into the SoC and explains that the phone MACHINE_FEATURE is needed to enable the required ModemManager userspace support.

lumag (changes requested):
> shikra-evk lacks WWAN connectivity support for the modem integrated
into the SoC.

Lacks, ok. Does it specify, that Shikra has internal modem? No. Which SKUs have it? Does it make sense to split configuration into two different machines (like it is done for QCM6490-IDP vs RB3 Gen2) or does it make sense to keep all of them enabled as a single machine entry?
Is the cellular modem supported by the published qdsp6sw firmware or does it require a separate firmware file, to be licenced from Qualcomm?

vkraleti:
> Which SKUs have it?

The CQM variant supports an integrated CAT4 cellular modem in addition to WLAN running on the Q6. The CQS variant supports Audio and WLAN on the Q6, but does not include cellular modem functionality.

> Does it make sense to split configuration into two different machines (like it is done for QCM6490-IDP vs RB3 Gen2) or does it make sense to keep all of them enabled as a single machine entry?

Maintaining separate machine configurations does not provide much value as the there is no difference in the publicly available firmware between the CQM and CQS variants. Even the CDT is shared by these variants. 

> Is the cellular modem supported by the published qdsp6sw firmware or does it require a separate firmware file, to be licenced from Qualcomm?

The qdsp6sw firmware published in linux-firmware.git does not include support for the cellular modem. Cellular modem functionality requires an additional modem firmware package that is provided separately to licensed customers.

lumag:
Thanks! Now you can turn it into a nice commit message.

ricardosalveti:
Yeah, please fix the commit message, we do have variants which do have such "machine feature", it is just that we are using one board conf supporting multiple variants, which is fine.

mdembla-bit:
Thanks. I've pushed an updated commit message covering the CQM integrated modem support and the shared shikra-evk configuration.

koenkooi (approved):

lumag (approved):

ricardosalveti:
Please rebase to clear out the yocto-run-checks failure.

mdembla-bit:
> Please rebase to clear out the yocto-run-checks failure.

Thanks, Ricardo. I've rebased onto the latest master and force-pushed the branch.

PR #2678, opened 2026-07-04
[Backport wrynose] Add xdp-tools and rtc-testbench recipes

raj-patel412:
Add xdp-tools and rtc-testbench recipes under the
openembedded-layer dynamic layer for wrynose.

These recipes are imported from meta-openembedded/meta-oe because they are
needed for Qualcomm Linux wrynose, but cannot be backported through
meta-openembedded for this stable branch.

rtc-testbench depends on libxdp from xdp-tools, so both recipes are carried
together in meta-qcom. The recipes are placed under
dynamic-layers/openembedded-layer so they are only exposed when meta-oe is
present in BBLAYERS.

Imported from meta-openembedded/meta-oe:
- xdp-tools current recipe state, including:
  - cdd53d4d0d xdp-tools: add recipe
  - b1f98d1a7a xdp-tools: upgrade 1.2.10 -> 1.6.3
- 1eebf6bd51 rtc-testbench: add recipe


lumag (changes requested):
Wrong layer and wrong place. It has nothing to do with the BSP layer. 
I
Yesterday I have pushed a correct implementation to https://github.com/qualcomm-linux/meta-qcom-distro/pull/386

Closing this one.

PR #2688, opened 2026-07-07
workflows/compile: mount kvm device on kas container

quaresmajose:
The KVM is necessary when we need to emulate a full systems, considerably improvising the use of qemu. This is necessary to efficiently execute the OE-core runqemu inside the kas container helper.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -61,8 +61,11 @@ jobs:
         shell: bash
         run: |
           KAS_CONTAINER=$GITHUB_WORKSPACE/kas-container
-          echo "KAS_CONTAINER=$KAS_CONTAINER" >> $GITHUB_ENV
           chmod +x $KAS_CONTAINER
+          if [ -c /dev/kvm ]; then
+            KAS_CONTAINER="$KAS_CONTAINER --kvm"
+          fi
+          echo "KAS_CONTAINER=$KAS_CONTAINER" >> $GITHUB_ENV
 
       - name: Setup build variables
         shell: bash
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -41,6 +41,7 @@ on:
 
 env:
   KAS_CLONE_DEPTH: 1
+  KAS_IMAGE_VERSION: "latest"
 
 jobs:
   reusable_compile_job:
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -68,8 +68,11 @@ jobs:
         shell: bash
         run: |
           KAS_CONTAINER=$RUNNER_TEMP/kas-container
-          echo "KAS_CONTAINER=$KAS_CONTAINER" >> $GITHUB_ENV
           chmod +x $KAS_CONTAINER
+          if [ -c /dev/kvm ]; then
+            KAS_CONTAINER="$KAS_CONTAINER --kvm"
+          fi
+          echo "KAS_CONTAINER=$KAS_CONTAINER" >> $GITHUB_ENV
 
       - name: Setup build variables
         shell: bash
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -28,6 +28,7 @@ concurrency:
 env:
   CACHE_DIR: /efs/qli/meta-qcom
   KAS_CLONE_DEPTH: 1
+  KAS_IMAGE_VERSION: "latest"
 
 jobs:
   kas-setup:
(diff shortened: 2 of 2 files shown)


quaresmajose:
Ideally, I shouldn't use `chmod`. But inside the container and outside, the group that has access to the /dev/kvm device is not the same.

.github/workflows/compile.yml
```diff
@@ -126,6 +126,12 @@ jobs:
         if:  inputs.target
         shell: bash
         run: |
+          if [ -e /dev/kvm ]; then
+            # default mod of /dev/kvm is 0660 and this works well on the host
+            # but host user group id don't exist inside the container
+            sudo chmod 0666 /dev/kvm
```
lumag: This changes access outside of the container. Can we fix the mode inside the container instead?

quaresmajose: I will analyze it.

koenkooi: Since it's a bind mount, mode changes propagate back to the host :/

.github/workflows/compile.yml
```diff
@@ -126,6 +126,12 @@ jobs:
         if:  inputs.target
         shell: bash
         run: |
+          if [ -e /dev/kvm ]; then
```
ricardosalveti: Do we only need KVM for this build step?

ricardosalveti: And we can probably use -c here.

quaresmajose: > Do we only need KVM for this build step?

It's only in this step, but perhaps it makes sense to make it available whenever it exists. 
I'll do that when creating the KAS_CONTAINER variable in the first step. I will also use the '-c'.

quaresmajose:
To test the feature, we need a machine with kvm, so it would be better to use https://github.com/qualcomm-linux/meta-qcom/pull/1900

.github/workflows/compile.yml
quaresmajose: > This changes access outside of the container. Can we fix the mode inside the container instead?

For that, I need to modify the container and save a copy to be used, which is what I would like to avoid.

quaresmajose: > Since it's a bind mount, mode changes propagate back to the host :/

I tested modifying it inside the container, but on the host I don't see the change, and the permissions remain the same. I also hope the modifications have an impact outside the container; the device is the same.

quaresmajose:
I will add support for this use case in kas.

quaresmajose:
Proposed "--kvm" to kas-container
https://groups.google.com/g/kas-devel/c/r8Po5O_u9RQ

.github/workflows/compile.yml
quaresmajose: Fixed with the introduction of "--kvm" in kas-container.

quaresmajose:
> Proposed "--kvm" to kas-container https://groups.google.com/g/kas-devel/c/r8Po5O_u9RQ

Landed in master branch.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/ffbb7b7b-1545-43ae-a269-5855e2e84635)**
---
**Test jobs for commit [`0c62cc5`](https://github.com/qualcomm-linux/meta-qcom/commit/0c62cc56b1c5649ec37f5c1f6e269bd3a00d4b0c)**
---
Pass: 9 | Fail: 0 | Total: 9

lumag (dismissed):

quaresmajose:
Everything indicates that a new version of kas will be released next week. If we wait, we can discard the second patch.

ricardosalveti:
I would prefer to wait for the released version if indeed next week.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/25952916-b3ef-44c5-a938-35b626bdbe54)**
---
**Test jobs for commit [`2bbb008`](https://github.com/qualcomm-linux/meta-qcom/commit/2bbb008b14114e68ebbdb300c3722a93cbbfb97a)**
---
Pass: 9 | Fail: 0 | Total: 9

quaresmajose:
> I would prefer to wait for the released version if indeed next week.

kas 5.5 was released https://groups.google.com/g/kas-devel/c/Kc_nRi8wnlQ/m/W4ogLshIAgAJ

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/4c4827b2-799b-4085-a34f-ce04698557de)**
---
**Test jobs for commit [`05b1407`](https://github.com/qualcomm-linux/meta-qcom/commit/05b140797a4e37a0924483b59c464e67a3bda19c)**
---
Pass: 9 | Fail: 0 | Total: 9

ricardosalveti (approved):

lumag (approved):

quaresmajose:
> Git push to origin failed for wrynose with exitcode 1

Proposed in https://github.com/qualcomm-linux/meta-qcom/pull/3016

PR #2693, opened 2026-07-08
ci: skip uploading redundant build artifacts to S3

ricardosalveti:
The artifact staging step publishes every symlinked file from the
deploy directory, which uploads several copies of the same content: the
versioned '*--<version>*' symlinks duplicate the machine-suffixed
names, and the '<type>-<machine>.bin' kernel symlinks duplicate Image,
Image.gz, zImage and vmlinux (vmlinux alone accounts for two 450 MB
copies per job). For machines flashed via the qcomflash tarball, the
raw rootfs '.ext4' (multiple GBs per image, uncompressed), the 512 MB
ESP '.vfat' and vmlinux are already shipped inside the tarball as
rootfs.img, efi.bin and vmlinux, so their standalone copies only
inflate the upload. On top of that, the world build deploys the same
initramfs images (firmware and test variants, around 120 files and
1.3 GiB) for every machine, duplicating near-identical content across
every job of a run. A single qcom-distro job currently uploads around
32 GiB, most of it redundant.

Skip the duplicated symlink names, skip the raw rootfs ext4, the ESP
vfat and vmlinux whenever a qcomflash tarball is part of the upload
(anyone needing those files can extract them from the tarball), and
publish the world-built initramfs images only from the generic
qcom-armv8a and qcom-armv7a machines. The initramfs-rootfs-image
artifacts are still published for every machine, as that image is part
of the machine's UKI boot chain. The qcom-armv8a artifacts consumed by
the LAVA fastboot jobs (boot images and the compressed rootfs) are
unaffected, as that machine does not produce a qcomflash tarball.

### DIFF
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -169,8 +169,25 @@ jobs:
           uploads_dir=./uploads/${INPUTS_DISTRO_NAME}${INPUTS_KERNEL_DIRNAME}/${INPUTS_MACHINE}
           mkdir -p $uploads_dir
           if [ -d $images_dir ]; then
+            # Skip symlink names whose content is also published under another
+            # name: the versioned '*--<version>*' links duplicate the
+            # machine-suffixed ones, and the '<type>-<machine>.bin' kernel
+            # links duplicate Image, Image.gz, zImage and vmlinux.
+            exclude=( ! -name '*--*' ! -name 'vmlinux-*.bin' ! -name 'Image-*.bin' ! -name 'Image.gz-*.bin' ! -name 'zImage-*.bin' )
+            # The qcomflash tarball already ships the rootfs (rootfs.img), the
+            # ESP (efi.bin) and vmlinux, so skip their standalone copies when
+            # the tarball is published (uncompress or extract them if needed)
+            if compgen -G "$images_dir/*.qcomflash.tar.gz" > /dev/null; then
+              exclude+=( ! -name '*.rootfs.ext4' ! -name 'esp-qcom-image-*.vfat' ! -name 'vmlinux' )
+            fi
+            # The world build deploys the same initramfs images for every
+            # machine, so publish them only from the generic machines
+            case "${INPUTS_MACHINE}" in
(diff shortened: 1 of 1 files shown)


lumag:
So, is the problem in the build time or in the upload size? In the latter case I'd prefer to fix only the second issue for now.

ricardosalveti:
> So, is the problem in the build time or in the upload size? In the latter case I'd prefer to fix only the second issue for now.

If we can optimize both build and upload, it would be better, but I can separate to optimize the upload first.

koenkooi (approved):
Both the abundance of duplicate symlinks as well as S3 mishandling of symlink has bothered me for years, so I can't click 'APPROVE' fast enough!

.github/workflows/compile.yml
```diff
@@ -169,8 +169,25 @@ jobs:
...
           if [ -d $images_dir ]; then
+            # Skip symlink names whose content is also published under another
+            # name: the versioned '*--<version>*' links duplicate the
+            # machine-suffixed ones, and the '<type>-<machine>.bin' kernel
+            # links duplicate Image, Image.gz, zImage and vmlinux.
+            exclude=( ! -name '*--*' ! -name 'vmlinux-*.bin' ! -name 'Image-*.bin' ! -name 'Image.gz-*.bin' ! -name 'zImage-*.bin' )
+            # The qcomflash tarball already ships the rootfs (rootfs.img), the
+            # ESP (efi.bin) and vmlinux, so skip their standalone copies when
+            # the tarball is published (uncompress or extract them if needed)
+            if compgen -G "$images_dir/*.qcomflash.tar.gz" > /dev/null; then
```
quaresmajose, line 180: What are the benefits of using this instead of `ls "$images_dir/*.qcomflash.tar.gz"` ?

quaresmajose (approved):

.github/workflows/compile.yml
ricardosalveti: bash built-in and handles better the case when there is no directory or files there.

PR #2695, opened 2026-07-08
thermald: install default Qualcomm thermal configuration

priyjain1:
Install thermal-conf.xml as the default thermald
configuration file for Qualcomm platforms.

### DIFF
--- a/recipes-bsp/thermald/files/thermal-conf.xml
+++ b/recipes-bsp/thermald/files/thermal-conf.xml
@@ -1,0 +1,66 @@
+<?xml version="1.0"?>
+
+<!-- BEGIN -->
+<ThermalConfiguration>
+
+  <!--
+    Platform-specific thermal mitigation rules for Qualcomm SoC can be added here.
+    Use <SoCModel> to match the platform via /sys/firmware/devicetree/base/model (prefix match).
+    Temperatures are in milli-degrees Celsius (e.g., 95000 = 95°C).
+  -->
+
+  <!--
+  <Platform>
+    <Name>Qualcomm SoC Thermal Policy</Name>
+    <SoCModel>Qualcomm Technologies, Inc. REPLACE_WITH_SOC_NAME</SoCModel>
+    <Preference>ENERGY_CONSERVE</Preference>
+    <ThermalSensors>
+      <ThermalSensor>
+        <Type>cpu-thermal</Type>
+        <Path>/sys/class/thermal/thermal_zone0/temp</Path>
+        <AsyncCapable>0</AsyncCapable>
+      </ThermalSensor>
+    </ThermalSensors>
+    <CoolingDevices>
+      <CoolingDevice>
+        <Index>0</Index>
+        <Type>cpu-freq</Type>
+        <Path>/sys/class/thermal/cooling_device0/cur_state</Path>
+        <MinState>0</MinState>
+        <MaxState>10</MaxState>
+        <IncDecStep>1</IncDecStep>
+      </CoolingDevice>
+    </CoolingDevices>
+    <ThermalZones>
+      <ThermalZone>
+        <Type>cpu-zone</Type>
+        <TripPoints>
+          <TripPoint>
+            <Temperature>85000</Temperature>
+            <Hyst>5000</Hyst>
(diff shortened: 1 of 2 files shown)


recipes-bsp/thermald/files/thermal-conf.xml
```diff
@@ -0,0 +1,66 @@
...
+<!-- BEGIN -->
+<ThermalConfiguration>
+
+  <!--
+    Platform-specific thermal mitigation rules for Qualcomm SoC can be added here.
+    Use <SoCModel> to match the platform via /sys/firmware/devicetree/base/model (prefix match).
+    Temperatures are in milli-degrees Celsius (e.g., 95000 = 95°C).
+  -->
+
+  <!--
```
lumag, line 12: So, you are adding a config file just to add a huuge comment? I don't think it's a good idea. Use documentation instead.

lumag (changes requested):
Please see https://github.com/qualcomm-linux/meta-qcom/blob/a43bf9fdf9796339ba527a34f0c56fe46763b6ac/CONTRIBUTING.md?plain=1#L30

Not to mention that your editor is misconfigured, wrapping commit messages too early.

recipes-bsp/thermald/files/thermal-conf.xml
koenkooi: We should make the end-user do the least amount of work possible, so we can do a few of the following things:

- Generate configs for all SoCs supported by meta-qcom, add then to meta-qcom
- Ship a generator script on-device (which no external dependencies, so most likely bash) that runs on first boot

priyjain1:
> Please see
> 
> https://github.com/qualcomm-linux/meta-qcom/blob/a43bf9fdf9796339ba527a34f0c56fe46763b6ac/CONTRIBUTING.md?plain=1#L30
> 
> Not to mention that your editor is misconfigured, wrapping commit messages too early.

Thanks for pointing the commit-text wrapping issue. I'll rework the commit message to follow the expected format and rewrap the body appropriately before respinning the patch.

recipes-bsp/thermald/files/thermal-conf.xml
priyjain1: > So, you are adding a config file just to add a huuge comment? I don't think it's a good idea. Use documentation instead.

The intention is not to add the file solely for the comment. The plan is to keep the configuration effectively empty for now and use it as the placeholder/default thermald configuration for Qualcomm platforms. Once the thermal mitigation rules are tuned and validated on supported SoCs, those rules will be added to this XML.
Regarding the documentation, thermald already has comprehensive documentation and examples upstream, so I do not see a need for separate meta-qcom documentation. The example rule in the comment was added only as a reference for future contributors who may add thermald policies for other Qualcomm chipsets.
If you feel that the example comment is unnecessary or too verbose, I am happy to remove it and keep the XML minimal until actual platform-specific rules are added.

priyjain1: > We should make the end-user do the least amount of work possible, so we can do a few of the following things:
> 
> * Generate configs for all SoCs supported by meta-qcom, add then to meta-qcom
> * Ship a generator script on-device (which no external dependencies, so most likely bash) that runs on first boot

I agree that we should minimize the amount of work required from end users.
In our case, the intention is to use a single thermal-conf.xml across Qualcomm platforms. I've already added Qualcomm SoC detection logic in thermald itself (PR (under review ): https://github.com/intel/thermal_daemon/pull/584) , so the policy can adapt based on the detected platform rather than requiring a separate XML per SoC.
The current XML is primarily a placeholder to establish the integration path. As we tune and validate thermal mitigation policies, the corresponding rules will be added incrementally. 
Also, installing thermal-conf.xml from meta-qcom would only provide a default configuration. Customers can still override it. Let me know your thoughts on this.

lumag: Well. Please don't commit placeholders. Once you have actual rules, please commit them here.

ricardosalveti: No need for a placeholder, add once we have something that brings value here.

PR #2698, opened 2026-07-08
tpm2: allow tpm2-abrmd D-Bus chat with initrc_t for adbd

khalidfaisal17:
On Qualcomm platforms adbd (Android Debug Bridge daemon) runs as initrc_t, confirmed on IQ-9075 EVK:

    # ps -eZ | grep adbd
    system_u:system_r:initrc_t:s0  1649  ?  00:00:00 adbd

tpm2-tools invoked from an ADB shell session run in initrc_t context. Without tpm2_dbus_chat_abrmd(initrc_t), tpm2_abrmd_t cannot send the D-Bus reply back to initrc_t -- the reply is silently dropped and the caller times out with 'Timeout was reached'. No AVC is logged for the dropped reply, making it non-obvious to diagnose.

This rule was submitted to upstream refpolicy but rejected as platform-specific. The upstream reviewer stated:
'we are not handling this in refpolicy. The rule will need to remain in your downstream policy until an adbd domain is added here.'

SELinuxProject PR for reference: [tpm2: fix tpm2-abrmd D-Bus communication under enforcing mode](https://github.com/SELinuxProject/refpolicy/pull/1170)

Upstream-Status: Inappropriate

WHY? platform-specific: adbd runs as initrc_t on Qualcomm platforms; upstream will accept this only when a proper adbd_t domain is contributed to refpolicy


lumag (changes requested):
There is nothing platform-specific here. It is not relevant to the BSP (meta-qcom).
And yes, instead of adding workarounds, please implement SELinux policy for adbd.

PR #2700, opened 2026-07-08
workflows/compile: drop buildstats github artifacts

quaresmajose:
They're already stored in S3, so we don't need to have them available here. This will considerably reduce the number of artifacts used because we have many builds.

### DIFF
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -87,12 +87,6 @@ runs:
 
     - uses: actions/upload-artifact@v6
       with:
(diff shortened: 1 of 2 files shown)


mwasilew (dismissed):

ricardosalveti:
Should we also remove from https://github.com/qualcomm-linux/meta-qcom/blob/master/.github/actions/compile/action.yml#L88?

quaresmajose:
> Should we also remove from https://github.com/qualcomm-linux/meta-qcom/blob/master/.github/actions/compile/action.yml#L88?

Yes, its better. This is not used anymore in meta-qcom but is in use in other repos.

ricardosalveti (approved):

koenkooi (approved):

PR #2717, opened 2026-07-09
linux-qcom: add pahole-native dependency

raj-patel412:
Add pahole-native as a kernel build dependency to support BTF generation when CONFIG_DEBUG_INFO_BTF and CONFIG_DEBUG_INFO_BTF_MODULES are enabled.

This is required for builds that enable BTF support, which is used by modern eBPF workflows.

### DIFF
--- a/recipes-kernel/linux/linux-qcom-next_git.bb
+++ b/recipes-kernel/linux/linux-qcom-next_git.bb
@@ -14,6 +14,10 @@ LINUX_VERSION ?= "7.1"
 
 PV = "${LINUX_VERSION}+git"
 
+KERNEL_PAHOLE ?= '${@oe.utils.vartrue("DEBUG_BUILD", bb.utils.contains("BBFILE_COLLECTIONS", "openembedded-layer", "1", "0", d), "0", d)}'
+do_configure[depends] += '${@oe.utils.vartrue("KERNEL_PAHOLE", "pahole-native:do_populate_sysroot", "", d)}'
+EXTRA_OEMAKE += '${@oe.utils.vartrue("KERNEL_PAHOLE", "", "PAHOLE=false", d)}'
+
 # tag: qcom-next-7.1-20260708
 SRCREV ?= "947408df47978cb86ffcd180a6244379e03301bf"
 
--- a/recipes-kernel/linux/linux-qcom_6.18.bb
+++ b/recipes-kernel/linux/linux-qcom_6.18.bb
@@ -16,6 +16,10 @@ PV = "${LINUX_VERSION}"
 
 FILESEXTRAPATHS:prepend := "${THISDIR}/linux-qcom-6.18:"
 
+KERNEL_PAHOLE ?= '${@oe.utils.vartrue("DEBUG_BUILD", bb.utils.contains("BBFILE_COLLECTIONS", "openembedded-layer", "1", "0", d), "0", d)}'
+do_configure[depends] += '${@oe.utils.vartrue("KERNEL_PAHOLE", "pahole-native:do_populate_sysroot", "", d)}'
+EXTRA_OEMAKE += '${@oe.utils.vartrue("KERNEL_PAHOLE", "", "PAHOLE=false", d)}'
+
 # tag:qcom-6.18.y-20260720
 SRCREV ?= "4ab384e41300a12c67d5c18c4e7bc1e3f546cc28"
 


lumag (changes requested):
Why are they being enabled here rather than in the linux-qcom tree itself?

raj-patel412:
CONFIG_DEBUG_INFO_BTF depends on a valid PAHOLE_VERSION during kernel configuration. Without pahole-native in the Yocto recipe, PAHOLE_VERSION is 0 and Kconfig drops the BTF options from the final .config. That is why I kept the config fragment together with the pahole-native dependency in this layer.

lumag:
> CONFIG_DEBUG_INFO_BTF depends on a valid PAHOLE_VERSION during kernel configuration. Without pahole-native in the Yocto recipe, PAHOLE_VERSION is 0 and Kconfig drops the BTF options from the final .config. That is why I kept the config fragment together with the pahole-native dependency in this layer.

So, it should be fine to enable Kconfig options in the qcom-next tree and enable `pahole-native` here.

ricardosalveti:
It should also cover both linux-qcom and linux-qcom-next, no?

raj-patel412:
Actually, yes. I will update the PR to enable pahole-native for both kernel recipes here. As for the kernel configs, I will move them to qualcomm-linux/kernel.

One question regarding the kernel configs: should I submit PRs for both kernel branches, or is there a primary branch where these changes should be made first?

quaresmajose:
> One question regarding the kernel configs: should I submit PRs for both kernel branches, or is there a primary branch where these changes should be made first?

This is correct as it is; all PRs must be sent to the master. The backport is done automatically if necessary.

raj-patel412:
Actually i was asking about the qualcomm-linux/kernel. i mean for the change to be raised for the kernel config enablement in the kernel tree should i raise the PR for both kernel branches (qcom-6.18.y and qcom-next) individually?

koenkooi (changes requested):
The commit message is being very coy about what this actually is/does, the only term not present in the patch body used to explain this is 'eBPF', which doesn't make it any clearer to me what this is.
Also what is the impact, is the kernel bigger/smaller/better proportioned now?

ricardosalveti:
> Actually i was asking about the qualcomm-linux/kernel. i mean for the change to be raised for the kernel config enablement in the kernel tree should i raise the PR for both kernel branches (qcom-6.18.y and qcom-next) individually?

I believe so, but I'm not following how the config changes are maintained there, if synchronized between branches or managed individually and separately, open one there and ask.

lumag:
> This branch has conflicts that must be resolved

ricardosalveti:
Rebase and fix the conflicts.

anujm1:
Please see oe-core commit:

```
commit 217a4db53edbd88001f6390bbff39e5dd3d137af
Author: Xiangyu Chen <xiangyu.chen@windriver.com>
Date:   Thu Oct 26 14:38:48 2023 +0800

    linux-yocto: make sure the pahole-native available before do_kernel_configme

    When using debug-btf.scc in a clean workspace, the CONFIG_MODULE_ALLOW_BTF_MISMATCH cannot
    apply to kernel until clean the kernel code(bitbake linux-yocto -c cleanall) and rebuild.

    After tracking the code, some options depend on CONFIG_PAHOLE_VERSION, it was generated by
    scripts/pahole-version.sh in kernel, but during do_kernel_configme step, the pahole-native
    is not available in sysroot-native, so need to wait pahole-native install to sysroot-native
    before do_kernel_configme.
```

and evaluate that test case and make changes if required.

raj-patel412:
Thanks for pointing this out.
The same situation applies here as well, because the BTF-related kernel
config options need pahole to be available during kernel configuration.
I can use a task-level dependency instead of adding
pahole-native as a full recipe dependency:

do_configure[depends] += "pahole-native:do_populate_sysroot"

This makes pahole-native available before linux-qcom runs do_configure,
which matches the intent of the oe-core linux-yocto change more closely.

One concern is that pahole-native is provided by meta-oe, while meta-qcom
should not take a direct dependency on meta-oe. Can you please suggest what can be done here.

lumag:
> Can you please suggest what can be done here.

Just follow the approach of OE-core.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -14,6 +14,8 @@ LINUX_VERSION ?= "7.1"
 
 PV = "${LINUX_VERSION}+git"
 
+DEPENDS += "pahole-native"
```
lumag: Start your commit messages by describing the problem.

recipes-kernel/linux/linux-qcom_6.18.bb
```diff
@@ -16,6 +16,8 @@ PV = "${LINUX_VERSION}"
 
 FILESEXTRAPATHS:prepend := "${THISDIR}/linux-qcom-6.18:"
 
+DEPENDS += "pahole-native"
```
lumag: Two separate commits

lumag (changes requested):

ricardosalveti:
```
ERROR: Nothing PROVIDES 'pahole-native'. Close matches:
  pango-native
  patch-native
  patchelf-native
ERROR: Required build target 'esp-qcom-image' has no buildable providers.
```

Also rebase to fix the conflicts.

recipes-kernel/linux/linux-qcom-next_git.bb
```diff
@@ -14,6 +14,8 @@ LINUX_VERSION ?= "7.1"
 
 PV = "${LINUX_VERSION}+git"
 
+do_configure[depends] += "pahole-native:do_populate_sysroot"
```
anujm1: These would need to be enabled based on a variable that is unset by default, similar to OE-core.

raj-patel412:
> One concern is that pahole-native is provided by meta-oe, while meta-qcom
> should not take a direct dependency on meta-oe.

As mentioned before my concern is that pahole-native is provided by the meta-oe layer from the meta-openembedded.

> ```
> ERROR: Nothing PROVIDES 'pahole-native'. Close matches:
>   pango-native
>   patch-native
>   patchelf-native
> ERROR: Required build target 'esp-qcom-image' has no buildable providers.
> ```
> 
> Also rebase to fix the conflicts.

As meta-qcom has only  oe-core as the dependency layer, we are getting the above error. How should I handle this?

anujm1:
> > Also rebase to fix the conflicts.
> 
> As meta-qcom has only oe-core as the dependency layer, we are getting the above error. How should I handle this?

https://github.com/qualcomm-linux/meta-qcom/pull/2717#discussion_r3618936269 OE-core does not always enable this. You can follow that approach.

lumag (dismissed):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/e3572a86-d669-4ddb-8b76-591d187ece5b)**
---
**Test jobs for commit [`6726988`](https://github.com/qualcomm-linux/meta-qcom/commit/6726988187d515b4548e1578b3f7322a2fcfdf4d)**
---
Pass: 31 | Fail: 6 | Total: 37

ricardosalveti:
The conditional is now following what was done by oe-core, but this now brings a layer dependency when DEBUG_BUILD is enabled, and a build with just meta-qcom + oe-core will now fail when this variable is set.

We could use something like:

```
KERNEL_PAHOLE ?= '${@oe.utils.vartrue("DEBUG_BUILD", bb.utils.contains("BBFILE_COLLECTIONS", "openembedded-layer", "1", "0", d), "0", d)}'
do_configure[depends] += '${@oe.utils.vartrue("KERNEL_PAHOLE", "pahole-native:do_populate_sysroot", "", d)}'
EXTRA_OEMAKE += '${@oe.utils.vartrue("KERNEL_PAHOLE", "", "PAHOLE=false", d)}'
```

lumag (approved):

ricardosalveti (approved):

anujm1 (approved):

raj-patel412:
Please backport to wrynose as well.

ricardosalveti:
@koenkooi mind another look?

koenkooi (approved):

PR #2721, opened 2026-07-10
Improvide packaging of Hexagon DSP binaries

lumag:
Providing empty Hexagon DSP packages for commercial devices doesn't play
 well with the packages for commercial devices. Users can end up with
empty packages still claiming to be covered by the EULA. Drop
ALLOW_EMPTY from those packages to prevent them from being created.


lumag:
Triggered by the errors in https://github.com/qualcomm-linux/meta-qcom-3rdparty/pull/29

koenkooi (approved):
This looks like it does what you describe :)

ricardosalveti (approved):

PR #2730, opened 2026-07-13
systemd-boot: add EFI_MEMORY_ATTRIBUTE_PROTOCOL warning downgrade patch

aswinm94:
Add backport systemd-boot patch that downgrades the
EFI_MEMORY_ATTRIBUTE_PROTOCOL missing message from warning to debug.
 
U-Boot currently does not implement EFI_MEMORY_ATTRIBUTE_PROTOCOL,
causing systemd-boot to display a warning during every boot on
EFI 2.10+ platforms. The warning triggers log_wait() behavior,
introducing an unnecessary boot-time delay even though the
condition is a known U-Boot limitation and requires no user action.
 
Include the patch in the systemd-boot recipe to suppress the
non-actionable warning and avoid the associated boot delay on
U-Boot-based systems.

upstream-status: Backport [https://github.com/systemd/systemd/pull/43017]

### DIFF
--- a/recipes-core/systemd/systemd-boot_%.bbappend
+++ b/recipes-core/systemd/systemd-boot_%.bbappend
@@ -2,4 +2,5 @@ FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/systemd:"
 
 SRC_URI:append:qcom = " \
     file://0001-boot-stub-honor-PE-SectionAlignment-when-loading-inn.patch \
+    file://0002-boot-downgrade-EFI_MEMORY_ATTRIBUTE_PROTOCOL-warning.patch \
 "
--- a/recipes-core/systemd/systemd/0002-boot-downgrade-EFI_MEMORY_ATTRIBUTE_PROTOCOL-warning.patch
+++ b/recipes-core/systemd/systemd/0002-boot-downgrade-EFI_MEMORY_ATTRIBUTE_PROTOCOL-warning.patch
@@ -1,0 +1,47 @@
+From 829c2f0e3dffe8b4484e730a3f3058b61e2133ad Mon Sep 17 00:00:00 2001
+From: Aswin Murugan <aswin.murugan@oss.qualcomm.com>
+Date: Tue, 14 Jul 2026 11:37:14 +0530
+Subject: [PATCH] boot: downgrade EFI_MEMORY_ATTRIBUTE_PROTOCOL warning
+
+U-Boot currently does not implement EFI_MEMORY_ATTRIBUTE_PROTOCOL
+even when reporting EFI version >= 2.10. Consequently, systemd-boot
+emits a warning on every boot when running on U-Boot firmware.
+
+The absence of EFI_MEMORY_ATTRIBUTE_PROTOCOL is a current
+U-Boot limitation and not a condition users can remedy.
+Furthermore, the EFI specification does not require all
+firmware advertising EFI 2.10 or newer to implement the
+protocol. As a result, the warning provides little value on
+U-Boot systems while causing log_wait() to impose a 2.5-second
+boot delay.
+
+Downgrade the message to LOG_DEBUG, this keeps the diagnostic
+available for debugging purposes without penalizing normal
+boot time.
+
+Upstream-Status: Backport [ https://github.com/systemd/systemd/pull/43017 ]
+
+Signed-off-by: Aswin Murugan <aswin.murugan@oss.qualcomm.com>
+---
+ src/boot/linux.c | 5 +----
+ 1 file changed, 1 insertion(+), 4 deletions(-)
+
+diff --git a/src/boot/linux.c b/src/boot/linux.c
+index 67ed859993..09fb63419d 100644
+--- a/src/boot/linux.c
++++ b/src/boot/linux.c
+@@ -267,10 +267,7 @@ EFI_STATUS linux_exec(
+                  * if required for NX_COMPAT */
+                 err = BS->LocateProtocol(MAKE_GUID_PTR(EFI_MEMORY_ATTRIBUTE_PROTOCOL), /* Registration= */ NULL, (void **) &memory_proto);
+                 if (err != EFI_SUCCESS)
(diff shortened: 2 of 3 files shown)


recipes-core/systemd/systemd/0002-boot-skip-NX_COMPAT-warning-on-U-Boot-firmware.patch
```diff
@@ -0,0 +1,52 @@
...
+Since log_wait() stalls for 2.5 seconds per warning message,
+systems booting via U-Boot incur a fixed 2.5-second startup
+delay.
+
+Treat this as a known U-Boot limitation by skipping the
+LOG_WARNING path when ST->FirmwareVendor is "Das U-Boot" and
+logging at LOG_DEBUG instead. Warnings are still emitted for
+other UEFI >= 2.10 firmware that lack the protocol.
+
+Upstream-Status: Pending [platform-specific: works around
```
lumag: No Pending patches, please. Post it and get reviewed.

lumag (changes requested):
Also you are not changing the "boot" recipe. Don't blindly c&p the description of the patch. Write a suitable commit message.

recipes-core/systemd/systemd/0002-boot-skip-NX_COMPAT-warning-on-U-Boot-firmware.patch
ricardosalveti: This issue is not specific to qualcomm, please fix in either u-boot or in systemd upstream, and once done backport to the respective projects (e.g. systemd -> oe-core, u-boot-qcom, here).

aswinm94: @lumag , @ricardosalveti 
The change has been merged in upstream
https://github.com/systemd/systemd/pull/43017
backported the submitted patch

ricardosalveti: Great, can you send to oe-core as well? Justifying that it is useful to reduce the boot time when booting with u-boot based firmware.

recipes-core/systemd/systemd-boot_%.bbappend
```diff
@@ -2,4 +2,5 @@ FILESEXTRAPATHS:prepend:qcom := "${THISDIR}/systemd:"
 
 SRC_URI:append:qcom = " \
     file://0001-boot-stub-honor-PE-SectionAlignment-when-loading-inn.patch \
+    file://0002-boot-downgrade-EFI_MEMORY_ATTRIBUTE_PROTOCOL-warning.patch \
```
ricardosalveti, line 5: Also, we need to patch both recipes-core/systemd/systemd-boot_%.bbappend and recipes-core/systemd/systemd_%.bbappend, since systemd-boot produces the efi binary that gets into the deploy folder, but the efi binary that is available as a package comes from systemd (it is an issue in oe-core itself).

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/6f680679-35c9-4a7c-95a4-d0cbf247bf81)**
---
**Test jobs for commit [`d69cf39`](https://github.com/qualcomm-linux/meta-qcom/commit/d69cf39ea2860c5e299cbf7095ff7b8c444739e4)**
---
Pass: 42 | Fail: 2 | Total: 44
Pass: 10 | Fail: 2 | Total: 12

aswinm94:
@ricardosalveti ,This change is now available as part of Systemd v261.2. The latest OE-Core updates have already moved to Systemd v261.2, so could we backport this patch to meta-qcom as well until we update the oe-core layer revision?

anujm1:
> Ricardo Salveti (@ricardosalveti) ,This change is now available as part of Systemd v261.2. The latest OE-Core updates have already moved to Systemd v261.2, so could we backport this patch to meta-qcom as well until we update the oe-core layer revision?

I think you could propose a bump in OE-core revision for the lock file instead.

ricardosalveti:
> Ricardo Salveti (@ricardosalveti) ,This change is now available as part of Systemd v261.2. The latest OE-Core updates have already moved to Systemd v261.2, so could we backport this patch to meta-qcom as well until we update the oe-core layer revision?

https://github.com/qualcomm-linux/meta-qcom/pull/2920 has the updated systemd, will try to land that instead.

aswinm94:
> > Ricardo Salveti (Ricardo Salveti (@ricardosalveti)) ,This change is now available as part of Systemd v261.2. The latest OE-Core updates have already moved to Systemd v261.2, so could we backport this patch to meta-qcom as well until we update the oe-core layer revision?
> 
> #2920 has the updated systemd, will try to land that instead.

Thanks @ricardosalveti,
For Wrynose, how should we bring in this change? Will this OE-Core layer version bump be backported to the Wrynose branch as well, or do we need to maintain it as a downstream patch?

ricardosalveti:
> > #2920 has the updated systemd, will try to land that instead.
> 
> Thanks Ricardo Salveti (@ricardosalveti), For Wrynose, how should we bring in this change? Will this OE-Core layer version bump be backported to the Wrynose branch as well, or do we need to maintain it as a downstream patch?

Send the backport to oe-core wrynose, it is a valid fix.

ricardosalveti:
Will close this pr as we're bumping meta-qcom lock to a newer hash which includes a recent enough systemd.

ricardosalveti:
> > > #2920 has the updated systemd, will try to land that instead.
> > 
> > 
> > Thanks Ricardo Salveti (Ricardo Salveti (@ricardosalveti)), For Wrynose, how should we bring in this change? Will this OE-Core layer version bump be backported to the Wrynose branch as well, or do we need to maintain it as a downstream patch?
> 
> Send the backport to oe-core wrynose, it is a valid fix.

It might be good to land via the systemd stable branch for the version used by wrynose first actually, then bump oe-core wrynose systemd to the latest stable release.

PR #2735, opened 2026-07-14
Inherit GI in gst-plugins-imsdk-common and add recipe for the Python bindings package

c-petarg:
Inherit gobject-introspection and add depends on PyGObject in order enable Python bindings from IMSDK base.


recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
```diff
@@ -23,11 +23,12 @@ DEPENDS += "\
 COMPATIBLE_MACHINE = "^$"
 COMPATIBLE_MACHINE:aarch64 = "(.*)"
 
-PACKAGECONFIG[base]         = "-DENABLE_GST_PLUGIN_BASE=ON, -DENABLE_GST_PLUGIN_BASE=OFF, json-glib virtual/egl virtual/libgbm virtual/libgles2 virtual/libgles3"
+PACKAGECONFIG[base]         = "-DENABLE_GST_PLUGIN_BASE=ON -DPYTHON_SITEPACKAGES_DIR=${PYTHON_SITEPACKAGES_DIR}, -DENABLE_GST_PLUGIN_BASE=OFF, \
```
quaresmajose: Why not create a dedicated PACKAGECONFIG[python]? so it could be optional. This could also be packaged in an additional package ${PN}-python.

c-petarg: Python binding are tightly coupled with base. They are set to compile with base. Right now we cannot make them separate package.

quaresmajose: I understand, but could it be optional?

c-petarg: No because the upcoming Python SDK and C++ SDK are stepping on the changes in base.

lumag: If somebody doesn't need Python IMSDK, they must be able to turn it off completely.

c-petarg: Alright, lets suppose that we make some hacky solution in the source so that GI related files are generated separately. How do you propose to resolve inherit of gobject-introspection ? I'm not acquanted enough to know how to resolve this.
As for REQUIRED_DISTRO_FEATURES I think we can use something like "${@bb.utils.contains('PACKAGECONFIG', 'python', 'gobject-introspection-data', '', d)}"
 
Not to mention that whatever hack/solution we do in source needs to work across multiple flavored build systems. So do we really need this at this stage?

I double checked with the source code of upstream Gstreamer. There the generated GI files are not in separate plugin or package. They are created along the compiled libraries.
 
The typelib and gir files  are placed in ${libdir}/girepository-1.0 and ${datadir}/gir-1.0 respectively and are part of the plugin-base. I do not see how we differ with upstream on this?
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-base/meson.build?ref_type=heads#L550
 
They are generated as part of the same compilation in gst-plugins-base and are not in any way part of gst-python
https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-base/meson.build?ref_type=heads#L532

These typelib files are directly usable in python scripts. On top we have some overrides as the GI tool is not perfect and creates ambiguous connections between Python and the underlying C implementation.

anujm1: g-i files are packaged separately: https://git.openembedded.org/openembedded-core/tree/meta/recipes-multimedia/gstreamer/gstreamer1.0-plugins-packaging.inc#n24 for gstreamer in oe-core. Perhaps we could do something similar in our packaging inc?

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
```diff
@@ -5,9 +5,9 @@ SECTION = "multimedia"
 LICENSE = "BSD-3-Clause-Clear"
 LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
 
-inherit cmake features_check pkgconfig
+inherit cmake features_check pkgconfig gobject-introspection
 
-REQUIRED_DISTRO_FEATURES = "opengl"
+REQUIRED_DISTRO_FEATURES = "opengl gobject-introspection-data"
```
anujm1: I don't think this should be a required feature? The class enables it based on whether it's in `DISTRO_FEATURES` but is it really mandatory for imsdk to work?

c-petarg: I can drop it from here but who is going to make sure it is enabled as DISTRO_FEATURES? Especially when user wants to use Python with gstreamer and our plugins. How are they going to enabled it? Is it always going to be enabled in DISTRO_FEATURES for all distros? If it is going to be only for some then for which ones? What will the strategy be?

anujm1: > I can drop it from here but who is going to make sure it is enabled as DISTRO_FEATURES? Especially when user wants to use Python with gstreamer and our plugins. How are they going to enabled it? Is it always going to be enabled in DISTRO_FEATURES for all distros? If it is going to be only for some then for which ones? What will the strategy be?

These would need to be answered even when this is present :) This right now just ensures recipe won't build if it is not enabled.

The class sets `GI_DATA_ENABLED` based on this feature and that can be used for any decision in the recipe. The distro layer should control whether it is enabled or not, or when it is enabled and not this layer.

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
c-petarg: > g-i files are packaged separately: https://git.openembedded.org/openembedded-core/tree/meta/recipes-multimedia/gstreamer/gstreamer1.0-plugins-packaging.inc#n24 for gstreamer in oe-core. Perhaps we could do something similar in our packaging inc?

I'm not able to find any meaningful usage of these lines in the upstream packaging. The lines are producing several typelib packages but they are completely empty as I do not see any instructions anywhere to fill them with anything.
Nothing is put in them so what is their point? Am I missing something?

The packages are:
- gstreamer1.0-plugins-base-allocators-typelib
- gstreamer1.0-plugins-base-app-typelib
- gstreamer1.0-plugins-base-audio-typelib
- gstreamer1.0-plugins-base-glegl-typelib
- gstreamer1.0-plugins-base-gl-typelib
- gstreamer1.0-plugins-base-glwayland-typelib
- gstreamer1.0-plugins-base-glx11-typelib
- gstreamer1.0-plugins-base-pbutils-typelib
- gstreamer1.0-plugins-base-rtp-typelib
- gstreamer1.0-plugins-base-rtsp-typelib
- gstreamer1.0-plugins-base-sdp-typelib
- gstreamer1.0-plugins-base-tag-typelib
- gstreamer1.0-plugins-base-video-typelib

In reality the generated .typelib files live inside the gstreamer1.0-plugins-base package.

```
./tmp/work/armv8-2a-qcom-linux/gstreamer1.0-plugins-base/1.28.2/packages-split/gstreamer1.0-plugins-base
└── usr
    └── lib
        └── girepository-1.0
            ├── GstAllocators-1.0.typelib
            ├── GstApp-1.0.typelib
            ├── GstAudio-1.0.typelib
            ├── GstGL-1.0.typelib
            ├── GstGLEGL-1.0.typelib
            ├── GstGLWayland-1.0.typelib
            ├── GstGLX11-1.0.typelib
            ├── GstPbutils-1.0.typelib
            ├── GstRtp-1.0.typelib
            ├── GstRtsp-1.0.typelib
            ├── GstSdp-1.0.typelib
            ├── GstTag-1.0.typelib
            └── GstVideo-1.0.typelib
```

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
c-petarg: Alright. Will remove it.

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
c-petarg: So, typelib files must remain part of base package. There is no way around that. But what I can do is move the source for the generation of the override Python extension module to separate location by restructuring the source code. Similar to what upstream GStreamer is doing with their gst-python subproject.
With this change to the source code we can trigger the compilation of the module separately. For that purpose a new gst-plugins-imsdk-python recipe has been created. It additionally includes REQUIRED_DISTRO_FEATURES for gobject-introspection-data as the Python extension module step on the GI typelib files.

Please check the update commit.

recipes-multimedia/imsdk/gst-plugins-imsdk-python_1.0.2.bb
```diff
@@ -0,0 +1,12 @@
+require gst-plugins-imsdk-common.inc
+require gst-plugins-imsdk-packaging.inc
+
+SUMMARY = "Qualcomm IMSDK GStreamer Python Libraries"
+DESCRIPTION = "Python binding overrides (complementing the bindings provided by python-gi) for IMSDK GStreamer base libraries."
+
+DEPENDS += "gst-plugins-imsdk-base python3-pygobject"
+RDEPENDS:${PN} += "gst-plugins-imsdk-base python3-pygobject"
```
quaresmajose: Is RDEPENDS really necessary? It should be inferred during packaging.

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
```diff
@@ -5,7 +5,7 @@ SECTION = "multimedia"
 LICENSE = "BSD-3-Clause-Clear"
 LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
 
-inherit cmake features_check pkgconfig
+inherit cmake features_check pkgconfig gobject-introspection
```
quaresmajose, line 8: It seems to me that this `inherit gobject-introspection` can be moved to the gst-plugins-imsdk-python recipe, It is only needed there.

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
```diff
@@ -52,3 +53,7 @@ RDEPENDS:libgstqtimqttadaptor += "mosquitto"
 
 RDEPENDS:${PN}-apps += "${@bb.utils.contains('PACKAGECONFIG', 'python-apps', 'python3-core gstreamer1.0-python python3-pygobject gtk+3 python3-opencv', '', d)} \
                         ${@bb.utils.contains('PACKAGECONFIG', 'webrtc', 'gstreamer1.0-plugins-bad-webrtc', '', d)}"
+
+# When the 'python' PACKAGECONFIG is enabled, include the Python site-packages directory
+# to ensure that the Python extension modules installed by the IMSDK python are packaged.
+FILES:${PN}:append = " ${@bb.utils.contains('PACKAGECONFIG', 'python', d.getVar('PYTHON_SITEPACKAGES_DIR'), '', d)}"
```
quaresmajose: This can also be moved to gst-plugins-imsdk-python.

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
c-petarg: No, it is not. It is needed in gst-plugins-qti-base as it generates the GI typelib files and in the future it my be needed for gst-plugins-qti-oss as there are internal discussions. So that is why I've put in common. Prefer to keep it in common as it does not interfere with anything but if insisted I can move to gst-plugins-qti-base.

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
c-petarg: Alright

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
quaresmajose: Rigth, thank you for explaining.

recipes-multimedia/imsdk/gst-plugins-imsdk-python_1.0.2.bb
c-petarg: Yeah, you are right. Not needed. Will remove

ricardosalveti:
```
ERROR: Nothing PROVIDES 'gst-plugins-imsdk-base' (but /work/build/../../repo/recipes-multimedia/imsdk/gst-plugins-imsdk-python_1.0.2.bb DEPENDS on or otherwise requires it)
gst-plugins-imsdk-base was skipped: using DISTRO 'nodistro', which is missing required DISTRO_FEATURES: 'opengl'
ERROR: Required build target 'gst-plugins-imsdk-python' has no buildable providers.
```

recipes-multimedia/imsdk/gst-plugins-imsdk-python_1.0.2.bb
```diff
@@ -0,0 +1,13 @@
...
+require gst-plugins-imsdk-packaging.inc
+
+SUMMARY = "Qualcomm IMSDK GStreamer Python Libraries"
+DESCRIPTION = "Python binding overrides (complementing the bindings provided by python-gi) for IMSDK GStreamer libraries."
+
+DEPENDS += "gst-plugins-imsdk-base python3-pygobject"
+
+PACKAGECONFIG = "python"
+
+REQUIRED_DISTRO_FEATURES = "gobject-introspection-data"
```
ricardosalveti: This should be += in order to include `REQUIRED_DISTRO_FEATURES = "opengl"` from imsdk.

c-petarg: Updated

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/56d7c2b4-27d6-4a5a-8ef6-4bcc566b1e85)**
---
**Test jobs for commit [`b5e6fda`](https://github.com/qualcomm-linux/meta-qcom/commit/b5e6fdab1d4baa0c139a460de504b3ff86d7a38e)**
---
Pass: 54 | Fail: 2 | Total: 56

lumag (approved):

ricardosalveti (approved):

PR #2743, opened 2026-07-15
QCS615-LYRA-EVK enablement

kranthi-peruri:
Added machine configuration, CI build support for QCS615-IQ-615-EVK.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -243,6 +243,7 @@ jobs:
           - qcm6490-idp
           - qcom-armv8a
           - qcs615-ride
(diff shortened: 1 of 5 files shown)


quaresmajose:
Please change the order of the commits so that they can all be used one by one without causing errors. The kas fragment must be the last one.

kranthi-peruri:
> Please change the order of the commits so that they can all be used one by one without causing errors. The kas fragment must be the last one.

Updated the change order

quaresmajose:
> > Please change the order of the commits so that they can all be used one by one without causing errors. The kas fragment must be the last one.
> 
> Updated the change order

Thanks

PR #2750, opened 2026-07-16
ci: base.lock: update layers to latest

ricardosalveti:
Also update patches/oe-core/0001-igt-gpu-tools-fix-build-on-non-x86-platforms.patch
to make it compatible with igt-gpu-tools 2.5, which is now in oe-core.

meta-openembedded is intentionally held at dc0ccaa27, the last commit
before the abseil-cpp 20260107.1 -> 20260526.0 upgrade (90603f7d4),
because the prebuilt libsensinghubapiprop.so.1.2.1 in qcom-sensors-
binaries links against the abseil 2601.0.0 sonames and fails the file-
rdeps QA check with the newer abseil.

meta-updater is likewise held at fff4a58, right before it moved the
ostree bbappend to 2026.2, since the held-back meta-openembedded still
ships ostree 2026.1.

Relevant changes for oe-core:
- 171af3b29 oeqa/selftest/devtool: cover update-recipe --initial-rev
- 62abb60b8 devtool: standard: fix update-recipe/finish --initial-rev override
- aa85baff9 package.bbclass: hardcode emit_pkgdata to run last
- b05a90340 oeqa/selftest/devtool: exercise devtool finish on an AUTOREV recipe
- ae7be93cf devtool: standard: fix finish/update-recipe for recipes using AUTOREV
- 497133784 python3-pyasn1: set CVE_PRODUCT
- bb80fef9a python3-ply: set CVE_PRODUCT
- b721019e0 python3-cryptography: set CVE_PRODUCT
- cf5859260 os-release: set BUILD_ID vardepsexclude with weak assignment
- 698e72cfe screen: upgrade 5.0.1 -> 5.0.2
- 7ea792798 bindgen-cli: set CLANG_PATH in native and nativesdk wrappers to matching clang
- de2e11e63 oe-selftest: fitimage: Add tests for KERNEL_DTBVENDORED
- 3bceb2dab kernel-fit-image: Add KERNEL_DTBVENDORED support for FIT_CONF_DEFAULT_DTB
- f77d8886f glibc: disable nscd by default
- 42b917f96 glibc: only inherit init script classes if nscd is enabled
- 6da10b262 glibc: only install nscd if it's been enabled
- 55bf6734b glibc: inherit update-rc.d
- aa0d9d1c1 gtk-doc: upgrade 1.35.1 -> 1.36.1
- c2bd43b37 openssh: set status for https://github.com/advisories/GHSA-wcpp-3x59-h8vp
[...]

PR #2754, opened 2026-07-16
qcom-sensors-binaries: drop all dependencies

quaresmajose:
This is a binary package, so it has no build dependencies.
Instead, it should have explicit runtime dependencies.


quaresmajose:
The missing libabsl-* comes from protobuf
```
sensinghub -> protobuf-camx
protobuf-camx -> protobuf
protobuf -> abseil-cpp
```

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/28aab695-3537-4391-9e00-457227805f58)**
---
**Test jobs for commit [`9cf9b7c`](https://github.com/qualcomm-linux/meta-qcom/commit/9cf9b7ce89576aecf173c7dd09c73a6c96ddf20e)**
---
Pass: 9 | Fail: 2 | Total: 11
Pass: 42 | Fail: 2 | Total: 44

lumag (changes requested):
I think this is wrong. Why do we need to circumvent the existing runtime dependencies discovery mechanism?

quaresmajose:
I'm coming to that conclusion too. Given that this is a pre-compiled package, we shouldn't need build dependencies, but if we do, the runtime dependencies will have to be set manually. Some of the runtime dependencies aren't obvious, like glibc, which isn't on the manual list and should be.

quaresmajose:
Changing this will also complicate the lives of colleagues who provide the binaries. I'm going to end up closing this PR; this was more of an experiment for me to see if I could improve what we have.

quaresmajose:
Add a new abseil-cpp-camx and pick the bump from https://github.com/qualcomm-linux/meta-qcom/pull/2750 to test

PR #2758, opened 2026-07-17
abseil-cpp-camx: add CamX local adapation on abseil-cpp 

quaresmajose:
The meta-oe layer has upgraded the abseil-cpp recipe, breaking CamX and
sensors recipes in meta-qcom. Vendor in the older version of the recipe
until the situation is sorted out by new binary uploads of those
binary-only recipes.

This needs to go along with protobuf-camx that we already have.


quaresmajose:
However, there is a problem with our current implementation. It is not possible to use `protobuf-camx` and `protobuf` at the same time when both have the same version. I believe the same will happen with `abseil-cpp-camx` and `abseil-cpp`.
```
ERROR: protobuf-camx-6.33.6-r0 do_packagedata: Recipe protobuf-camx is trying to install files into a shared area when those files already exist. The files and the manifests listing them are:
  /work/build/tmp/pkgdata/qcom-armv8a/runtime-reverse/libprotobuf-lite33.6.0
    (matched in manifest-qcom_armv8a-protobuf.packagedata)
  /work/build/tmp/pkgdata/qcom-armv8a/runtime-reverse/libprotoc33.6.0
    (matched in manifest-qcom_armv8a-protobuf.packagedata)
Please adjust the recipes so only one recipe provides a given file. 
ERROR: Logfile of failure stored in: /work/build/tmp/work/armv8a-qcom-linux/protobuf-camx/6.33.6/temp/log.do_packagedata.2925239
ERROR: Task (/work/build/../../repo/dynamic-layers/openembedded-layer/recipes-support/protobuf-camx/protobuf-camx_6.33.6.bb:do_packagedata) failed with exit code '1'
```

Locally, I can compile qcom-sensors-binaries by deleting protobuf and abseil-cpp.

@lumag ^^

lumag:
@quaresmajose it's by design, you should not be using two different versions of the same library in the dependency tree, it's difficult to predict symbol resolution between them, if they are not using symbol versioning.

lumag:
Also, see the modifications that we have been making to those recipes.

quaresmajose:
Replaced by https://github.com/qualcomm-linux/meta-qcom/pull/2881

PR #2759, opened 2026-07-20
Add kas config for qcom-console-image

vkraleti:
qcom-console-image is a lightweight image that exercises the base system without the multimedia stack. Without a dedicated KAS fragment, there is no standardized way for developers and CI systems to build and validate this image independently from the heavier qcom-distro targets.

Add a qcom-distro-console-image KAS fragment that inherits the full qcom-distro layer setup but limits the build target to qcom-console-image only, enabling focused CI coverage for the minimal console use case.

### DIFF
--- a/ci/qcom-distro-console-image.yml
+++ b/ci/qcom-distro-console-image.yml
@@ -1,0 +1,9 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+  includes:
+   - ci/qcom-distro.yml
+
+target:
+  - qcom-console-image


lumag (changes requested):
Just include it into the normal qcom-distro.yaml

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/b40968cf-d6be-4e15-8c66-4e772fba576f)**
---
**Test jobs for commit [`5ff8402`](https://github.com/qualcomm-linux/meta-qcom/commit/5ff8402438c671f04c67708929b9e94254628ca9)**
---
Pass: 54 | Fail: 4 | Total: 58

ricardosalveti:
Users can easily just build this target with kas build ... --target qcom-console-image, I don't think we need a dedicated fragment for this necessarily.

If we want to produce these images in our CI, best would indeed be to extend qcom-distro to also build this image, along with others.

quaresmajose:
> Users can easily just build this target with kas build ... --target qcom-console-image, I don't think we need a dedicated fragment for this necessarily.
> 
> If we want to produce these images in our CI, best would indeed be to extend qcom-distro to also build this image, along with others.

I also think it's better not to create another fragment just for that.

PR #2774, opened 2026-07-22
linux-qcom-next: Remove patch that is part of tree

sgaud-quic:


lumag:
This should be done as a part of the tag bump. CLosing.

PR #2779, opened 2026-07-22
docs: document A/B slot selection via OsTrialBootStatus

royzah:
## What

Adds `docs/ab-slot-selection.md`: how the active slot is chosen on the `PBL -> XBL -> UEFI` boot chain.

## Why

Nothing in the tree documents this. XBL selects the slot from the `OsTrialBootStatus` UEFI variable and ignores GPT slot attributes, so values written by `qbootctl` do not switch slots on their own. The slot B ESP must also carry the `SystemPartB` type GUID or slot B is invisible.

## How to verify

Write the variable for slot B on an unfused device, reboot, confirm XBL mounts `efi_b` and the slot B rootfs is used. Repeat with bit 16 set to exercise trial boot and rollback.

### DIFF
--- a/docs/ab-slot-selection.md
+++ b/docs/ab-slot-selection.md
@@ -1,0 +1,84 @@
+# A/B slot selection
+
+On platforms whose boot chain is `PBL -> XBL -> UEFI -> systemd-boot`, the
+active slot is selected by XBL from the `OsTrialBootStatus` UEFI variable.
+GPT slot attributes are not consulted: values written by `qbootctl` are
+ignored by XBL, so they cannot switch slots on their own.
+
+XBL reads the variable in `MountRequiredFatPartitions` and mounts the ESP of
+the resulting slot:
+
+    ActiveSlot 0 -> gEfiPartTypeSystemPartGuid   (standard ESP GUID)
+    ActiveSlot 1 -> gEfiPartTypeSystemPartBGuid
+
+If the variable is absent, XBL falls back to the primary ESP and returns before
+the slot logic, so the mechanism stays dormant until something writes it.
+
+## Partition requirement
+
+The slot B ESP must carry the `SystemPartB` type GUID, not the standard ESP
+GUID and not an arbitrary one:
+
+| Slot | Partition type GUID                    |
+| ---- | -------------------------------------- |
+| A    | `C12A7328-F81F-11D2-BA4B-00A0C93EC93B` |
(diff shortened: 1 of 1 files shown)


ricardosalveti:
Thanks for the doc, this is good content, but I believe it would fit better at https://dragonwingdocs.qualcomm.com/qualcomm-linux-qli instead (see https://github.com/dragonwing-docs/dragonwing-docs), since this behavior is specific to the firmware and not really just meta-qcom (e.g. the same flow can also be used by debian and ubuntu).

Would you mind proposing there instead?

royzah:
@ricardosalveti happy to move it. `github.com/dragonwing-docs/dragonwing-docs` 404s for me though, might be private. Can you point me at the right repo? Agreed on the reasoning, it is firmware behaviour, not meta-qcom.

ricardosalveti:
> Ricardo Salveti (@ricardosalveti) happy to move it. `github.com/dragonwing-docs/dragonwing-docs` 404s for me though, might be private. Can you point me at the right repo? Agreed on the reasoning, it is firmware behaviour, not meta-qcom.

Yeah, sorry, as I have access I didn't see it was private, let me check internally how the doc team is expecting to accept contributions there, will get back.

ricardosalveti:
> Can you point me at the right repo? Agreed on the reasoning, it is firmware behaviour, not meta-qcom.

While we don't yet have an easy way to get the contribution done via a git repository, would you mind sending the feedback directly at https://dragonwingdocs.qualcomm.com/System/Boot/boot-addendum-overview?

At 'Was this page helpful?' click at No, then request 'Update this documentation', explaining what you want to add and probably just pointing to this PR.

I reported this internally as well, so people are aware of the ask.

Thanks for the contribution and sorry if the process is not fully in place yet, things were migrated to this new doc system recently.

royzah:
@ricardosalveti no problem, thanks for chasing it internally. Will send it through the feedback form pointing at this PR. Leaving this open until the docs side picks it up, unless you prefer otherwise.

ricardosalveti:
> Ricardo Salveti (@ricardosalveti) no problem, thanks for chasing it internally. Will send it through the feedback form pointing at this PR. Leaving this open until the docs side picks it up, unless you prefer otherwise.

Thanks, I will try to track internally, but please let us know if you get stuck in the process, we can keep this open until it lands there.

PR #2789, opened 2026-07-23
ci: test-pr: recreate the results comment so runs notify again

ricardosalveti:
Tagging the PR results comment with comment-tag made the action update a single comment in place (the default upsert mode). GitHub only sends notifications when a comment is created, never when one is edited, so test re-runs stopped generating any email and updated results went unnoticed until someone manually revisited the PR.

Switch the comment action to mode: recreate, which deletes the tagged comment and posts a fresh one. Every (re-)run's results now trigger a notification again, while the PR still carries exactly one results comment per head instead of accumulating stale ones.

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -110,14 +110,17 @@ jobs:
           echo "pr_number=$PR_NUMBER" >> $GITHUB_OUTPUT
 
       - name: Comment on PR
-        uses: thollander/actions-comment-pull-request@24bffb9b452ba05a4f3f77933840a6a841d1b32b # v3
+        uses: marocchino/sticky-pull-request-comment@5770ad5eb8f42dd2c4f34da00c94c5381e49af88 # v3.0.5
         with:
-          file-path: pr-comment.txt
-          pr-number: ${{ steps.pr_comment_prep.outputs.pr_number }}
-          # Update a single tagged comment per PR instead of posting a new one on
-          # every run, so the PR shows one current result instead of accumulating
-          # stale comments (one per run / per superseded commit).
-          comment-tag: test-results
+          path: pr-comment.txt
+          number_force: ${{ steps.pr_comment_prep.outputs.pr_number }}
+          # Post a fresh comment on every run (GitHub only notifies on
+          # comment creation) and minimize the previous tagged comment as
+          # outdated instead of deleting it, so the retry history stays
+          # visible (collapsed) in the PR timeline.
+          header: test-results
+          hide_and_recreate: true
+          hide_classify: OUTDATED
 
   publish-test-results:
     uses: ./.github/workflows/publish-results.yml


ricardosalveti:
Another possibility is using marocchino/sticky-pull-request-comment, which supports hiding the previous comment instead of removing it, but it would require moving to it (have to approve it first in the org / repo).

```
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -110,14 +110,14 @@ jobs:
           echo "pr_number=$PR_NUMBER" >> $GITHUB_OUTPUT

       - name: Comment on PR
-        uses: thollander/actions-comment-pull-request@24bffb9b452ba05a4f3f77933840a6a841d1b32b # v3
+        uses: marocchino/sticky-pull-request-comment@5770ad5eb8f42dd2c4f34da00c94c5381e49af88 # v3.0.5
         with:
-          file-path: pr-comment.txt
-          pr-number: ${{ steps.pr_comment_prep.outputs.pr_number }}
-          # Keep a single tagged results comment per PR, but recreate it
-          # (delete + post new) instead of editing in place: GitHub only
-          # notifies on comment creation, so this makes every (re-)run's
-          # results generate an email while still avoiding stale-comment
-          # accumulation.
-          comment-tag: test-results
-          mode: recreate
+          path: pr-comment.txt
+          number_force: ${{ steps.pr_comment_prep.outputs.pr_number }}
+          # Post a fresh comment on every run (GitHub only notifies on
+          # comment creation) and minimize the previous tagged comment as
+          # outdated instead of deleting it, so the retry history stays
+          # visible (collapsed) in the PR timeline.
+          header: test-results
+          hide_and_recreate: true
+          hide_classify: OUTDATED
```

anujm1:
> Tagging the PR results comment with comment-tag made the action update a single comment in place (the default upsert mode). GitHub only sends notifications when a comment is created, never when one is edited, so test re-runs stopped generating any email and updated results went unnoticed until someone manually revisited the PR.
> 
> Switch the comment action to mode: recreate, which deletes the tagged comment and posts a fresh one. Every (re-)run's results now trigger a notification again, while the PR still carries exactly one results comment per head instead of accumulating stale ones.

Why do we need to use `comment-tag` at all? I think I prefer how it's for wrynose, and the old comments stay. It helps with comparison without extra steps.

ricardosalveti:
> > Tagging the PR results comment with comment-tag made the action update a single comment in place (the default upsert mode). GitHub only sends notifications when a comment is created, never when one is edited, so test re-runs stopped generating any email and updated results went unnoticed until someone manually revisited the PR.
> > Switch the comment action to mode: recreate, which deletes the tagged comment and posts a fresh one. Every (re-)run's results now trigger a notification again, while the PR still carries exactly one results comment per head instead of accumulating stale ones.
> 
> Why do we need to use `comment-tag` at all? I think I prefer how it's for wrynose, and the old comments stay. It helps with comparison without extra steps.

Right, I was just thinking if there is a better way than just keeping all the individual comments (which can be several when we have an infra issue), as that kind of pollutes the github web interface, and it is kind of annoying on long running PRs (as github will also hide comments over a certain amount).

We basically have two options for this, deleting the old comments or hiding them (which requires this other action), this PR is just deleting the old ones (as then we are notified by email, it is always added as the last comment and the older ones are not causing any noise).

anujm1:
> > Why do we need to use `comment-tag` at all? I think I prefer how it's for wrynose, and the old comments stay. It helps with comparison without extra steps.
> 
> Right, I was just thinking if there is a better way than just keeping all the individual comments (which can be several when we have an infra issue), as that kind of pollutes the github web interface, and it is kind of annoying on long running PRs (as github will also hide comments over a certain amount).
> 
> We basically have two options for this, deleting the old comments or hiding them (which requires this other action), this PR is just deleting the old ones (as then we are notified by email, it is always added as the last comment and the older ones are not causing any noise).

I think in that case, hiding the old ones would be better. We still get notifications and, the comments are present as well.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/21a7d3e9-e88e-4a0a-abcf-8c43b3beaa43)**
---
**Test jobs for commit [`d36e8cb`](https://github.com/qualcomm-linux/meta-qcom/commit/d36e8cb5e20d8e4efddba667b6d97b5f0c77a629)**
---
Pass: 9 | Fail: 1 | Total: 10
Pass: 48 | Fail: 6 | Total: 54

koenkooi:
Hiding sounds indeed better than plain deleting, but I suspect we'll need to actually try it to see how it works for us.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/e5921829-1b5d-4155-8bac-17c61dc480d3)**
---
**Test jobs for commit [`cc28ff9`](https://github.com/qualcomm-linux/meta-qcom/commit/cc28ff95c749db1036e0cbb4b53a35faf2051ed8)**
---
Pass: 48 | Fail: 0 | Total: 48
Pass: 9 | Fail: 0 | Total: 9

PR #2808, opened 2026-07-27
u-boot-qcom: enable libubootenv environment access tool

aswinm94:
Enable userspace access to the U-Boot environment variable on U-Boot based builds using libubootenv (fw_printenv/fw_setenv) tools.

Changes
1. Enable fw_env.config support in u-boot-qcom
- Install u-boot-fw-utils and u-boot-qcom-env.
- Add fw_env.config for the U-Boot environment partition.
- Include fw_env.config in SRC_URI for automatic packaging.

2. Standardize initial environment naming
- Re-enable UBOOT_INITIAL_ENV generation for Qualcomm machines.
- Use the standard name u-boot-initial-env for compatibility with libubootenv.

### DIFF
--- a/ci/u-boot-qcom.yml
+++ b/ci/u-boot-qcom.yml
@@ -4,3 +4,5 @@ header:
 local_conf_header:
   bootloaderprovider: |
     PREFERRED_PROVIDER_virtual/bootloader = "u-boot-qcom"
+  libubootenv: |
+    IMAGE_INSTALL:append = " u-boot-fw-utils u-boot-qcom-env"
--- a/recipes-bsp/u-boot/files/fw_env.config
+++ b/recipes-bsp/u-boot/files/fw_env.config
@@ -1,0 +1,3 @@
+# Configuration file for fw_(printenv/setenv) utility (libubootenv).
+# Device path             Offset   Size
+/dev/disk/by-partlabel/logfs   -0x4000   0x4000
--- a/conf/machine/include/qcom-u-boot-common.inc
+++ b/conf/machine/include/qcom-u-boot-common.inc
@@ -6,7 +6,7 @@
 UBOOT_CONFIG ?= ""
 UBOOT_CONFIG_DEFAULT ?= "${UBOOT_CONFIG}"
 
-UBOOT_INITIAL_ENV = ""
+UBOOT_INITIAL_ENV = "u-boot-initial-env"
 
 # Don't forget to add config to qcom-armv8a.conf when adding it here
 UBOOT_CONFIG[iq-615-evk] = "qcom_qcs615_defconfig"
--- a/recipes-bsp/u-boot/u-boot-qcom_git.bb
+++ b/recipes-bsp/u-boot/u-boot-qcom_git.bb
@@ -14,6 +14,7 @@ SRC_URI = "git://github.com/qualcomm-linux/u-boot.git;${SRCBRANCH};protocol=http
 SRC_URI += " \
     file://disable-eficapsule-tool.cfg \
     file://efi-rt-volatile-store.cfg \
+    file://fw_env.config \
     ${@bb.utils.contains('MACHINE_FEATURES', 'optee', 'file://tfa-optee.cfg', '', d)} \
     ${@bb.utils.contains('MACHINE_FEATURES', 'kvm', 'file://gunyah-exit.cfg', '', d)} \
 "


koenkooi (changes requested):
Drop the 'meta-qcom: ' from the leader, and start explaining the problem you're solving, not describe what the commit does.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/3f06bc94-6953-4a90-99e9-3df445f80241)**
---
**Test jobs for commit [`00e996f`](https://github.com/qualcomm-linux/meta-qcom/commit/00e996ff88e7e24e80abb385d57d372114ab863d)**
---
Pass: 42 | Fail: 4 | Others: 1 | Total: 47
Pass: 9 | Fail: 0 | Total: 9

recipes-bsp/u-boot/files/fw_env.config
```diff
@@ -0,0 +1,3 @@
+# Configuration file for fw_(printenv/setenv) utility (libubootenv).
+# Device path             Offset   Size
+/dev/disk/by-partlabel/logfs   -0x4000   0x4000
```
ricardosalveti: Add proper end of line here.

ricardosalveti: Also, why logfs? Where is that set to be the partition used for the env variables?

ci/u-boot-qcom.yml
```diff
@@ -4,3 +4,5 @@ header:
 local_conf_header:
   bootloaderprovider: |
     PREFERRED_PROVIDER_virtual/bootloader = "u-boot-qcom"
+  libubootenv: |
+    IMAGE_INSTALL:append = " u-boot-fw-utils u-boot-qcom-env"
```
ricardosalveti, line 8: This should be in a separated commit.

recipes-core/systemd/systemd/0002-boot-skip-NX_COMPAT-warning-on-U-Boot-firmware.patch
```diff
@@ -0,0 +1,52 @@
...
+systems booting via U-Boot incur a fixed 2.5-second startup
+delay.
+
+Treat this as a known U-Boot limitation by skipping the
+LOG_WARNING path when ST->FirmwareVendor is "Das U-Boot" and
+logging at LOG_DEBUG instead. Warnings are still emitted for
+other UEFI >= 2.10 firmware that lack the protocol.
+
+Upstream-Status: Pending [platform-specific: works around
+U-Boot's lack of EFI_MEMORY_ATTRIBUTE_PROTOCOL support]
```
ricardosalveti: Didn't you send the same change upstream and to oe-core? Why are you changing systemd here?

ricardosalveti:
> Drop the 'meta-qcom: ' from the leader, and start explaining the problem you're solving, not describe what the commit does.

@aswinm94 you didn't update the commits, just the github PR description.

recipes-core/systemd/systemd/0002-boot-skip-NX_COMPAT-warning-on-U-Boot-firmware.patch
aswinm94: Seems the commit is squashed by mistake, will remove this change

recipes-bsp/u-boot/files/fw_env.config
aswinm94: @ricardosalveti , Since the logfs partition is currently not being used by any component, U-Boot utilizes it for storing environment variables. At present, there is no dedicated partition available for U-Boot environment storage.

ci/u-boot-qcom.yml
aswinm94: @ricardosalveti , made as a separate commit.

recipes-bsp/u-boot/files/fw_env.config
aswinm94: @ricardosalveti , added

recipes-core/systemd/systemd/0002-boot-skip-NX_COMPAT-warning-on-U-Boot-firmware.patch
aswinm94: @ricardosalveti , removed these systemd changes

aswinm94:
> > Drop the 'meta-qcom: ' from the leader, and start explaining the problem you're solving, not describe what the commit does.
> 
> aswinm94 you didn't update the commits, just the github PR description.

@ricardosalveti , removed from the commits also

recipes-bsp/u-boot/files/fw_env.config
ricardosalveti: Now there is an extra empty line :-)

ricardosalveti: That is not ideal, can't we just use a file instead, which is more generic? We can store the environment as a file under the ESP partition, for example.

I would really prefer to not 'abuse' one of the partitions here, as they could later be changed or removed.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/c34089c6-b4d3-4687-ad52-505091a28395)**
---
**Test jobs for commit [`3bb6837`](https://github.com/qualcomm-linux/meta-qcom/commit/3bb6837cde3ed6c3a49f0c8470c24bc5bad26d86)**
---
Pass: 9 | Fail: 0 | Total: 9

recipes-bsp/u-boot/files/fw_env.config
aswinm94: @ricardosalveti , An environment file will not be suitable here because in FIT based kernel boot up case, the FIT image resides in a raw partition, while file-based environment storage requires a filesystem partition (EXT4/FAT). This would require an additional partition solely for environment storage

Earlier there was a discussion & concluded that partition layout changes are not feasible for commercialized products. LogFS is currently used only by UEFI, when U-Boot replaces UEFI, LogFS is no longer utilized by any component. Therefore, U-Boot can use the LogFS partition for environment storage without requiring any partition layout changes.

ricardosalveti: > Ricardo Salveti (@ricardosalveti) , An environment file will not be suitable here because in FIT based kernel boot up case, the FIT image resides in a raw partition, while file-based environment storage requires a filesystem partition (EXT4/FAT). This would require an additional partition solely for environment storage

The current design is storing the FIT kernel in ESP as well, not in a separated partition (we don't even have that defined in qcom-ptool, and I don't think we should).

The environment file could also be stored in ESP.

ricardosalveti: > LogFS is currently used only by UEFI, when U-Boot replaces UEFI, LogFS is no longer utilized by any component.

Is this going to stay true? If logfs is only supposed to be used by UEFI, then it could be reused here, but this needs to be documented properly as I'm sure users / customer will be confused by this.

lumag: I'd second a suggestion to put both the kernel and the environment as files in the ESP partition. If you are limited to the raw images for some reason, then putting the environment to the end of the partition would be good enough option.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/ae5007be-61ad-42b7-9c31-d7e5dc116b9c)**
---
**Test jobs for commit [`eff506f`](https://github.com/qualcomm-linux/meta-qcom/commit/eff506f41b729381ad499433f70afa46770c1649)**
---
Pass: 51 | Fail: 1 | Total: 52
Pass: 9 | Fail: 0 | Total: 9

recipes-bsp/u-boot/files/fw_env.config
aswinm94: @ricardosalveti , @lumag ,
In the U-Boot SPL case, filesystem support is not enabled, so we cannot use a file-based environment. Given that, can we proceed with storing the U-Boot environment in the logfs partition, since it is not used by any other component?

lumag: Put the environment at the end of the partition with U-BOot.

aswinm94: @lumag , The environment is stored at the end of the partition. In fw_env.config, the offset is specified as -0x4000, which is interpreted relative to the end of the partition. This places the U-Boot environment in the last of the logfs partition.

ricardosalveti: @aswinm94 why exactly do you want to use environment support in SPL?

lumag: At the end of the _boot_ partition (instead of logfs).

aswinm94: @lumag, having a dedicated partition for U-Boot environment variables is the preferred approach.

In case of A/B setup, each slot stores its environment within its respective boot partition. However, OTA updates typically replace the boot image without updating the corresponding environment. Depending on how the update is performed, the inactive slot may end up with either a cleared environment or an outdated one that no longer matches the newly flashed image. When that slot later becomes active, this can lead to inconsistencies between the boot image and its environment.

Currently, we use the logfs partition for storing the U-Boot environment since it is not used by any other component. For future SoCs, the plan is to introduce a dedicated U-Boot environment partition

lumag: > having a dedicated partition for U-Boot environment variables is the preferred approach.

Is it documented so?

> However, OTA updates typically replace the boot image without updating the corresponding environment. 

Then OTA update scripts must be fixed accordingly

> we use the logfs partition for storing the U-Boot environment since it is not used by any other component. 

If it is unsed, remove it from the partition table

For future SoCs, the plan is to introduce a dedicated U-Boot environment partition

Why only for the future SoCs? You can introduce it now too.

aswinm94: @lumag ,
The logfs partition was originally used by UEFI and therefore cannot be removed from the partition table in existing products. Since UEFI has been replaced by U-Boot in the current boot flow, the logfs partition is no longer used.

However, dedication uboot env partition table changes cannot be made to products that are already in production. Such changes can only be considered for upcoming SoC platforms.

ricardosalveti: Is logfs really not used by the current UEFI downstream releases? If so, please remove the partition from qcom-ptool, as this won't cause issues on reflash.

For users performing OTA the partition will still be there and won't cause any issue.

Now for u-boot env, I don't think we should be creating a new partition for it, as we want to share a single partition table (if possible) between u-boot end edk2, otherwise it is a mess to maintain.

lumag: > However, dedication uboot env partition table changes cannot be made to products that are already in production. Such changes can only be considered for upcoming SoC platforms.

Then stop playing with it and, as I wrote earlier, just put environment at the end of the U-Boot partition. If there is anything to be handled by the OTA, it... must be handled by the OTA scripts rather than affecting the rest of the system design.
Dixi.

PR #2813, opened 2026-07-27
gst-plugins-imsdk-base: add opencv to base PACKAGECONFIG deps

skesava-qcom:
## Summary
- libgstqtivideobase's video-converter-engine only registers the "ocv" backend enum value when opencv4 is present in the recipe's own sysroot at configure time. The `base` PACKAGECONFIG never pulled in opencv, so gpu plugins only ever exposed the gles backend despite opencv runtime libs being installed on target.
- Add opencv to `PACKAGECONFIG[base]`'s deps so `ocv-video-converter.cc` is compiled in.


anujm1 (changes requested):
```
2026-07-28 01:37:32 - ERROR    - Command "/work/build$ /work/bitbake/bin/bitbake -c build world" failed:
ERROR: Nothing PROVIDES 'opencv' (but /work/build/../../repo/recipes-multimedia/imsdk/gst-plugins-imsdk-base_1.0.2.bb DEPENDS on or otherwise requires it)
ERROR: Required build target 'gst-plugins-imsdk-python' has no buildable providers.
Missing or unbuildable dependency chain was: ['gst-plugins-imsdk-python', 'gst-plugins-imsdk-base', 'opencv']
```

skesava-qcom:
> ```
> 2026-07-28 01:37:32 - ERROR    - Command "/work/build$ /work/bitbake/bin/bitbake -c build world" failed:
> ERROR: Nothing PROVIDES 'opencv' (but /work/build/../../repo/recipes-multimedia/imsdk/gst-plugins-imsdk-base_1.0.2.bb DEPENDS on or otherwise requires it)
> ERROR: Required build target 'gst-plugins-imsdk-python' has no buildable providers.
> Missing or unbuildable dependency chain was: ['gst-plugins-imsdk-python', 'gst-plugins-imsdk-base', 'opencv']
> ```

updated package will only build with opencv when meta open embedded layer is  present.

dynamic-layers/openembedded-layer/recipes-multimedia/imsdk/gst-plugins-imsdk-base_%.bbappend
```diff
@@ -0,0 +1,2 @@
+# opencv4 in the sysroot is what enables the "ocv" video engine at build time.
+DEPENDS += "opencv"
```
anujm1: umm, I think the `PACKAGECONFIG` from `gst-plugins-imsdk-base` should move to `dynamic-layers/openembedded-layer` and `opencv` should correctly be added as a dependency in `PACKAGECONFIG` for `base` in `gst-plugins-imsdk-common.inc`

lumag: I think, PACKAGECONFIG should be kept here (with the dependency). But the enablement of the PACKAGECONFIG should happen in the bbappend inside dynamic-layers.

koenkooi: @lumag agreed

skesava-qcom: None of `base`’s other dependencies come from `meta-openembedded`.

If we move enabling `base` into a dynamic-layers `.bbappend`, then `base` becomes conditional on `meta-openembedded`. In configurations without that layer, `PACKAGECONFIG` will use the disable option (`-DENABLE_GST_PLUGIN_BASE=OFF`), so `gst-plugin-base` will not be built and the base libraries will not be produced.

That would be a behavior change for setups that do not include `meta-openembedded`, because `gst-plugins-imsdk-oss`, `gst-plugins-imsdk-prop`, and `gst-plugins-imsdk-python` all depend on `gst-plugins-imsdk-base` and link against those libraries.

`opencv`  absence does not cause a build failure. Upstream detects it with `pkg_check_modules(OPEN_CV opencv4)` and only enables the additional `ocv` engine when it is present, alongside the existing `gles` engine. Keeping `opencv` as a plain `DEPENDS` in the dynamic layer adds that optional engine where `opencv` is available, without changing behavior where it is not.

lumag: I've written something completely different. Add PACKAGECONFIG for opencv.

dynamic-layers/openembedded-layer/recipes-multimedia/imsdk/gst-plugins-imsdk-base_%.bbappend
```diff
@@ -0,0 +1,2 @@
+# opencv4 in the sysroot is what enables the "ocv" video engine at build time.
+PACKAGECONFIG:append = " ocv"
```
lumag, line 2: Separate commit, drop unparseable comment.

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
```diff
@@ -31,6 +31,7 @@ PACKAGECONFIG[camera]       = "-DENABLE_GST_PLUGIN_CAMERA_BASE=ON -DENABLE_GST_P
 PACKAGECONFIG[camera-apps]  = "-DENABLE_GST_SAMPLE_APPS_CAMERA=1, -DENABLE_GST_SAMPLE_APPS_CAMERA=0, camera-service"
 PACKAGECONFIG[messaging]    = "-DENABLE_GST_MESSAGING_PLUGINS=1, -DENABLE_GST_MESSAGING_PLUGINS=0, librdkafka mosquitto"
 PACKAGECONFIG[ml]           = "-DENABLE_GST_ML_PLUGINS=1, -DENABLE_GST_ML_PLUGINS=0, cairo json-glib libeigen opencv"
+PACKAGECONFIG[ocv]          = ",, opencv"
```
lumag, line 34: Why is it `ocv`? `opencv` is more readable.

lumag (changes requested):

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
quaresmajose: Also please add something like "ENABLE_GST_OPENCV" on the IMSDK cmake build system, otherwise this will be host dependent and can be enabled inadvertently when the user has opencv on their system. Therefore it will break the reproducibility of IMSDK.

skesava-qcom: Adding an `ENABLE_GST_OPENCV` CMake option would require coordinated updates across multiple supported SPs and their existing build configurations. Introducing this option as part of the current change could unintentionally modify existing defaults or impact OpenCV availability for other consumers.

For this change, we propose retaining the existing CMake detection logic and controlling OpenCV support through the layer-gated `PACKAGECONFIG[opencv]` and the corresponding `.bbappend`. This ensures that OpenCV is added to the recipe-specific Yocto sysroot only when the feature is explicitly enabled, keeping the build behavior deterministic and preventing any influence from OpenCV installations on the host system.

Would this approach be acceptable for the current change?

lumag: >  CMake option would require coordinated updates across multiple supported SPs and their existing build configurations.

Why / how?

skesava-qcom: The same IMSDK source and CMake files are also used by downstream SP-specific recipes that do not use this meta-qcom recipe or its .bbappend.

Adding ENABLE_GST_OPENCV here would change the CMake interface for those downstream builds as well. If the option defaults to OFF, their OpenCV backend would be disabled until each corresponding .bb file passes -DENABLE_GST_OPENCV=ON. If the option is enabled and OpenCV is required, recipes without the matching dependency could fail during configuration.

Supporting this CMake option therefore requires coordinated changes across all downstream SP recipes. For now, we would like to limit this change to PACKAGECONFIG[opencv] and the layer-specific .bbappend

lumag: You can keep it ON by default. The recipe will turn it off.

quaresmajose: I don't see how adding a new knob could break anything. When downstream SP-specific recipes version is updated, simply adapt the new option as needed.

When ENABLE_GST_OPENCV is not specified in the input, it can internaly by cmake be enabled or disabled when opencv is available in sysroot, so downstream will still function the same way. But this way we fix the host contamination and reproducibility problem in the new versions.

ricardosalveti: Just make the default behave similarly as today, then our updates here will just do the right thing and enable/disable based on the dependencies.

skesava-qcom: Revisited the alternatives and updated the change to follow the suggested approach. Added ENABLE_GST_OPENCV (default ON) and PACKAGECONFIG[opencv] to explicitly control OpenCV enablement and dependencies.

dynamic-layers/openembedded-layer/recipes-multimedia/imsdk/gst-plugins-imsdk-base_%.bbappend
quaresmajose: This comment has not yet been fixed.

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
quaresmajose: This [comment](https://github.com/qualcomm-linux/meta-qcom/pull/2813#discussion_r3950351450) has not yet been fixed.

dynamic-layers/openembedded-layer/recipes-multimedia/imsdk/gst-plugins-imsdk-base_%.bbappend
skesava-qcom: separated commits now.

quaresmajose (dismissed):

quaresmajose:
Please rebase to fix conflicts.

quaresmajose (approved):

anujm1 (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/2a812038-f28c-4d89-a017-0bdc85165663)**
---
**Test jobs for commit [`a5c6b82`](https://github.com/qualcomm-linux/meta-qcom/commit/a5c6b82d6ce0e0c3c99512548b5b7d10bd12749d)**
---
Pass: 9 | Fail: 0 | Total: 9

ievlogie:
Why was a new CMake option introduced to enable OpenCV support?

Adding OpenCV as a dependency in the QTI plugins recipe should be sufficient, as this automatically propagates the OpenCV package configuration, headers, and libraries to the GStreamer plugins' CMake build system.

The intended design appears to be that the Qt/GStreamer CMake files automatically detect and use OpenCV whenever the package is available. The same mechanism is already applied to other external dependencies.

Based on this approach, the additional CMake enable flag seems unnecessary, as it introduces extra configuration complexity without providing any clear functional benefit.

quaresmajose:
> Why was a new CMake option introduced to enable OpenCV support?
> 
> Adding OpenCV as a dependency in the QTI plugins recipe should be sufficient, as this automatically propagates the OpenCV package configuration, headers, and libraries to the GStreamer plugins' CMake build system.
> 
> The intended design appears to be that the Qt/GStreamer CMake files automatically detect and use OpenCV whenever the package is available. The same mechanism is already applied to other external dependencies.
> 
> Based on this approach, the additional CMake enable flag seems unnecessary, as it introduces extra configuration complexity without providing any clear functional benefit.

Without this, the package cannot be reproducible. In other words, the same build system will produce different results depending on the host where it is executed, and this is unacceptable.

ievlogie:
> Without this, the package cannot be reproducible. In other words, the same build system will produce different results depending on the host where it is executed, and this is unacceptable.

This does not depend on the host; it depends on the recipes that are added as dependencies. The same codebase must be able to adapt to different build systems, each of which may provide a different compute backend, buffer manager, or other underlying components.

For this reason, the decision was made to rely entirely on the configuration provided by the package. This approach ensures flexibility and compatibility across various build environments.

If a consistent build configuration is required, OpenCV should be added as a recipe dependency in all relevant projects.

quaresmajose:
> > Without this, the package cannot be reproducible. In other words, the same build system will produce different results depending on the host where it is executed, and this is unacceptable.
> 
> This does not depend on the host; it depends on the recipes that are added as dependencies. The same codebase must be able to adapt to different build systems, each of which may provide a different compute backend, buffer manager, or other underlying components.

It depends on the host sysroot, yes, because it will be enabled or disabled depending on whether the dependencies are available.

Following this line of thought, we could remove all the other options present in the recipe, and everything would be controlled through dependencies. Why do you think only OpenCV should be controlled like that?

> 
> For this reason, the decision was made to rely entirely on the configuration provided by the package. This approach ensures flexibility and compatibility across various build environments.
> 
> If a consistent build configuration is required, OpenCV should be added as a recipe dependency in all relevant projects.

I didn't understand that point; it should be a dependency whenever it's used by the component in question. If dlopen is used, we need to add it explicitly to rdepends; otherwise, it's not necessary.

ievlogie:
> > > Without this, the package cannot be reproducible. In other words, the same build system will produce different results depending on the host where it is executed, and this is unacceptable.
> > 
> > 
> > This does not depend on the host; it depends on the recipes that are added as dependencies. The same codebase must be able to adapt to different build systems, each of which may provide a different compute backend, buffer manager, or other underlying components.
> 
> It depends on the host sysroot, yes, because it will be enabled or disabled depending on whether the dependencies are available.
> 
> Following this line of thought, we could remove all the other options present in the recipe, and everything would be controlled through dependencies. Why do you think only OpenCV should be controlled like that?
> 
> > For this reason, the decision was made to rely entirely on the configuration provided by the package. This approach ensures flexibility and compatibility across various build environments.
> > If a consistent build configuration is required, OpenCV should be added as a recipe dependency in all relevant projects.
> 
> I didn't understand that point; it should be a dependency whenever it's used by the component in question. If dlopen is used, we need to add it explicitly to rdepends; otherwise, it's not necessary.

The other dependencies enabled through the package configuration correspond to different components within the QTI GStreamer Plugins repository. These components are controlled through CMake options because various Yocto build configurations may require different sets of plugins, applications, or tools. This behavior is expected.

OpenCV, however, is not part of the QTI GStreamer Plugins source tree. Instead, it serves as a compute backend for some of the plugins. The same plugins may support alternative compute backends as well.

OpenCV support is enabled automatically when the corresponding development package is available. Therefore, there is no need for an additional CMake option to explicitly enable the OpenCV backend.

In general, the CMake options are intended to control the build of QTI GStreamer plugin components, test applications, and tools. They are not intended to manage external dependencies that those components may utilize.

quaresmajose:
> OpenCV, however, is not part of the QTI GStreamer Plugins source tree. Instead, it serves as a compute backend for some of the plugins. The same plugins may support alternative compute backends as well.

So why do you need it to be available in the sysroot for compilation?

> OpenCV support is enabled automatically when the corresponding development package is available. Therefore, there is no need for an additional CMake option to explicitly enable the OpenCV backend.

As already explained, this can break reproducibility. Have you tried checking if this same method works when you don't have OpenCV listed as a dependency in the recipe but you do have it installed on your host machine? Will BitBake consider the dependency to exist or not in this case?

ievlogie:
> So why do you need it to be available in the sysroot for compilation?

H files are needed ;)

> As already explained, this can break reproducibility. Have you tried checking if this same method works when you don't have OpenCV listed as a dependency in the recipe but you do have it installed on your host machine? Will BitBake consider the dependency to exist or not in this case?

BitBake is designed to configure an isolated sysroot that is independent of the host machine's sysroot. If the build is relying on host system libraries or development packages, that indicates a problem with the BitBake/Yocto configuration rather than with this project.

The expected behavior is that all required dependencies are provided through the build environment and its configured sysroot. If that is not happening, the BitBake configuration should be corrected to ensure proper isolation from the host system.

quaresmajose:
> > So why do you need it to be available in the sysroot for compilation?
> 
> H files are needed ;)

So, in the end, it's part of the code.

> 
> > As already explained, this can break reproducibility. Have you tried checking if this same method works when you don't have OpenCV listed as a dependency in the recipe but you do have it installed on your host machine? Will BitBake consider the dependency to exist or not in this case?
> 
> BitBake is designed to configure an isolated sysroot that is independent of the host machine's sysroot. If the build is relying on host system libraries or development packages, that indicates a problem with the BitBake/Yocto configuration rather than with this project.

So you think a build with Bitbake works the same regardless of the OS distribution used? Unfortunately, that's not true.

So why do you think the Yocto documentation says to install gcc (and all the other tools) on the host system as a dependency? According to your theory, it shouldn't be necessary because the host has no influence whatsoever.

> 
> The expected behavior is that all required dependencies are provided through the build environment and its configured sysroot. If that is not happening, the BitBake configuration should be corrected to ensure proper isolation from the host system.

If you can do this without using a container, please propose it upstream; the community will appreciate it.

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
```diff
@@ -30,6 +30,7 @@ PACKAGECONFIG[camera]       = "-DENABLE_GST_PLUGIN_CAMERA_BASE=ON -DENABLE_GST_P
 PACKAGECONFIG[camera-apps]  = "-DENABLE_GST_SAMPLE_APPS_CAMERA=1, -DENABLE_GST_SAMPLE_APPS_CAMERA=0, camera-service"
 PACKAGECONFIG[messaging]    = "-DENABLE_GST_MESSAGING_PLUGINS=1, -DENABLE_GST_MESSAGING_PLUGINS=0, librdkafka mosquitto"
 PACKAGECONFIG[ml]           = "-DENABLE_GST_ML_PLUGINS=1, -DENABLE_GST_ML_PLUGINS=0, cairo json-glib libeigen opencv"
+PACKAGECONFIG[opencv]       = "-DENABLE_GST_OPENCV=ON, -DENABLE_GST_OPENCV=OFF, opencv"
```
quaresmajose, line 33: I checked and opencv is not required in sysroot to build with -DENABLE_GST_OPENCV=ON.
```diff
--- a/recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
+++ b/recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
@@ -30,7 +30,7 @@ PACKAGECONFIG[camera]       = "-DENABLE_GST_PLUGIN_CAMERA_BASE=ON -DENABLE_GST_P
 PACKAGECONFIG[camera-apps]  = "-DENABLE_GST_SAMPLE_APPS_CAMERA=1, -DENABLE_GST_SAMPLE_APPS_CAMERA=0, camera-service"
 PACKAGECONFIG[messaging]    = "-DENABLE_GST_MESSAGING_PLUGINS=1, -DENABLE_GST_MESSAGING_PLUGINS=0, librdkafka mosquitto"
 PACKAGECONFIG[ml]           = "-DENABLE_GST_ML_PLUGINS=1, -DENABLE_GST_ML_PLUGINS=0, cairo json-glib libeigen opencv"
-PACKAGECONFIG[opencv]       = "-DENABLE_GST_OPENCV=ON, -DENABLE_GST_OPENCV=OFF, opencv"
+PACKAGECONFIG[opencv]       = "-DENABLE_GST_OPENCV=ON, -DENABLE_GST_OPENCV=OFF"
 PACKAGECONFIG[python]       = "-DENABLE_GST_PYTHON=ON -DENABLE_APP_BUILDER_PYTHON=1 -DPYTHON_SITEPACKAGES_DIR=${PYTHON_SITEPACKAGES_DIR}, \
                                -DENABLE_GST_PYTHON=OFF -DENABLE_APP_BUILDER_PYTHON=0, , gstreamer1.0-python python3-pyyaml gst-plugins-imsdk-base"
 PACKAGECONFIG[python-apps]  = "-DENABLE_GST_PYTHON_EXAMPLES=1, -DENABLE_GST_PYTHON_EXAMPLES=0"

```
I managed to compile the package anyway. so -DENABLE_GST_OPENCV is not working.

quaresmajose (changes requested):

recipes-multimedia/imsdk/gst-plugins-imsdk-common.inc
quaresmajose: There is no reference to ENABLE_GST_OPENCV in the source code! @skesava-qcom  can you explain what's going on here?

skesava-qcom: > There is no reference to ENABLE_GST_OPENCV in the source code! Siva Kesava (@skesava-qcom) can you explain what's going on here?

This recipe change is paired with a companion IMSDK CMake update that adds ENABLE_GST_OPENCV; meta-qcom will need to reference that source revision before this change is finalized.

The flag is mainly required for the disable path. Enabling the OCV backend remains dependency-driven: adding opencv to PACKAGECONFIG stages opencv4.pc, headers, and libraries into the recipe sysroot, allowing CMake to detect and build the backend.

This keeps existing downstream SP behavior unchanged, as OpenCV continues to be auto-detected when the flag is not provided.

Also, even with the source change, -DENABLE_GST_OPENCV=ON alone does not provide OpenCV. Without the opencv dependency, the build still succeeds because the lookup is optional, but the OCV backend is not compiled.

quaresmajose: Then a bump will be needed in SRCREV. In this case, you should have marked PR as a draft because it's not ready to be reviewed.

ievlogie:
> > > So why do you need it to be available in the sysroot for compilation?
> > 
> > 
> > H files are needed ;)
> 
> So, in the end, it's part of the code.
> 
> > > As already explained, this can break reproducibility. Have you tried checking if this same method works when you don't have OpenCV listed as a dependency in the recipe but you do have it installed on your host machine? Will BitBake consider the dependency to exist or not in this case?
> > 
> > 
> > BitBake is designed to configure an isolated sysroot that is independent of the host machine's sysroot. If the build is relying on host system libraries or development packages, that indicates a problem with the BitBake/Yocto configuration rather than with this project.
> 
> So you think a build with Bitbake works the same regardless of the OS distribution used? Unfortunately, that's not true.
> 
> So why do you think the Yocto documentation says to install gcc (and all the other tools) on the host system as a dependency? According to your theory, it shouldn't be necessary because the host has no influence whatsoever.
> 
> > The expected behavior is that all required dependencies are provided through the build environment and its configured sysroot. If that is not happening, the BitBake configuration should be corrected to ensure proper isolation from the host system.
> 
> If you can do this without using a container, please propose it upstream; the community will appreciate it.

Testing in two separate build environments on different machines shows that source code cannot access headers located in /usr/include, which matches the expected behavior.

Which environment allows source code to access headers that are not present in the recipe sysroot? If such access occurs, it invalidates the current assumptions about sysroot isolation and requires a review, and potentially a redesign, of the CMake configuration across all QTI GStreamer plugin projects to ensure correct dependency management and reproducible builds.

quaresmajose:
> > > > So why do you need it to be available in the sysroot for compilation?
> > > 
> > > 
> > > H files are needed ;)
> > 
> > 
> > So, in the end, it's part of the code.
> > > > As already explained, this can break reproducibility. Have you tried checking if this same method works when you don't have OpenCV listed as a dependency in the recipe but you do have it installed on your host machine? Will BitBake consider the dependency to exist or not in this case?
> > > 
> > > 
> > > BitBake is designed to configure an isolated sysroot that is independent of the host machine's sysroot. If the build is relying on host system libraries or development packages, that indicates a problem with the BitBake/Yocto configuration rather than with this project.
> > 
> > 
> > So you think a build with Bitbake works the same regardless of the OS distribution used? Unfortunately, that's not true.
> > So why do you think the Yocto documentation says to install gcc (and all the other tools) on the host system as a dependency? According to your theory, it shouldn't be necessary because the host has no influence whatsoever.
> > > The expected behavior is that all required dependencies are provided through the build environment and its configured sysroot. If that is not happening, the BitBake configuration should be corrected to ensure proper isolation from the host system.
> > 
> > 
> > If you can do this without using a container, please propose it upstream; the community will appreciate it.
> 
> Testing in two separate build environments on different machines shows that source code cannot access headers located in /usr/include, which matches the expected behavior.
> 
> Which environment allows source code to access headers that are not present in the recipe sysroot? If such access occurs, it invalidates the current assumptions about sysroot isolation and requires a review, and potentially a redesign, of the CMake configuration across all QTI GStreamer plugin projects to ensure correct dependency management and reproducible builds.

All of CMake options must also be manually controlled independently of automatic detection. This is how most open-source projects work, especially the most reputable ones in terms of quality, and there's no reason why we shouldn't do the same.

ievlogie:
> > > > > So why do you need it to be available in the sysroot for compilation?
> > > > 
> > > > 
> > > > H files are needed ;)
> > > 
> > > 
> > > So, in the end, it's part of the code.
> > > > > As already explained, this can break reproducibility. Have you tried checking if this same method works when you don't have OpenCV listed as a dependency in the recipe but you do have it installed on your host machine? Will BitBake consider the dependency to exist or not in this case?
> > > > 
> > > > 
> > > > BitBake is designed to configure an isolated sysroot that is independent of the host machine's sysroot. If the build is relying on host system libraries or development packages, that indicates a problem with the BitBake/Yocto configuration rather than with this project.
> > > 
> > > 
> > > So you think a build with Bitbake works the same regardless of the OS distribution used? Unfortunately, that's not true.
> > > So why do you think the Yocto documentation says to install gcc (and all the other tools) on the host system as a dependency? According to your theory, it shouldn't be necessary because the host has no influence whatsoever.
> > > > The expected behavior is that all required dependencies are provided through the build environment and its configured sysroot. If that is not happening, the BitBake configuration should be corrected to ensure proper isolation from the host system.
> > > 
> > > 
> > > If you can do this without using a container, please propose it upstream; the community will appreciate it.
> > 
> > 
> > Testing in two separate build environments on different machines shows that source code cannot access headers located in /usr/include, which matches the expected behavior.
> > Which environment allows source code to access headers that are not present in the recipe sysroot? If such access occurs, it invalidates the current assumptions about sysroot isolation and requires a review, and potentially a redesign, of the CMake configuration across all QTI GStreamer plugin projects to ensure correct dependency management and reproducible builds.
> 
> All of CMake options must also be manually controlled independently of automatic detection. This is how most open-source projects work, especially the most reputable ones in terms of quality, and there's no reason why we shouldn't do the same.

Please provide instructions for reproducing a Yocto build environment in which headers from /usr/include are considered during a recipe build.

A recipe build is expected to remain isolated within the recipe sysroot and use only the dependencies explicitly provided through that sysroot. This behavior is consistent across all existing build environments, and as a result, the original issue cannot be reproduced.

The dependency model in this repository intentionally relies on automatic detection of available dependencies. The same codebase targets multiple platforms and build configurations, and automatic detection of dependent headers is a deliberate design decision that enables the code to adapt to the capabilities available in each environment.

This design will remain unchanged.

quaresmajose:
> > > > > > So why do you need it to be available in the sysroot for compilation?
> > > > > 
> > > > > 
> > > > > H files are needed ;)
> > > > 
> > > > 
> > > > So, in the end, it's part of the code.
> > > > > > As already explained, this can break reproducibility. Have you tried checking if this same method works when you don't have OpenCV listed as a dependency in the recipe but you do have it installed on your host machine? Will BitBake consider the dependency to exist or not in this case?
> > > > > 
> > > > > 
> > > > > BitBake is designed to configure an isolated sysroot that is independent of the host machine's sysroot. If the build is relying on host system libraries or development packages, that indicates a problem with the BitBake/Yocto configuration rather than with this project.
> > > > 
> > > > 
> > > > So you think a build with Bitbake works the same regardless of the OS distribution used? Unfortunately, that's not true.
> > > > So why do you think the Yocto documentation says to install gcc (and all the other tools) on the host system as a dependency? According to your theory, it shouldn't be necessary because the host has no influence whatsoever.
> > > > > The expected behavior is that all required dependencies are provided through the build environment and its configured sysroot. If that is not happening, the BitBake configuration should be corrected to ensure proper isolation from the host system.
> > > > 
> > > > 
> > > > If you can do this without using a container, please propose it upstream; the community will appreciate it.
> > > 
> > > 
> > > Testing in two separate build environments on different machines shows that source code cannot access headers located in /usr/include, which matches the expected behavior.
> > > Which environment allows source code to access headers that are not present in the recipe sysroot? If such access occurs, it invalidates the current assumptions about sysroot isolation and requires a review, and potentially a redesign, of the CMake configuration across all QTI GStreamer plugin projects to ensure correct dependency management and reproducible builds.
> > 
> > 
> > All of CMake options must also be manually controlled independently of automatic detection. This is how most open-source projects work, especially the most reputable ones in terms of quality, and there's no reason why we shouldn't do the same.
> 
> Please provide instructions for reproducing a Yocto build environment in which headers from /usr/include are considered during a recipe build.

The reproducible issue I mentioned isn't a symptom of gst-plugins-imsdk-base; consequently, I'm not even sure if it's affected by that.

What I mentioned is a generic problem that can easily be avoided by explicitly specifying the options we intend to use.

From your perspective, build options aren't necessary at all because dependencies handle that. However as I’ve already pointed out, that approach is problematic.

> 
> A recipe build is expected to remain isolated within the recipe sysroot and use only the dependencies explicitly provided through that sysroot. This behavior is consistent across all existing build environments, and as a result, the original issue cannot be reproduced.
> 
> The dependency model in this repository intentionally relies on automatic detection of available dependencies. The same codebase targets multiple platforms and build configurations, and automatic detection of dependent headers is a deliberate design decision that enables the code to adapt to the capabilities available in each environment.
> 
> This design will remain unchanged.

Once again:

_All of CMake options must also be manually controlled independently of automatic detection. This is how most open-source projects work, especially the most reputable ones in terms of quality, and there's no reason why we shouldn't do the same._

Please take a look on [gstreamer](https://git.openembedded.org/openembedded-core/tree/meta/recipes-multimedia/gstreamer) receipes in openembedded-core and how options are used.

ricardosalveti (changes requested):
I agree with Jose, and I don't think we should merge this with a flag that CMake ignores.

The reproducibility problem is not bitbake picking up headers from /usr/include. It is the same source and the same cmake line producing a different library depending on what happens to be in the sysroot, and that is already happening. The qimsdk CI ([run 35827347941](https://github.com/qualcomm/qimsdk/actions/runs/35827347941)) builds gst-plugin-base with just `-DENABLE_GST_PLUGIN_BASE=1` against our wrynose nightly SDK and gets three converters:

```
-- Found qcom-fastcv-binaries, version 1.8
-- Found opencv4, version 4.13.0
[ 46%] Building CXX object .../gstqtivideobase.dir/gles-video-converter.cc.o
[ 48%] Building C object .../gstqtivideobase.dir/fcv-video-converter.c.o
[ 50%] Building CXX object .../gstqtivideobase.dir/ocv-video-converter.cc.o
```

Our recipe on master only builds `gles-video-converter.cc`. Nothing failed, it was only noticed on target, and that's why this PR exists. FastCV still differs today, and whether we want it or not should be a decision in the recipe, not a side effect of the sysroot.

With this PR the flag is a no-op in both directions, since no qimsdk branch or tag defines it:

```
CMake Warning (unused-cli):
  Manually-specified variables were not used by the project:
    ENABLE_GST_OPENCV
```

So `PACKAGECONFIG:remove = "opencv"` only works while nothing else brings opencv into this sysroot. In a master build the sysroot of this recipe already has 101 target recipes and only 8 of them come from its own DEPENDS, so DEPENDS can add opencv but can never keep it out. 

What we need from qimsdk is small and doesn't change anything for other users: an option per external backend, defaulting to AUTO (today's behaviour), where ON fails if the dependency is missing and OFF skips the check. That's how GStreamer handles its own `opencv` option (`auto` by default), and oe-core pins it with `-Dopencv=enabled/disabled`. Once that's in a qimsdk release, this PR can go in on top of that bump.

Example for gst-plugin-base/CMakeLists.txt:

```cmake
set(ENABLE_GST_OPENCV AUTO CACHE STRING "OpenCV converter backend: AUTO, ON or OFF")
if(ENABLE_GST_OPENCV)
  pkg_check_modules(OPEN_CV opencv4)
  if(NOT ENABLE_GST_OPENCV STREQUAL "AUTO" AND NOT OPEN_CV_FOUND)
    message(FATAL_ERROR "ENABLE_GST_OPENCV=ON but opencv4 was not found")
  endif()
else()
  set(OPEN_CV_FOUND FALSE)
endif()
```

ievlogie:
> I agree with Jose, and I don't think we should merge this with a flag that CMake ignores.
> 
> The reproducibility problem is not bitbake picking up headers from /usr/include. It is the same source and the same cmake line producing a different library depending on what happens to be in the sysroot, and that is already happening. The qimsdk CI ([run 35827347941](https://github.com/qualcomm/qimsdk/actions/runs/35827347941)) builds gst-plugin-base with just `-DENABLE_GST_PLUGIN_BASE=1` against our wrynose nightly SDK and gets three converters:
> 
> ```
> -- Found qcom-fastcv-binaries, version 1.8
> -- Found opencv4, version 4.13.0
> [ 46%] Building CXX object .../gstqtivideobase.dir/gles-video-converter.cc.o
> [ 48%] Building C object .../gstqtivideobase.dir/fcv-video-converter.c.o
> [ 50%] Building CXX object .../gstqtivideobase.dir/ocv-video-converter.cc.o
> ```
> 
> Our recipe on master only builds `gles-video-converter.cc`. Nothing failed, it was only noticed on target, and that's why this PR exists. FastCV still differs today, and whether we want it or not should be a decision in the recipe, not a side effect of the sysroot.
> 
> With this PR the flag is a no-op in both directions, since no qimsdk branch or tag defines it:
> 
> ```
> CMake Warning (unused-cli):
>   Manually-specified variables were not used by the project:
>     ENABLE_GST_OPENCV
> ```
> 
> So `PACKAGECONFIG:remove = "opencv"` only works while nothing else brings opencv into this sysroot. In a master build the sysroot of this recipe already has 101 target recipes and only 8 of them come from its own DEPENDS, so DEPENDS can add opencv but can never keep it out.
> 
> What we need from qimsdk is small and doesn't change anything for other users: an option per external backend, defaulting to AUTO (today's behaviour), where ON fails if the dependency is missing and OFF skips the check. That's how GStreamer handles its own `opencv` option (`auto` by default), and oe-core pins it with `-Dopencv=enabled/disabled`. Once that's in a qimsdk release, this PR can go in on top of that bump.
> 
> Example for gst-plugin-base/CMakeLists.txt:
> 
> ```cmake
> set(ENABLE_GST_OPENCV AUTO CACHE STRING "OpenCV converter backend: AUTO, ON or OFF")
> if(ENABLE_GST_OPENCV)
>   pkg_check_modules(OPEN_CV opencv4)
>   if(NOT ENABLE_GST_OPENCV STREQUAL "AUTO" AND NOT OPEN_CV_FOUND)
>     message(FATAL_ERROR "ENABLE_GST_OPENCV=ON but opencv4 was not found")
>   endif()
> else()
>   set(OPEN_CV_FOUND FALSE)
> endif()
> ```

The design of this codebase intentionally limits CMake options to components that are part of the QTI GStreamer Plugins repository. All such components are disabled by default, and individual plugins, tools, or applications can be enabled through recipe configuration, DISTRO_FEATURES, or other build-specific settings.

Some of these plugins support optional dependencies such as FastCV, OpenCL, OpenGL, OpenCV, and others. These are not repository components but external dependencies. The build system automatically detects their availability based on the headers and libraries present in the recipe sysroot.

It is also important to note that multiple versions of the same dependency may exist. The current CMake implementation already handles this through automatic detection and selection of the available dependency version. As a result, there is no need for additional recipe flags to enable a particular backend. Once the dependency is added to the recipe dependencies and becomes available in the sysroot, the corresponding headers and libraries become visible to CMake, which then enables the supported functionality automatically.

Introducing an additional recipe flag would duplicate information that is already available through dependency management and CMake dependency detection, without providing any additional benefit.

The same approach is used throughout the QIM SDK and other QTI codebases for many optional dependencies, not only OpenCV. Furthermore, this code is deployed across multiple build environments, where this mechanism has been used successfully and consistently.

When recipe dependencies are defined correctly, the build remains fully reproducible. The presence or absence of optional functionality derives solely from the dependencies available in the recipe sysroot, which is the expected and intended behavior.

koenkooi:
I had an unrelated meeting with @ievlogie today and discussed this. What I think we don't explicitly mention here is openCV getting dragged in by a transitive dependency (e.g. CamX has a mandatory fastcv plugin which drags in openCV), leading to openCV being present, even when we explicitly don't list in the DEPENDS in gst-plugins-imsdk-base.

ievlogie:
> I agree with Jose, and I don't think we should merge this with a flag that CMake ignores.

Following a discussion with Koen, the technical issue related to transitive dependencies has been clarified. A CMake change is indeed required, and there is agreement on the need to address it.

However, the change must be applied consistently across all indirect QIM SDK dependencies, making the scope of the work significantly broader than a localized fix. It will require corresponding updates to recipes, build scripts, and dependency handling throughout the affected components.

Given the extent of the changes, thorough validation across all supported build environments will also be necessary to ensure compatibility and avoid regressions.

As a result, reviewing, implementing, testing, and approving this work will require additional time.

PR #2814, opened 2026-07-27
qcom-common: force PCIe root port D3 during suspend for validation

krishnachaitanya-linux:
Any PCIe Root Port that is a native OS-handled hotplug-capable bridge stays in D0 across system suspend instead of transitioning to D3hot.

drivers/pci/pci.c:pci_bridge_d3_possible() explicitly rejects D3 for this class of port:

    if (bridge->is_hotplug_bridge && !bridge->is_pciehp)
            return false;
    if (bridge->is_pciehp && !pciehp_is_native(bridge))
            return false;
    ...
    if (bridge->is_pciehp)
            return false;

so pci_dev->bridge_d3 is left cleared. pci_power_manageable() (drivers/pci/pci.h) in turn reports the bridge as not power manageable whenever it has a subordinate bus and bridge_d3 isn't set, and the suspend path in drivers/pci/pci-driver.c gates pci_prepare_to_sleep() behind that same check:

    if (!pci_dev->skip_bus_pm && pci_power_manageable(pci_dev))
            pci_prepare_to_sleep(pci_dev);

so pci_prepare_to_sleep() is skipped entirely and the Root Port never leaves D0, observed as:

    bridge_d3 = 0
    pci_power_manageable = 0

The kernel already has a command line hook for this: pcie_port_pm=force maps to pci_bridge_d3_force in pci_bridge_d3_possible(), which short- circuits the hotplug checks above and always allows D3:

    if (pci_bridge_d3_force)
            return true;

Add it for all qcom targets so bridge_d3 and pci_power_manageable get set to 1, allowing hotplug-capable Root Ports to transition to D3hot during suspend. This is meant to validate that the D0 behavior is caused by the generic hotplug bridge D3 policy rather than a platform-specific issue.

### DIFF
--- a/conf/machine/include/qcom-common.inc
+++ b/conf/machine/include/qcom-common.inc
@@ -95,3 +95,5 @@ RT_KERNEL_CMDLINE = "${RT_ARGS_ISOL} ${RT_ARGS_IRQ} ${RT_ARGS_NOCBS} \
 KERNEL_CMDLINE_EXTRA:append = "${@bb.utils.contains_any('PREFERRED_PROVIDER_virtual/kernel',\
                               'linux-qcom-rt linux-qcom-next-rt', \
                               ' ${RT_KERNEL_CMDLINE}', '', d)}"
+
+KERNEL_CMDLINE_EXTRA:append = " pcie_port_pm=force"


qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/11f85fba-8272-4a0a-837d-9b3986214cc0)**
---
**Test jobs for commit [`5575bf1`](https://github.com/qualcomm-linux/meta-qcom/commit/5575bf16bab8768a380e3b575476fe6ba0271cb5)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 47 | Fail: 1 | Total: 48

Mani-Sadhasivam:
@krishnachaitanya-linux I don't think we should pursue this path of adding kernel command line. First of all, cmdline params will make life harder for distro developers as they would need to add this special cmdline only for Qcom host platforms.

We should try to fix this issue in the kernel itself. The core issue here is that, ever since we enabled hotplug support for our RPs, pci_bridge_d3_possible() is returning false due to the opaque check:

```
if (bridge->is_pciehp)
        return false;
```

The comment about this check indicates that the code is trying to be conservative here as enabling D3 resulted in some crash on Intel platforms back in 2018. We need to take it to the PCI community and figure out a way to allow D3 for hotplug capable RPs on DT platforms.

krishnachaitanya-linux:
> krishnachaitanya-linux I don't think we should pursue this path of adding kernel command line. First of all, cmdline params will make life harder for distro developers as they would need to add this special cmdline only for Qcom host platforms.
> 
> We should try to fix this issue in the kernel itself. The core issue here is that, ever since we enabled hotplug support for our RPs, pci_bridge_d3_possible() is returning false due to the opaque check:
> 
> ```
> if (bridge->is_pciehp)
>         return false;
> ```
> 
> The comment about this check indicates that the code is trying to be conservative here as enabling D3 resulted in some crash on Intel platforms back in 2018. We need to take it to the PCI community and figure out a way to allow D3 for hotplug capable RPs on DT platforms.

Ack, I will close this.

PR #2815, opened 2026-07-27
Pre-populate the Sstate Cache folder for the upcoming month

sampra2025:
This will help avoid job slowness during the monthly cache setup. On meta-qcom, each build uses a cache from a monthly folder, so on the first day of the month there is no cache for that month and most builds run as clean builds, which takes longer to build the target and can potentially increase stabilization issues as well as cost.

This change will pre-populate the upcoming month folder with the existing cache and clean up cache older than 30days.

### DIFF
--- a/.github/workflows/monthly.yml
+++ b/.github/workflows/monthly.yml
@@ -1,8 +1,8 @@
-name: Monthly job
+name: Monthly job - pre-poulate cache
 
 on:
   schedule:
-  - cron: "22 10 5 * *"   # montly job - at "random" time - top of hour can be busy in github
+  - cron: "22 10 28 * *"   # montly job - at "random" time - top of hour can be busy in github
 
 jobs:
   sstate-cache-cleanup:
@@ -10,6 +10,10 @@ jobs:
     runs-on: [self-hosted, qcom-sm-u2404, sm-amd64]
     timeout-minutes: 720
     steps:
+      - name: Setup Sstate-cache for upcoming month
+        run: |
+          NEXT_MONTH=$(date -d "$(date +%Y-%m-01) +1 month" +%Y-%m)
+          cp -r -v /efs/qli/meta-qcom/sstate-cache-$(date +%Y-%m)/* /efs/qli/meta-qcom/sstate-cache-${NEXT_MONTH}
       - name: Clean up the persistant sstate-cache dir
         run: |
-          rm -rf $(find /efs/qli/meta-qcom -name 'sstate-cache-*' -maxdepth 1 -type d -ctime +30)
+          rm -rf $(find /efs/qli/meta-qcom/sstate-cache-${NEXT_MONTH}/ -maxdepth 2 -type d -ctime +30 -name '*')


anujm1:
> This will help avoid job slowness during the monthly cache setup. On meta-qcom, each build uses a cache from a monthly folder, so on the first day of the month there is no cache for that month and most builds run as clean builds, which takes longer to build the target and can potentially increase stabilization issues as well as cost.

Can you share more about the stabilization issues? I do not think there should be any issues because of presence or absence of sstate.

Not related to this change, I am wondering why we have monthly sstate directories if we can clean up 'older than 30 days' files in the same directory to save space.

ndechesne:
> > This will help avoid job slowness during the monthly cache setup. On meta-qcom, each build uses a cache from a monthly folder, so on the first day of the month there is no cache for that month and most builds run as clean builds, which takes longer to build the target and can potentially increase stabilization issues as well as cost.
> 
> Can you share more about the stabilization issues? I do not think there should be any issues because of presence or absence of sstate.
> 

Agreed here. Also the reason why we created a 'fresh' sstate which is empty is to clean up everything at once. So if we make a copy then this is going against the original idea of doing the cleanup. 

> Not related to this change, I am wondering why we have monthly sstate directories if we can clean up 'older than 30 days' files in the same directory to save space.

the -ctime 30 is here to remove files 'created' more than 1 month ago, to ensure that we delete old files. We've seen cases where the monthly job timed out and left 'old sstate' in there, so they will be deleted 'next' month. 

Regardless we should not merge this patch, for 2 reasons:
1. we've learned recently that bitbake is 'touching' all files it is using during a build. so we could largely simplify the sstate management and delete unused files with a single -mtime 15 find command, and removes the 'per month' folder. I believe @quaresmajose is looking into that
2. @ricardosalveti is looking at using S3 instead of EFS, as another alternative

With either 2 or 3 , we don't need this patch anyway.

koenkooi:
Regardless, please adhere to the commit policy, have a look at CONTRIBUTING.md. The easiest way is to have an agent ingest AGENTS.md and advice you on the commit message.

quaresmajose:
> > > This will help avoid job slowness during the monthly cache setup. On meta-qcom, each build uses a cache from a monthly folder, so on the first day of the month there is no cache for that month and most builds run as clean builds, which takes longer to build the target and can potentially increase stabilization issues as well as cost.
> > 
> > 
> > Can you share more about the stabilization issues? I do not think there should be any issues because of presence or absence of sstate.
> 
> Agreed here. Also the reason why we created a 'fresh' sstate which is empty is to clean up everything at once. So if we make a copy then this is going against the original idea of doing the cleanup.
> 
> > Not related to this change, I am wondering why we have monthly sstate directories if we can clean up 'older than 30 days' files in the same directory to save space.
> 
> the -ctime 30 is here to remove files 'created' more than 1 month ago, to ensure that we delete old files. We've seen cases where the monthly job timed out and left 'old sstate' in there, so they will be deleted 'next' month.
> 
> Regardless we should not merge this patch, for 2 reasons:
> 
> 1. we've learned recently that bitbake is 'touching' all files it is using during a build. so we could largely simplify the sstate management and delete unused files with a single -mtime 15 find command, and removes the 'per month' folder. I believe Jose Quaresma (@quaresmajose) is looking into that
> 2. Ricardo Salveti (@ricardosalveti) is looking at using S3 instead of EFS, as another alternative
> 
> With either 2 or 3 , we don't need this patch anyway.

Yes, I have planned to do a daily cleanup using  `-mtime 15 find` or somethin similar.
With this, we will no longer need to create new folders monthly and we can start storing them in a folder without dates.

ricardosalveti:
> > 1. we've learned recently that bitbake is 'touching' all files it is using during a build. so we could largely simplify the sstate management and delete unused files with a single -mtime 15 find command, and removes the 'per month' folder. I believe Jose Quaresma (Jose Quaresma (@quaresmajose)) is looking into that
> > 2. Ricardo Salveti (Ricardo Salveti (@ricardosalveti)) is looking at using S3 instead of EFS, as another alternative
> > 
> > With either 2 or 3 , we don't need this patch anyway.
> 
> Yes, I have planned to do a daily cleanup using `-mtime 15 find` or somethin similar. With this, we will no longer need to create new folders monthly and we can start storing them in a folder without dates.

We noticed that s3 ends up being as expensive as efs, so we're now looking at moving to FSx, so with that we should indeed just do a weekly (daily might be too much) run to perform a cleanup using -mtime 15 (on a shared sstate-cache folder, not separated per month as we have now).

We should also be able to land both in parallel, since mtime can already be used with efs.

quaresmajose:
Yes, I will give this task some attention as soon as possible.

ricardosalveti:
Created https://github.com/qualcomm-linux/meta-qcom/pull/2971 as we won't be going the s3 route anymore (we're now experimenting with FSx, which behaves similarly to EFS).

ricardosalveti:
Will close this one in favor of https://github.com/qualcomm-linux/meta-qcom/pull/2971.

PR #2833, opened 2026-07-29
selinux: Fix urm avc denials for capabilities and process setsched

varunsinghal29:
URM needs sys_ptrace to read /proc/<pid>/cmdline, sys_admin for sysfs and cgroup management, perfmon for performance monitoring, and setsched for moving processes to cgroups via cgroup.procs.

### DIFF
--- a/dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted/0003-Update-SELinux-policy-for-URM.patch
+++ b/dynamic-layers/selinux/recipes-security/refpolicy/refpolicy-targeted/0003-Update-SELinux-policy-for-URM.patch
@@ -1,0 +1,37 @@
+From 9beea5f87542e4c83cc1297a6c13a47a6f10c3d7 Mon Sep 17 00:00:00 2001
+From: Varun Singhal <varusing@qti.qualcomm.com>
+Date: Wed, 29 Jul 2026 11:36:32 +0530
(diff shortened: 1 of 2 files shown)


koenkooi:
Can this live in meta-selinux instead of here? URM is qcom specific, but the patch was submitted upstream, so we should try to keep it in an upstream layer as well.

varunsinghal29:
@koenkooi 
This PR fixes issue https://orbit/cr/4588185 i.e Observing Fps drops in GFX-Manhattan use case

The upstream PR https://github.com/SELinuxProject/refpolicy/pull/1188 is pending merge and will take time to flow into meta-qcom via meta-selinux.
This change is a temporary fix until the upstream PR is merged and picked up by meta-selinux into meta-qcom, at which point this patch will be removed.

koenkooi:
> Koen Kooi (@koenkooi) This PR fixes issue https://orbit/cr/4588185 i.e Observing Fps drops in GFX-Manhattan use case
> 
> The upstream PR [SELinuxProject/refpolicy#1188](https://github.com/SELinuxProject/refpolicy/pull/1188) is pending merge and will take time to flow into meta-qcom via meta-selinux. This change is a temporary fix until the upstream PR is merged and picked up by meta-selinux into meta-qcom, at which point this patch will be removed.

What's stopping you from putting this patch in meta-selinux?

PR #2802, opened 2026-07-27
Fix licence handling

lumag:
OE-Core is has switched to use SPDX licence expression. Follow the change, removing the warnings generated by the layer. Also a lot of recipes here were not able to specify the licence for the repackaged firmware, so drop them from this layer. They will be moved to a separate layer later on.

PR #2836, opened 2026-07-29
systemd: drop SectionAlignment patch merged upstream

ndechesne:
The patch 0001-boot-stub-honor-PE-SectionAlignment-when-loading-inn.patch was merged upstream (commit 90541abf6ed) and is now part of systemd v261.1, so it no longer applies after the oe-core upgrade.

Drop the patch.

Please do not merge until we update the local kas lock files, which is currently locked prior to the systemd upgrade


vkraleti:
We first need to merge https://github.com/qualcomm-linux/meta-qcom/pull/2802 then bump oe-core to the latest. As part of that PR will pick this revert.

ndechesne:
ah, right. I searched first to see if it was fixed somewhere, and missed it.. this PR can be closed then.

PR #2738, opened 2026-07-15
camx: revision update for Hamoa, Lemans, Talos,Kodiak

gkhose-qipl:
camx: Rename cax library to avoid clash with open source libcamera library.

camxcommon-headers: Update to the 1.0.12 revision.

- Source/header mismatch causes CamX compilation failure.
- Update the headers tar to align with the sources.
- Change SRC_URI to use the prebuilt_yocto_master artifact path, which is the current location of the published tarball.

camxlib-kodiak: Update to the 1.0.25 revision.

- Change SRC_URI to use the prebuilt_yocto_master artifact path, which is the current location of the published tarball.
- Rename libcamera_metadata, libcamera_hardware to libcamx_metadata, libcamx_hardware to avoid clash with open source libcamera library.

common.inc: Change SRC_URI and update to the 1.0.28 revision
common:

- Change SRC_URI to use the prebuilt_yocto_master artifact path, which is he current location of the published tarball.
- Rename libcamera_metadata, libcamera_hardware to libcamx_metadata, libcamx_hardware to avoid clash with open source libcamera library.
- Update RPATH to include target architectures for debian builds.

camxlib-lemans:

-  Enable SW-MCTF(Software Motion Compensated Temporal Filter) support.

camxlib-talos:

-  Add AR0231 GMSL sensor driver support
- Add OX03F10 RAW/YUV GMSL sensor driver support
- Add SerDes support for max9296a (deserializer) and max9295a/max96717 (serializers)
- Enable the common Hidrx (High Dynamic Range X) algorithm library for Talos (QCS615) to support BayerFusion and YUVFusion SHDR usecases.
- Update camxlib-talos recipe to include sensor binaries, Hidrx library and EIS/LDC feature library.

camxlib-hamoa:

- Add IMX688 sensor driver support for Hamoa and Purwa platforms
- Enable OG0VA1B sensor integration for Hamoa and Purwa

camera-service: upgrade 1.0.2 -> 1.0.4
Upgrade camera-service recipes tag from 1.0.2 to 1.0.4

Highlights:
- Add missing copyrights
- Correctly track and assign callback buffer FDs
- Use CMAKE_INSTALL_LIBDIR for destination install paths
- Support CamX 2.0 lib names and header paths
- Preserve run directory during service restart
- Added SocId and string name for glymur

PR #2849, opened 2026-07-30
camera-service: upgrade 1.0.2 -> 1.0.4

shukulk:
Upgrade camera-service recipes tag from 1.0.2 to 1.0.4

Highlights:
- Add missing copyrights
- Correctly track and assign callback buffer FDs
- Use CMAKE_INSTALL_LIBDIR for destination install paths
- Support CamX 2.0 lib names and header paths
- Use existing install path for 1.x for debian
- Preserve run directory during service restart
- Added SocId and string name for glymur
-  Added camxlib-talos camxlib-hamoa runtime dependencies
    to ${PN}-server-lib

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-multimedia/camera-service/camera-service_1.0.4.bb
+++ b/dynamic-layers/openembedded-layer/recipes-multimedia/camera-service/camera-service_1.0.4.bb
@@ -7,7 +7,7 @@ HOMEPAGE = "https://github.com/qualcomm/camera-service"
 
 SRC_URI = "git://github.com/qualcomm/camera-service;protocol=https;nobranch=1;tag=${PV}"
 
-SRCREV = "666344262e69a6c2c5e1df9e9e52b3e74c2f4a33"
+SRCREV = "9f50e49824ed112593e6643a9197e03922cbc807"
 
 # Limit this recipe to ARMv8 (aarch64) only, because it depends
 # on camxcommon-headers which is explicitly restricted to ARMv8 (aarch64).


anujm1 (changes requested):
Commit sha: [7c0bf79](https://github.com/qualcomm-linux/meta-qcom/pull/2849/commits/7c0bf7921be8343fe4f67c8053f85f60ed6da756), Author: shukulk, Committer: shukulk; Expected "shukulk [shukulk@qti.qualcomm.com](mailto:shukulk@qti.qualcomm.com)", but got "Shubhankar Kulkarni [shukulk@qti.qualcomm.com](mailto:shukulk@qti.qualcomm.com)".

shukulk:
Depends on #2738

lumag (changes requested):
> From: shukulk <shukulk@qti.qualcomm.com>

Fix your Git configuration to stop using login name instead of a full name.

dynamic-layers/openembedded-layer/recipes-multimedia/camera-service/camera-service_1.0.4.bb
```diff
@@ -67,7 +67,7 @@ RDEPENDS:${PN} += " \
 "
 
 # Add camx libraries as runtime deps since they are dlopened
-RDEPENDS:${PN}-server-lib += "camxlib-lemans"
+RDEPENDS:${PN}-server-lib += "camxlib-lemans camxlib-talos camxlib-hamoa"
```
lumag: Do we depend on all of them? Do we need to install all of them for example on Talos?

lumag (changes requested):
Fails to build

dynamic-layers/openembedded-layer/recipes-multimedia/camera-service/camera-service_1.0.4.bb
shukulk: Yes, this change is required because we use dlopen() to load target-specific libraries with different library names (for example, a _targetname suffix) based on the detected SOC_ID.
In configurations such as config1 + camera, if the camera service is installed on a Talos device, the Talos-specific CAMX libraries are not automatically pulled in.
To ensure the required libraries are available at runtime, we need to add explicit RDEPENDS for the Talos CAMX libraries. Please let me know if there is a better or more appropriate way to handle this dependency management.

lumag: No, you need to RDEPEND on the one, specific to your platform. Use virtual runtime providers, define those in the machine files.

shukulk: > No, you need to RDEPEND on the one, specific to your platform. Use virtual runtime providers, define those in the machine files.

I agree with this approach and have removed the dependency from this PR. I'll work on a follow-up PR implementing the virtual runtime provider mechanism as you suggested.

> Fails to build.

PR currently depends on the changes from PR #2738 due to some header path changes from camx, which has not been merged yet. In this case, what would be the preferred approach?
1.Wait for PR #2738 to be merged, then rebase this PR and push the updated changes.
2.Add the required commit(s) to PR #2738 instead.

lumag: Broken quoting. Please fix your message.

shukulk: Fixed it.

lumag: Yes, wait for the dependencies to be merged. Then rebase this PR on top of the merged tree.

Raja-Ganapathi-Busam:
This PR can be closed since the latest tag (1.0.5) already updated with https://github.com/qualcomm-linux/meta-qcom/pull/2980

PR #2865, opened 2026-08-04
qcom-raw-partitions-udev-rules: skip blkid probing for raw partitions

wenwfu:
Qualcomm platforms expose many bootloader, firmware, metadata, modem NV, and dump partitions as raw block devices. During coldplug, systemd's `60-persistent-storage.rules` still runs `blkid` on these partitions even though they are not filesystems and will never be mounted. This delays `sysinit.target`.                                                                                                                                                                                                                
                                                                                                                                                                                                                                   
Add an allarch `qcom-raw-partitions-udev-rules` recipe that generates a udev rule from the partition layouts provided by qcom-ptool. The generated rule:

  - skips `blkid` probing for known Qualcomm raw partitions;
  - matches partitions by GPT `PARTNAME`, rather than unstable kernel device names;
  - preserves `/dev/disk/by-partlabel`, `/dev/disk/by-partuuid`, and related partition metadata links;
  - leaves filesystem partitions such as `rootfs`, `efi`, `persist`, and `userdata` under the normal systemd handling.

The rule generator and its reviewed filesystem-partition policy are provided by qcom-ptool. Update the qcom-ptool revision in this layer and package the generated rule separately, so the rule stays synchronized with the supported  Qualcomm partition layouts.

The package is intentionally independent from the systemd recipe. Its selection is a distro/image-level decision.

On `rb3gen2-core-kit` with QLI-2.0 `wrynose-260627.1`, this reduces the time to reach `sysinit.target` by approximately 0.38 seconds:

  ```text
  Default:             5.74s
  With optimization:   5.37s
```

### DIFF
--- a/recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
+++ b/recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
@@ -1,0 +1,51 @@
+SUMMARY = "udev rules for Qualcomm raw partitions"
+DESCRIPTION = "Machine-specific udev rules that skip filesystem probing for reviewed Qualcomm raw GPT partitions"
+
+require qcom-ptool.inc
+
+DEPENDS = "qcom-ptool-native"
+
+QCOM_PARTITION_FILES_SUBDIR ??= ""
+QCOM_PARTITION_FILES_SUBDIR_SPINOR ??= ""
+
+QCOM_RAW_PARTITIONS_RULES = "${B}/55-qcom-raw-partitions-noblkid.rules"
+QCOM_RAW_PARTITION_LAYOUTS = " \
+    ${QCOM_PARTITION_FILES_SUBDIR} \
+    ${QCOM_PARTITION_FILES_SUBDIR_SPINOR} \
+"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+do_compile() {
+    rm -f ${QCOM_RAW_PARTITIONS_RULES}
+
+    set --
+    for layout in ${QCOM_RAW_PARTITION_LAYOUTS}; do
+        layout=${layout#partitions/}
+        set -- "$@" --input "${S}/platforms/$layout/partitions.conf"
+    done
+
+    if [ "$#" -gt 0 ]; then
+        if ${STAGING_BINDIR_NATIVE}/qcom-ptool \
+                gen_udev_rules --help >/dev/null 2>&1; then
+            ${STAGING_BINDIR_NATIVE}/qcom-ptool gen_udev_rules \
+                --output ${QCOM_RAW_PARTITIONS_RULES} \
+                "$@"
+        else
+            bbwarn "qcom-ptool ${SRCREV} has no gen_udev_rules support; skipping raw partition rules"
+        fi
+    fi
+}
+
+do_install() {
+    if [ -f ${QCOM_RAW_PARTITIONS_RULES} ]; then
+        install -Dm 0644 ${QCOM_RAW_PARTITIONS_RULES} \
+            ${D}${nonarch_libdir}/udev/rules.d/55-qcom-raw-partitions-noblkid.rules
+    fi
+}
+
+FILES:${PN} = " \
+    ${nonarch_libdir}/udev/rules.d/55-qcom-raw-partitions-noblkid.rules \
+"
+
+ALLOW_EMPTY:${PN} = "1"
--- a/recipes-core/systemd/systemd_%.bbappend
+++ b/recipes-core/systemd/systemd_%.bbappend
@@ -16,3 +16,6 @@ do_install:append:qcom() {
 }
 
 FILES:${PN}-udev-rules:append:qcom = " ${nonarch_libdir}/udev/rules.d/99-dma-heap.rules"
+
+# Install the machine-specific raw partition rules whenever udev is installed.
+RRECOMMENDS:udev:append:qcom = " qcom-raw-partitions-udev-rules"

### CODE REVIEW

File: recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules [hunk 1]
```diff
@@ -0,0 +1,45 @@
...
+ENV{PARTNAME}=="abl|abl_[ab]|aop|aop_[ab]|aop_config|aop_config_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="apdp|apdp_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="tz|tz_[ab]|tzapps|tzapps_[ab]|TZAPPS|TZAPPS_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="hyp|hyp_[ab]|devcfg|devcfg_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="keymaster|keymaster_[ab]|cmnlib|cmnlib_[ab]|cmnlib64|cmnlib64_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="uefi|uefi_[ab]|uefisecapp|uefisecapp_[ab]|imagefv|imagefv_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="catecontentfv|quantumfv|questdatafv|toolsfv", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="qupfw|qupfw_[ab]|storsec|storsec_[ab]|shrm|shrm_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="cpucp|cpucp_[ab]|featenabler|featenabler_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="modem|modem_[ab]|bluetooth|bluetooth_[ab]|dsp|dsp_[ab]", GOTO="qcom_raw_noblkid"
```
lumag: Drop, these should not be used.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
```diff
@@ -0,0 +1,45 @@
...
+ENV{PARTNAME}=="dtb|dtb_[ab]|dtbo|dtbo_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="vbmeta|vbmeta_[ab]|vbmeta_system|vbmeta_system_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="cdt|cdt_[ab]|ddr|ddr_[ab]|multiimgqti|multiimgqti_[ab]|multiimgoem|multiimgoem_[ab]", GOTO="qcom_raw_noblkid"
```
lumag: There definitely are no `cdt_a` or `ddr_b` partitions. Drop the superfluous partitions from the list.

lumag (changes requested):
You list too many partitions which never existed. Drop them.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
wenwfu: Thanks. I will drop the modem/bluetooth/dsp entries and remove the superfluous cdt_[ab] match.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
wenwfu: For ddr_b, I observed a ddr_b partition on rb3gen2-core-kit and iq-9075-evk. Do you still prefer dropping ddr_b from this whitelist?

**root@iq-9075-evk:/# ls /dev/disk/by-partlabel/**
ALIGN_TO_128K_1  TZAPPS  cdt  ddr_a  **_ddr_b_**  diag_log  dtb_a  dtb_b  efi  emac  gearvm_a  gearvm_b  gvm_log  keymaster_a  keymaster_b  multiimgqti_a  multiimgqti_b  persist  pvm_log  rootfs  softsku  splash  spunvm  xbl_a  xbl_b  xbl_config_a  xbl_config_b  xbl_logs

**root@rb3gen2-core-kit:/# ls /dev/disk/by-partlabel/**
ALIGN_TO_128K_1  aop_b   catecontentfv  cpucp_b  devcfg_a  dip    efi  hyp_a      imagefv_b    logdump   modemst2       qmcs        questdatafv  qweslicstore_a  rootfs   shrm_b   tz_a    uefi_b        vm-data  xbl_config_a
SYSFW_VERSION    apdp_a  cdt            **_ddr_a_**    devcfg_b  dtb_a  fsc  hyp_b      limits       logfs     multiimgoem_a  quantumfv   qupfw_a      qweslicstore_b  secdata  splash   tz_b    uefisecapp_a  xbl_a    xbl_ramdump_a    
aop_a            apdp_b  cpucp_a        ddr_b    devinfo   dtb_b  fsg  imagefv_a  limits-cdsp  modemst1  multiimgoem_b  quantumsdk  qupfw_b      recoveryinfo    shrm_a   toolsfv  uefi_a  uefisecapp_b  xbl_b    xbl_ramdump_b

lumag: Hmm. okay.

wenwfu:
Updated per review comments:                                                                                                                                                                                                     
  - Dropped modem/bluetooth/dsp.                                                                                                                                                                                                   
  - Removed cdt_[ab].                                                                                                                                                                                                              
  - Kept ddr_b explicitly since it is present on the tested rb3gen2-core-kit and                                                                                                                                                   
    iq-9075-evk partition layouts.                                                                                                                                                                                                 
                                                                                                                                                                                                                                   
  Please review again when you get a chance.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
```diff
@@ -0,0 +1,44 @@
...
+# such as rootfs, system, userdata, persist, or an ESP from the same storage
+# device. Keep those partitions under the stock persistent-storage rules.
+#
+# 60-persistent-storage.rules creates by-partlabel/by-partuuid links after the
+# blkid import. Since this rule skips that import for raw partitions, emit
+# the partition links here from kernel-provided PARTNAME/PARTUUID properties.
+ACTION=="remove", GOTO="qcom_raw_noblkid_end"
+SUBSYSTEM!="block", GOTO="qcom_raw_noblkid_end"
+ENV{DEVTYPE}!="partition", GOTO="qcom_raw_noblkid_end"
+ENV{PARTNAME}=="", GOTO="qcom_raw_noblkid_end"
```
lumag: I'd say, that for paritions with an empty name we should create all symlinks.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
```diff
@@ -0,0 +1,44 @@
...
+
+ENV{PARTNAME}=="xbl|xbl_[ab]|xbl_config|xbl_config_[ab]|xbl_ramdump|xbl_ramdump_[ab]", GOTO="qcom_raw_noblkid"
```
lumag: I really asked to review the list. Which platform has non-slot xbl_config? Likewise I know only one platform with non-slotted xbl, but it's easier to ignore it.
Please go through your list and then apply common sense

lumag (changes requested):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/83e30962-8163-423c-81a3-8dc7d411b19a)**
---
**Test jobs for commit [`476ad31`](https://github.com/qualcomm-linux/meta-qcom/commit/476ad313a24da65207d98bd08f06c8d084a4c8dc)**
---
Pass: 9 | Fail: 0 | Total: 9

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/5816e775-59e6-4c5f-b044-90a156996073)**
---
**Test jobs for commit [`73cfa60`](https://github.com/qualcomm-linux/meta-qcom/commit/73cfa60f5bee0b05b091c396208e79c97189abdf)**
---
Pass: 43 | Fail: 9 | Total: 52
Pass: 9 | Fail: 0 | Total: 9

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
wenwfu: **Fixed.**
Empty `PARTNAME` now falls through to a shared `qcom_raw_links` label instead of exiting early.、
  For partitions without a name, this still creates the non-name-dependent partition metadata links:                                                                                                                               

  - `disk/by-partuuid/...`、
  - `disk/by-path/.../by-partnum/...`、
  - `disk/by-path/.../by-partuuid/...`、

  The `by-partlabel` links remain guarded by `PARTNAME=="?*"`, so we do not create invalid empty-label symlinks.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
wenwfu: **Fixed. I went through the list again and trimmed it against the generated partition XML.**、
                                                                                                                                                                                                                                   
  I checked the generated `qcom-partition-conf` `rawprogram*.xml` from the deployed partition artifacts and removed entries that were not backed by those partition labels.                                                        
                                                                                                                                                                                                                                   
  Notably, I dropped the non-slotted `xbl`, `xbl_config`, and `xbl_ramdump` entries, and also removed other unsupported bare names such as `abl`, `aop`, `aop_config`, `uefi`, `imagefv`, `qupfw`, `shrm`, `cpucp`, `featenabler`, 
  `dtbo`, `vbmeta`, and `gearvm`.                                                                                                                                                                                                  
                                                                                                                                                                                                                                   
  I kept bare names that do exist in the generated partition tables, such as `apdp`, `tz`, `hyp`, `devcfg`, `keymaster`, `cmnlib`, `cmnlib64`, `ddr`, `multiimgqti`, `multiimgoem`, and `qweslicstore`.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
```diff
@@ -0,0 +1,49 @@
...
+ENV{PARTNAME}=="qupfw_[ab]|storsec|storsec_[ab]|shrm_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="cpucp_[ab]|featenabler_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="dtb|dtb_[ab]|dtbo_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="vbmeta_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="cdt|ddr|ddr_a|ddr_b|multiimgqti|multiimgqti_[ab]|multiimgoem|multiimgoem_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="gearvm_[ab]|spunvm|softsku|splash|emac", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="fsc|fsg|modemst[12]|qweslicstore|qweslicstore_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="devinfo|dip|limits|limits-cdsp|qmcs|quantumsdk|recoveryinfo|secdata|SYSFW_VERSION", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="diag_log|gvm_log|pvm_log|logdump|xbl_logs|ALIGN_TO_128K_*", GOTO="qcom_raw_noblkid"
+GOTO="qcom_raw_noblkid_end"
```
ricardosalveti, line 42: This is a long list, we will have to keep track of the updates based on the changes available via https://github.com/qualcomm-linux/qcom-ptool.

ricardosalveti: Can you add a comment saying how you got to these partitions here? Just so we know how to keep this file in sync with all latest targets.

wenwfu: Currently the list was derived from labels observed on the tested devices and from generated rawprogram*.xml artifacts. I agree that is not enough as a maintenance rule. 

Updated. I added a maintenance comment above the whitelist pointing to qcom-ptool platforms and explaining how to refresh the list for new targets or  layout changes.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/a2bb3e34-059b-4c68-8155-192561ea1939)**
---
**Test jobs for commit [`64dc0a3`](https://github.com/qualcomm-linux/meta-qcom/commit/64dc0a3e437ff57078dc3bf08004dc62f47cfbfd)**
---
Pass: 9 | Fail: 0 | Total: 9

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/65a115fe-cdd8-4172-a90c-ebebb9299081)**
---
**Test jobs for commit [`a4903bd`](https://github.com/qualcomm-linux/meta-qcom/commit/a4903bd404c1907958866794525ad59f7d180d13)**
---
Pass: 41 | Fail: 5 | Others: 6 | Total: 52
Pass: 9 | Fail: 0 | Total: 9

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
```diff
@@ -0,0 +1,58 @@
...
+# platforms/*/*/partitions.conf files for --partition --name entries,
+# or the rawprogram*.xml files generated from those configs. Add only
+# raw firmware, metadata, and log partitions that should skip blkid
+# probing. Do not add partitions that can contain mountable filesystems,
+# such as rootfs, system, userdata, persist, or EFI system partitions.
+ENV{PARTNAME}=="xbl_[ab]|xbl_config_[ab]|xbl_ramdump_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="abl_[ab]|aop_[ab]|aop_config_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="apdp|apdp_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="tz|tz_[ab]|TZAPPS", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="hyp|hyp_[ab]|devcfg|devcfg_[ab]", GOTO="qcom_raw_noblkid"
```
lumag: Ignore non-slotted tz/hyp/devcfg/keymaster/cmnlib/cmnlib64, they are used only for the very old platforms (which we probably should fix).

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
```diff
@@ -0,0 +1,58 @@
...
+ENV{PARTNAME}=="keymaster|keymaster_[ab]|cmnlib|cmnlib_[ab]|cmnlib64|cmnlib64_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="uefi_[ab]|uefisecapp|uefisecapp_[ab]|imagefv_[ab]", GOTO="qcom_raw_noblkid"
```
lumag: I don't see bare uefisecapp

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
```diff
@@ -0,0 +1,58 @@
...
+ENV{PARTNAME}=="catecontentfv|quantumfv|questdatafv|toolsfv", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="qupfw_[ab]|storsec|storsec_[ab]|shrm_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="cpucp_[ab]|featenabler_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="dtb|dtb_[ab]|dtbo_[ab]", GOTO="qcom_raw_noblkid"
```
lumag: I don't see dtb / dtb_a / dtb_b partitions at all

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
```diff
@@ -0,0 +1,58 @@
...
+ENV{PARTNAME}=="vbmeta_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="cdt|ddr|ddr_a|ddr_b|multiimgqti|multiimgqti_[ab]|multiimgoem|multiimgoem_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="gearvm_[ab]|spunvm|softsku|splash|emac", GOTO="qcom_raw_noblkid"
```
lumag: I don't see gearvm

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
```diff
@@ -0,0 +1,58 @@
...
+ENV{PARTNAME}=="fsc|fsg|modemst[12]|qweslicstore|qweslicstore_[ab]", GOTO="qcom_raw_noblkid"
+ENV{PARTNAME}=="devinfo|dip|limits|limits-cdsp|qmcs|quantumsdk|recoveryinfo|secdata|SYSFW_VERSION", GOTO="qcom_raw_noblkid"
```
lumag: limits-cdsp doesn't exist. SYSFW_VERSION doesn't exist

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
```diff
@@ -0,0 +1,58 @@
...
+ENV{PARTNAME}=="diag_log|gvm_log|pvm_log|logdump|xbl_logs|ALIGN_TO_128K_*", GOTO="qcom_raw_noblkid"
```
lumag: I don't see `diag_log`, `gvm_log`, `pvm_log`.

lumag (changes requested):

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
wenwfu: Agreed. In qcom-ptool these bare (non-slotted) names appear only on apq8016-sbc and apq8096-db820c — the legacy platforms. I'll change these to slot-only:

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
wenwfu: Correct. Bare uefisecapp exists only on iq-x7181-evk; every other target uses uefisecapp_[ab]. I'll drop the bare form and keep uefi_[ab]|uefisecapp_[ab]|imagefv_[ab].

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
wenwfu: dtb_[ab] is actually defined on 16 platforms in qcom-ptool, including the ones I tested (rb3gen2, iq-9075-evk), I also confirmed them live in /dev/disk/by-partlabel/ on both boards.  Bare dtb only exists on iq-x7181-evk, and dtbo only on qrb5165-rb5.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
wenwfu: gearvm_[ab] is defined on iq-8275-evk, iq-9075-evk, qcs8275-monza, qcs8300-ride-sx, qcs9100-ride-sx in qcom-ptool, and I see it live on iq-9075-evk.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
wenwfu: Both exist in qcom-ptool: limits-cdsp on qcm6490-idp, qcs6490-rb3gen2, qcs6490-thundercomm-rubikpi3, qrb5165-rb5;

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
wenwfu: All three are defined on iq-8275-evk, iq-9075-evk, qcs8275-monza, qcs8300-ride-sx, qcs9100-ride-sx

wenwfu:
@ricardosalveti @lumag  — on long-term maintenance (keeping this in sync with qcom-ptool):
I'd propose generating this file from qcom-ptool rather than hand-maintaining it. A small script scans platforms/*/*/partitions.conf for --name= entries across all supported targets, subtracts the known mountable labels (rootfs/system/userdata/persist/efi/vm-data), and emits this rule with sorted/de-duped, slot-collapsed (foo_[ab]) entries. The generated file would carry a header: # Generated from qcom-ptool @ <commit>; regenerate with <cmd>.

Two points I'd want your call on:

1. Generate offline and commit the result (script + generated .rules both in-tree), rather than generating at build time — this keeps the build free of a qcom-ptool dependency and lets you review the actual final file in the diff. Building it at do_compile would always be current but hides the final content from review and adds a cross-repo build dependency.
2. Optionally add a CI check that re-runs the generator and fails if the committed file is stale — so a qcom-ptool layout change that isn't regenerated is caught pre-merge. This stays fail-safe: a stale list only costs a little boot-time optimization, never a mount/boot failure, since it still only adds raw partitions to the skip set.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
lumag: Agree, sorry.

recipes-core/systemd/systemd/55-qcom-raw-partitions-noblkid.rules
lumag: Ack. Let's kill dtbo_[ab] here and a raw dtb.

lumag:
Sure, let's have a script. How do sort out the one-offs (like db410c / db820c / iq-x7181)?

wenwfu:
> Sure, let's have a script. How do sort out the one-offs (like db410c / db820c / iq-x7181)?

Added a generator. It excludes the legacy db410c/db820c layouts, while iq-x7181 remains scanned. Labels found on fewer than two platforms are treated as one-offs and require explicit per-pattern approval.

lumag:
If you have written a generator, what about using it to generate the rules? Extend the qcom-ptool recipe, add separate udev-rules package and generate the rules while building.

wenwfu:
> If you have written a generator, what about using it to generate the rules? Extend the qcom-ptool recipe, add separate udev-rules package and generate the rules while building.

Yes, that makes sense. I’ll move the generator into the qcom-partition-conf recipe, generate the rules from the recipe’s qcom-ptool source during the build, and install them through a separate udev-rules package.

ricardosalveti:
Could this generator be part of qcom-ptool upstream instead? The same benefit we get here could also later be applied for other distros, since this is just a set of generic qcom udev rules.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/ab9d78f4-ac65-4a97-a816-4775883e5439)**
---
**Test jobs for commit [`3f5434b`](https://github.com/qualcomm-linux/meta-qcom/commit/3f5434b896893b0b5650a4e6b68d6eb6b29f0ec1)**
---
Pass: 9 | Fail: 0 | Total: 9

wenwfu:
> Could this generator be part of qcom-ptool upstream instead? The same benefit we get here could also later be applied for other distros, since this is just a set of generic qcom udev rules.

Thanks for the suggestion. I rechecked both qcom-ptool and how it is consumed by meta-qcom.                                                                                                                                    

 qcom-ptool currently provides build-time partitioning utilities and platform descriptions. Its outputs are GPT binaries, partition XML, contents XML, and QDL flashing files; it does not currently provide runtime udev or systemd policy.                                                                                                                                                                                                                

This rule serves a different purpose. It is a systemd-udev runtime optimization that sets UDEV_DISABLE_PERSISTENT_STORAGE_BLKID_FLAG for a reviewed subset of raw partitions and preserves the partition symlinks that would  otherwise depend on the blkid import. Whether a partition is safe to exclude from probing cannot be inferred solely from its presence in partitions.conf, so the generator also contains a deliberately reviewed policy.       

meta-qcom owns the mapping from an OE MACHINE to the relevant qcom-ptool layouts through QCOM_PARTITION_FILES_SUBDIR and QCOM_PARTITION_FILES_SUBDIR_SPINOR. The new recipe uses that mapping and the same pinned qcom-ptool  revision to generate a machine-specific runtime package.                                                                                                                                                                       

Other distributions would still need their own layout selection, policy review, packaging, and rule installation even if the generator were moved to qcom-ptool. Therefore, I would prefer to keep the generator and systemd-specific template in meta-qcom for this implementation. A generalized upstream interface can be reconsidered later if another distribution wants to consume the same runtime policy.

wenwfu:
@lumag @ricardosalveti -- This PR has been reworked from a hand-maintained partition allowlist to a generated, per-machine approach, addressing the maintainability concern @ricardosalveti raised earlier and the accuracy issues @lumag found in the previous partition-name list (several entries didn't exist on the platforms I'd tested, others existed but weren't in the reviewed list).

Could you take another look when you have a chance?

**What changed:**

1. The udev rule template (55-qcom-raw-partitions-noblkid.rules.in) and the raw-partition match list are now generated at build time by generate_qcom_raw_partitions.py from each machine's qcom-ptool partition layout (QCOM_PARTITION_FILES_SUBDIR), instead of a single hand-written list shared across all platforms.
2. A new approved-raw-partition-patterns.list holds the policy: patterns reviewed as safe to exclude from blkid, filtered against qcom-ptool at generation time so a machine only gets the patterns for partitions it actually has. The generated section records the qcom-ptool commit it was built from.
3. Moved into a standalone recipe (qcom-raw-partitions-udev-rules_git.bb) recommended via RRECOMMENDS:udev on qcom machines, rather than living inside the systemd bbappend.
 
This means adding a new machine or a new qcom-ptool partition just requires updating the **_approved-patterns policy_** (if needed) and regenerating — no more per-machine list edits or arguing over whether partition X exists on platform Y.

recipes-bsp/partition/qcom-raw-partitions-udev-rules/approved-raw-partition-patterns.list
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+#
+# SPDX-License-Identifier: MIT
+
+# Partition-name patterns reviewed as safe to exclude from blkid probing.
+# Keep one udev glob per line. A new qcom-ptool label is not implicitly safe.
```
lumag: This is not better than the previous iteration. You still hardcode the full list.

lumag (changes requested):
I think I'd second the question of moving it to qcom-ptool. @lool, would you want something similar in Debian images?

lool:
I would have preferred to use the data directly rather than resorting to heuristics and generators. As you note, it's hard to exclude entire block devices because LUNs mix HLOS and NHLOS partitions.

The ideal next level would have been partition GUIDs, but currently these are not reported by the kernel. Could we send a patch to allow this? That sounds like a more useful general mechanism.

Lastly, I hate the very long-list of weird partition names. We know it will keep growing. Should simply flip this around? That is, just run blkid on a few partitions that we expect, and nothing else. It would mean:
- shorter, simpler rules
- no generator needed
- breaks if we forget to list a partition (this implementation silently misses it)

The list of partitions we want to mount from HLOS is quite small: efi, rootfs, persist, userdata (UNOQ).

ricardosalveti:
> Lastly, I hate the very long-list of weird partition names. We know it will keep growing. Should simply flip this around? That is, just run blkid on a few partitions that we expect, and nothing else. It would mean:

This could cause bad side effect on users customizing the partition table as well, which is a use case we also want to support, that is why an exclude list seems better in the end.

recipes-bsp/partition/qcom-raw-partitions-udev-rules/approved-raw-partition-patterns.list
wenwfu: You're right. The per-machine filtering avoids generating rules for nonexistent partitions, but it does not eliminate the hardcoded policy list.

Would moving the policy, generator, rule template, and tests to qcom-ptool address your concern, or should the raw/no-blkid classification be explicit metadata in each partition definition? I want to avoid simply relocating the same hardcoded list to another repository.

wenwfu:
The implementation has been reworked based on the previous review.

  The generic generator, reviewed policy, and udev template have been moved to qcom-ptool:                                                                                                                                                                                                        

  https://github.com/qualcomm-linux/qcom-ptool/pull/156

  This PR now only contains the Yocto machine integration and packaging. It capability-checks gen_udev_rules so existing builds remain functional until the qcom-ptool dependency is available.                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                        
  Please take another look. This PR must not be merged until qcom-ptool #156 is merged and SRCREV is updated to its final upstream commit.

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
```diff
@@ -0,0 +1,55 @@
...
+
+    ${STAGING_BINDIR_NATIVE}/qcom-ptool gen_udev_rules \
+        --output ${QCOM_RAW_PARTITIONS_RULES} \
+        "$@"
+}
+
+do_install() {
+    if [ ! -f ${QCOM_RAW_PARTITIONS_RULES} ]; then
+        exit 0
+    fi
```
quaresmajose, line 44: please drop this check, bitbake internal takes care of this

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
```diff
@@ -0,0 +1,55 @@
...
+do_compile() {
+    rm -f ${QCOM_RAW_PARTITIONS_RULES}
+
+    set --
+    for layout in ${QCOM_RAW_PARTITION_LAYOUTS}; do
+        layout=${layout#partitions/}
+        set -- "$@" --input "${S}/platforms/$layout/partitions.conf"
+    done
+    if [ "$#" -eq 0 ]; then
+        exit 0
```
quaresmajose: please don't use exit in bitbake task, this brakes any do_compile:append that may exist

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
```diff
@@ -0,0 +1,55 @@
...
+    fi
+
+    if ! ${STAGING_BINDIR_NATIVE}/qcom-ptool \
+            gen_udev_rules --help >/dev/null 2>&1; then
+        bbwarn "qcom-ptool ${SRCREV} has no gen_udev_rules support; skipping raw partition rules"
+        exit 0
```
quaresmajose: same

quaresmajose (changes requested):

lool:
Generator feels overkill to me; why not a single list with all names? We have to support that list for generic images that work on all targets anyway

Anyone on the kernel changes to send part GUID as part of uevent?

lumag:
> Generator feels overkill to me; why not a single list with all names? We have to support that list for generic images that work on all targets anyway

Earlier revisions had a list of all partitions. It was a nightmare to review and/or change later.

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
wenwfu: fixed

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
wenwfu: fixed

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
wenwfu: I kept the file check intentionally, but removed the early exit.
                                                                                                                                                                                                                        
  `gen_udev_rules` may legitimately produce no output when:                                                                                                                                                             
                                                                                                                                                                                                                        
  - the machine has no selected qcom-ptool layout;
  - no approved raw partitions exist in the selected layout;
  - the transitional qcom-ptool revision lacks `gen_udev_rules`.
                                                                                                                                                                                                                        
  `ALLOW_EMPTY` allows BitBake to create an empty package, but it does not prevent                                                                                                                                      
  `install` from failing when its source file does not exist.                                                                                                                                                           
                                                                                                                                                                                                                        
  With the conditional install, `do_install:append` still runs normally. Dropping
  the check would require qcom-ptool to always emit an empty rules file and would
  remove the current optional/compatibility behaviour. Please let me know if that                                                                                                                                       
  is the preferred direction.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/3c4f3f8a-7e0b-486e-8b46-3fb2dc0234a0)**
---
**Test jobs for commit [`bffc4ce`](https://github.com/qualcomm-linux/meta-qcom/commit/bffc4ce09854126bcb2e01f87055031d32b52224)**
---
Pass: 9 | Fail: 0 | Total: 9

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/39ba0e1f-378f-4d6e-ae1c-8307fafa2b53)**
---
**Test jobs for commit [`4b9afcf`](https://github.com/qualcomm-linux/meta-qcom/commit/4b9afcf61be398eef3b15a0ad22c62642e9aa5eb)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 51 | Fail: 0 | Total: 51

wenwfu:
The related qcom-ptool PR has been updated based on the review feedback:
https://github.com/qualcomm-linux/qcom-ptool/pull/156

The meta-qcom change is currently waiting for the qcom-ptool code-owner review. Please take another look when you have time, as this PR has been marked stale and is scheduled for automatic closure.

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
```diff
@@ -0,0 +1,46 @@
+SUMMARY = "udev rules for Qualcomm raw partitions"
+DESCRIPTION = "Machine-specific udev rules that skip filesystem probing for reviewed Qualcomm raw GPT partitions"
```
lumag: They are no longer machine-specific. So, please rework the recipe.

lumag (changes requested):

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
wenwfu: fixed

recipes-core/systemd/systemd_%.bbappend
```diff
@@ -15,3 +15,6 @@ do_install:append:qcom() {
 }
 
 FILES:${PN}-udev-rules:append:qcom = " ${nonarch_libdir}/udev/rules.d/99-dma-heap.rules"
+
+# Install the common Qualcomm raw partition rules whenever udev is installed.
+RRECOMMENDS:udev:append:qcom = " qcom-raw-partitions-udev-rules"
```
quaresmajose, line 20: It will be better to have just this change in a separete commit.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/f3b570ea-3ca7-440b-83de-b48170780d7f)**
---
**Test jobs for commit [`31db906`](https://github.com/qualcomm-linux/meta-qcom/commit/31db9064fcd98fd75a49eca7a8a02abe44e78604)**
---
Pass: 9 | Fail: 0 | Total: 9

recipes-core/systemd/systemd_%.bbappend
wenwfu: done

wenwfu:
The change has been split into three logical commits: the qcom-ptool update, the generated udev-rules recipe, and the systemd integration. The PR is ready for another review.

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
quaresmajose: The way you have it makes sense, then. Thanks for the explanation.

But if I understand correctly, we'll get an error when the file doesn't exist because ALLOW_EMPTY
 is not defined.

wenwfu: Fixed in the latest update by adding ALLOW_EMPTY:${PN} = "1" next to FILES:${PN}.

quaresmajose (dismissed):

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
quaresmajose: Thanks.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/20aad16c-4540-43c1-826b-54144fa5bdc5)**
---
**Test jobs for commit [`b79f372`](https://github.com/qualcomm-linux/meta-qcom/commit/b79f372aaf6f0f15a49832366c78a2e77c1b1796)**
---
Pass: 9 | Fail: 0 | Total: 9

ricardosalveti:
Please rebase.

wenwfu:
> Please rebase.

done

quaresmajose (dismissed):

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
```diff
@@ -0,0 +1,29 @@
...
+        install -Dm 0644 ${QCOM_RAW_PARTITIONS_RULES} \
+            ${D}${nonarch_libdir}/udev/rules.d/55-qcom-raw-partitions-noblkid.rules
+    fi
+}
+
+FILES:${PN} = " \
+    ${nonarch_libdir}/udev/rules.d/55-qcom-raw-partitions-noblkid.rules \
+"
+
+ALLOW_EMPTY:${PN} = "1"
```
ricardosalveti, line 29: Do we expect this package to be empty?

wenwfu: Yes, for layouts without any approved raw partitions. gen_udev_rules intentionally produces no file in that case, so the conditional install and ALLOW_EMPTY:${PN} are required. Supported machines with raw partitions will still get a non-empty package.

ricardosalveti (dismissed):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/5e767e31-8177-4f19-9202-1cea46584808)**
---
**Test jobs for commit [`2389825`](https://github.com/qualcomm-linux/meta-qcom/commit/23898256ae9a430839632d728443a7951cd83daf)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 40 | Fail: 0 | Total: 40

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
lumag: The recipe is now machine-independent. Please drop ALLOW_EMPTY.

recipes-core/systemd/systemd_%.bbappend
lumag: No, it should be a distro-level decision. Please mention the package implicitly in the qcom images.

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
wenwfu: done

wenwfu:
Addressed both comments:

  - Dropped `ALLOW_EMPTY:${PN}` from the raw-partitions recipe.
  - Removed the `udev` recommendation from the systemd bbappend. The package selection belongs at the distro/image level and added to `meta-qcom-distro`.

recipes-core/systemd/systemd_%.bbappend
wenwfu: fixed

lumag (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/ffee71c0-54b9-4317-bca6-e0193d5faaeb)**
---
**Test jobs for commit [`16d4f41`](https://github.com/qualcomm-linux/meta-qcom/commit/16d4f415f8fd8c8c9b3276df427194676314d4b0)**
---
Pass: 34 | Fail: 1 | Others: 5 | Total: 40
Pass: 9 | Fail: 0 | Total: 9

koenkooi:
"qcom-raw-partitions-udev-rules: generate raw partition rules
Add a standalone allarch package that invokes qcom-ptool at build
time and installs the generated rule file. This keeps the generated
udev rules separate from the systemd recipe implementation.

Assisted-by: Codex:GPT-5
Signed-off-by: Wenwen Fu <wenwfu@qti.qualcomm.com>"

That completely lacks any information on what this is supposed to accomplish and what problem it's trying to fix. Also: it's recipe, not  package.

koenkooi (changes requested):
See previous comment

wenwfu:
> "qcom-raw-partitions-udev-rules: generate raw partition rules Add a standalone allarch package that invokes qcom-ptool at build time and installs the generated rule file. This keeps the generated udev rules separate from the systemd recipe implementation.
> 
> Assisted-by: Codex:GPT-5 Signed-off-by: Wenwen Fu [wenwfu@qti.qualcomm.com](mailto:wenwfu@qti.qualcomm.com)"
> 
> That completely lacks any information on what this is supposed to accomplish and what problem it's trying to fix. Also: it's recipe, not package.

fixed

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
quaresmajose: So, what will happen when the package is empty?

lumag (changes requested):
PR title and description also need to be updated.

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
wenwfu: > So, what will happen when the package is empty?

 With the qcom-ptool revision currently pinned by this recipe, gen_udev_rules() always renders the complete udev rule template when partition layouts are found. If no raw partitions match, the generated file still contains  the common rule handling, so the package is non-empty. If no partition layouts are found, qcom-ptool returns an error and do_compile() fails instead of producing an empty package. Therefore, ALLOW_EMPTY is not required.

wenwfu:
> PR title and description also need to be updated.

Updated the PR title and description.  Please take another look.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/95be4434-ea1f-4029-8906-0850dd1a9429)**
---
**Test jobs for commit [`fc2d91b`](https://github.com/qualcomm-linux/meta-qcom/commit/fc2d91b7704f19192cdad90a6655466d29bfbb78)**
---
Pass: 8 | Fail: 0 | Others: 2 | Total: 10
Pass: 35 | Fail: 6 | Total: 41

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
quaresmajose: Got it. Thanks

recipes-bsp/partition/qcom-raw-partitions-udev-rules_git.bb
```diff
@@ -0,0 +1,27 @@
...
+QCOM_RAW_PARTITIONS_RULES = "${B}/55-qcom-raw-partitions-noblkid.rules"
+
+do_compile() {
+    cd ${S}
+    ${STAGING_BINDIR_NATIVE}/qcom-ptool gen_udev_rules \
+        --output ${QCOM_RAW_PARTITIONS_RULES}
+}
+
+do_install() {
+    if [ -f ${QCOM_RAW_PARTITIONS_RULES} ]; then
```
quaresmajose: Since the file will always be there you could drop that if now.

wenwfu: Fixed. Removed the redundant file-existence check from `do_install()`.

lumag (approved):

quaresmajose (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/5bcd540d-f88b-4288-ac36-a5d1c47575ca)**
---
**Test jobs for commit [`17700d0`](https://github.com/qualcomm-linux/meta-qcom/commit/17700d0d78fa4335fab5cccd57abe9680872a217)**
---
Pass: 7 | Fail: 0 | Others: 4 | Total: 11
Pass: 23 | Fail: 0 | Total: 23

ricardosalveti (approved):

ricardosalveti:
@koenkooi mind another look?

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/4447a68a-f083-4603-a6ea-901ce6fa8d5a)**
---
**Test jobs for commit [`17700d0`](https://github.com/qualcomm-linux/meta-qcom/commit/17700d0d78fa4335fab5cccd57abe9680872a217)**
---
Pass: 7 | Fail: 0 | Others: 4 | Total: 11
Pass: 34 | Fail: 0 | Others: 3 | Total: 37

koenkooi (approved):

ricardosalveti:
Known test failures.

PR #2874, opened 2026-08-04
linux-qcom-dtbbin: run qcom_dtbbin_deploy before qcom_img_deploy

quaresmajose:
There is race condiction when do_qcom_img_deploy is executed and sometimes the kernel image is not available when the task is executed. If we run qcom_dtbbin_deploy before qcom_img_deploy, we guarantee we'll fix this problem.

Fix https://github.com/qualcomm-linux/meta-qcom/issues/2871

PR #2884, opened 2026-08-05
classes: make kernel deploy-time tasks safe in partially cached builds

### LINKED ISSUE #2871: classes/linux-qcom-bootimg.bbclass: do_qcom_img_deploy race condition
```
  ERROR: linux-yocto-6.18.35+git-r0 do_qcom_img_deploy: Error executing a python function in exec_func_python() autogenerated:
  
  The stack trace of python calls that resulted in this exception/failure was:
  File: 'exec_func_python() autogenerated', lineno: 2, function: <module>
       0001:
   *** 0002:do_qcom_img_deploy(d)
       0003:
  File: '/work/build/../meta-qcom/classes/linux-qcom-bootimg.bbclass', lineno: 117, function: do_qcom_img_deploy
       0113:        consoles = ' '.join(map(lambda c: "console=%(tty)s,%(rate)sn8" % dict(zip(("rate", "tty"), c.split(';'))), getVarDTB("SERIAL_CONSOLES").split()))
       0114:
       0115:        # prepare kernel image with appended dtb
       0116:        with open(kernel, 'wb') as wfd:
   *** 0117:            with open(os.path.join(kernel_output_dir, kernel_name), 'rb') as rfd:
       0118:                shutil.copyfileobj(rfd, wfd)
       0119:            with open(os.path.join(D, kernel_dtbdest, dtb), 'rb') as rfd:
       0120:                shutil.copyfileobj(rfd, wfd)
       0121:
  Exception: FileNotFoundError: [Errno 2] No such file or directory: 'arch/arm64/boot/Image.gz'
  
  ERROR: Logfile of failure stored in: /work/build/tmp/work/qcom_armv8a-qcom-linux/linux-yocto/6.18.35+git/temp/log.do_qcom_img_deploy.328295
  ERROR: Task (/work/build/../oe-core/meta/recipes-kernel/linux/linux-yocto_6.18.bb:do_qcom_img_deploy) failed with exit code '1'

```
https://github.com/qualcomm-linux/meta-qcom-distro/actions/runs/30829768008/job/91937360125?pr=409

discovered in https://github.com/qualcomm-linux/meta-qcom-distro/pull/409

ricardosalveti:
do_qcom_dtbbin_deploy, do_qcom_img_deploy (#2871) and
do_generate_qcom_fitimage read the kernel ${D}/${B} work directories, which
are empty when the kernel tasks were restored from sstate. If the custom task
itself misses the cache, it fails deterministically — this is what has been
failing the same 12 wrynose nightly compile jobs since the monthly sstate
cache rotation on Aug 1.

Move the three tasks after do_deploy and source their inputs from
DEPLOY_DIR_IMAGE, which is populated whether do_deploy executed or was
restored from sstate — a single code path for every build scenario. The
qcomflash/capsule consumers now depend on these tasks explicitly, since they
leave the do_deploy chain. Supersedes the f18d7eff workaround, fixes #2871.

### DIFF
--- a/classes/linux-qcom-dtbbin.bbclass
+++ b/classes/linux-qcom-dtbbin.bbclass
@@ -12,6 +12,14 @@ DTBBIN_SIZE ?= "4096"
 do_qcom_dtbbin_deploy[depends] += "dosfstools-native:do_populate_sysroot mtools-native:do_populate_sysroot"
 do_qcom_dtbbin_deploy[cleandirs] = "${DTBBIN_DEPLOYDIR}"
 do_qcom_dtbbin_deploy() {
+    # Source the DTBs from what do_deploy published: unlike ${D}, which
+    # only exists when do_install ran in this build, DEPLOY_DIR_IMAGE is
+    # also populated when do_deploy is restored from sstate.
+    deployDir="${DEPLOY_DIR_IMAGE}"
+    if [ -n "${KERNEL_DEPLOYSUBDIR}" ]; then
+        deployDir="${DEPLOY_DIR_IMAGE}/${KERNEL_DEPLOYSUBDIR}"
+    fi
+
     for dtbf in ${KERNEL_DEVICETREE}; do
         bbdebug 1 " combining: $dtbf"
         dtb=`normalize_dtb "$dtbf"`
@@ -20,7 +28,7 @@ do_qcom_dtbbin_deploy() {
         [ "$dtb_ext" = "dtbo" ] && continue
         dtb_base_name=`basename $dtb .$dtb_ext`
         mkdir -p ${DTBBIN_DEPLOYDIR}/$dtb_base_name
-        cp ${D}/${KERNEL_DTBDEST}/$dtb_base_name.dtb ${DTBBIN_DEPLOYDIR}/$dtb_base_name/combined-dtb.dtb
+        cp $deployDir/$dtb_base_name.dtb ${DTBBIN_DEPLOYDIR}/$dtb_base_name/combined-dtb.dtb
         mkfs.vfat -S ${QCOM_VFAT_SECTOR_SIZE} -C ${DTBBIN_DEPLOYDIR}/dtb-${dtb_base_name}-image.vfat ${DTBBIN_SIZE}
         mcopy -i "${DTBBIN_DEPLOYDIR}/dtb-${dtb_base_name}-image.vfat" -vsmpQ ${DTBBIN_DEPLOYDIR}/$dtb_base_name/* ::/
         rm -rf ${DTBBIN_DEPLOYDIR}/$dtb_base_name
@@ -32,7 +40,7 @@ do_qcom_dtbbin_deploy() {
         mcopy -i "${DTBBIN_DEPLOYDIR}/dtb-multi-dtb-image.vfat" -vsmpQ ${DEPLOY_DIR_IMAGE}/qclinuxfitImage ::/qclinux_fit.img
     fi
 }
-addtask qcom_dtbbin_deploy after do_populate_sysroot do_packagedata before do_deploy
+addtask qcom_dtbbin_deploy after do_deploy before do_build
 
 # Setup sstate, see deploy.bbclass
 SSTATETASKS += "do_qcom_dtbbin_deploy"
--- a/classes/linux-qcom-dtbbin.bbclass
+++ b/classes/linux-qcom-dtbbin.bbclass
@@ -32,7 +32,7 @@ do_qcom_dtbbin_deploy() {
         mcopy -i "${DTBBIN_DEPLOYDIR}/dtb-multi-dtb-image.vfat" -vsmpQ ${DEPLOY_DIR_IMAGE}/qclinuxfitImage ::/qclinux_fit.img
     fi
 }
-addtask qcom_dtbbin_deploy after do_populate_sysroot do_packagedata before qcom_img_deploy do_deploy
+addtask qcom_dtbbin_deploy after do_populate_sysroot do_packagedata before do_deploy
 
 # Setup sstate, see deploy.bbclass
 SSTATETASKS += "do_qcom_dtbbin_deploy"
--- a/conf/machine/qcom-armv7a.conf
+++ b/conf/machine/qcom-armv7a.conf
@@ -24,6 +24,10 @@ KERNEL_IMAGETYPE ?= "zImage"
 # To build an Android boot image
 KERNEL_CLASSES += "linux-qcom-bootimg"
 
+# Make every image build provide the Android boot images; the initramfs
+# image bundled into them is skipped to avoid a circular dependency.
+EXTRA_IMAGEDEPENDS += "${@'virtual/kernel:do_qcom_img_deploy' if d.getVar('PN') != d.getVar('INITRAMFS_IMAGE') else ''}"
+
 KERNEL_DEVICETREE ?= " \
     qcom/qcom-apq8064-asus-nexus7-flo.dtb \
     qcom/qcom-apq8064-ifc6410.dtb \
--- a/classes-recipe/dtb-fit-image.bbclass
+++ b/classes-recipe/dtb-fit-image.bbclass
@@ -14,18 +14,14 @@ require conf/image-fitimage.conf
 LINUX_QCOM_FIT_DTB_COMPATIBLE ?= "conf/machine/include/fit-dtb-compatible.inc"
 require ${LINUX_QCOM_FIT_DTB_COMPATIBLE}
 
-DEPENDS += "\
-    u-boot-tools-native \
-"
-
 MKIMAGE ?= "${STAGING_BINDIR_NATIVE}/mkimage"
 
 QCOMFIT_DEPLOYDIR = "${WORKDIR}/qcom_fitimage_deploy-${PN}"
 
-do_generate_qcom_fitimage[depends] += "qcom-dtb-metadata:do_deploy"
+do_generate_qcom_fitimage[depends] += "qcom-dtb-metadata:do_deploy u-boot-tools-native:do_populate_sysroot"
 do_generate_qcom_fitimage[cleandirs] += "${QCOMFIT_DEPLOYDIR}"
 python do_generate_qcom_fitimage() {
-    import os, shutil
+    import os
     from qcom.dtb_only_fitimage import QcomItsNodeRoot
 
     fit_dir = d.getVar('QCOMFIT_DEPLOYDIR')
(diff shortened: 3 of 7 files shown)


ricardosalveti:
@quaresmajose I'm still validating this as I don't have a warm cache locally, but I believe the issue we found in wrynose is just a cache issue in the end.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/cb69ca53-08fd-492f-81ee-7f539c722742)**
---
**Test jobs for commit [`69354f8`](https://github.com/qualcomm-linux/meta-qcom/commit/69354f8caabbaeda6a0b781a530850f9326cb79c)**
---
Pass: 9 | Fail: 0 | Total: 9

quaresmajose:
> Jose Quaresma (@quaresmajose) I'm still validating this as I don't have a warm cache locally, but I believe the issue we found in wrynose is just a cache issue in the end.

Yes, that makes sense. To avoid situations like this, I think we should have a weekly job without using sstate-cache

classes/linux-qcom-dtbbin.bbclass
```diff
@@ -32,7 +40,7 @@ do_qcom_dtbbin_deploy() {
         mcopy -i "${DTBBIN_DEPLOYDIR}/dtb-multi-dtb-image.vfat" -vsmpQ ${DEPLOY_DIR_IMAGE}/qclinuxfitImage ::/qclinux_fit.img
     fi
 }
-addtask qcom_dtbbin_deploy after do_populate_sysroot do_packagedata before qcom_img_deploy do_deploy
+addtask qcom_dtbbin_deploy after do_deploy before do_build
```
quaresmajose, line 43: You should revert the https://github.com/qualcomm-linux/meta-qcom/pull/2874 here before making the change, otherwise the wrynose backport will fail.

ricardosalveti: Done.

ricardosalveti:
Validating wrynose as part of https://github.com/qualcomm-linux/meta-qcom/pull/2895.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/14cdb584-ccdd-4c4d-9201-5db600cd62db)**
---
**Test jobs for commit [`bedbae8`](https://github.com/qualcomm-linux/meta-qcom/commit/bedbae8f532861e99b843e56b88f650a024b5acb)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 31 | Fail: 0 | Others: 3 | Total: 34

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/96d69ddb-2df5-4ff4-bdf4-db8569f9ee73)**
---
**Test jobs for commit [`71a3f53`](https://github.com/qualcomm-linux/meta-qcom/commit/71a3f5356179390d1520288c3b8643ccc19154f0)**
---
Pass: 9 | Fail: 0 | Total: 9

conf/machine/qcom-armv7a.conf
```diff
@@ -24,6 +24,10 @@ KERNEL_IMAGETYPE ?= "zImage"
 # To build an Android boot image
 KERNEL_CLASSES += "linux-qcom-bootimg"
 
+# Make every image build provide the Android boot images; the initramfs
+# image bundled into them is skipped to avoid a circular dependency.
+EXTRA_IMAGEDEPENDS += "${@'virtual/kernel:do_qcom_img_deploy' if d.getVar('PN') != d.getVar('INITRAMFS_IMAGE') else ''}"
```
lumag, line 29: Hmm, why?

conf/machine/include/qcom-base.inc
```diff
@@ -11,6 +11,9 @@ KERNEL_IMAGETYPES ?= "Image.gz"
 # For dtb.bin generation
 KERNEL_CLASSES += "linux-qcom-dtbbin"
 
+# Make every image build provide the dtb.bin vfat images
+EXTRA_IMAGEDEPENDS += "virtual/kernel:do_qcom_dtbbin_deploy"
```
lumag: Why? The qcomflash images should depend on them, nothing else.

lumag (changes requested):

conf/machine/include/qcom-base.inc
ricardosalveti: We don't have a CI use for dtb bins outside qcomflash but I don't see them necessarily as just a qcomflash dependency, as you could consume all the artifacts without building with qcomflash.

Not strictly required, fine to drop, but it makes sense to have the dependency outside qcomflash from my perspective.

conf/machine/qcom-armv7a.conf
ricardosalveti: armv7a/armv8a don't use qcomflash, and by moving do_qcom_img_deploy out of the kernel do_deploy chain (required in order to fix the partial sstate state), we need to create a dependency in order to build them.

conf/machine/include/qcom-base.inc
lumag: The kernel will be built completely, so by the end of the build the dtb.bin should be there. But adding extra dependencies would force dtb.bin to be built before all the images (including all the extra initramfs images).

conf/machine/qcom-armv7a.conf
lumag: Hmm, Can we do it slightly differently then? Turn `do_qcom_img_deploy` into `do_deploy:append`. WDYT?

conf/machine/include/qcom-base.inc
ricardosalveti: It won't in the partial sstate case, since all it will look for is the deploy / sysroot state. dtb.bin was built as part of the kernel because it was ordered before do_deploy, moving it after do_deploy (needed due the possible partial sstate cache state) caused it to not be built so we now require a proper task dependency (which can be qcomflash, but it makes sense to also not depend on it).

conf/machine/qcom-armv7a.conf
ricardosalveti: It could potentially be done, but there is an additional cost which I don't think that would be worth it. When initramfs is bundled it would add a do_image_complete dependency, and the deploy signature would also have to take into account the variables that could only be configured in qcom_img, such as kernel command line, bootimg rootfs and serial console (a single qcom_img change would also force a kernel deploy re-run).

Since this is at the image level, extending EXTRA_IMAGEDEPENDS seems simpler.

ricardosalveti:
This series should fix the current broken wrynose CI state once backported (which is blocking all our approved wrynose prs).

conf/machine/qcom-armv7a.conf
lumag: I think that's fine. We don't change rootfs or serial console frequently.

conf/machine/include/qcom-base.inc
lumag: Sorry, it will be be built completely at the end of the full BB build. However I don't think anything during the build depends on dtb.bin. It's not required for running do_rootfs of the core-image.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/e8ace7e1-1972-4b39-a5db-98823e615a64)**
---
**Test jobs for commit [`64ad9a3`](https://github.com/qualcomm-linux/meta-qcom/commit/64ad9a3d39d3f18a02045fbea6e5db88ac9ce21e)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 50 | Fail: 1 | Others: 2 | Total: 53

conf/machine/qcom-armv7a.conf
ricardosalveti: My main concern here is the image-completion dependency it also implies.

As qcom_img builds the boot-initramfs variants, the kernel deploy would depend on the initramfs do_image_complete task, which would expand the dependency chain for the kernel itself for an image that can be built independently from it (as a post deploy, as done by this series).

lumag: I'm not sure that I follow you.

conf/machine/include/qcom-base.inc
ricardosalveti: Nothing really depends or uses it besides qcomflash, I removed the dependency to make it linked to qcomflash only.

conf/machine/qcom-armv7a.conf
ricardosalveti: With the current changes do_qcom_img_deploy is executed after the kernel do_deploy, and when qcom_img also builds an initrd variant, the initrd do_image_complete requirement is carried by do_qcom_img_deploy itself as an explicit task dependency, letting the the kernel do_deploy to not depend on the initrd image. A change to the initrd or to the boot image arguments (as we discussed above) re-runs only do_qcom_img_deploy, nothing else of the kernel.

If we instead fold do_qcom_img_deploy into the kernel do_deploy (as a postfunc, since it is a mix of python and shell), both of those move onto do_deploy itself, and every consumer of do_deploy (images do_build, qcomflash, the capsule and the dtb tasks) would also have to wait for the initrd do_image_complete even though nothing in their dependency chain needs it.

The end result is the same, but the task dependencies would be decoupled if we move it outside the kernel do_deploy.

lumag: Thanks for the explanation.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/b8ef0e08-98d1-435a-9605-3549d0978892)**
---
**Test jobs for commit [`6f9c512`](https://github.com/qualcomm-linux/meta-qcom/commit/6f9c512ec477956eef2f6d31bc2126b3c0a19041)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 51 | Fail: 0 | Total: 51

quaresmajose (approved):

quaresmajose:
We need to apply https://github.com/qualcomm-linux/meta-qcom/pull/2877 first so that the backport of this can then work automatically.

conf/machine/qcom-armv7a.conf
ricardosalveti: Are you ok keeping the current logic from this PR in place? I wanted to land this to unblock wrynose, which is currently blocked by the sstate cache issue.

anujm1 (approved):

vkraleti (approved):

lumag (approved):

PR #2906, opened 2026-08-07
qcom-common-binary: Add include with dependencies of binaries

quaresmajose:
Here are defined the versions necessary for the correct functioning
of the binaries available in the layer.

Normally, the versions made available upstream are used. However, when
these versions are updated upstream, we need to temporarily copy the old
versions until new binaries are available.


ricardosalveti:
Add a better commit message for your second patch, just 'drop protobuf-camx' is not aligned with our commit message guidelines.

conf/machine/include/qcom-common-binary.inc
```diff
@@ -0,0 +1,17 @@
+# This file specifies the necessary versions for the binary recipes.
+
+VERSION_abseil-cpp = ""
```
ricardosalveti: We should use a soft assignment here (?=).

conf/machine/include/qcom-common-binary.inc
```diff
@@ -0,0 +1,17 @@
...
+PREFERRED_VERSION_abseil-cpp = "${VERSION_abseil-cpp}"
+PREFERRED_VERSION_abseil-cpp-native = "${VERSION_abseil-cpp}"
+PREFERRED_VERSION_nativesdk-abseil-cpp = "${VERSION_abseil-cpp}"
+
+VERSION_protobuf = ""
```
ricardosalveti: Same, and it might be good to do the same for all preferred version settings.

conf/machine/include/qcom-common-binary.inc
ricardosalveti: And it might be good to name it specifically for this layer, like QCOM_ABSEIL_CPP_PREFERRED_VERSION or similar.

quaresmajose: I thought about that when I was testing, but since specifying the version can only be defined when downgrading, I thought it wasn't necessary.

However, thinking about it more now, it might make sense when the downgrade is specified. In this case, to produce a new binary, the upstream version will have to be used again, and then whoever produces the binary needs to restore the `VERSION_abseil-cpp = "0"` externally again.

conf/machine/include/qcom-common-binary.inc
quaresmajose: I think for all preferred version settings. it doesn't make sense, we'll always use the version in VERSION\__recipe_

conf/machine/include/qcom-common-binary.inc
ricardosalveti: Right, we need to allow users to change the preferred version via local.conf in order to support camx rebuilds with the updated dependencies, for example.

quaresmajose: > And it might be good to name it specifically for this layer, like QCOM_ABSEIL_CPP_PREFERRED_VERSION or similar.

I belive this is not needed, the VERSION\__recipe_ variable concept dont exist in bitbake/oe-core.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/4bd2600d-90f5-4237-8fb5-633d56537dd9)**
---
**Test jobs for commit [`8b7a97a`](https://github.com/qualcomm-linux/meta-qcom/commit/8b7a97a64902ed2833d84ddd5ed52785621f9db1)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 47 | Fail: 2 | Total: 49

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/4dbc5040-a7bf-4b07-b1ef-7342877d29fe)**
---
**Test jobs for commit [`1e3119f`](https://github.com/qualcomm-linux/meta-qcom/commit/1e3119f45f3585f8693097e0dccf2e0f111294c9)**
---
Pass: 51 | Fail: 1 | Total: 52
Pass: 9 | Fail: 0 | Total: 9

lumag (approved):

ricardosalveti (approved):

PR #2908, opened 2026-08-07
firmware-qcom-boot: package and ship the boot firmware build metadata

ricardosalveti:
Qualcomm boot firmware releases embed their build information (meta build ID,
per-image build IDs such as TZ, BOOT, AOP) as the zip archive comment, which
is discarded today, leaving images with no record of the boot firmware
release they ship alongside.

Extract it (with a generated recipe/version/archive header, header-only when
no comment is available) into
`/usr/share/qcom-boot-firmware/<board>/build-info.txt` in the recipes'
previously-empty main package. Board machines install their own board's
package via `QCOM_BOOT_FIRMWARE`. The file is also deployed next to
the boot binaries so the qcomflash bundle self-describes.

Can be used as a replacement for /usr/lib/firmware/qcom/qcs9100/Ver_Info.txt
that is available in QLI 1.x (produced by the meta build, which we don't build anymore
in QLI 2.x).

### DIFF
--- a/recipes-bsp/firmware-boot/firmware-qcom-boot-common.inc
+++ b/recipes-bsp/firmware-boot/firmware-qcom-boot-common.inc
@@ -4,12 +4,62 @@ S = "${UNPACKDIR}/${BOOTBINARIES}"
 
 QCOM_BOOT_IMG_SUBDIR ?= ""
 
+# Rootfs directory holding the per-board boot firmware build metadata
+QCOM_BOOT_INFO_DIR ?= "${datadir}/qcom-boot-firmware"
+
 INHIBIT_DEFAULT_DEPS = "1"
 
 do_configure[noexec] = "1"
 do_compile[noexec] = "1"
 
 inherit deploy
+
+python do_install() {
+    import os
+    import zipfile
+
+    # Prefer the SRC_URI entry named "bootbinaries"; legacy recipes don't
+    # name their sole archive, so fall back to the first .zip entry.
+    fetch = bb.fetch2.Fetch(d.getVar('SRC_URI').split(), d)
+    bootbin = None
+    for url in fetch.urls:
+        ud = fetch.ud[url]
+        if ud.parm.get('name') == 'bootbinaries':
+            bootbin = ud
+            break
+        if bootbin is None and ud.localpath.endswith('.zip'):
+            bootbin = ud
+
+    # Qualcomm releases embed build info (Meta_Build_ID, per-image build
+    # IDs) as the zip archive comment; releases in other archive formats
+    # or without a comment carry no such metadata.
+    comment = ''
+    if bootbin is not None and zipfile.is_zipfile(bootbin.localpath):
+        with zipfile.ZipFile(bootbin.localpath) as z:
+            comment = z.comment.decode('utf-8', errors='replace')
+        comment = comment.replace('\r\n', '\n').replace('\r', '\n').strip()
+
+    info = 'Recipe: %s\n' % d.getVar('PN')
+    info += 'Version: %s\n' % d.getVar('PV')
+    if bootbin is not None:
+        info += 'Archive: %s\n' % os.path.basename(bootbin.localpath)
+        sha256 = getattr(bootbin, 'sha256_expected', None)
+        if sha256:
+            info += 'Archive-SHA256: %s\n' % sha256
+    if comment:
+        info += '\n' + comment + '\n'
+
+    board = (d.getVar('QCOM_BOOT_IMG_SUBDIR') or '').strip('/').split('/')[0] \
+        or d.getVar('BPN')
+    destdir = os.path.join(d.getVar('D') + d.getVar('QCOM_BOOT_INFO_DIR'), board)
+    bb.utils.mkdirhier(destdir)
+    dest = os.path.join(destdir, 'build-info.txt')
+    with open(dest, 'w') as f:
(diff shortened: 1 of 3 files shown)


ricardosalveti:
Example on ride:

```
root@qcs9100-ride-sx:~# cat /usr/share/qcom-boot-firmware/qcs9100/build-info.txt
Recipe: firmware-qcom-boot-qcs9100
Version: 00135
Archive: QCS9100_bootbinaries_00135.zip
Archive-SHA256: 2db1108f41ea9feddff0303be7e0b0af4fa66ce7405ee6e874749a9f4f266e82

=== Build Information ===
Meta_Build_ID: QCS9100.LE.2.0-00135-STD.PROD-5
Timestamp: 2026-07-17 02:21:23

=== Selected Image Build IDs ===
TZ: TZ.XF.5.29.1-00171.1-KODIAKAAAAANAAZT-1
BOOT: BOOT.MXF.1.0.c1-00532-KODIAKLA-1
AOP: AOP.HO.3.6.1-00027-LEMANSAU_E-1
SAIL: SAIL.SI.1.2-00125-AU.LEMANS-1
CPUCP: CPUCP.FW.1.0-00022-K2C.EXT-2
TZ_APPS: TZ.APPS.1.29-00391-KODIAKAAAAANAZT-1
```

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/aeafe834-bc96-4b80-9c2e-660984c77a79)**
---
**Test jobs for commit [`6069900`](https://github.com/qualcomm-linux/meta-qcom/commit/6069900cfef2f0e5a3442c4c834cd008bacfdfc1)**
---
Pass: 51 | Fail: 0 | Others: 2 | Total: 53
Pass: 9 | Fail: 0 | Total: 9

recipes-bsp/firmware-boot/firmware-qcom-boot-common.inc
```diff
@@ -4,13 +4,63 @@ S = "${UNPACKDIR}/${BOOTBINARIES}"
...
+        or d.getVar('BPN')
+    destdir = os.path.join(d.getVar('D') + d.getVar('QCOM_BOOT_INFO_DIR'), board)
+    bb.utils.mkdirhier(destdir)
+    dest = os.path.join(destdir, 'build-info.txt')
+    with open(dest, 'w') as f:
+        f.write(info)
+    os.chmod(dest, 0o644)
+}
+
+FILES:${PN} += "${QCOM_BOOT_INFO_DIR}"
```
lumag, line 62: What is the licence for the generated files? Currently they'd fall into the qcom-2 bin, which, I think, is not correct.

lumag (changes requested):
But, what happens if the user flashes the rootfs generated with one set of boot firmware onto the board with the older (or newer) firmware. The files would become out of sync.
I think, if we want to generate these reports, we should gather data from the debugfs/socinfo in runtime.

ricardosalveti:
> But, what happens if the user flashes the rootfs generated with one set of boot firmware onto the board with the older (or newer) firmware. The files would become out of sync. I think, if we want to generate these reports, we should gather data from the debugfs/socinfo in runtime.

That would be ideal, but this is currently not easily supported I would say, at least until we switch to an open boot firmware implementation, as it is not trivial to coordinate changes on the multiple different proprietary releases we have to maintain (not even the build info ID is available for all). We also have a problem of identifying one common component that could report the version id for all the other components (as we want to report cpucp, sail, tz, edk2, etc), so I believe the ideal here would be for all the individual components to have a way to report their respective version ids, then we could have a script that would gather the info from all the components.

The idea here was just to provide enough information for the firmware which is supposed to be used along with the rootfs, which is something CE often requires when debugging customer related issues (a mix of runtime and rootfs based information).

recipes-bsp/firmware-boot/firmware-qcom-boot-common.inc
ricardosalveti: Yeah, we don't need the same license here, as in the end it is just a metadata file, MIT/BSD would work fine I would say.

lumag:
> but this is currently not easily supported I would say,

```
# cat /sys/kernel/debug/qcom_socinfo/*/name
12:ADSP.HT.5.6.c2-00013-MAKENA-1.75542.4
00:BOOT.MXF.1.1.c1-00018-MAKENA-1
16:CDSP.HT.2.6-00735.4-MAKENA-1
15:SLPI.HY.4.1-00061-MAKENA-1
01:TZ.XF.5.13-96796-16
```

ricardosalveti:
> > but this is currently not easily supported I would say,
> 
> ```
> # cat /sys/kernel/debug/qcom_socinfo/*/name
> 12:ADSP.HT.5.6.c2-00013-MAKENA-1.75542.4
> 00:BOOT.MXF.1.1.c1-00018-MAKENA-1
> 16:CDSP.HT.2.6-00735.4-MAKENA-1
> 15:SLPI.HY.4.1-00061-MAKENA-1
> 01:TZ.XF.5.13-96796-16
> ```

Oh, this is great, didn't know we were already storing that info as part of qcom_socinfo, then a simple agent skill should be enough, let me take a look at that.

Thanks!

ricardosalveti:
Closing in favor of extracting the boot firmware information at runtime.

PR #2707, opened 2026-07-09
trusted-firmware-a-qcom: upgrade to qcom-next-2.15-20260804

royzah:
Upgrade TF-A to qcom-next-2.15-20260804 for all machines. The tag carries the lemans native platform drivers, so qcs9100 builds with no qtiseclib; qcm6490 keeps its qc_blobs qtiseclib.

PR #2910, opened 2026-08-10
trusted-firmware-a-qcom, optee-os-qcom: split into per-platform recipes

royzah:
qcom-armv8a cannot build kodiak and lemans firmware together: TF-A is gated on MACHINE and builds one platform, and optee-os-qcom picks its platform the same way.

Split both per platform, each TF-A recipe depending on its own OP-TEE, then widen the machine gate once nothing is tied to a single machine. Machines opt in through EXTRA_IMAGEDEPENDS as before.

Renames recipes, so layers with a trusted-firmware-a-qcom or optee-os-qcom bbappend or DEPENDS need updating.

Still machine-scoped on a multi-platform build: BL33 (`u-boot.bin`) and the shared `optee/` deploy dir.

Grew out of #2707.


dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -1,15 +1,19 @@
 DEPENDS += "qtestsign-native optee-os-qcom u-boot-qcom"
 
+# Per-SoC bits key on TFA_PLATFORM; underscores are invalid in overrides.
+TFA_PLATFORM_OVERRIDE = "${@(d.getVar('TFA_PLATFORM') or '').replace('_', '-')}"
+OVERRIDES .= ":tfa-${TFA_PLATFORM_OVERRIDE}"
+
 SRCREV_FORMAT_qcom ?= ""
 
-SRC_URI:append:qcm6490 = " git://github.com/coreboot/qc_blobs.git;protocol=https;name=qc-blobs;subdir=qc_blobs;branch=main"
+SRC_URI:append:tfa-rb3gen2 = " git://github.com/coreboot/qc_blobs.git;protocol=https;name=qc-blobs;subdir=qc_blobs;branch=main"
```
lumag: Is it specific to RB3 Gen2 or to all Kodiak platforms? In the latter case, this line is not correct.

lumag (changes requested):

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
royzah: Kodiak-wide, @lumag is right

`plat/qti/hoya/kodiak` has `rb3gen2` and `sc7280_chrome`, and `qc_blobs` is per SoC. Keyed blob on `QTISECLIB_SOC` now, both kodiak platforms map to sc7280. lemans stays blob-free.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -1,15 +1,24 @@
 DEPENDS += "qtestsign-native optee-os-qcom u-boot-qcom"
 
+# Per-SoC bits key on TFA_PLATFORM; underscores are invalid in overrides.
+TFA_PLATFORM_OVERRIDE = "${@(d.getVar('TFA_PLATFORM') or '').replace('_', '-')}"
+OVERRIDES .= ":tfa-${TFA_PLATFORM_OVERRIDE}"
+
+# qtiseclib is per-SoC and one SoC serves several TF-A platforms.
+QTISECLIB_SOC ?= ""
+QTISECLIB_SOC:tfa-rb3gen2 = "sc7280"
```
lumag: This should be a part of the recipe rather than an include file.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -23,14 +32,12 @@ EXTRA_OEMAKE:append = " \
...
-EXTRA_OEMAKE:append:qcm6490 = " \
-    QTISECLIB_PATH=${UNPACKDIR}/qc_blobs/sc7280/qtiseclib/libqtisec.a \
-   "
+EXTRA_OEMAKE:append = "${@' QTISECLIB_PATH=${UNPACKDIR}/qc_blobs/%s/qtiseclib/libqtisec.a' % d.getVar('QTISECLIB_SOC') if d.getVar('QTISECLIB_SOC') else ''}"
 
 do_compile[depends] += " u-boot-qcom:do_deploy"
 
-FIP_ELF_ADDR:qcm6490 = "0x9fc00000"
-FIP_ELF_ADDR:qcs9100 = "0xaf000000"
+FIP_ELF_ADDR:tfa-rb3gen2 = "0x9fc00000"
```
lumag: Again, if you are now overriding per the device, move it to the device recipe.

lumag (changes requested):

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
royzah: Moved both to the recipe; the .inc keeps only the mechanism. Also dropped the now-dead `SRCREV_FORMAT_qcom ?= ""`.

`TFA_PLATFORM` and `COMPATIBLE_MACHINE` are the same category and belong there too, but I kept this diff on qtiseclib. Say the word and I fold them in.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -1,15 +1,21 @@
 DEPENDS += "qtestsign-native optee-os-qcom u-boot-qcom"
 
-SRCREV_FORMAT_qcom ?= ""
+# Per-SoC bits key on TFA_PLATFORM; underscores are invalid in overrides.
+TFA_PLATFORM_OVERRIDE = "${@(d.getVar('TFA_PLATFORM') or '').replace('_', '-')}"
+OVERRIDES .= ":tfa-${TFA_PLATFORM_OVERRIDE}"
```
lumag: I think we don't need this anymore.

lumag (changes requested):

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
```diff
@@ -1,15 +1,21 @@
 DEPENDS += "qtestsign-native optee-os-qcom u-boot-qcom"
 
-SRCREV_FORMAT_qcom ?= ""
```
quaresmajose, line 3: Given that SRCREV_FORMAT_qcom no longer uses overrides and will not use them again, you can directly use SRCREV_FORMAT. With this, `SRCREV_FORMAT_qcom` can be dropped.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
royzah: Dropped the translation helper: underscores are fine in override names with the `:` syntax, so it is just `OVERRIDES .= ":tfa-${TFA_PLATFORM}"` and the overrides read `tfa-lemans_evk`, `tfa-rb3gen2`. My comment claiming otherwise was wrong, and it is gone.

The `OVERRIDES` line itself is load-bearing: meta-arm only sets `TFA_PLATFORM ?= ${FIRMWARE_PLATFORM}` and never puts the platform into `OVERRIDES`, so without this line nothing keys per platform. If you want it gone the alternatives are back to MACHINE overrides, or a recipe per device. Which would you rather have?

lumag: A recipe per device (or per SoC). The end-goal: let `qcom-armv8a` build both kodiak and lemans binaries at the same time.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
royzah: Agreed, dropped. `SRCREV_FORMAT .= "${@'_qc-blobs' if d.getVar('QTISECLIB_SOC') else ''}"` is the only line left, no other user of the variable.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom.inc
royzah: Done, split it. `trusted-firmware-a-qcom-rb3gen2` and `-lemans-evk`, each holding its platform, FIP address and qtiseclib SoC; no OVERRIDES left. Both are compatible with qcom-armv8a and deploy under their own FIRMWARE_DIR, so they build side by side.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom-rb3gen2_git.bb
```diff
@@ -0,0 +1,7 @@
+require trusted-firmware-a-qcom.inc
+
+TFA_PLATFORM = "rb3gen2"
+QTISECLIB_SOC = "sc7280"
+FIP_ELF_ADDR = "0x9fc00000"
+
+COMPATIBLE_MACHINE = "qcm6490|qcom-armv8a"
```
lumag: Isn't it compatible with any ARMv8 machine?

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom-lemans-evk_git.bb
```diff
@@ -0,0 +1,6 @@
+require trusted-firmware-a-qcom.inc
+
+TFA_PLATFORM = "lemans_evk"
+FIP_ELF_ADDR = "0xaf000000"
+
+COMPATIBLE_MACHINE = "qcs9100|qcom-armv8a"
```
lumag: Same, isn't it build-compatible with any ARMv8a machine?

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom-rb3gen2_git.bb
royzah: Yes. Dropped the machine list and moved it to the include as `COMPATIBLE_MACHINE:aarch64 = "(qcom)"`, matching u-boot-qcom, so any aarch64 qcom machine can build either platform. Note meta-arm defaults it to "invalid", so it cannot just be omitted.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom-lemans-evk_git.bb
royzah: Same here

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom-rb3gen2_git.bb
lumag: `(qcom)` isn't correct, since it'd match 32-bit machines too.

royzah: `:aarch64` already covers that one, meta-arm defaults `COMPATIBLE_MACHINE` to invalid so 32-bit gets skipped. 

It leans on another layer though, so made it explicit with `^$`. Also moved `qtiseclib` into the rb3gen2 recipe, include is blob-free now.

One catch for armv8a goal: optee-os-qcom is still `qcm6490|qcs9100`, so nothing provides it there, and BL33 is whichever u-boot config deployed last. Happy to split OP-TEE next.

royzah:
Added the OP-TEE split so the `(qcom)` claim actually holds: each TF-A recipe now depends on its own optee-os-qcom-kodiak / -lemans. Devkit and optee-test stay machine-scoped.

dynamic-layers/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a-qcom-rb3gen2_git.bb
lumag: Sorry, missed the `:aarch64` part. Then it's ok.

lumag (commented):
I think you might want to split/squash the commits.

royzah:
Restructured and rebased on master.

1. trusted-firmware-a-qcom: split into per-device recipes
2. optee-os-qcom: split into per-platform recipes
3. build on any aarch64 qcom machine

royzah:
CI on the previous head caught one real failure: `rb3gen2-core-kit-open-fw` fails the image licence gate, because meta-qcom-distro keys its `INCOMPATIBLE_LICENSE_EXCEPTIONS` on the old recipe name.

Fix is one line there: qualcomm-linux/meta-qcom-distro#422. Needs to merge before this can go green (CI tracks that repo at `main`, no pin).

Rest of that run was clean: `iq-9075-evk-open-fw` passed end to end, and rb3gen2 built TF-A with qtiseclib plus both OP-TEE platform recipes before hitting the licence check.

@lumag @ricardosalveti @b49020 @quaresmajose heads up, the rename touches that list.

royzah:
Rebased on master, picks up the `LicenseRef-` rename from #2802. Commits restructured as you suggested and the OP-TEE split is in. @lumag PTAL when you get a chance.

lumag:
"This branch has conflicts that must be resolved"

royzah:
Rebased on master, conflicts resolved (picked up the OP-TEE 20260721 bump). @lumag PTAL

lumag (approved):

royzah:
Thanks @lumag! Build is green (67/67) and the licence gate passes now that meta-qcom-distro#422 landed. Only red is the workflow security scan, which fails on other PRs too. Good to merge?

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/aadab72d-7d7c-4bfb-a454-581733b10af4)**
---
**Test jobs for commit [`a5c633a`](https://github.com/qualcomm-linux/meta-qcom/commit/a5c633ae0806edf17e754742c9e83b92ade3024a)**
---
Pass: 9 | Fail: 0 | Total: 9

lumag:
@royzah it requires a second review.

ricardosalveti (approved):

PR #2916, opened 2026-08-11
userspace-resource-manager: Update to 0.4.5

kartnema:
CHANGES in v0.4.5

- As part of this release URM-Extensions codebase is pulled in as part of the URM project itself.
- No functional changes, same set of target artifacts continue to be built


recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.5.bb
```diff
@@ -37,3 +50,5 @@ FILES:${PN}-tests += " \
     ${libdir}/libRestuneTestUtils.so* \
     ${libdir}/urm/libUrmTestPlugin.so* \
 "
+
+RDEPENDS:${PN} += "${PN}-extensions"
```
lumag: Does this not cause conflicts with the urm-extensions recipe?

lumag (changes requested):

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.5.bb
kartnema: Plan is to remove the separate recipe for urm-extensions entirely. We'll keep a single recipe.

lumag: And it does cause conflicts:
```
ERROR: userspace-resource-manager-extensions-0.3.8-r0 do_packagedata: Recipe userspace-resource-manager-extensions is trying to install files into a shared area when those files already exist. The files and the manifests listing them are:
```

kartnema: We can rename the package

lumag: Why? You are obsoleting -extensions. Drop the recipe.

kartnema: Added a commit to delete the recipe, so that we can keep the same package name: ${PN}-extensions

I had a query, do we need to create a PR to remove userspace-resource-manager-extensions from CORE_IMAGE_BASE_INSTALL as well (https://github.com/qualcomm-linux/meta-qcom-distro/blob/main/recipes-products/images/qcom-multimedia-image.bb#L36)

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/99dc87b5-aa57-4926-a6c9-5e8857b53b7d)**
---
**Test jobs for commit [`4e8d6f5`](https://github.com/qualcomm-linux/meta-qcom/commit/4e8d6f5d7e96cb04bae1b2d5e297554e7c244f39)**
---
Pass: 42 | Fail: 1 | Others: 1 | Total: 44
Pass: 8 | Fail: 1 | Total: 9

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.5.bb
```diff
@@ -8,27 +8,40 @@ LICENSE = "BSD-3-Clause-Clear"
...
 SYSTEMD_SERVICE:${PN} = "urm.service"
-FILES:${PN}-dev += "${libdir}/urm/libUrmTestPlugin.so"
+FILES:${PN}-dev += " \
+    ${libdir}/urm/libUrmTestPlugin.so \    
+    ${libdir}/urm/libUrmPlugin.so \
+"
+
 FILES:${PN} += "${sysconfdir}/urm/*"
 
+PACKAGE_BEFORE_PN += "${PN}-extensions"
```
lumag, line 38: Why do we need a separate -extensions package now?

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.5.bb
```diff
@@ -8,27 +8,40 @@ LICENSE = "BSD-3-Clause-Clear"
...
 
 PACKAGECONFIG ??= "\
     classifier \
+    extensions \
     ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'state-detector systemd', '', d)} \
     tests \
 "
 
 PACKAGECONFIG[classifier] = "-DBUILD_CLASSIFIER=ON,-DBUILD_CLASSIFIER=OFF"
+PACKAGECONFIG[extensions] = "-DBUILD_EXTENSIONS=ON,-DBUILD_EXTENSIONS=OFF"
```
lumag, line 25: Which builds are going to disable it?

lumag (changes requested):

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.5.bb
lumag: Yes.

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.5.bb
rajulup: Separating extensions from the base implementation helps avoid a monolithic architecture, improves maintainability, and provides a cleaner path for scaling and customization without impacting the core package. These features are already targeting LE and Android, and we expect the plugin ecosystem to grow over time. That is the reason initially we had used separate project itself but agree with your point on same code base and maintaining separately.

lumag: It's really all nice from the software architecture point of view. Here I'm questioning the packaging. URM doesn't make a lot of sense without the extensions (please correct me if I'm wrong).

kartnema: URM can run standalone as well, without Extensions.
Extensions are just customizations on top of the base behavior.

lumag: Why were we always installing them in the image?

rajulup: As target needs different tunings we're installing them for targets, if targets generic urm can serve alone.

Having a separate URM-Extensions package not only enables target-specific tunings to be delivered independently of the core URM package, but also provides clear separation of concerns. The URM core remains generic, stable, and upstream-friendly, while target/BU-specific YAMLs, plugins, and custom resources can evolve independently. This reduces validation effort, enables faster tuning updates, supports multi-target scalability, allows customer/SKU-specific customization, minimizes package update size, and prevents the core framework from becoming a large monolithic package as more target-specific features are added.

lumag: How is this all related to the meta-qcom / meta-qcom-distro integration, which always installs both URM and extensions?

kartnema: From meta-qcom-distro, we will remove userspace-resource-manager-extensions from CORE_IMAGE_BASE_INSTALL going forward.

In the current (updated) URM recipe, the packages are not tightly coupled at the packaging level. urm does not have an RDEPENDS on urm-extensions; it only has an RRECOMMENDS.

Both packages are currently installed together because most targets benefit from the extensions, this is not a package dependency requirement. urm remains independently installable and functional without urm-extensions and using RRECOMMENDS provides flexibility to omit the extensions where appropriate.

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.5.bb
kartnema: There are no immediate plans to disable it for existing targets. The reason for the split is not based on a specific build configuration, but on the dependency relationship itself: urm can operate independently, whereas urm-extensions provides target-specific tunings, YAMLs, and plugins layered on top of the core framework.

Modeling them as separate packages with an RRECOMMENDS better reflects that relationship and avoids making target-specific content a mandatory dependency of the core package.

Since it is always built, the CMake flag DBUILD_EXTENSIONS does not serve any purpose here, we have removed this flag in URM version v0.4.7, which we can sync here next.

lumag (approved):

ricardosalveti (approved):

Issue #1839, opened 2026-03-30
Boot time higher by 0.5 secs due to initramfs overhead


ctheegal:
Boot time higher by 0.5 secs due to initramfs overhead. 

Need to further to dig inside the initramfs. 
Unable to instrument the initramfs code. Please share steps for the same.

ctheegal:
Device used: https://www.qualcomm.com/developer/hardware/rb3-gen-2-development-kit 
Build Tag: https://github.com/qualcomm-linux/meta-qcom-releases/releases/tag/qli-2.0-rc2 

Boot time (Kerenel + initramfs) was always poor since the beginning of this upstream branch itself due to bulky initramfs cpio image used here (vs. the simple vanilla initramfs used for other/older reference builds on other code branches ).

**Timeline markers for dmesg logs**

 grep -E "unpack|initrd|init process|unused|System time" dmesg.txt
[    0.445076] Freeing initrd memory: 8852K
[    0.596326] clk: Disabling unused clocks
[    0.596574] PM: genpd: Disabling unused power domains
[    0.754571] Freeing unused kernel memory: 1792K
[    0.754817] Run /init as init process
[    2.003154] systemd[1]: System time advanced to built-in epoch: Wed 2026-03-04 13:21:03 UTC

**systemd_analyze log**  systemd_plot.svg
Startup finished in 2.072s (firmware) + 194ms (loader) + 1.991s (kernel) 

**Duplicate udev enumerations** 
Observing, udev rules/enumerations are invoked twice here (once in initramfs and next again in systemd ) which is an overkill by 1 sec for rb3-gen2 device and others.

Ftrace log (post processed) for better systemic visualization. 

[dmesg.txt](https://github.com/user-attachments/files/26684944/dmesg.txt)

![Image](https://github.com/user-attachments/assets/7958bb79-b4fd-435e-8931-c72067a9122e)

[ftrace_postProcess_20260413_173939.zip](https://github.com/user-attachments/files/26685403/ftrace_postProcess_20260413_173939.zip)

<img width="1181" height="329" alt="Image" src="https://github.com/user-attachments/assets/a5becc1d-fe1c-4f74-9bb2-20e32a25a25f" />

[initramfs-rootfs-image-rb3gen2-core-kit.cpio.gz](https://github.com/user-attachments/files/26685579/initramfs-rootfs-image-rb3gen2-core-kit.cpio.gz)

ctheegal:
[...]

PR #2917, opened 2026-08-11
ci/performance: enable root-only initramfs udev trigger

wenwfu:
## Summary

Follow up on #1839 by reducing unnecessary udev processing in `initramfs-rootfs-image`.

The initramfs udev module currently replays add events for the complete device tree and waits for all resulting events. The regular rootfs initramfs only needs the root block device before mounting the real root filesystem, while the remaining devices will be enumerated again after switching to systemd.

Add an opt-in `initramfs.udev-root-only=1` kernel parameter that triggers udev only for the root partition. Enable it through `ci/performance.yml` for `initramfs-rootfs-image`.

## Implementation

For supported root specifications, the initramfs extracts the corresponding kernel uevent property:

- `PARTLABEL=<label>` is matched using `PARTNAME=<label>`;
- `PARTUUID=<uuid>` is matched using `PARTUUID=<uuid>`.

A dry-run is performed first because `udevadm trigger` succeeds even when no device matches. The targeted trigger is used only when the root partition can be identified unambiguously.

## Scope and fallback

This is narrower than the original proposal in #1839, which triggered the entire block subsystem.
The implementation preserves the existing full coldplug path when:

- the root specification is unsupported;
- the root partition is not yet present in sysfs;
- no device matches the requested partition;
- the targeted udev trigger fails.

This also addresses the USB and network-root concerns raised in #1839:

- a USB-backed root is optimized only when its root partition is already discoverable; otherwise full coldplug is used;
- network roots do not match the partition selectors and retain full coldplug.

No udev rules are removed. Without the kernel parameter, the existing OE-Core behavior remains unchanged.

The parameter is enabled only by the meta-qcom performance configuration when `INITRAMFS_IMAGE` is `initramfs-rootfs-image`. SOTA and other specialized initramfs images retain their existing device discovery behavior.

## Performance results

Test configuration:

- Machine: `rb3gen2-core-kit` (RB3 Gen 2)
- KAS configuration:
  `ci/rb3gen2-core-kit.yml:ci/qcom-distro.yml:ci/performance.yml`
- Initramfs: `initramfs-rootfs-image`
- Root specification: `PARTLABEL=rootfs`
- Number of measured boots: 3 (full coldplug), 2 (root-only)

| Measurement | Full coldplug | Root-only trigger | Improvement |
|---|---:|---:|---:|
| initramfs udev trigger + settle | 981.6 ms | 109.8 ms | −871.8 ms |
| `Run /init` → rootfs `EXT4-fs mounted` | 1340.2 ms | 441.1 ms | −899.1 ms |

### DIFF
--- a/ci/performance.yml
+++ b/ci/performance.yml
@@ -4,3 +4,4 @@ header:
 local_conf_header:
   cmdline-perf: |
     KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=linux"
+    KERNEL_CMDLINE_EXTRA:append = "${@' initramfs.udev-root-only=1' if d.getVar('INITRAMFS_IMAGE') == 'initramfs-rootfs-image' else ''}"


lumag (changes requested):
The initramfs changes must go to OE-Core

wenwfu:
> The initramfs changes must go to OE-Core

 Thanks for the review. I’ll submit this change to OE-Core first and follow up here once it has been accepted.

ricardosalveti:
Not really qcom specific.

wenwfu:
hi @lumag  @ricardosalveti  - The initramfs change is now upstream in OE-Core as https://github.com/openembedded/openembedded-core/commit/d3a196f5742425fb24cfc21bb380f3e96c49253c. This PR has been rebased and now only enables the feature for initramfs-rootfs-image in the performance configuration.

lumag (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/ce4d466e-332b-4ca6-8d8f-d769a7074b34)**
---
**Test jobs for commit [`ad365db`](https://github.com/qualcomm-linux/meta-qcom/commit/ad365db7bb366a25b61719bf31d4825853d16be2)**
---
Pass: 21 | Fail: 4 | Total: 25
Pass: 9 | Fail: 0 | Total: 9

ci/performance.yml
```diff
@@ -4,3 +4,4 @@ header:
 local_conf_header:
   cmdline-perf: |
     KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=linux"
+    KERNEL_CMDLINE_EXTRA:append = "${@' initramfs.udev-root-only=1' if d.getVar('INITRAMFS_IMAGE') == 'initramfs-rootfs-image' else ''}"
```
ricardosalveti, line 7: This won't work for the catchall/sota build, as that uses a different INITRAMFS_IMAGE. The real dependency is the availability of initramfs-module-udev in the initramfs image.

Maybe just add it by default instead?

wenwfu: The catchall/sota builds use initramfs-ostree-image with root=LABEL=, which is not yet support by the current OE-core implement. I have submitted LABEL support here: https://patchwork.yoctoproject.org/project/oe-core/patch/20260826092522.1618546-1-wenwfu@qti.qualcomm.com/. 

Could we merge this PR with the current scope? One the LABEL support is merged into OE-core, I will submit a follow-up Fr to enable  initramfs.udev-root-only=1 uncoditionally.

vjetty-qcom:
Observing camera regression on RB8_Prop

[Info] Test Summary:

Total Tests: 3
Passed: 1
Failed: 2
Skipped: 0
Failed Tests:

DUMP_DIR does not exist: /var/cache/camera/nativehaltest
Camera_NHX FAIL NHX reported FAILED=1 no dumps detected dump checksum validation failed

Axiom Job : https://axiom.qualcomm.com/#/reports/job/38762413

lumag:
> Observing camera regression on RB8_Prop

Please stop spamming.

ci/performance.yml
lumag: @ricardosalveti I think it will still work, It will pass an empty string, meaning no performance improvements.

ricardosalveti: Oh sure, my point was that this won't be used for the catchall build, just for qcom-distro, but yeah, won't cause any harm.

ricardosalveti (approved):

koenkooi (approved):

quaresmajose (approved):

ricardosalveti:
Removed backport wrynose tag as this is not yet available in oe-core wrynose, needs backporting there first.

ricardosalveti:
Known failures, merging.

PR #2935, opened 2026-08-13
Gunyah boot support: el2gh compatibles and VendorDtbOverlays pre-seeding

vkraleti:
On Gunyah-capable SoCs (lemans, monaco, talos) the plain DT compatible string
selects KVM and carries the `<soc>-el2` overlay, while the `el2gh` variant
selects the Gunyah hypervisor and drops those overlays. This series completes
that split and makes a Gunyah boot selectable at build time:

- provide the `el2gh` compatible strings that were missing, so every KVM
string on a Gunyah-capable SoC has a counterpart.

- add an EBBR-style `VolatileVars.bin` to the ESP so firmware sees the
`VendorDtbOverlays` UEFI variable on first boot, instead of waiting for
firmware-side generation to create it.

### DIFF
--- a/conf/machine/include/fit-dtb-compatible-linux-qcom.inc
+++ b/conf/machine/include/fit-dtb-compatible-linux-qcom.inc
@@ -14,180 +14,21 @@
 #
 # <compatible-encoded> is the DT compatible string with every comma replaced
 # by an underscore (BitBake flag names cannot contain commas).
+#
+# Add entries for combinations whose overlays are available only in
+# LINUX_QCOM_KERNEL_DEVICETREE. Move these to fit-dtb-compatible.inc once
+# the overlays are available in linux-yocto.
 
 require conf/machine/include/fit-dtb-compatible.inc
+
+# ---------- glymur ----------
+FIT_DTB_COMPATIBLE[qcom_glymur-crd-staging] = "glymur-crd glymur-staging"
+FIT_DTB_COMPATIBLE[qcom_glymur-crd-camx] = "glymur-crd glymur-crd-camx"
 
 # ---------- hamoa  ----------
 FIT_DTB_COMPATIBLE[qcom_hamoa-evk-camx] = "hamoa-iot-evk hamoa-evk-camx"
 FIT_DTB_COMPATIBLE[qcom_hamoa-evk-camx-staging] = "hamoa-iot-evk hamoa-evk-camx hamoa-staging"
 FIT_DTB_COMPATIBLE[qcom_hamoa-evk-camx-el2kvm] = "hamoa-iot-evk hamoa-evk-camx x1-el2 hamoa-camx-el2"
-
-# ---------- purwa -----------
-FIT_DTB_COMPATIBLE[qcom_purwa-evk-camx] = "purwa-iot-evk purwa-evk-camx"
-FIT_DTB_COMPATIBLE[qcom_purwa-evk-camx-el2kvm] = "purwa-iot-evk purwa-evk-camx x1-el2 purwa-camx-el2"
-
-# ---------- lemans ----------
-
-# Add entries for combinations whose overlays are available only in
-# LINUX_QCOM_KERNEL_DEVICETREE. Move these to fit-dtb-compatible.inc once
-# the overlays are available in linux-yocto.
-
-FIT_DTB_COMPATIBLE[qcom_qcs9075-iot] = \
-    "lemans-evk lemans-evk-camera-csi1-imx577 lemans-el2 lemans-evk-sd-card"
-FIT_DTB_COMPATIBLE[qcom_qcs9075-iot-staging] = \
-    "lemans-evk lemans-evk-camera-csi1-imx577 lemans-staging lemans-el2 lemans-evk-sd-card"
-FIT_DTB_COMPATIBLE[qcom_qcs9075-iot-el2gh-staging] = \
-    "lemans-evk lemans-evk-camera-csi1-imx577 lemans-staging lemans-evk-sd-card"
-FIT_DTB_COMPATIBLE[qcom_qcs9075-iot-camx] = \
-    "lemans-evk lemans-evk-camx lemans-el2 lemans-camx-el2 lemans-evk-sd-card"
-FIT_DTB_COMPATIBLE[qcom_qcs9075-iot-camx-staging] = \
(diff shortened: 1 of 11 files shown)


vkraleti:
VendorDtbOverlays pre-seeding script is replicated from https://github.com/qualcomm-linux/qcom-build-utils/pull/203, where a similar script bootloader/gen-volatile-vars.py was added.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/b604bbd4-77f5-468e-8bef-53149a0a733f)**
---
**Test jobs for commit [`610f28e`](https://github.com/qualcomm-linux/meta-qcom/commit/610f28ead330996641c5beb302cc25237a9efc62)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 50 | Fail: 0 | Others: 2 | Total: 52

lumag:
Can we please make edk2 set either el2kvm or el2gh, depending on the XBL config that was flashed?

conf/machine/include/fit-dtb-compatible-linux-qcom.inc
```diff
@@ -139,24 +139,30 @@ FIT_DTB_COMPATIBLE[qcom_sa8775p-qam-r2.0-camx-staging] = \
...
 FIT_DTB_COMPATIBLE[qcom_qcs9100-qam-r2.0-el2gh-staging] = \
     "qcs9100-ride-r3 lemans-staging"
 
+FIT_DTB_COMPATIBLE[qcom_sa8775p-qam-el2gh] = "sa8775p-ride"
 FIT_DTB_COMPATIBLE[qcom_sa8775p-qam-el2gh-staging] = \
     "sa8775p-ride lemans-staging"
 FIT_DTB_COMPATIBLE[qcom_sa8775p-qam-el2gh-camx] = \
     "sa8775p-ride sa8775p-ride-camx"
-FIT_DTB_COMPATIBLE[qcom_sa8775p-qam-camx-el2gh-staging] = \
+FIT_DTB_COMPATIBLE[qcom_sa8775p-qam-el2gh-camx-staging] = \
```
lumag, line 163: Why? No, it was agreed that the extra strings are sorted.

vkraleti: Sure, I'll retain the sort order.

vkraleti:
> Can we please make edk2 set either el2kvm or el2gh, depending on the XBL config that was flashed?

Presently UEFI is not supporting this. @anushka-qcom  can you confirm if this can be supported?

ricardosalveti:
The concept of having VolatileVars.bin is fine and valid, but I don't think this should be used for deciding kvm x gunyah, as the decision about the hypervisor to use is critical for the device to boot successfully, so erasing this file or the environment will already be enough to put the image in a bad state (camx is different as it won't break the boot behavior).

Also, this won't work on closed devices, as then the firmware will use RPMB for variable storage.

anushka-qcom:
> > Can we please make edk2 set either el2kvm or el2gh, depending on the XBL config that was flashed?
> 
> Presently UEFI is not supporting this. anushka-qcom can you confirm if this can be supported?

This is not scalable - UEFI should not update VendorDtbOverlays. This will also not work when we move to UEFI in EL2, the hypervisor would execute after EDK2. Ideally, OEM that is using Gunyah should configure the multi-dtb with Gunyah overlays by default and vice-versa. This should not be a runtime decision. 
For current UEFI in EL1 flows - user needs to update xbl_config as well as multi-dtb. 
For UEFI in EL2 flows, user would need to update UKI as well as multi-dtb.

lumag:
@anushka-qcom the multi-dtb is already common, it contains DTs for both KVM and Gunyah (selected via el2kvm or el2gh, selected via the UEFI variable). Currently that variable is modified by the user manually. Question is if the el2kvm vs el2gh can be detected by the UEFI on its own, sto that we don't require manual intervention after reflashing.

anushka-qcom:
> anushka-qcom the multi-dtb is already common, it contains DTs for both KVM and Gunyah (selected via el2kvm or el2gh, selected via the UEFI variable). Currently that variable is modified by the user manually. Question is if the el2kvm vs el2gh can be detected by the UEFI on its own, sto that we don't require manual intervention after reflashing.

Downstream UEFI may be able to detect, however, we need to consider system ready boot flows where UEFI will not be able to detect which hypervisor would run after UEFI and hence it will not be able to choose the overlay for the hypervisor. We cannot design solutions with UEFI in EL1 as an assumption.

lumag:
@anushka-qcom that's exactly the SystemReady-targeting point. The UEFI should detect and provide the information about the flashed XBL config, whether Linux will be started in EL2 targeting KVM or EL1 and use Gunyah.

anushka-qcom:
@lumag When UEFI runs in EL2 (in ARM standard way), then UEFI doesn't have information on which Hypervisor would run next - Gunyah, Xen, KVM etc. Any of these is a possibility and xbl_config will not hold any information on what hypervisor is needed. In that scenario, UEFI cannot detect and apply overlays for hypervisors. Please note that UEFI is expected to move to EL2 boot flow for upcoming targets and we need to ensure design supports UEFI in EL2 and UEFI in EL1 without assumption that UEFI is always aware of the Hypervisor to be used.

lumag:
@anushka-qcom can we export any information from XBL?

anushka-qcom:
XBL will also not be aware of the hypervisor.. Hypervisor will be loaded by systemd-boot..

lumag:
> XBL will also not be aware of the hypervisor.. Hypervisor will be loaded by systemd-boot..

How? I don't see this happening. The systemd-boot loads Linux. In the end, we mostly care about the Linux being loaded in EL1 or EL2.

anushka-qcom:
When EDK2 executes in EL2, XBL doesn;t load HYP (this is boot flow on Glymur). EDK2 will load OS Loader which can load Hyp+Linux or just Linux. If Hyp is loaded, then hyp will execute and transition to Linux in EL1. If hyp is not loaded, then Linux will get control in EL2. In either scenario, EDK2 or XBL doesn't know what execution env Linux is going to start in.

lumag:
@anushka-qcom you've described Glymur boot flow. However the el2kvm / el2gh also applies to other platforms.

lumag:
@anushka-qcom and for the Glymur situation, would it be possible to toggle the UEFI var from HYP?

PR #2936, opened 2026-08-13
ci/ci: enable ccache

### LINKED ISSUE #1051: Enable ccache on CI builds

quaresmajose:
The ccache can improve significantly our builds speeds when we can't take
advantage of the bitbake sstate cache, in sutch scenario we need to compile
the code. The cache storage needs to be available from the previous builds
so we need to store these ccache artifacts temporay on our permanent storage.

Fixes https://github.com/qualcomm-linux/meta-qcom/issues/1051

### DIFF
--- a/ci/ccache.yml
+++ b/ci/ccache.yml
@@ -11,3 +11,21 @@ local_conf_header:
     # required to build '-native' target
     HOSTTOOLS += "ccache"
     ASSUME_PROVIDED += "ccache-native"
+
+  ccache-recipe-contol: |
+    # disable all recipes
+    CCACHE_DISABLE = "1"
+    # table with enabled recipes
+    CCACHE_DISABLE:pn-clang = "0"
+    CCACHE_DISABLE:pn-clang-native = "0"
+    CCACHE_DISABLE:pn-gcc = "0"
+    CCACHE_DISABLE:pn-gcc-cross-aarch64 = "0"
+    CCACHE_DISABLE:pn-gcc-cross-arm = "0"
+    CCACHE_DISABLE:pn-glibc = "0"
+    CCACHE_DISABLE:pn-llvm = "0"
+    CCACHE_DISABLE:pn-llvm-native = "0"
+    CCACHE_DISABLE:pn-linux-qcom = "0"
+    CCACHE_DISABLE:pn-linux-qcom-next = "0"
+    CCACHE_DISABLE:pn-linux-yocto = "0"
+    CCACHE_DISABLE:pn-rust = "0"
+    CCACHE_DISABLE:pn-rust-native = "0"
--- a/ci/ccache.yml
+++ b/ci/ccache.yml
@@ -1,0 +1,13 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/siemens/kas/master/kas/schema-kas.json
+
+header:
+  version: 14
+
+local_conf_header:
+  ccache: |
+    INHERIT += "ccache"
+    CCACHE_MAXSIZE = "200G"
+    CCACHE_TOP_DIR = "${SSTATE_DIR}/ccache"
+    # required to build '-native' target
+    HOSTTOOLS += "ccache"
+    ASSUME_PROVIDED += "ccache-native"
--- a/ci/ci.yml
+++ b/ci/ci.yml
@@ -4,6 +4,7 @@ header:
   version: 14
   includes:
   - ci/mirror.yml
+  - ci/ccache.yml
 
 local_conf_header:
   ci: |
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -29,7 +29,7 @@ env:
   CACHE_DIR: /efsx/qli/meta-qcom
   KAS_CLONE_DEPTH: 1
   KAS_CONTAINER: ${{ github.workspace }}/kas-container
-  KAS_CONTAINER_IMAGE: ghcr.io/siemens/kas/kas:5.5
+  KAS_CONTAINER_IMAGE: ghcr.io/siemens/kas/kas@sha256:4b210b0be3c887875973c4f4c35470b4ecf3edd2346c9de1bdd33feae5b86e94 # 5.5
 
 jobs:
   kas-setup:
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -29,7 +29,7 @@ env:
   CACHE_DIR: /efsx/qli/meta-qcom
   KAS_CLONE_DEPTH: 1
   KAS_CONTAINER: ${{ github.workspace }}/kas-container
-  KAS_CONTAINER_IMAGE: ghcr.io/siemens/kas/kas@sha256:4b210b0be3c887875973c4f4c35470b4ecf3edd2346c9de1bdd33feae5b86e94 # 5.5
+  KAS_CONTAINER_IMAGE: ghcr.io/siemens/kas/kas@sha256:3f3d3257e8d4ca3adfcb2e0025e6da24339fb8d68a6999ca36c37381351582d5 # 5.5-9-g0555188
 
 jobs:
   kas-setup:
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -46,7 +46,7 @@ on:
 env:
   KAS_CLONE_DEPTH: 1
   KAS_CONTAINER: ${{ github.workspace }}/kas-container
-  KAS_CONTAINER_IMAGE: ghcr.io/siemens/kas/kas:5.5
+  KAS_CONTAINER_IMAGE: ghcr.io/siemens/kas/kas@sha256:4b210b0be3c887875973c4f4c35470b4ecf3edd2346c9de1bdd33feae5b86e94 # 5.5
 
 jobs:
   reusable_compile_job:
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -46,7 +46,7 @@ on:
 env:
   KAS_CLONE_DEPTH: 1
   KAS_CONTAINER: ${{ github.workspace }}/kas-container
-  KAS_CONTAINER_IMAGE: ghcr.io/siemens/kas/kas@sha256:4b210b0be3c887875973c4f4c35470b4ecf3edd2346c9de1bdd33feae5b86e94 # 5.5
+  KAS_CONTAINER_IMAGE: ghcr.io/siemens/kas/kas@sha256:3f3d3257e8d4ca3adfcb2e0025e6da24339fb8d68a6999ca36c37381351582d5 # 5.5-9-g0555188
 
 jobs:
   reusable_compile_job:
--- a/.github/workflows/sstate-cleanup.yml
+++ b/.github/workflows/sstate-cleanup.yml
@@ -20,5 +20,5 @@ jobs:
         run: |
           # bitbake refreshes the mtime of every sstate file a build uses, so
           # a file older than 15 days was not used by any build in 15 days
-          find "${CACHE_DIR}/sstate-cache" -type f -mtime +15 -delete
-          find "${CACHE_DIR}/sstate-cache" -mindepth 1 -type d -empty -delete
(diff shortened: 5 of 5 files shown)


ci/ccache.yml
```diff
@@ -8,3 +8,11 @@ local_conf_header:
     INHERIT += "ccache"
     CCACHE_MAXSIZE = "200G"
     CCACHE_TOP_DIR = "${SSTATE_DIR}/ccache"
+    CCACHE_DISABLE = "1"
```
lumag, line 17: Can we just enable ccache for all the recipes? What are the drawbacks?

ci/ccache.yml
```diff
@@ -14,5 +14,5 @@ local_conf_header:
     CCACHE_DISABLE:pn-gcc = "0"
     CCACHE_DISABLE:pn-glibc = "0"
     CCACHE_DISABLE:pn-llvm = "0"
-    CCACHE_DISABLE:pn-linux-yocto = "0"
+    #CCACHE_DISABLE:pn-linux-yocto = "0"
```
lumag: So, what is in 2026-08-14 09:59:21 - INFO     - | ERROR: Details can be found at: /work/build/tmp/work-shared/qcom-armv8a/kernel-source/.kernel-meta/cfg/merge_config_build.log ? Why does it fail?

lumag (changes requested):

ci/ccache.yml
quaresmajose: In internal discussions, we concluded that there wasn't much value in enabling it for everyone. However, the heavier packages in the compilation would always have some benefit. In reality, it's a trade-off between space occupied and benefit utilized.

ci/ccache.yml
quaresmajose: I am in the evaluation process and I don't yet know the reasons.

ci/ccache.yml
lumag: > there wasn't much value in enabling it for everyone

Why? What was the main argument for deciding it this way? Just to document it here. Do we have any stats?

ci/ccache.yml
```diff
@@ -8,3 +8,11 @@ local_conf_header:
...
     CCACHE_MAXSIZE = "200G"
     CCACHE_TOP_DIR = "${SSTATE_DIR}/ccache"
+    CCACHE_DISABLE = "1"
+    # table with enabled recipes
+    CCACHE_DISABLE:pn-clang = "0"
+    CCACHE_DISABLE:pn-gcc = "0"
+    CCACHE_DISABLE:pn-glibc = "0"
+    CCACHE_DISABLE:pn-llvm = "0"
+    CCACHE_DISABLE:pn-linux-yocto = "0"
+    CCACHE_DISABLE:pn-rust = "0"
```
lumag, line 30: Also, does it affect -native packages? Should llvm-native still use ccache?

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/dd5001ed-ba8c-4f8c-8d72-0945d3dfb428)**
---
**Test jobs for commit [`28464f9`](https://github.com/qualcomm-linux/meta-qcom/commit/28464f962fd56e81449a830a9b6118532844e61f)**
---
Pass: 9 | Fail: 0 | Total: 9

ci/ccache.yml
quaresmajose: We don't have stats and that conclusion came from a internal slack discussion.

Basically, we'll have to divide the available space among all existing components. Focusing a large portion of the allocated space on components that are fast enough even without ccache. We then conclude that it would be more efficient if we allocated space only to the components that take the most advantage of using ccache.

I'll add it to the commit.

ci/ccache.yml
quaresmajose: Unfortunately, it's not possible to use it for `-native` targets because it requires ccache on the container and it is not installed by default. I was thinking of proposing this upstream.

ci/ccache.yml
quaresmajose: It fails becuas it also requires the ccache-native for the native tools. After install ccache on the container and add the required configs it works

ci/ccache.yml
lumag: Can't we cache those through `CCACHE_NATIVE_RECIPES_ALLOWED`?

ci/ccache.yml
lumag: Isn't it a missing dependency somewhere?

ci/ccache.yml
quaresmajose: With `CCACHE_DISABLE:pn-linux-yocto = "0"` the `ccache-native` was added in DEPENDS of the linux-yocto regardless of CCACHE_NATIVE_RECIPES_ALLOWED but it still fails. perhaps because the linux-yocto is not purely native.

Finding the variable CCACHE_NATIVE_RECIPES_ALLOWED quite difficult to use, you use it to control native recipes but the same can be done simply as CCACHE_DISABLE, It's only useful when ccache is no in HOSTTOOLS.

It works with ccache installed on the host machine, for us in the container, and with: 
```
ASSUME_PROVIDED += "ccache-native"
HOSTTOOLS += "ccache"
```

ci/ccache.yml
quaresmajose: The missing dependencie is `HOSTTOOLS += "ccache"`

lumag: That makes linux-yocto depend on the host's ccache, doesn't it? My point is that the recipe should be buildable even withou it, by using ccache-native.

quaresmajose:
Proposed ccache to be installed on kas-container and it is accepted and available in `:next` image.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/f0255b54-4530-4cfb-b82f-42355a95b553)**
---
**Test jobs for commit [`2ca5572`](https://github.com/qualcomm-linux/meta-qcom/commit/2ca55720702bfb4823a6b4fe8645936bc1c0f5d9)**
---
Pass: 9 | Fail: 0 | Total: 9

ci/ccache.yml
lumag: As ccache is now a part of the kas-container, should we also enable ccache for major native packages?

lumag (dismissed):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/be5cf30a-9024-40c6-8872-f3328ecfc2c1)**
---
**Test jobs for commit [`5989d85`](https://github.com/qualcomm-linux/meta-qcom/commit/5989d856e1761fb287743972d0324ffb2ef2503e)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 30 | Fail: 4 | Total: 34

ci/ccache.yml
quaresmajose: You're right, we need to list specifically everything we want. Added the "-native" target, gcc-cross-* and linux-qcom[-next]

lumag (dismissed):

ricardosalveti (dismissed):
LGTM, but I believe we want to wait for kas 5.6, no?

quaresmajose:
While it's still in the next branch, it can still float. It's best to wait until it land in the master branch and then it will be safe to do the https://github.com/qualcomm-linux/meta-qcom/pull/2936/commits/5429c71bca7e0d7852f9facf673cf6964286317a bump but with pin.

```
5429c71bc github/workflows: bump KAS_CONTAINER_IMAGE to get ccache

Proposed ccache to be installed on kas-container and it is accepted
and available in :next image.

FIXME: Wait until it lands on the master and do the final pin.
```

quaresmajose:
> While it's still in the next branch, it can still float. It's best to wait until it land in the master branch and then it will be safe to do the [5429c71](https://github.com/qualcomm-linux/meta-qcom/commit/5429c71bca7e0d7852f9facf673cf6964286317a) bump but with pin.
> 
> ```
> 5429c71bc github/workflows: bump KAS_CONTAINER_IMAGE to get ccache
> 
> Proposed ccache to be installed on kas-container and it is accepted
> and available in :next image.
> 
> FIXME: Wait until it lands on the master and do the final pin.
> ```

Landed in master https://github.com/siemens/kas/commit/055518846c3963114efbd0107bf3f2e59b4b3ba5 ` 05551884 Dockerfile: install ccache`

lumag (approved):
Thanks!

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/14475ec9-ff4d-4186-b78b-a3ae364827d4)**
---
**Test jobs for commit [`f16a384`](https://github.com/qualcomm-linux/meta-qcom/commit/f16a384db18679bb69f0f9af4e89a09ce344895f)**
---
Pass: 9 | Fail: 0 | Total: 9

ci/ci.yml
```diff
@@ -4,6 +4,7 @@ header:
   version: 14
   includes:
   - ci/mirror.yml
+  - ci/ccache.yml
```
koenkooi, line 7: Nit: sutch -> such in the commit message

koenkooi:
I'm missing numbers in the commit about how much time it saves vs the amount of storage it adds

ci/ci.yml
quaresmajose: fixed

quaresmajose:
> I'm missing numbers in the commit about how much time it saves vs the amount of storage it adds

I don't have the exact numbers, but the more components that share the total space, the less space each one will have. Since less space results in less cache, then the more components, the less efficient it is.

ricardosalveti:
Should we wait for kas 5.6?

quaresmajose:
We don't need to wait. There is a cosmetic update in kas-container and the essential package list in Dockerfile was updated to mach Yocto 6. The rest has no impact on us.
```diff
$ git diff --stat 5.5 origin/master 
 .github/actions/docker-init/action.yml   |  2 +-
 .github/actions/perform-tests/action.yml | 24 +++++++++++++-----------
 .github/workflows/next.yml               |  3 +--
 Dockerfile                               | 54 +++++++++++++++++++++++++++++++-----------------------
 docs/format-changelog.rst                |  2 +-
 kas-container                            |  4 ++--
 scripts/build-container.sh               |  4 ++--
```

ricardosalveti:
Known failures.

ricardosalveti (approved):

ricardosalveti:
@quaresmajose mind having a look at backporting to wrynose?

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/fe8b3002-45f8-44b8-b4af-13fcef2d773e)**
---
**Test jobs for commit [`39cf185`](https://github.com/qualcomm-linux/meta-qcom/commit/39cf1856083fe252efdaa789f9c696c032e10897)**
---
Pass: 9 | Fail: 0 | Total: 9

quaresmajose:
> Jose Quaresma (@quaresmajose) mind having a look at backporting to wrynose?

Proposed in https://github.com/qualcomm-linux/meta-qcom/pull/3144

PR #2944, opened 2026-08-14
camxlib-kodiak: Raise temporary changes for customer.

kripal-qipl:
camxlib-kodiak: Raise temporary changes for customer.
- Add temporary changes to consume the DeCommand feature on the OX03F10 sensor.
- Changes are based on the QLI2.0 GA release for customer validation requirements.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.24.bb
+++ b/dynamic-layers/openembedded-layer/recipes-multimedia/camx/camxlib-kodiak_1.0.24.bb
@@ -8,15 +8,15 @@ LICENSE = "LICENSE.qcom-2"
 LIC_FILES_CHKSUM = "file://usr/share/doc/${BPN}/LICENSE.QCOM-2.txt;md5=165287851294f2fb8ac8cbc5e24b02b0 \
                     file://usr/share/doc/${BPN}/NOTICE;md5=04facc2e07e3d41171a931477be0c690"
 
-PBT_BUILD_DATE = "260617"
+PBT_BUILD_DATE = "260817"
 SRC_URI = " \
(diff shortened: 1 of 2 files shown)


ricardosalveti:
Why are you raising a temporary customer-specific change here? It should go as a normal change to master first, then backported to wrynose.

kripal-qipl:
> Why are you raising a temporary customer-specific change here? It should go as a normal change to master first, then backported to wrynose.

The changes will go into master, but this PR is to support the dot release. It includes only the fix on the GA tip, and we need to share a new tarball with just those fixes. So raised a PR so we can add it to doc/SBA and get the fix on top of GA for customer who reported issue.

could you confirm whether this is the right process, or if there’s another way to share the fix with the customer?

ricardosalveti:
We should discuss as we might not required a full PR for this scenario. You could open a bug, describe the behavior and point to a github tree and branch (which would the one consumed as part of the PR) containing the temporary fix for the issue, without necessarily making it a customer specific fix (in the end it is just a temporary bug fix). Then that same branch / tree could be documented as a workaround for the issue.

lumag:
No, the changes must go through the master branch. We can't land fixes in the wrynose and keep master without the fix. Also, the whole concept of 'temporary changes' is broken and forbidden here.
Last, but not least, you have completely ignored all the docs describing the contribution process.

Closing the PR. Please do it properly and submit against master.

manishd-qcom:
This is a temporary PR created solely to share engineering changes with the customer through the Service Task process. It is not intended to be merged.
For the actual merge, a new PR will first be raised against master, and the changes will then be propagated to Wrynose.

lumag:
> This is a temporary PR created solely to share engineering changes with the customer through the Service Task process. It is not intended to be merged.

Don't do that. Don't open PRs which are not supposed to be merged.

PR #2901, opened 2026-08-07
ci: pass step state via GITHUB_OUTPUT instead of GITHUB_ENV

anujm1:
zizmor's github-env audit flags writes to GITHUB_ENV as a potential code-execution vector. Replace the GITHUB_ENV writes in the compile and lava-test-plans composite actions with step outputs.

In compile/action.yml the kas variables (KAS_CONTAINER, KAS_WORK_DIR, KAS_YAMLS, DL_DIR, SSTATE_DIR) are now emitted as step outputs and re-supplied to the downstream kas steps via per-step env blocks, keeping each step's shell environment identical so kas-container still forwards DL_DIR/SSTATE_DIR into the build. The INPUTS_* re-exports are dropped in favour of referencing the action inputs directly.

In lava-test-plans/action.yml JOBS_OUT_PATH becomes a step output consumed by the upload step.

PR #2455, opened 2026-06-15
ci: compile/action: drop github action

quaresmajose:
The action is no longer necessary because we moved to reusable workflow.

PR #2956, opened 2026-08-18
ci: fix remaining zizmor warnings / issues

ricardosalveti:
The org-required "GitHub Actions Security Scan (zizmor)" check currently fails on every pull request due to pre-existing high-severity findings in .github/ (#2901 fixed the github-env set; this series clears the rest).

- template-injection: expand step/job outputs via step env: blocks instead of interpolating ${{ }} into run scripts (actions/compile, build-yocto.yml, compile.yml, test.yml). The build-url artifact consumed by LAVA is unchanged.
- excessive-permissions: drop workflow-level checks/pull-requests: write in pr.yml, push.yml, nightly-build.yml, test-pr*.yml; each job now gets only what its steps and called reusable workflows declare.
- github-app: scope the test-reporting app token in publish-results.yml with permission-* inputs instead of inheriting the full installation.
- dangerous-triggers: annotate the by-design pull_request_target (backport) and workflow_run (test chain) triggers with the reviewed inline ignore the enterprise policy documents, with justification comments.

The unused actions/compile is fixed in place rather than dropped (#2455): the fixed action will first be copied to meta-qcom-distro (update to meta-qcom-distro#417), and then removed from this repository.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,6 +1,10 @@
 name: Backport merged pull request
 on:
-  pull_request_target:
+  # pull_request_target is required here: the workflow must run with a write
+  # token after the pull request is closed. It only acts on merged pull
+  # requests (see the job condition) and checks out the base repository, so
+  # no pull-request-controlled code is ever executed.
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
     types: [closed]
 permissions:
   contents: write # so it can comment
--- a/.github/workflows/publish-results.yml
+++ b/.github/workflows/publish-results.yml
@@ -57,6 +57,12 @@ jobs:
         with:
           client-id: 2291458
           private-key: ${{ secrets.TEST_REPORTING_APP_TOKEN }}
+          # Scope the token to what publish-unit-test-result-action uses
+          # instead of inheriting every permission of the app installation
+          permission-checks: write
+          permission-pull-requests: write
+          permission-contents: read
+          permission-issues: read
 
       - name: Publish Test Results
         uses: EnricoMi/publish-unit-test-result-action@c950f6fb443cb5af20a377fd0dfaa78838901040 # v2.23.0
--- a/.github/actions/compile/action.yml
+++ b/.github/actions/compile/action.yml
@@ -147,12 +147,13 @@ runs:
         KERNEL_DIRNAME="${INPUTS_KERNEL_DIRNAME}"
         BUILDNAME="${INPUTS_MACHINE}_${INPUTS_DISTRO_NAME}${KERNEL_DIRNAME}"
         FILENAME="build-url_${BUILDNAME}"
-        echo "${{ steps.upload_s3_artifacts.outputs.url }}" > "${FILENAME}"
+        echo "${UPLOAD_URL}" > "${FILENAME}"
         echo "filename=${FILENAME}" >> $GITHUB_OUTPUT
       env:
         INPUTS_KERNEL_DIRNAME: ${{ inputs.kernel_dirname }}
         INPUTS_MACHINE: ${{ inputs.machine }}
         INPUTS_DISTRO_NAME: ${{ inputs.distro_name }}
+        UPLOAD_URL: ${{ steps.upload_s3_artifacts.outputs.url }}
     - name: Upload build URL
       uses: actions/upload-artifact@v6
       with:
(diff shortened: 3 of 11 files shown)


ricardosalveti:
Next running at https://github.com/qualcomm-linux/meta-qcom/actions/runs/32092517527

ricardosalveti:
https://github.com/qualcomm-linux/meta-qcom/pull/2455 should be merged first as well.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/b1189c65-dd7f-4e9f-ba0a-bc37bfedead2)**
---
**Test jobs for commit [`e469188`](https://github.com/qualcomm-linux/meta-qcom/commit/e469188b259ea957be28553fc9a2803482632224)**
---
Pass: 9 | Fail: 0 | Total: 9

ricardosalveti:
> #2455 should be merged first as well.

Instead of removing it directly, better to fix the remaining issue first, so we can then update https://github.com/qualcomm-linux/meta-qcom-distro/pull/417.

ricardosalveti:
> Next running at https://github.com/qualcomm-linux/meta-qcom/actions/runs/32092517527

Worked correctly (test failure is not related to the change), moving to ready.

lumag (approved):

anujm1 (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/c4576ac8-2114-4990-a88a-f8784de7d80a)**
---
**Test jobs for commit [`b6bb710`](https://github.com/qualcomm-linux/meta-qcom/commit/b6bb710741c53f667bbd238354f8fb03ba6f24b6)**
---
Pass: 26 | Fail: 0 | Total: 26
Pass: 9 | Fail: 0 | Total: 9

.github/workflows/publish-results.yml
```diff
@@ -57,6 +57,12 @@ jobs:
         with:
           client-id: 2291458
           private-key: ${{ secrets.TEST_REPORTING_APP_TOKEN }}
+          # Scope the token to what publish-unit-test-result-action uses
```
mwasilew, line 60: this might bite us when running for PR. I'm not sure these permissions cover commenting on a PR.

.github/workflows/backport.yml
```diff
@@ -1,6 +1,10 @@
 name: Backport merged pull request
 on:
-  pull_request_target:
+  # pull_request_target is required here: the workflow must run with a write
```
mwasilew, line 3: This is silly. What is stopping anyone from using a similar technique on actually unsecure code?

ricardosalveti: Sorry, you mean because we're ignoring it?

The action itself requires it, we would have to switch to a github app to not have to rely on pull_request_target here.

.github/workflows/publish-results.yml
ricardosalveti: In this PR itself it worked ok, but not sure if it would potentially change the behavior after the merged, but we can fix that if needed as a follow up.

.github/workflows/backport.yml
mwasilew: The use of pull_request_target here is as recommended by github. I don't understand why zizmor doesn't recognize it. Adding a comment for zizmor to ignore the issue is something I have problem with. I don't see any mechanism stopping an attacker from using the same trick to mask any other warning.

.github/workflows/publish-results.yml
mwasilew: To test this you need to merge it (somewhere) and then create PR. The PR should use the merged change for running tests and publishing results. As far as I know this can't be tested without merging it first. It has to be merged to the "main" repository branch. Github takes the workflows from "main" branch when running workflows against PR. So for this PR the workflow version was taken from `master`. If you push this change to `next` you will test the `push` trigger. There is no easy mechanism to test `pull_request` trigger.

.github/workflows/backport.yml
ricardosalveti: Oh sure, this is by "design".

ricardosalveti:
Will merge and perform the final validation on a PR, as we can't validate it fully here.

PR #2962, opened 2026-08-18
Add eDP display overlays for hamoa and purwa IoT EVKs

anan-tiwa:
Both Hamoa and Purwa IoT EVKs are available in headed and headless
variants. The headed variants use an eDP display panel connected via
the DP3 controller, while the headless variants do not include a
display.

The existing DT is causing the eDP-panel driver to probe an unavailable
panel, resulting in AUX probe timeouts and missing display modes.
Userspace then attempts display commits through the eDP connector,
causing repeated DPU vblank, frame-done, and IRQ timeout errors.

Move the eDP-specific nodes out of the base DTS files into separate
DTBO overlays for each board. This allows the base DTBs to be used
as-is for headless variants, while the composite DTBs enable eDP
support on headed variants.

CRs-Fixed: 4543794, 4543797


ricardosalveti:
Kernel changes should go first to the kernel tree instead.

lumag (changes requested):
Get kernel changes merged in qcom-6.18 / qcom-next, then rebase this PR once corresponding changes are available inside meta-qcom

PR #2971, opened 2026-08-19
ci: replace the monthly sstate rotation with an mtime-based cleanup

ricardosalveti:
Rotating the sstate cache into an empty per-month folder makes the first run of every month build the whole matrix cold. The rotation only exists as a cleanup mechanism, and bitbake already provides a better signal: it refreshes the mtime of every sstate file a build uses — the availability check, the setscene unpack and the re-publish of an existing package all touch the object and its .sig and .siginfo (see sstate.bbclass) — so a file whose mtime is older than N days was not used by any build for N days.

Switch to a single shared sstate-cache folder and expire unused files with a weekly cleanup workflow (find -mtime +15 plus empty directory pruning), replacing the monthly job. The 15-day window is at least twice the cadence of the broadest scheduled builds, so an object is only deleted after two consecutive weekly passes missed it, and a build racing the deletion just rebuilds the task from an ordinary setscene miss.

Assisted-by: Claude Code:claude-fable-5


lumag (dismissed):

mwasilew (dismissed):

.github/workflows/sstate-cleanup.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  sstate-cache-cleanup:
+    if: github.repository_owner == 'qualcomm-linux'
+    runs-on: [self-hosted, qcom-sm-u2404, sm-amd64]
+    timeout-minutes: 720
+    steps:
+      - name: Expire unused files from the shared sstate-cache dir
+        run: |
+          # bitbake refreshes the mtime of every sstate file a build uses:
+          # the availability check, the setscene unpack and the publish of an
```
ndechesne: this is very verbose here. the first sentence is key/important and something we missed months ago. the other lines bring very less value.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/c5079b75-eea2-4f46-b217-f14958fa51fc)**
---
**Test jobs for commit [`666511a`](https://github.com/qualcomm-linux/meta-qcom/commit/666511ad9f3374ca2506256bd544ac5ad7a2bfb0)**
---
Pass: 9 | Fail: 0 | Total: 9

.github/workflows/sstate-cleanup.yml
ricardosalveti: Shorter now.

ndechesne (approved):

ricardosalveti:
Heads up that this will have to be merged with quic-yocto-ci, I will take care of that.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/be79559e-38b1-4ce8-a3f4-f300e2c83db3)**
---
**Test jobs for commit [`7593f5b`](https://github.com/qualcomm-linux/meta-qcom/commit/7593f5bac44d9c3b248a53c43d9fcd6c6d8b2dc0)**
---
Pass: 9 | Fail: 0 | Total: 9

lumag (approved):

.github/workflows/sstate-cleanup.yml
```diff
@@ -0,0 +1,21 @@
...
+    if: github.repository_owner == 'qualcomm-linux'
+    runs-on: [self-hosted, qcom-sm-u2404, sm-amd64]
+    timeout-minutes: 720
+    steps:
+      - name: Expire unused files from the shared sstate-cache dir
+        run: |
+          # bitbake refreshes the mtime of every sstate file a build uses, so
+          # a file older than 15 days was not used by any build in 15 days
+          find /efs/qli/meta-qcom/sstate-cache -type f -mtime +15 -delete
+          find /efs/qli/meta-qcom/sstate-cache -mindepth 1 -type d -empty -delete
```
quaresmajose, line 21: I have a WIP branch with this implementation also in my side. I haven't proposed it yet because I haven't resolved the concurrency job issues which also exists in this implementation. 

It's possible that sstate-cache files may disappear when builds are running. My solution consists of creating an action helper to execute all kas commands and then add a `concurrency groups` on this action.

ricardosalveti: What would be the issue if the sstate cache file gets removed during the build process, won't it just generate a new one anyway?

Also, we have daily / nightly runs for both branches, so I don't expect necessarily for things to be used in parallel with the removal process, as the cache will be old enough for files to be removed without impacting anyone.

quaresmajose: Yes, that's true. Perhaps the probability and possibility of that happening is almost zero.

PR #3000, opened 2026-08-25
Split compile matrix into two batches

vkraleti:
The compile job's reusable-workflow strategy now produces more configurations than the 256-configuration limit for a single matrix. Split it into compile_batch1 and compile_batch2 by machine, since every exclude/include entry pins to one machine, so partitioning preserves exact build coverage across both batches.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -220,7 +220,11 @@ jobs:
       kernel_dirname: ''
       cache_dir: ${{ needs.compile-env.outputs.CACHE_DIR }}
 
-  compile:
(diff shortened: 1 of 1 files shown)


qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/70756790-5b54-4aaf-86c3-8224435fa1a8)**
---
**Test jobs for commit [`608b7b6`](https://github.com/qualcomm-linux/meta-qcom/commit/608b7b61c9feb4852648e1e320d865ec75e5f337)**
---
Pass: 9 | Fail: 0 | Total: 9

PR #3019, opened 2026-08-26
image_types_qcom: cut the cost of building the qcomflash tarball

ricardosalveti:
create_qcomflash_pkg assembles the flashable package in a staging
directory that exists only so it can be tarred. It installs the rootfs
image into that directory, copying 4-9 GB that nothing ever modifies,
and then compresses the tarball at pigz -9, which spends about three
times the CPU of -6 to produce a tarball half a percent smaller. Both
costs are paid once per image, so a qcom-distro-catchall build pays them
three times.

Hard link the rootfs image instead of copying it, and lower the
compression level to -6.

Measured on a qcs9100-ride-sx qcom-distro-catchall build over sixteen
threads, do_image_qcomflash drops from 1,701 to 707 CPU seconds and from
21.9 to 7.1 GB of writes, while the tarballs grow from 5,226 to 5,251
MiB in total.

### DIFF
--- a/classes-recipe/image_types_qcom.bbclass
+++ b/classes-recipe/image_types_qcom.bbclass
@@ -201,7 +201,7 @@ create_qcomflash_pkg() {
     ln -rsf ${QCOMFLASH_DIR} ${IMGDEPLOYDIR}/${IMAGE_LINK_NAME}.qcomflash
 
     # Create qcomflash tarball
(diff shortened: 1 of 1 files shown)


koenkooi (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/265ca297-b4dd-4701-81ae-435915052862)**
---
**Test jobs for commit [`bcd43dd`](https://github.com/qualcomm-linux/meta-qcom/commit/bcd43dd4107c1c7114a468271dfb6edd21a7a599)**
---
Pass: 9 | Fail: 0 | Total: 9
Pass: 21 | Fail: 3 | Total: 24

lumag (approved):

PR #3026, opened 2026-08-27
userspace-resource-manager: Update to v0.4.9

kartnema:
userspace-resource-manager: Update to v0.4.9

CHANGES in v0.4.9
- URM Executable moved to /usr/sbin
- post-boot configs moved from /etc/urm/initscripts to /usr/libexec/initscripts
- Test configs moved to /usr/share/urm/tests, instead of /etc/urm/tests
- Additional Configs for Seraph
- Added the retuneSignal API, to extend duration of existing signal configurations
- Removed the BUILD_EXTENSIONS CMake flag, to ensure the extensions
  package is always generated alongside URM, users retain the flexibility
  to install / remove the extensions package according to their requirements.
- Update project licensing from BSD-3-Clause-Clear to BSD-3-Clause
- Updated URM test framework, so that the test nodes are correctly placed in
  a rw directory prior to testing. This is achieved by using the mktemp system
  call in the test runner and passing the path returned by the system call to
  the URM test binaries UrmIntegrationTests and UrmComponentTests, appropriate
  CLI handling options have been added to the above binaries.

### DIFF
--- a/recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.9.bb
+++ b/recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.9.bb
@@ -4,11 +4,11 @@ DESCRIPTION = "Userspace Resource Manager(URM) is a lightweight userspace \
 daemon that monitors system resources and enforces policies using \
 Linux kernel interfaces such as cgroups and sysfs."
 
-LICENSE = "BSD-3-Clause-Clear"
-LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=2998c54c288b081076c9af987bdf4838"
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://LICENSE.txt;md5=9701d0ef17353f1d05d7b74c8712ebbd"
 
 SRC_URI = "git://github.com/qualcomm/userspace-resource-manager.git;protocol=https;branch=main;tag=v${PV}"
-SRCREV = "58375d896a828e47504b2770131a8357e1341dd1"
+SRCREV = "21380f4c0150d2df2eef2e1e45338fbe9438a1a4"
 
 inherit cmake pkgconfig systemd
 
@@ -16,20 +16,18 @@ DEPENDS += "libyaml"
 
 PACKAGECONFIG ??= "\
     classifier \
-    extensions \
     ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'state-detector systemd', '', d)} \
     tests \
 "
 
 PACKAGECONFIG[classifier] = "-DBUILD_CLASSIFIER=ON,-DBUILD_CLASSIFIER=OFF"
-PACKAGECONFIG[extensions] = "-DBUILD_EXTENSIONS=ON,-DBUILD_EXTENSIONS=OFF"
 PACKAGECONFIG[state-detector] = "-DBUILD_STATE_DETECTOR=ON,-DBUILD_STATE_DETECTOR=OFF"
 PACKAGECONFIG[systemd] = ",,systemd"
 PACKAGECONFIG[tests] = "-DBUILD_TESTS=ON,-DBUILD_TESTS=OFF"
 
 SYSTEMD_SERVICE:${PN} = "urm.service"
 FILES:${PN}-dev += " \
-    ${libdir}/urm/libUrmTestPlugin.so \    
+    ${libdir}/urm/libUrmTestPlugin.so \
     ${libdir}/urm/libUrmPlugin.so \
 "
 
@@ -38,13 +36,14 @@ FILES:${PN} += "${sysconfdir}/urm/*"
 PACKAGE_BEFORE_PN += "${PN}-extensions"
 FILES:${PN}-extensions += " \
     ${sysconfdir}/urm/target/* \
-    ${sysconfdir}/urm/initscripts/post_boot/* \
+    ${libexecdir}/urm/initscripts/post_boot/* \
     ${libdir}/urm/libUrmPlugin.so* \
 "
 
 PACKAGE_BEFORE_PN += "${PN}-tests"
 FILES:${PN}-tests += " \
-    ${sysconfdir}/urm/tests/* \
+    ${datadir}/urm/tests/configs/* \
+    ${datadir}/urm/tests/nodes/* \
     ${bindir}/UrmComponentTests \
     ${bindir}/UrmIntegrationTests \
     ${libdir}/libRestuneTestUtils.so* \


qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/e91c139e-50f8-40a0-abb3-a4417532ac45)**
---
**Test jobs for commit [`218ae37`](https://github.com/qualcomm-linux/meta-qcom/commit/218ae3793fa0df5639499ce16add143994e70a72)**
---
Pass: 9 | Fail: 0 | Total: 9

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/d23ce2bc-995d-44f0-bd46-d79d2a5faa4c)**
---
**Test jobs for commit [`e07c6fe`](https://github.com/qualcomm-linux/meta-qcom/commit/e07c6fe0d2525250aeea85d9c561f4dea3e67415)**
---
Pass: 9 | Fail: 0 | Total: 9

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.9.bb
```diff
@@ -8,28 +8,26 @@ LICENSE = "BSD-3-Clause-Clear"
...
 
 PACKAGECONFIG ??= "\
     classifier \
-    extensions \
     ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'state-detector systemd', '', d)} \
     tests \
 "
 
 PACKAGECONFIG[classifier] = "-DBUILD_CLASSIFIER=ON,-DBUILD_CLASSIFIER=OFF"
-PACKAGECONFIG[extensions] = "-DBUILD_EXTENSIONS=ON,-DBUILD_EXTENSIONS=OFF"
```
ricardosalveti, line 25: You removed this but it is not in your commit message, please explain why.

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.7.bb
```diff
@@ -38,13 +36,14 @@ FILES:${PN} += "${sysconfdir}/urm/*"
...
-    ${sysconfdir}/urm/initscripts/post_boot/* \
+    ${libexecdir}/urm/initscripts/post_boot/* \
     ${libdir}/urm/libUrmPlugin.so* \
 "
 
 PACKAGE_BEFORE_PN += "${PN}-tests"
 FILES:${PN}-tests += " \
-    ${sysconfdir}/urm/tests/* \
+    ${datadir}/urm/tests/configs/* \
+    ${localstatedir}/lib/urm/tests/nodes/* \
```
ricardosalveti: This won't work for SOTA builds (with ostree), since it won't really be saved.

Also, you should not really package contents in /var.

ricardosalveti (changes requested):
> - Test configs moved to /usr/share and /var/lib, instead of /etc/urm/tests

This breaks SOTA, you should not install & package content in /var.

kartnema:
Hi @ricardosalveti

The files installed in /var/lib are test assets used to simulate sysfs nodes and are only installed when the optional test package is explicitly installed by the user. They are not runtime-generated state.

However, if the requirement is to not maintain any files in /var/lib we can potentially move them to /usr/share

lumag:
> However, if the requirement is to not maintain any files in /var/lib we can potentially move them to /usr/share

Yes, please

ricardosalveti:
> The files installed in /var/lib are test assets used to simulate sysfs nodes and are only installed when the optional test package is explicitly installed by the user. They are not runtime-generated state.
> 
> However, if the requirement is to not maintain any files in /var/lib we can potentially move them to /usr/share

Yes, anything that can be installed should not land in /var, move to /usr/share.

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.8.bb
```diff
@@ -5,31 +5,29 @@ daemon that monitors system resources and enforces policies using \
 Linux kernel interfaces such as cgroups and sysfs."
 
 LICENSE = "BSD-3-Clause-Clear"
```
lumag: So still -Clear?

lumag (changes requested):

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.7.bb
ricardosalveti:
```
The test sources and test YAML still hard-code /var/lib/urm/tests/nodes:

┌───────────────────────────────────────────┬────────────┐
│                   File                    │ References │
├───────────────────────────────────────────┼────────────┤
│ tests/Integration/IntegrationTests.cpp    │ 54         │
├───────────────────────────────────────────┼────────────┤
│ tests/Configs/ResourcesConfig.yaml        │ 12         │
├───────────────────────────────────────────┼────────────┤
│ tests/Integration/StateDetectionTests.cpp │ 4          │
├───────────────────────────────────────────┼────────────┤
│ tests/Configs/ResourcesConfigAddOn.yaml   │ 3          │
├───────────────────────────────────────────┼────────────┤
│ tests/Component/ParserTests.cpp           │ 3          │
├───────────────────────────────────────────┼────────────┤
│ tests/Integration/CCIntegrationTests.cpp  │ 2          │
└───────────────────────────────────────────┴────────────┘
```

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/fc27f511-a111-480c-98e9-66b740499465)**
---
**Test jobs for commit [`9668678`](https://github.com/qualcomm-linux/meta-qcom/commit/9668678f11424458d1fd5ca3e76e095c809184ca)**
---
Pass: 34 | Fail: 0 | Total: 34
Pass: 9 | Fail: 0 | Total: 9

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.7.bb
kartnema: Yes, the reason is that the URM test-runner is expected to stage (copy) all of these nodes to /var/lib/urm/tests prior to starting the test-cases and remove them once done. Hence tests will see the nodes in /var/lib/urm/tests/nodes itself.

We are doing this, because the files are being written to during the course of the tests, since /usr is supposed to be for read-only files, hence a better location would be /var (as direct installation to /var creates issues with SOTA, we came up with the above workaround).

ricardosalveti: Then /tmp/ or /var/tmp/ should be used instead, as they are removed once done.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/a8d5f712-1d79-4e4b-926b-294c90904607)**
---
**Test jobs for commit [`cd32566`](https://github.com/qualcomm-linux/meta-qcom/commit/cd32566dec85517d8fc67b9d1786ba4e0692f62b)**
---
Pass: 9 | Fail: 0 | Total: 9

recipes-support/userspace-resource-manager/userspace-resource-manager_0.4.7.bb
lumag: Neither. It should use `mktemp()` to create a temporary directory.

kartnema: The tests currently rely on a fixed path for the nodes, so switching to a directory created via mkdtemp() would require making the path configurable and updating the tests. As an alternative, would /run/urm/tests be a suitable location? The data is runtime-only, writable during test execution, and does not need to persist across reboots.

lumag: Does that require a user-writeable directory? Why do we allow users to influence tests by using the satic directory?

kartnema: Yes, because as part of some of the test-case executions we write values to those nodes for certain durations and eventually reset them. Hence we are temporarily copying the files from a read-only dir to either /tmp or /run. Once tests are completed they can be removed.

I am not entirely sure about the ask in the second part of the question, these nodes are just text files which serve no greater purpose and only created when the URM tests package is installed, we don't expect users to access or directly modiy them.

lumag: Okay. No, use `mktemp()`, don't use a predefined directory for tests.

kartnema: Changes are completed, using mktemp() now

lumag (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/292a7ec7-8637-48d7-adcf-3a434b1eab51)**
---
**Test jobs for commit [`0e6c2b5`](https://github.com/qualcomm-linux/meta-qcom/commit/0e6c2b5ef1178307a7b64bf7c87cdac5d93bb7e0)**
---
Pass: 9 | Fail: 0 | Total: 9

kartnema:
Hi @ricardosalveti kindly request to review the PR

ricardosalveti (approved):

lumag:
KNown failures.

PR #3029, opened 2026-08-28
ci/performance.yml: skip GPT auto-discovery for non-SOTA builds

wenwfu:
systemd-gpt-auto-generator probes the partition table during early boot even when the root device is explicitly specified. Skip this probing in performance builds to reduce generator time.

SOTA relies on the generator to create boot.mount and boot.automount for OSTree services that access /boot. Retain GPT discovery for SOTA builds.

### DIFF
--- a/ci/performance.yml
+++ b/ci/performance.yml
@@ -3,4 +3,4 @@ header:
 
 local_conf_header:
   cmdline-perf: |
-    KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=linux"
+    KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=linux ${@bb.utils.contains('DISTRO_FEATURES', 'sota', '', 'systemd.gpt_auto=no', d)}"


qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/e6f5841d-7fdc-46c9-8ab8-fbaa9f3e253f)**
---
**Test jobs for commit [`9cbd101`](https://github.com/qualcomm-linux/meta-qcom/commit/9cbd10144196589cb5fbbea4535be920e63a0106)**
---
Pass: 9 | Fail: 0 | Total: 9

lumag (commented):
> Skip
this probing in performance builds to reduce generator time.

What for? Also, what does it have to do with the BSP layer?

wenwfu:
> > Skip
> > this probing in performance builds to reduce generator time.
> 
> What for? Also, what does it have to do with the BSP layer?

Thanks for the review, Dmitry.

**What for**: to reduce systemd init time. systemd-gpt-auto-generator scans the GPT table on every boot even when root= is already explicit on the cmdline — that scan is pure overhead when nothing consumes what it produces. On non-SOTA builds, nothing does: there's no static fstab entry for /boot, and no service reads it. SOTA builds are different — ostree-finalize-staged-hold.service/ostree-boot-complete.service need /boot mounted, and only this generator provides that (via synthesized boot.mount), so it must keep running there.

Benefit: On rb3gen2-core-kit (non-SOTA) with QLI.2.0 tag-wrynose-260627.1q
  - Generators phase: 55.9ms → 35.6ms (-20ms, the generator's own probing cost)
  - Units-load phase: 320.0ms → 303.9ms (-16ms, since boot.mount/boot.automount are no longer synthesized and don't need to be loaded)

~36ms total off the systemd init window, for a build where nothing consumes the synthesized boot.mount anyway. 

**Why BSP layer**: systemd.gpt_auto=no is a stock systemd switch, no code change needed. What's BSP-specific is whether it's safe — SOTA builds on this platform depend on this generator synthesizing boot.mount for  ostree-finalize-staged-hold.service/ostree-boot-complete.service to reach /boot (no static fstab entry as fallback), so the flag is gated on sota being absent from DISTRO_FEATURES.

ci/performance.yml
```diff
@@ -3,4 +3,4 @@ header:
 
 local_conf_header:
   cmdline-perf: |
-    KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=linux"
+    KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=linux ${@bb.utils.contains('DISTRO_FEATURES', 'sota', '', 'systemd.gpt_auto=no', d)}"
```
quaresmajose, line 6: sota is a distrooverride variable so we can also use like above, which, in my opinion, is easier to read.
```
KERNEL_CMDLINE_EXTRA:append:sota = " systemd.gpt_auto=no"
```

ricardosalveti:
I prefer not disabling auto discovery and use it instead, see https://github.com/qualcomm-linux/meta-qcom-distro/pull/349.

auto discovery can and should also be used to automatically mount the efi partition, not only root.

lumag:
@wenwfu this kind of optimisations (as you notices) are distro-specific. As such, they don't belong to the BSP layer. If you want to perform it, it should be done from the qcom-disto layer. I'm going to close this PR.

PR #3042, opened 2026-08-31
Known failures

mwasilew:
In attempt to make failed test signal significant, this series adds support for "known issues" file. The file is kept in the repository that uses test-distro.yml reusable workflow. Know issues are evaluated at the reporting stage. Raw results can still be retrieved from LAVA logs.

### DIFF
--- a/.github/actions/test-job-summary/action.yml
+++ b/.github/actions/test-job-summary/action.yml
@@ -15,6 +15,13 @@ inputs:
   summary_title:
     description: "Title for the summary part of the report"
     required: false
+  known_failures_file:
+    description: >
+      Path to the YAML file listing the known failures of the tested build
+      variant. Relative to the workspace. Missing file means no known failure.
+      See .github/known-failures/README.md for the file format.
+    required: false
+    default: ""
 outputs:
   artifact_id:
     description: "ID of the uploaded artifact"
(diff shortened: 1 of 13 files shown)


qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/ca43e24b-2223-4171-ad8f-77cdbffdfa83)**
---
**Test jobs for commit [`7f73080`](https://github.com/qualcomm-linux/meta-qcom/commit/7f730808be6a979829e9d684ed2ade6d7753f122)**
---
Pass: 9 | Fail: 0 | Total: 9

lumag:
I guess the main question from my side is if the known failures are just taken from the master branch or if they come from the PR merged tree. We really need the latter case, otherwise it will be a nightmare to handle them.

mwasilew:
> I guess the main question from my side is if the known failures are just taken from the master branch or if they come from the PR merged tree. We really need the latter case, otherwise it will be a nightmare to handle them.

I assume you're trying to cover the situation when PR fixes a bug. In such case the list of known issues should also be updated to keep the CI "green". I don't think it's the case with this PR. This approach makes sense. Let me try to implement it.

mwasilew:
This PR is obsolete because of #3137

PR #3055, opened 2026-09-02
ci/performance: skip console DSR terminal-size probing

wenwfu:
Set systemd.tty.term.console=dumb to skip systemd's early ANSI DSR-based terminal-size probing on the console. This avoids the cursor-position query and response wait in terminal_get_size_by_dsr(), reducing PID 1 early initialization time on slow serial consoles. 

The setting only affects systemd's console terminal handling. It does not change the TERM value of the ttyMSM0 serial login session, which continues to use its default terminal type.

Three measurements of this window on `rb3gen2-core-kit` with the `QLI.2.0 wrynose-260627.1` catchall build show:
- Default: `1.259 s`, `1.222 s`, `1.202 s` (average `1.228 s`)
- With this change: `0.850 s`, `0.876 s`, `0.805 s` (average `0.844 s`)

This saves `384 ms (31.3%)` on average, letting PID 1 proceed to manager startup and unit loading sooner so userspace services can start earlier.

### DIFF
--- a/ci/performance.yml
+++ b/ci/performance.yml
@@ -3,4 +3,4 @@ header:
 
 local_conf_header:
   cmdline-perf: |
-    KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=linux initramfs.udev-root-only=1"
+    KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=dumb initramfs.udev-root-only=1"


qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/9092c2f9-e375-453d-966f-4821172065cf)**
---
**Test jobs for commit [`b2bda09`](https://github.com/qualcomm-linux/meta-qcom/commit/b2bda09798ba6dc1adf8c803dd00ecfeee23ddb6)**
---
Pass: 9 | Fail: 0 | Total: 9

ci/performance.yml
```diff
@@ -3,5 +3,5 @@ header:
 
 local_conf_header:
   cmdline-perf: |
-    KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=linux"
+    KERNEL_CMDLINE_EXTRA:append = " quiet systemd.tty.term.console=dumb"
```
ricardosalveti: This will also affect the login shell, no?

wenwfu: Yes, that is expected. It may affect terminal features such as colors, cursor control, and screen clearing, but this configuration is only used for CI boot-performance measurements. The impact on the interactive login shell is acceptable in this context.

lumag: Will it prevent vi/ vim from working on the serial console? If so, no, it's not acceptable.

wenwfu: I tested this through picocom. With TERM=dumb, BusyBox vi starts but full-screen handling is degraded and :wq did not work reliably over the serial console. Since this affects interactive serial-console usage, we should avoid setting systemd.tty.term.console=dumb globally and restrict it to the CI performance test environment.

wenwfu: I confirmed that TERM=dumb affects full-screen applications such as vi/vim on the serial console.
One possible solution is: `systemd.tty.term.console=dumb systemd.setenv=TERM=linux`

This keeps TERM=dumb for PID 1, avoiding the synchronous ANSI reset during early initialization, while restoring TERM=linux for spawned services and the serial login session. The downside is that systemd.setenv=TERM=linux applies to all system services, not only the serial getty. Since this configuration is limited to the CI performance image, that may be acceptable, but I would like to get your opinion before updating the patch. 
A cleaner long-term solution would require systemd to separate early console ANSI-reset policy from the TERM value exposed to login sessions.

wenwfu:
Heads-up on a side effect and its upstream fix:
Setting `systemd.tty.term.console=dumb` makes PID 1 run with `TERM=dumb`, and any service attached to `/dev/console` (e.g. getties) then inherits dumb as well — even when the console is backed by a concrete device like `ttyS0` for which one would want a richer TERM. Full-screen apps such as vi/vim degrade in that case.                                                                                                                                        
                                                                                                                                                                                                                                   I've raised an upstream systemd PR to address this: https://github.com/systemd/systemd/pull/43643 — it resolves `/dev/console` to its backing device and lets an explicit `systemd.tty.term.<device>= (e.g. systemd.tty.term.ttyS0=linux)` take precedence, so you can keep the fast console=dumb boot path while still giving login shells a usable terminal type. It's under review/discussion.

wenwfu:
Updated the implementation based on the review feedback. The new approach:
- keeps PID 1 on `TERM=dumb` to avoid the early console ANSI reset;
- sets serial getty sessions to `TERM=linux` through a systemd drop-in;


could @lumag @ricardosalveti please take another look?

wenwfu:
Use systemd.tty.term.console=dumb to avoid ANSI terminal reset during early systemd initialization on slow serial consoles. Set systemd.tty.term.ttyMSM0=linux so serial login shells retain the normal terminal type required by applications such as vi and vim.

lumag (changes requested):
> delay boot on slow serial consoles

So, what is the gain?

wenwfu:
> > delay boot on slow serial consoles
> 
> So, what is the gain?

The gain is a reduction in systemd's early initialization window on slow serial consoles.

Three measurements of this window on `rb3gen2-core-kit` with the `QLI.2.0 wrynose-260627.1` catchall build show:
- Default: `1.259 s`, `1.222 s`, `1.202 s` (average `1.228 s`)
- With this change: `0.850 s`, `0.876 s`, `0.805 s` (average `0.844 s`)

This saves 384 ms (31.3%) on average

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/57a792ce-ac59-43f6-a250-35f446356cc7)**
---
**Test jobs for commit [`4478724`](https://github.com/qualcomm-linux/meta-qcom/commit/44787243b894fe363699bf888124e81566cf54e4)**
---
Pass: 9 | Fail: 0 | Total: 9

ci/performance.yml
ricardosalveti: Maybe worth trying a systemd patch instead?

wenwfu: The combination of

- systemd.tty.term.console=dumb systemd.tty.term.ttyMSM0=linux

already satisfies the required behavior using existing upstream systemd kernel command-line options. The first option keeps PID 1's console TERM as `dumb` during early initialization, while the second restores `TERM=linux` for the `ttyMSM0` serial login session.

Therefore, no additional upstream systemd patch is required

lumag:
> > > delay boot on slow serial consoles
> > 
> > 
> > So, what is the gain?
> 
> The gain is a reduction in systemd's early initialization window on slow serial consoles.
> 
> Three measurements of this window on `rb3gen2-core-kit` with the `QLI.2.0 wrynose-260627.1` catchall build show:
> 
> * Default: `1.259 s`, `1.222 s`, `1.202 s` (average `1.228 s`)
> * With this change: `0.850 s`, `0.876 s`, `0.805 s` (average `0.844 s`)
> 
> This saves 384 ms (31.3%) on average

Is it captured in the commit message? See CONTRIBUTING.md

lumag (changes requested):
SoB should be the last tag. No empty lines between tags.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/8526588b-d390-41ef-9867-b5debca9a475)**
---
**Test jobs for commit [`541b835`](https://github.com/qualcomm-linux/meta-qcom/commit/541b8353c844332209aa85860f97cffc010681b0)**
---
Pass: 9 | Fail: 0 | Total: 9

ci/performance.yml
ricardosalveti: You are just fixing 'ttyMSM0' here, that is why my ask to see if it is worth trying to find a more generic solution upstream that could even potentially work without any specific command line here.

wenwfu: Yes, I already explored this direction in systemd PR #43643: https://github.com/systemd/systemd/pull/43643

That PR tried to resolve /dev/console to its effective backing device and apply a per-device TERM setting. It was closed because the maintainers were not convinced about the use case and questioned changing the semantics and precedence of systemd.tty.term.console=.

Therefore, there is currently no accepted generic systemd solution for separating the TERM used by PID 1 console output from the TERM inherited by the serial getty.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/9b9070e8-0715-42a6-8598-f7159ca65e20)**
---
**Test jobs for commit [`3b1ff61`](https://github.com/qualcomm-linux/meta-qcom/commit/3b1ff61de5cf103d4aef5d0d0a7926cdec324a05)**
---
Pass: 9 | Fail: 0 | Total: 9

lumag (commented):
Quoting Lennart's review:
> hmm, what? you want that your gettys and pid1 have different terminal settings even if they ultimately use the same effective device (i.e. after resolution)? And you want to use systemd.tty.term.* for making this arbitration?
> 
> sorry, but i am not convinced this is the right away. I am pretty sure that systemd.tty.term.* should be honoured the same way, regardless which program processes, i.e. if it's for the getty service or for pid 1 itself.

From my point of view this means several things.
- The console and tty* are expected to have the same settings
- This change doesn't belong to meta-qcom (it's a generic BSP layer)
- Including this change in the meta-qcom-bsp would be questionable too
- The PR fails to explain, why terminal reset takes almost 0.4s.

ci/performance.yml
wenwfu: > Will it prevent vi/ vim from working on the serial console? If so, no, it's not acceptable.

I reconfirmed this on the serial console : with only `systemd.tty.term.console=dumb`, the login session still has `TERM=vt220`  and vi works normally. I will remove the unnecessary  `systemd.tty.term.ttyMSM0=linux` override.

wenwfu:
> Quoting Lennart's review:
> 
> > hmm, what? you want that your gettys and pid1 have different terminal settings even if they ultimately use the same effective device (i.e. after resolution)? And you want to use systemd.tty.term.* for making this arbitration?
> > sorry, but i am not convinced this is the right away. I am pretty sure that systemd.tty.term.* should be honoured the same way, regardless which program processes, i.e. if it's for the getty service or for pid 1 itself.
> 
> From my point of view this means several things.
> 
> * The console and tty* are expected to have the same settings
> * This change doesn't belong to meta-qcom (it's a generic BSP layer)
> * Including this change in the meta-qcom-bsp would be questionable too
> * The PR fails to explain, why terminal reset takes almost 0.4s.

I rechecked this more closely. The measured delta comes from PID-1's terminal_get_size_by_dsr() path: unless the terminal is dumb, systemd sends DSR/CPR queries and a large cursor-position sequence, then waits for the terminal responses. Setting the console to dumb makes this path return early. I had incorrectly described this as a console reset.

Although the current serial getty still ends up with TERM=vt220, relying on different terminal settings for PID 1 and the resolved tty is not a valid use of systemd.tty.term.*. The ttyMSM0 override is therefore wrong.

wenwfu:
I have opened systemd/systemd#43735 to discuss the ANSI terminal-size probe timeout and the appropriate generic solution.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/f2758968-afc7-4b60-9c4c-e93023429955)**
---
**Test jobs for commit [`4f69ef8`](https://github.com/qualcomm-linux/meta-qcom/commit/4f69ef8ceaeb69f0ce07679b2fdbf301f3e72e89)**
---
Pass: 9 | Fail: 0 | Total: 9

lumag (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/e8c4a512-e330-4d78-b571-2f10c373f6f2)**
---
**Test jobs for commit [`a60563a`](https://github.com/qualcomm-linux/meta-qcom/commit/a60563a7eba7b2444570c3e247954f242d348210)**
---
Pass: 9 | Fail: 0 | Total: 9

ricardosalveti (approved):

ricardosalveti:
Known failures.

PR #3056, opened 2026-09-02
iq-615-evk: Enable m2connector MACHINE_FEATURE

DaizhuangBai:
Declare external M.2 connector capability for iq-615-evk.

This feature flag is used by image recipes to decide whether to pull optional M.2 wireless firmware packagegroups. Machines without this hardware should not install those add-on firmware packages.


DaizhuangBai:
some mismatch

PR #3068, opened 2026-09-02
image-adbd: install the USB gadget modules adbd needs

angolini:
When trying to install adbd to get serial console on Uno Q we faced a missing module on the target image.

With this it will work when the kernel has the modules built-in or when it's enabled as a module. But it still requires that the kernel enables it somehow (otherwise adbd will not work on the board).

### DIFF
--- a/classes-recipe/image-adbd.bbclass
+++ b/classes-recipe/image-adbd.bbclass
@@ -12,6 +12,8 @@ IMAGE_FEATURES[validitems] += "enable-adbd"
 
 PACKAGE_INSTALL:append = " ${@bb.utils.contains("BBFILE_COLLECTIONS", "openembedded-layer", "android-tools-adbd android-tools-adbd-cmdline", "", d)}"
 
+PACKAGE_INSTALL_ATTEMPTONLY:append = " ${@bb.utils.contains("BBFILE_COLLECTIONS", "openembedded-layer", "kernel-module-libcomposite kernel-module-usb-f-fs", "", d)}"
+
 enable_adbd_at_boot () {
     touch ${IMAGE_ROOTFS}/etc/usb-debugging-enabled
 }


qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/40b720c5-b4a2-47f0-a5bb-4a108689f7ee)**
---
**Test jobs for commit [`da06ddf`](https://github.com/qualcomm-linux/meta-qcom/commit/da06ddf18014ff095999b1461ace1d0158284da8)**
---
Pass: 9 | Fail: 0 | Total: 9

angolini:
Closing this in favour of fixing it in meta-oe directly, per @ricardosalveti's suggestion on qualcomm-linux/meta-qcom-3rdparty#61.

The dependency isn't Qualcomm-specific. `android-gadget-setup` modprobes `libcomposite` before it touches configfs, and creating `functions/ffs.usb0` pulls `usb_f_fs` in through `request_module`. Both live in meta-oe, and `android-tools-adbd` is the only consumer of the gadget, so the recommendation belongs on that package:

```
RRECOMMENDS:${PN}-adbd += "kernel-module-libcomposite kernel-module-usb-f-fs"
```

That covers any image installing adbd on any kernel or board, instead of only images using `image-adbd.bbclass`. Recommendations also tolerate a kernel that builds the gadget support in and therefore ships no `kernel-module-*` packages, which is the same tolerance `PACKAGE_INSTALL_ATTEMPTONLY` gave here.

Patch is prepared against meta-oe master for openembedded-devel.

Unrelated to the change: the Test Results failures here were lab issues, not this patch. They were qcs8300-ride-sx and sm8750-mtp boot/login timeouts, iq-8275-evk and rb3gen2-core-kit display-gfx (KMSCube, core_auth), and iq-9075-evk BT_SCAN. Every open PR in the repo was showing Test Results failures at the time.

PR #2057, opened 2026-04-28
rb1: switch to the new edk2-based downstream boot firmware release

ricardosalveti:
Include the new edk2-based qrb2210 boot firmware release (compatible with rb1 and uno-q) and switch rb1 to the new boot firmware implementation.

PR #3077, opened 2026-09-03
[Backport Wrynose] rb1: switch to the new edk2-based downstream boot firmware release

lumag:
Backport #2057 to wrynose


lumag:
Converted to draft, there seems to be issues with the new boot firmware.

PR #3083, opened 2026-09-06
android-tools-conf: allow pinning the adbd UDC

shawngsc:
On boards with two equally capable OTG ports, `android-gadget-start` has no
way to say which one adb should use. It ranks `/sys/class/udc` entries by
`maximum_speed` and breaks ties lexicographically, so on nord-ride-embedded —
where both controllers are `dr_mode = "otg"` behind a Type-C role switch and
both use a QMP USB3+DP PHY — `a400000.usb` (usb_1) always wins over
`a600000.usb` (usb_0). adb comes up on usb_1 and there is no supported way to
move it to usb_0.

This series adds an explicit override, preceded by two fixes to problems found
in the same script along the way.

- **android-tools-conf: honour ANDROID_GADGET_UDC_DELAY** — the script sleeps a
  hardcoded 10s, ignoring the `ANDROID_GADGET_UDC_DELAY=0` set by meta-oe's
  `10-adbd-configfs.conf`. Every adbd start pays 10s before adb is usable, and
  the drop-in's comment misleads anyone reading the unit. Default stays 10s.
- **android-tools-conf: exit instead of return on error** — three top-level
  error paths use `return 1`, but the script is executed as `ExecStartPost`,
  not sourced. In dash that is invalid, so configfs setup failures surface as a
  shell diagnostic instead of a clean service failure.
- **android-tools-conf: allow pinning the adbd UDC** — accept an explicit UDC
  from the `adbd.udc=<name>` kernel command line argument or `ADBD_UDC` in
  `/etc/default/android-tools-adbd`, command line winning. An unknown name logs
  a warning and falls back to the speed-based scan, so a typo or renamed node
  cannot leave a board without adb.

Boards that set neither override are unaffected.

### DIFF
--- a/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf/android-gadget-start
+++ b/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf/android-gadget-start
@@ -5,8 +5,12 @@ set -e
 # Wait before touching configfs, unless the unit tells us it is not needed.
 sleep "${ANDROID_GADGET_UDC_DELAY:-10}"
 
-# Speed-only selection:
-# 1) Pick UDC with highest maximum_speed
+# UDC selection:
+# 0) An explicit UDC always wins, taken from the 'adbd.udc=<name>' kernel
+#    command line argument. Needed on boards with several equally capable OTG
+#    ports, where the speed-based selection below cannot tell which one is
+#    wired for debug.
+# 1) Otherwise pick UDC with highest maximum_speed
 # 2) Tie-breaker: lexicographic order (first one found via sort)
 #
 # Recognized speeds:
@@ -51,21 +55,33 @@ read_max_speed() {
 best_udc=""
 best_max_rank=-1
 
-# Iterate all UDCs and select the best one by maximum_speed.
-for udc in $(ls -1 "$UDC_FOLDER" 2>/dev/null | sort); do
-    udc_path="$UDC_FOLDER/$udc"
-    [ -d "$udc_path" ] || continue
-
-    max_text="$(read_max_speed "$udc_path")"
-    max_rank="$(speed_rank "$max_text")"
-
-    # Initialize or update if we found a faster controller
-    # Strict inequality (>): maintains lexicographical order (first one wins) on ties
-    if [ -z "$best_udc" ] || [ "$max_rank" -gt "$best_max_rank" ]; then
-        best_udc="$udc"
-        best_max_rank="$max_rank"
+# An explicit UDC overrides the speed-based selection below.
+udc_override="$(sed -n -e '/adbd\.udc=/ s/.*adbd\.udc=\([^ ]*\).*/\1/p' /proc/cmdline)"
+if [ -n "$udc_override" ]; then
+    if [ -d "$UDC_FOLDER/$udc_override" ]; then
+        best_udc="$udc_override"
+    else
+        echo "Warning: requested UDC '$udc_override' not present, falling back to autodetection." >&2
     fi
-done
+fi
+
+# Iterate all UDCs and select the best one by maximum_speed.
+if [ -z "$best_udc" ]; then
+    for udc in $(ls -1 "$UDC_FOLDER" 2>/dev/null | sort); do
+        udc_path="$UDC_FOLDER/$udc"
+        [ -d "$udc_path" ] || continue
+
+        max_text="$(read_max_speed "$udc_path")"
+        max_rank="$(speed_rank "$max_text")"
+
+        # Initialize or update if we found a faster controller
+        # Strict inequality (>): maintains lexicographical order (first one wins) on ties
+        if [ -z "$best_udc" ] || [ "$max_rank" -gt "$best_max_rank" ]; then
+            best_udc="$udc"
+            best_max_rank="$max_rank"
+        fi
+    done
+fi
 
 if [ -z "$best_udc" ]; then
     echo "Error: No UDC controller found in $UDC_FOLDER" >&2
--- a/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf/android-gadget-start
+++ b/dynamic-layers/openembedded-layer/recipes-devtools/android-tools/android-tools-conf/android-gadget-start
@@ -2,7 +2,8 @@
 
 set -e
 
-sleep 10
+# Wait before touching configfs, unless the unit tells us it is not needed.
+sleep "${ANDROID_GADGET_UDC_DELAY:-10}"
 
 # Speed-only selection:
 # 1) Pick UDC with highest maximum_speed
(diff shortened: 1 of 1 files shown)


lumag (commented):
I think this was discussed several times. Please modify the adbd to bind to all suitable UDC devices.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/78a89267-9fa7-4f75-b336-38764df49b4c)**
---
**Test jobs for commit [`fa91be1`](https://github.com/qualcomm-linux/meta-qcom/commit/fa91be16b2d0ad14a34b0f962c90379f94578009)**
---
Pass: 9 | Fail: 0 | Total: 9

shawngsc:
> I think this was discussed several times. Please modify the adbd to bind to all suitable UDC devices.

I don't disagree that binding adb to every suitable UDC would be the nicer end state. But that isn't something `android-gadget-start` can do, and it's well beyond the scope of this change:

- `usb_gadget/adb/UDC` accepts exactly one UDC name, so N controllers means N gadget directories.
- A functionfs instance cannot be linked into two gadgets, so each gadget needs its own `ffs.adbN` and its own mount point.
- adbd serves a single hardcoded ffs path, so this requires modifying adbd itself plus templated units — that is meta-oe/android-tools work, not a BSP script change.
- Two instances would advertise the same `androidboot.serialno`, so the host would see duplicate serials; that needs per-instance serial handling before it is usable.
- Binding a gadget on every UDC is not free either: it claims peripheral role on that port, which conflicts with boards using the second port for host/DP-only or for a different gadget, and with Type-C role switching.

Redesigning adbd is not something I'm in a position to take on, and I don't think a board with two identical OTG ports should stay without a way to select the debug port until someone does. What this patch adds is an opt-in name check that boards setting nothing at all never see; if adbd later gains multi-UDC support, both the override and the speed-ranking heuristic it sits on top of can simply be removed.

Separately, patches 1 and 2 are independent bug fixes in the same script — `ANDROID_GADGET_UDC_DELAY` is ignored, so every adbd start pays a fixed 10s, and three top-level error paths use `return` in a script that is executed rather than sourced (invalid in dash). Can those land regardless of how the UDC-selection discussion goes? Happy to split them into their own PR if that helps.

lumag:
Maybe I missed something. Do you have two funcionally-equal USB ports with the script being unable to select between those? How would user select, which port to use?

shawngsc:
> Maybe I missed something. Do you have two funcionally-equal USB ports with the script being unable to select between those? How would user select, which port to use?

Yes, that's exactly the situation. On nord-ride-embedded both controllers are `dr_mode = "otg"` behind a Type-C role switch and both sit on a QMP USB3+DP PHY, so `/sys/class/udc` reports the same `maximum_speed` for `a600000.usb` (usb_0) and `a400000.usb` (usb_1). Nothing in the script can distinguish them, so the lexicographic tie-break decides and usb_1 silently wins.

On "how would the user select": they don't, and that's the point. The boot firmware already exposes **fastboot on usb_0**, so usb_0 is the port a developer has a cable in. adb landing on usb_1 means fastboot and adb are on different connectors on the same board, which is why we want adb on usb_0 too. That is a fixed property of the board and its firmware, not a per-plug decision — so there is nothing to auto-detect at runtime, only a fact to record.

Which raises the question of where to record it. `adbd.udc=` is convenient for testing, but since this is board description rather than a user preference, would you prefer it come from the machine conf — i.e. nord-ride-embedded ships an `ADBD_UDC=a600000.usb` fragment in `/etc/default/android-tools-adbd` — and the cmdline knob stays only as an override? The patch already supports both, so I'm happy to drop the cmdline path entirely if you think it invites misuse.

lumag:
I don't like encoding board-specifics into the config files: it suggest having a single-machine rootfs, while for meta-qcom and especially for QLI Debian images we try to endorse shared rootfs. What if we take a fancier step: declare the port with the `role-switch-default-mode = "peripheral"` and prefer those ports?

shawngsc:
> What if we take a fancier step: declare the port with the `role-switch-default-mode = "peripheral"` and prefer those ports?

The problem with marking usb_0 `peripheral` is that it doesn't change anything: an absent `role-switch-default-mode` already means peripheral, since dwc3 only special-cases `host` and falls back to device role otherwise. So usb_1 is peripheral-default too, and a "prefer peripheral-default ports" rule matches both controllers — leaving the same lexicographic tie-break that picks usb_1 today.

To make the preference actually discriminate, the property has to go on usb_1 as `host`, so I tested that on nord-ride-embedded, where both ports sit behind Type-C port controllers and switch role at runtime.

`role-switch-default-mode = "host"` on `&usb_1` does take effect while the port is idle — dwc3 comes up in host role, `a400000.usb` is absent from `/sys/class/udc`, and the existing scan picks usb_0. But as soon as a host PC is attached to usb_1, TCPM flips the port to device role, the UDC reappears, and adb binds to it again. The property only sets the idle role, so it cannot express which connector is the debug port on a board where both ports are genuine dual-role.

It also makes the outcome depend on what happens to be plugged in when adbd starts, which is worse than today: the current choice is stable and wrong, this one would be unstable. And on a symmetric dual-role port, declaring one side `host` states a policy preference as if it were a hardware fact.

Stepping back: "which port is the debug port" isn't a property of the SoC or the board. Both controllers work as adb and either would function. What singles out usb_0 is that the **boot firmware** exposes fastboot there, so a developer already has a cable in that connector and expects adb on the same one. The fact originates in firmware.

Which is why I think the kernel command line is the right carrier, and why it doesn't conflict with shared rootfs: `adbd.udc=` comes from the per-board boot configuration, not from the image, so one rootfs still serves every machine. Your shared-rootfs objection does apply to the `/etc/default` fragment I floated earlier — I'm happy to drop that path and keep only the cmdline override if that settles it.

lumag:
Yes, "peripheral" is a default, that's why I was proposing to use it (to reaffirm the default). Hower I agree with your comments. Let's drop the `/etc/defaut` part, I'd be happy with the rest.

shawngsc:
Dropped the `/etc/default` path in v2 — the override now comes only from the `adbd.udc=` kernel command line argument. `ADBD_UDC` and the sourcing of `/etc/default/android-tools-adbd` are gone, and the comment header and commit message are updated to match. The presence check and the warning-plus-fallback to the speed-based scan are unchanged.

Patches 1 and 2 are untouched. Range is fa91be1 -> cdcd21c.

lumag (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/f4a691da-9ed4-4e61-96f6-50d275337870)**
---
**Test jobs for commit [`cdcd21c`](https://github.com/qualcomm-linux/meta-qcom/commit/cdcd21cf787852419bad0c4e92c1fbbf87c73f51)**
---
Pass: 9 | Fail: 0 | Total: 9

koenkooi (approved):

ricardosalveti (approved):

lumag:
Known issues.

PR #2968, opened 2026-08-19
qcom-sv-binaries: Release initial sv binaries recipe

sasingh-cv:
Add a recipe for the Snapdragon Vision API runtime libraries.

The SV API provides a client interface to Qualcomm EVA
hardware-accelerated vision features. It enables Camera, Video,
and CV applications to configure EVA features, submit buffers,
communicate with the EVA driver and firmware stack, and retrieve
hardware-generated outputs and metadata.

The SV API complements OpenCV and FastCV. OpenCV provides
general-purpose computer-vision algorithms, while FastCV
provides Qualcomm-optimized CV primitives. Neither provides the
complete EVA-specific hardware, buffer, metadata, and firmware
integration required by EVA workloads.

PR #3086, opened 2026-09-07
eva-driver: Add EVA kernel driver recipe

sasingh-cv:
Add a recipe for the Qualcomm EVA (Engine for Visual Analytics) kernel driver.

The EVA driver provides the kernel-space interface to Qualcomm's vision acceleration hardware. It manages device
initialization, memory handling, firmware interaction, buffer management, and communication between userspace clients and EVA hardware.

The recipe builds and installs the msm-eva kernel module along with the associated UAPI headers required by userspace components to interact with the driver.

The driver forms the foundation of the EVA software stack and is consumed by higher-level vision runtimes such as Snapdragon Vision (SV), enabling applications to access hardware-accelerated vision capabilities on supported Qualcomm platforms.

Note: This PR is for review only and it will be merged only after getting the eva-driver overlay exception.

### DIFF
--- a/recipes-multimedia/eva/eva-driver_1.0.1.bb
+++ b/recipes-multimedia/eva/eva-driver_1.0.1.bb
@@ -1,0 +1,33 @@
+DESCRIPTION = "Qualcomm EVA (CVP) driver"
+HOMEPAGE = "https://github.com/qualcomm-linux/eva-driver"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://Kbuild;beginline=1;endline=1;md5=829e563511c9a1d6d41f17a7a4989d6a"
+
+SRC_URI = " \
+    git://github.com/qualcomm-linux/eva-driver.git;protocol=https;branch=eva-kernel.qclinux.0.0;tag=${PV} \
+"
+
+SRCREV = "6e20debbe82a3939ad3ec5e09ddf6662224ff7bd"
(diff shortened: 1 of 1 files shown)


lumag:
It doesn't even build.

lumag:
Where is the firmware? Why is it this  PR being submitted separately from #2968 ? Also, the question there is unanswered.

lumag:
Also, is firmware loaded on the first use or on the driver load? The latter is a much worse option for the system integration.

sasingh-cv:
> It doesn't even build.

Actually, it has missing dependencies like evacc headers in kernel source. That's why it might not be building.

sasingh-cv:
> Where is the firmware? Why is it this PR being submitted separately from #2968 ? Also, the question there is unanswered.

For firmware bins , once the FW bins are uploaded and hosted in JFrog QArtifactory , we will raise the FW bb recipe meta-qcom PR for the same. 
We chose to keep this PR focused on the EVA driver recipe and submit it separately from the SV recipe to maintain modularity and simplify the review process.

> Also, is firmware loaded on the first use or on the driver load? The latter is a much worse option for the system integration.

Firmware loading will happen on first use.

lumag:
> > It doesn't even build.
> 
> Actually, it has missing dependencies like evacc headers in kernel source. That's why it might not be building.

Please don't submit the work which can't be merged.

PR #3087, opened 2026-09-07
shikra-evk: fix CQM/CQS IMX577 DTBO entry

Komal-Bajaj:
The machine configuration still listed shikra-cqm-evk-imx577-camera.dtbo in KERNEL_DEVICETREE. The camera overlay name used for this target is shikra-cqm-cqs-evk-imx577-camera.dtbo, so the stale entry points to a non-existent artifact.

Update KERNEL_DEVICETREE to use the current combined CQM/CQS camera DTBO name.


lumag (changes requested):
Doesn't build. If it's something that got changes in the kernel, it must be submitted as a part of the corresponding PR (and even squashed in the corresponding commit to prevent breaking the build).

PR #2699, opened 2026-07-08
images: add support to store the images artifacts on sstate-cache

quaresmajose:

PR #3088, opened 2026-09-07
ci: use specific variable for build-id

quaresmajose:
The variable BUILD_ID should be excluded from the dependencies. To do this more easily, the variable QCOM_BUILD_ID is introduced.

The content of BUILD_ID is usually always different each time it's runs or parsed. This happens because the variable contains the value of DATETIME or the ${{ github.run_id }}-${{ github.run_attempt }} github ci action. Therefore, we should rule out running any task again if it is solely due to changes in the BUILD_ID.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -69,7 +69,9 @@ local_conf_header:
     # Use a weak assignment so the CI override wins regardless of how kas
     # orders the merged local.conf fragments.
     OS_RELEASE_FIELDS:append = " BUILD_ID"
-    BUILD_ID ?= "local-${DATETIME}"
+    QCOM_BUILD_ID ?= "local-${DATETIME}"
+    BUILD_ID ?= "${QCOM_BUILD_ID}"
+    BUILD_ID[vardepsexclude] ?= "QCOM_BUILD_ID"
 
 machine: unset
 
--- a/.github/workflows/compile.yml
+++ b/.github/workflows/compile.yml
@@ -90,7 +90,7 @@ jobs:
             version: 14
           local_conf_header:
             build-id: |
-              BUILD_ID = "${BUILD_ID}"
+              QCOM_BUILD_ID = "${BUILD_ID}"
           EOF
         env:
           INPUTS_CACHE_DIR: ${{inputs.cache_dir}}


lumag (commented):
> The variable BUILD_ID should be excluded from the dependencies.

Why? Commit messages should explain the reason for the changes.

quaresmajose:
> > The variable BUILD_ID should be excluded from the dependencies.
> 
> Why? Commit messages should explain the reason for the changes.

I added it to the commit:

The content of BUILD_ID is usually always different each time it's runs or parsed. This happens because the variable contains the value of DATETIME or the ${{ github.run_id }}-${{ github.run_attempt }} github ci action. Therefore, we should rule out running any task again if it is solely due to changes in the BUILD_ID.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/7f6fb62d-a918-4e17-ac6d-223b0c2659b9)**
---
**Test jobs for commit [`44b6a3e`](https://github.com/qualcomm-linux/meta-qcom/commit/44b6a3e426e89d4808e83ff3b3230131b828f671)**
---
Pass: 9 | Fail: 0 | Total: 9

mwasilew:
I have a strange feeling this will brake all LAVA jobs. Let me check more thoroughly.

quaresmajose:
> I have a strange feeling this will brake all LAVA jobs. Let me check more thoroughly.

I believe this can be verified forcing a new ci rebuild/retest using the [Re-run all jobs] button of a ci job that has already been completed.

mwasilew:
I pushed this change to `next` https://github.com/qualcomm-linux/meta-qcom/actions/runs/34359170198

mwasilew:
I think this is OK from testing perspective. The jobs are still running but they look OK. Example here:
https://lava.infra.foundries.io/scheduler/job/406800/definition#defline15

mwasilew (approved):

quaresmajose:
> I think this is OK from testing perspective. The jobs are still running but they look OK. Example here: https://lava.infra.foundries.io/scheduler/job/406800/definition#defline15

Thanks for checking it.

ci/base.yml
```diff
@@ -69,7 +69,9 @@ local_conf_header:
     # Use a weak assignment so the CI override wins regardless of how kas
     # orders the merged local.conf fragments.
     OS_RELEASE_FIELDS:append = " BUILD_ID"
-    BUILD_ID ?= "local-${DATETIME}"
+    QCOM_BUILD_ID ?= "local-${DATETIME}"
+    BUILD_ID ?= "${QCOM_BUILD_ID}"
+    BUILD_ID[vardepsexclude] ?= "QCOM_BUILD_ID"
```
ricardosalveti, line 74: I don't get the change, this will result in builds reusing the build_id from the cache, which will produce a mismatch in os-release / uki / etc, and it goes against the main purpose of the change that added this.

ricardosalveti: And some additional findings from AI:

- DATETIME exclusion is lost. BUILD_ID[vardepsexclude] ?= "QCOM_BUILD_ID" in local.conf is parsed before the recipe, so oe-core's ?= "DATETIME" becomes a no-op. Anyone setting BUILD_ID = "${DATETIME}-tag" (the upstream-documented pattern) now rebuilds os-release and every image on each invocation. Fix is "QCOM_BUILD_ID DATETIME", or drop the flag entirely.
- Both comments are now false. ci/base.yml lines 66-70 still say CI overrides BUILD_ID via the fragment, and compile.yml line 86 still promises os-release maps back to the S3 folder. Neither is true after this change.
- Weak binding regression. BUILD_ID ?= "${QCOM_BUILD_ID}" means any downstream hard BUILD_ID = unconditionally disconnects CI's run id. No in-tree fragment does this today.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/49f24fcb-620d-4d71-8a6a-5dce5759d842)**
---
**Test jobs for commit [`b277562`](https://github.com/qualcomm-linux/meta-qcom/commit/b27756242ec02cf256b761e4a5e79d3a051700f5)**
---
Pass: 9 | Fail: 0 | Total: 9

ci/base.yml
quaresmajose: The idea behind this change is to enable the reuse of images from the sstate-cache. Something I've been cooking up for a while now in https://github.com/qualcomm-linux/meta-qcom/pull/2699

It may happen that the image in question does not correspond to the **${{ github.run_id }}-${{ github.run_attempt }}**  of the executed version in CI. But if this is the case, it's because the image met the conditions to be reused.

So for the execution **AAAA-2** in CI we can have **AAAA-1** inside the rootfs of the image. This means that the image was reused from _${{ github.run_attempt }}=_**1**.

ricardosalveti: I understand the idea, but while this will save build time and increase our reproducibility in our CI, it might also confuse users flashing our images for development and debugging purposes, as the runid won't necessarily match.

Unfortunately I don't really have a better solution as anything that touches the image will cause it to be created from scratch.

quaresmajose: I'm going to move this change to https://github.com/qualcomm-linux/meta-qcom/pull/2699 and leave it there while I implement the functionality.

Issue #2484, opened 2026-06-16
Make CDT selection more user friendly 


lumag:
Currently we are shipping multiple cdts as part of the qcomflash tarball, and the user has to select the right one at flash time.

This is what we already have today.

_Originally posted by @ricardosalveti in https://github.com/qualcomm-linux/meta-qcom/issues/2475#issuecomment-4720534812_

vkraleti:
@ishnajain-qcom can you provide the latest on open sourcing CDT generation script?

ishnajain-qcom:
> [ishnajain (@ishnajain-qcom)](https://github.com/ishnajain-qcom) can you provide the latest on open sourcing CDT generation script?

Hi @vkraleti, script is ready,  please let us know where it needs to be hosted?

lumag:
@ishnajain-qcom could you please share it for the review?

ishnajain-qcom:
> [ishnajain (@ishnajain-qcom)](https://github.com/ishnajain-qcom) could you please share it for the review?

sure.. sharing internally

lumag:
@ishnajain-qcom another pointer / request from the CI point of view: https://github.com/qualcomm-linux/meta-qcom/issues/3002#issuecomment-5636732846

PR #3091, opened 2026-09-08
firmware-qcom-boot-qcs615: add CDT for IQ615 LYRA EVK

gummadiv-qcom:
Add IQ-615-LYRA-EVK_V2.zip as CDT source for IQ615 LYRA EVK board

### DIFF
--- a/recipes-bsp/firmware-boot/firmware-qcom-cdt-qcs615.bb
+++ b/recipes-bsp/firmware-boot/firmware-qcom-cdt-qcs615.bb
@@ -3,9 +3,11 @@ DESCRIPTION = "CDT (Configuration Data Table) Firmware for Qualcomm QCS615 platf
 SRC_URI = " \
     https://${CDT_ARTIFACTORY}/QCS615/cdt/ADP_AIR_SA6155P_V2.zip;downloadfilename=cdt-qcs615-adp-air_${PV}.zip;name=qcs615-adp-air \
     https://${CDT_ARTIFACTORY}/QCS615/cdt/EVK_SA6150P.zip;downloadfilename=cdt-iq-615-evk_${PV}.zip;name=iq-615-evk \
+    https://${CDT_ARTIFACTORY}/QCS615/cdt/IQ-615-LYRA-EVK_V2.zip;downloadfilename=cdt-iq-615-lyra-evk_${PV}.zip;name=iq-615-lyra-evk \
     "
 SRC_URI[qcs615-adp-air.sha256sum] = "37d99eb113e286400bce0d70aa12a74d05f93d01f045bf67e7a46b3c606c8fd0"
 SRC_URI[iq-615-evk.sha256sum] = "e8ea707a27090e4338c709cabdad46159c06cd91e7c2fc5348f5d98db7bbb802"
+SRC_URI[iq-615-lyra-evk.sha256sum] = "c6ac093dbb5ad52bd159fe5e81aef9c0825489d8d3992f65d329acc7de0f1e6e"
 
 QCOM_CDT_SUBDIR = "qcs615"
 


lumag (changes requested):
> Signed-off-by: gummadiv <gummadiv@qti.qualcomm.com>

Please fix your Git configuration. Also, you've added the Lyra CDT. How is it going to be used?

lumag:
And it doesn't build.

lumag:
No stray merge commits.

ishnajain-qcom:
> > Signed-off-by: gummadiv [gummadiv@qti.qualcomm.com](mailto:gummadiv@qti.qualcomm.com)
> 
> Please fix your Git configuration. Also, you've added the Lyra CDT. How is it going to be used?

Hi @lumag, it will be placed in flat build created during qcs615 compilation. And customers can simply rename it if they want this CDT to be flashed as per the current procedure for other form factors as well

ishnajain-qcom:
Also, please help to add 2.1 milestone release tag to this PR

lumag:
> > > Signed-off-by: gummadiv [gummadiv@qti.qualcomm.com](mailto:gummadiv@qti.qualcomm.com)
> > 
> > 
> > Please fix your Git configuration. Also, you've added the Lyra CDT. How is it going to be used?
> 
> Hi Dmitry Baryshkov (@lumag), it will be placed in flat build created during qcs615 compilation. And customers can simply rename it if they want this CDT to be flashed as per the current procedure for other form factors as well

This was already raised up for https://github.com/qualcomm-linux/meta-qcom/pull/2475#issuecomment-4720534812 and then we have https://github.com/qualcomm-linux/meta-qcom/issues/2484. Users should not need to rename anything. If the machine conf supports multiple machines, each requiring the separate CDT, then we need to build a sepate flashing directory / archive for each of those machines. Be user friendly (and scripting-friendly).

ishnajain-qcom:
#2484  was concluded to open source CDT generation script, we are internally waiting for an update from respective team, but until then can we follow the same procedure

lumag:
> #2484 was concluded to open source CDT generation script, we are internally waiting for an update from respective team, but until then can we follow the same procedure

How will CDT generator script help with this case?

ricardosalveti:
The problem is that so far we only had CDTs covering variations of the hardware, and here it seems to be a different EVK. Would the standard EVK CDT also work on lyra when not copied over by the user?

ishnajain-qcom:
Responding to both queries together

The standard EVK CDT is not expected to work on the Lyra EVK, which is why a separate Lyra CDT is being added.

My reference to #2484 was mainly around current solution being discussed around CDT and not as a solution to the packaging problem. I agree that the CDT generator script by itself doesn't address the issue of automatically selecting the correct CDT for a given machine.

For this change, we are following the current approach where CDT variants are included in the build artifacts. Since Lyra requires a different CDT, we wanted to ensure users have access to the correct one.

I also agree that requiring users to rename or manually select CDT files is not ideal. A machine-specific flashing directory/archive that automatically contains the correct CDT would be a much cleaner solution from both user and scripting perspectives.

If there's a preferred mechanism in meta-qcom today for handling machine-specific CDT packaging/artifacts, please point me in that direction and I'm happy to rework this patch accordingly. 

My goal here is primarily to ensure Lyra support is available in 2.1 release while staying consistent with the current flow

lumag:
> If there's a preferred mechanism in meta-qcom today for handling machine-specific CDT packaging/artifacts, please point me in that direction and I'm happy to rework this patch accordingly.
> 
> My goal here is primarily to ensure Lyra support is available in 2.1 release while staying consistent with the current flow

Then only mechanism available today is "create separate machine configuration". This PR in its current way can't be landed. Please work either on a proper generation of qflash folders or split Lyra to a separate machine.

vkraleti:
> > If there's a preferred mechanism in meta-qcom today for handling machine-specific CDT packaging/artifacts, please point me in that direction and I'm happy to rework this patch accordingly.
> > My goal here is primarily to ensure Lyra support is available in 2.1 release while staying consistent with the current flow
> 
> Then only mechanism available today is "create separate machine configuration". This PR in its current way can't be landed. Please work either on a proper generation of qflash folders or split Lyra to a separate machine.

The only difference between the Tessolve and Lyra EVK images is the CDT. Introducing a separate machine configuration for Lyra would significantly increase the build matrix, resulting in additional builds, a substantial CI overhead. Generating separate `qcomflash ` folders for different CDT variants is more scalable approach. 

But, considering  2.1 release timelines, could this enhancement be implemented in parallel as a follow-up PR? The existing `recipes-bsp/firmware-boot/firmware-qcom-cdt-qcs615.bb` recipe already supports 2 CDTs, so adding the Lyra EVK CDT is essentially an extension to the current model of handling CDTs.

lumag:
> But, considering 2.1 release timelines, could this enhancement be implemented in parallel as a follow-up PR? The existing `recipes-bsp/firmware-boot/firmware-qcom-cdt-qcs615.bb` recipe already supports 2 CDTs, so adding the Lyra EVK CDT is essentially an extension to the current model of handling CDTs.

We can't really claim that we support Lyra if the default image that is generated programs wrong CDT.

ricardosalveti:
> The only difference between the Tessolve and Lyra EVK images is the CDT.

And is this just to load a different DTB?

vkraleti:
> > The only difference between the Tessolve and Lyra EVK images is the CDT.
> 
> And is this just to load a different DTB?

Yes, Lyra DTB is going to be part of dtb.bin. (https://github.com/qualcomm-linux/meta-qcom/pull/3103), CDT would help UEFI to load this DTB.

ricardosalveti:
Are we going to add lyra evks to lava as well?

vkraleti:
> Are we going to add lyra evks to lava as well?

I checked with the team. There are no immediate plans to add Lyra EVKs to LAVA. But, this is expected to happen, as the Tessolve boards are expected to be replaced with Lyra.

lumag:
> > Are we going to add lyra evks to lava as well?
> 
> I checked with the team. There are no immediate plans to add Lyra EVKs to LAVA. But, this is expected to happen, as the Tessolve boards are expected to be replaced with Lyra.

We can't really support Lyra if it's not in the CI.

vkraleti:
> > > Are we going to add lyra evks to lava as well?
> > 
> > 
> > I checked with the team. There are no immediate plans to add Lyra EVKs to LAVA. But, this is expected to happen, as the Tessolve boards are expected to be replaced with Lyra.
> 
> We can't really support Lyra if it's not in the CI.

@gummadiv-qcom @nkumarsi @akdwived-dev do work with @nareshkamboju to add Lyra EVKs to LAVA.

PR #3107, opened 2026-09-09
ci: move debug to the weekly profile and trim the catchall variants

ricardosalveti:
The debug rows are 62 of the nightly's 129 machine-hours and, at 77 minutes
each, the last jobs to finish. On top of that, the rt-6.18-distro-kvm kernel
entry turns the debug and performance rows into qcom-distro builds the
qcom-distro row already produces, so 32 jobs per run publish a duplicate
under the wrong name.

Drop the hijacked rt-kvm rows of debug and performance, move debug to the
weekly profile (which had no rows of its own), compress the debugfs tarball
with zstd instead of bzip2, and skip the image-level SBOM on debug and
performance through a new ci/nospdx.yml fragment.

The matrix goes from 221 to 189 configurations. Nightly and push runs lose
all debug rows; the weekly keeps them, at a fraction of the cost. The
published debug tarball is renamed from .rootfs-dbg.tar.bz2 to
.rootfs-dbg.tar.zst.

### DIFF
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -281,6 +281,14 @@ jobs:
                 name: qcom-distro-catchall
             kernel:
                 type: rt-6.18-distro-kvm
+          - distro:
+                name: debug
+            kernel:
+                type: rt-6.18-distro-kvm
+          - distro:
+                name: performance
+            kernel:
+                type: rt-6.18-distro-kvm
           # Exclude jobs from compile_warm_up
           - machine: glymur-crd
             distro:
--- a/.github/workflows/build-yocto.yml
+++ b/.github/workflows/build-yocto.yml
@@ -257,7 +257,7 @@ jobs:
             yamlfile: ':ci/qcom-distro-catchall.yml'
           - name: debug
             yamlfile: ':ci/qcom-distro-catchall.yml:ci/debug.yml'
-            profile: full
+            profile: weekly
           - name: performance
             yamlfile: ':ci/qcom-distro-catchall.yml:ci/performance.yml'
             profile: full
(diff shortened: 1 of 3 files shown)


ricardosalveti:
This is a move to push our most expensive runs to weekly, as I don't really think it is worth building debug and performance every day, we don't really test and use that often, and it is the most expensive part of our builds.

lumag (approved):

koenkooi (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/8e16afb0-8800-459f-bdf2-e56b09f0106f)**
---
**Test jobs for commit [`93bd74e`](https://github.com/qualcomm-linux/meta-qcom/commit/93bd74e27a2ff9d022cdc1ba70d39633642bf358)**
---
Pass: 9 | Fail: 0 | Total: 9

.github/workflows/build-yocto.yml
```diff
@@ -256,10 +256,10 @@ jobs:
           - name: qcom-distro-catchall
             yamlfile: ':ci/qcom-distro-catchall.yml'
           - name: debug
-            yamlfile: ':ci/qcom-distro-catchall.yml:ci/debug.yml'
+            yamlfile: ':ci/qcom-distro-catchall.yml:ci/debug.yml:ci/nospdx.yml'
             profile: weekly
           - name: performance
-            yamlfile: ':ci/qcom-distro-catchall.yml:ci/performance.yml'
+            yamlfile: ':ci/qcom-distro-catchall.yml:ci/performance.yml:ci/nospdx.yml'
```
quaresmajose, line 262: I thought we could use drop the create-spdx-image-3.0 directly on ci/performance.yml and ci/debug.yml. That way we wouldn't even need the ci/nospdx.yml.

quaresmajose (approved):

.github/workflows/build-yocto.yml
ricardosalveti: I think it is still valuable to have that fragment, let's keep this for now and we can revisit if needed.

ricardosalveti:
Usual known failures.

PR #3109, opened 2026-09-09
ci/kas-container-shell-helper: run with the build's kas configuration

ricardosalveti:
The helper always starts kas with ci/base.yml, so the buildstats step between a job's world and image builds rewrites bblayers.conf for a smaller layer set. bitbake then treats its parse cache as stale and the image build parses all 3851 recipes again instead of loading them, 40 s against about 10 s for a cache hit, on each of the 36 nightly jobs that build world.

Use KAS_YAMLS when the caller exports it, as compile.yml does, and keep ci/base.yml as the default for the layer checks.

### DIFF
--- a/ci/kas-container-shell-helper.sh
+++ b/ci/kas-container-shell-helper.sh
@@ -17,4 +17,4 @@ SCRIPT=${SCRIPT#$TOPDIR/}
 # on ci the kas-container is not on the default path
 KAS_CONTAINER=${KAS_CONTAINER:-$(which kas-container)}
 
-exec $KAS_CONTAINER shell $TOPDIR/ci/base.yml --command "/repo/$SCRIPT /repo /work"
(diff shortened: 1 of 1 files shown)


koenkooi (approved):
The commit message needed a 3rd read before I caught on, but the change is good.

quaresmajose (approved):

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/3dbe94d5-1688-4aae-b6d0-86c8eaf7e957)**
---
**Test jobs for commit [`1a392b2`](https://github.com/qualcomm-linux/meta-qcom/commit/1a392b2bd841462e95fdd4b2fc24aa3f9911ae7b)**
---
Pass: 40 | Fail: 0 | Total: 40
Pass: 9 | Fail: 0 | Total: 9

anujm1 (approved):

PR #3110, opened 2026-09-09
image_types_qcom: hard link the ESP image into the qcomflash directory

ricardosalveti:
create_qcomflash_pkg installs the ESP image into a directory that exists only so it can be tarred, and never modifies it there, the same situation rootfs.img was in. The copy reads and writes the 512 MiB image once more per image, on a task that is already the heaviest disk writer of the image build, for a file whose only reader is the tar that follows.

Hard link it instead, so the tarball is built from the deployed file and nothing is written.

### DIFF
--- a/classes-recipe/image_types_qcom.bbclass
+++ b/classes-recipe/image_types_qcom.bbclass
@@ -53,7 +53,7 @@ deploy_partition_files() {
 
 create_qcomflash_pkg() {
     # esp image
-    [ -n "${QCOM_ESP_FILE}" ] && install -m 0644 ${QCOM_ESP_FILE} efi.bin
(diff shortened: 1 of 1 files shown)


koenkooi (approved):

lumag (approved):

ricardosalveti:
Known failures.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/8ccc2381-dad0-4c50-843f-53e4e5f1a60c)**
---
**Test jobs for commit [`77ff467`](https://github.com/qualcomm-linux/meta-qcom/commit/77ff467b89ca9ef31da9d79ed3ee162e988c0139)**
---
Pass: 9 | Fail: 0 | Total: 9

PR #3160, opened 2026-09-15
linux-qcom: build IPv6 in for early systemd startup

wenwfu:
systemd modprobes ipv6.ko during PID 1 initialization when it is configured as a module. Build IPv6 in so it is ready before userspace starts and remove this boot-critical module-load path.


recipes-kernel/linux/linux-qcom-6.18/configs/bsp-additions.cfg
```diff
@@ -1,6 +1,7 @@
 # for UFS enumeration
 CONFIG_PHY_QCOM_QMP=y
 CONFIG_SCSI_UFS_QCOM=y
+CONFIG_IPV6=y
```
quaresmajose, line 4: This fit better on the distro side, IPV6 is not critical and reqired for the BSP and should considerably increase the kernel size.

wenwfu: thanks, will move it to meta-qcom-distro

wenwfu: https://github.com/qualcomm-linux/meta-qcom-distro/pull/476  made a new PR on the distro side

PR #3162, opened 2026-09-15
ci/yocto-buildstats: add ccache report

quaresmajose:
It is nice to have a report about the space used by ccache. This makes it easier to understand how the space is distributed among the components.

The output will be similar to the one below:

```
| Ccache size (0.81GB) [/sstate/ccache]:
|  47%   0.38 gcc                  #########################################
|  18%   0.15 llvm-native          #################
|  17%   0.14 glibc                ###############
|  16%   0.13 linux-yocto          ###############
```


lumag:
I like the idea

PR #3165, opened 2026-09-16
update boot firmware to 00143

ishnajain-qcom:
This PR updates the QCS615,QCS9100, QCS8300 and QCS6490 boot firmware to tag 00143 and adapts to a new upstream delivery format where KVM is now the default XBL configuration (xbl_config.elf), and Gunyah is explicitly renamed (xbl_config_gunyah.elf)."


vkraleti:
@ishnajain-qcom you also need to swap QCOM_XBL_CONFIG selection in [qcom-common.inc](https://github.com/qualcomm-linux/meta-qcom/blob/master/conf/machine/include/qcom-common.inc#L39).

vkraleti:
@ishnajain-qcom  Just now noticed that the commit 7c9711df4d60cb709649285528765d834051e386 is adding an unwanted QCS615_bootbinaries_00143.zip file. You must drop it.

ricardosalveti:
> ishnajain (@ishnajain-qcom) Just now noticed that the commit [7c9711d](https://github.com/qualcomm-linux/meta-qcom/commit/7c9711df4d60cb709649285528765d834051e386) is adding an unwanted QCS615_bootbinaries_00143.zip file. You must drop it.

Why unwanted?

ricardosalveti:
I think renaming xbl_config, and changing the hypervisor as a side effect, is a bad change, as it will cause unexpected behavior for users consuming this update (suddenly the default xbl_config will switch from gunyah to kvm).

Also, we can only do this rename for all targets, so it needs to be an all or nothing. It would be terribly confusing to have kvm enable via xbl_config for some but gunyah for others, using the same file name.

We should probably just stick with different files, and have a variable defining the default one to be used as part of the machine conf.

vkraleti:
> > ishnajain (ishnajain (@ishnajain-qcom)) Just now noticed that the commit [7c9711d](https://github.com/qualcomm-linux/meta-qcom/commit/7c9711df4d60cb709649285528765d834051e386) is adding an unwanted QCS615_bootbinaries_00143.zip file. You must drop it.
> 
> Why unwanted?

There is no need check-in binary .zips into the layer, they are supposed to be fetched from a remote location.
 
image

lumag (changes requested):
Why? What is wrong with the current naming scheme? If you  want to be explicit, don't add "default" config. Let's have two separate entries, `xbl_config_kvm.mbn` and `xbl_config_gunyah.mbn`.

quaresmajose:
> > > ishnajain (ishnajain (ishnajain (@ishnajain-qcom))) Just now noticed that the commit [7c9711d](https://github.com/qualcomm-linux/meta-qcom/commit/7c9711df4d60cb709649285528765d834051e386) is adding an unwanted QCS615_bootbinaries_00143.zip file. You must drop it.
> > 
> > 
> > Why unwanted?
> 
> There is no need check-in binary .zips into the layer, they are supposed to be fetched from a remote location.
> 
> image

I believe the zipper was left out by mistake; there is no reference to its location in the recipes.

qcomlnxci:
**Test Coral run [workflow](https://qswat.qualcomm.com/coral/qli-open-development/workflows/meta-qcom-extensive-test/runs/b2fc0041-67e2-448f-98ea-988a3a32d678)**
---
**Test jobs for commit [`d706152`](https://github.com/qualcomm-linux/meta-qcom/commit/d7061528f4876cfea88dd32db7d6d524be52ab0b)**
---
Pass: 16 | Fail: 1 | Others: 23 | Total: 40
Pass: 6 | Fail: 0 | Others: 3 | Total: 9

ldts:
for QCS6490, how is cpucp.elf being shipped now? it is being referenced in rawprogram4.xml so I cant flash it.

lumag (changes requested):
Fails to boot on Lemans and Monaco EVKs.

PR #3202, opened 2026-09-21
refpolicy: allow TPM2 abrmd DBus communication to fix command failures

khalidfaisal17:
TPM2 commands were failing due to SELinux policy restrictions blocking communication between clients, the DBus daemon, and the TPM2 access broker/resource manager (tpm2-abrmd).

This change introduces the necessary SELinux permissions to allow:

- init and unconfined domains to communicate with tpm2-abrmd over DBus
- dbus daemon (system_dbusd_t) to establish unix stream socket connections with tpm2-abrmd
- a new interface (tpm2_dbus_stream_connect) to standardize this access

**Key changes:**

- Added tpm2_dbus_chat_abrmd() for init and unconfined domains
- Introduced tpm2_dbus_stream_connect() interface to allow DBus ↔ abrmd socket communication
- Enabled system_dbusd_t to connect to tpm2_abrmd_t over unix_stream_socket

Without these permissions, TPM2 tools and services fail due to AVC denials when attempting to use abrmd via DBus.


koenkooi:
The stack of patches keeps growing, what is the plan for getting these into upstream OE layers?

lumag:
> Added tpm2_dbus_chat_abrmd() for init and unconfined domains

This sounds wrong.

ricardosalveti:
This is not really BSP material.

Even if we decide to take it (until available via the upstream layer), it should go to meta-qcom-distro.

